PR #4, opened 2019-04-02
initial implementation of vae

bayesian:
@emailweixu

### DIFF
--- a/alf/algorithms/vae.py
+++ b/alf/algorithms/vae.py
@@ -1,0 +1,64 @@
+import tensorflow as tf
+
+
+class VariationalAutoEncoder:
+    """
+    VariationalAutoEncoder encodes data into diagonal multivariate gaussian, do sampling with
+    reparametrization trick, and return kl divergence between posterior and prior.
+    """
+    def __init__(self,
+                 hidden_dims,
+                 prior_network=None,
+                 preprocess_layers=None):
+
+        """Create an instance of `VariationalAutoEncoder`
+
+        Args:
+             hidden_dims (int): dimension of latent dimension
+             prior_network:     network to compute the priors (mean, log_var)
+             preprocess_layers: layers to preprocess input data before project into (mean, log_var)
+
+        """
+        if preprocess_layers is None:
+            # according to appendix in deepmind merlin paper
+            self._preprocess_layers = tf.keras.Sequential([
+                tf.keras.layers.Dense(2 * hidden_dims, activation='tanh') for i in range(2)])
+        else:
+            self._preprocess_layers = preprocess_layers
+
+        self._hidden_dims = hidden_dims
+        self._prior_network = prior_network
+        self._z_mean = tf.keras.layers.Dense(hidden_dims)
+        self._z_log_var = tf.keras.layers.Dense(hidden_dims)
+
+    def sampling_forward(self, inputs):
+        """Encoding the data into latent space then do sampling.
+
+        Args:
+            inputs (Tensor or Tuple of Tensor): data to be encoded. If it has a prior network, then the argument
+            is a tuple of (prior_input, new_observation).
+        """
+        if self._prior_network:
+            prior_input, new_obs = inputs
+            prior_z_mean, prior_z_log_var = self._prior_network(prior_input)
+            inputs = tf.concat([prior_input, new_obs, prior_z_mean, priror_z_log_var], -1)
+
+        latents = self._preprocess_layers(inputs)
+        z_mean = self._z_mean(latents)
+        z_log_var = self._z_log_var(latents)
+
+        if self._prior_network:
+            z_mean += prior_z_mean
+            z_log_var += prior_z_log_var
+
+            kl_div_loss = tf.square(z_mean - prior_z_mean) / tf.exp(prior_z_log_var) + \
+                          tf.exp(z_log_var - prior_z_log_var) + \
+                          prior_z_log_var - z_log_var - 1.0
+        else:
+            kl_div_loss = tf.square(z_mean) + tf.exp(z_log_var) - 1.0 - z_log_var
+
+        kl_div_loss = 0.5 * tf.reduce_sum(kl_div_loss, axis=-1)
+        # reparameterization sampling: z = u + var ** 0.5 * eps
+        eps = tf.random.normal(tf.shape(z_mean), dtype=tf.float32, mean=0., stddev=1.0)
+        z = z_mean + tf.exp(z_log_var / 2) * eps
+        return z, kl_div_loss
--- a/alf/algorithms/vae.py
+++ b/alf/algorithms/vae.py
@@ -1,10 +1,22 @@
 import tensorflow as tf
 from tf_agents.networks import network
 
-class VariationalAutoEncoder(network.Network):
+class VariationalAutoEncoder:
     """
     VariationalAutoEncoder encodes data into diagonal multivariate gaussian, do sampling with
     reparametrization trick, and return kl divergence between posterior and prior.
+
+    mathematically:
+
+    log p(x) >= E_z log P(x|z) - KL(q(z|x) || prior(z))
+
+    sampling_forward method return sampled z and KL, it is up to user of this class to use returned z to
+    decode and compute reconstructive loss to combine with kl loss returned here to optimize the whole network.
+
+    See vae_test.py for example usages to train vanilla vae, conditional vae and vae with prior network
+    on mnist dataset.
+
+    comments: this class is not subclass of tf_agents.network, thus not copible.
     """
     def __init__(self,
                  hidden_dim,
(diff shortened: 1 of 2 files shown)


alf/algorithms/vae.py
```diff
@@ -0,0 +1,81 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import tensorflow as tf
+
+
+class VariationalAutoEncoder:
```
emailweixu: need to be subclass of tf_agents.networks.Network

alf/algorithms/vae.py
```diff
@@ -0,0 +1,81 @@
...
+        z_mean = self._z_mean(latents)
+        z_log_var = self._z_log_var(latents)
+
+        if self._prior_network:
+            z_mean += prior_z_mean
+            z_log_var += prior_z_log_var
+
+            kl_div_loss = tf.square(z_mean - prior_z_mean) / tf.exp(prior_z_log_var) + \
+                          tf.exp(z_log_var - prior_z_log_var) + \
+                          prior_z_log_var - z_log_var - 1.0
```
emailweixu: The computation can be simplified a lot if you do this before line 66 and it can even be unified with line 73.

alf/algorithms/vae.py
```diff
@@ -0,0 +1,81 @@
...
+        """Create an instance of `VariationalAutoEncoder`
+
+        Args:
+             hidden_dims (int): dimension of latent dimension
+             prior_network:     network to compute the priors (mean, log_var)
+             preprocess_layers: layers to preprocess input data before project into (mean, log_var)
+
+        """
+        if preprocess_layers is None:
+            # according to appendix in deepmind merlin paper
```
emailweixu: better to use a more explicit reference such as arXiv:...

alf/algorithms/vae.py
```diff
@@ -0,0 +1,81 @@
...
+    VariationalAutoEncoder encodes data into diagonal multivariate gaussian, do sampling with
+    reparametrization trick, and return kl divergence between posterior and prior.
+    """
+
+    def __init__(self, hidden_dims, prior_network=None,
+                 preprocess_layers=None):
+        """Create an instance of `VariationalAutoEncoder`
+
+        Args:
+             hidden_dims (int): dimension of latent dimension
```
emailweixu: dimension of the latent vector

alf/algorithms/vae.py
```diff
@@ -0,0 +1,81 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import tensorflow as tf
+
+
+class VariationalAutoEncoder:
+    """
+    VariationalAutoEncoder encodes data into diagonal multivariate gaussian, do sampling with
```
emailweixu: Because we are essential doing conditional VAE here, please comment the formulation of conditional VAE:  log p(x|y) = E_{z~q(z|x,y)} log P(x|z,y) - KL(q(z|x,y)||p(z|y)) and note that this class only compute z and  KL divergence.

alf/algorithms/vae.py
```diff
@@ -0,0 +1,81 @@
...
+        self._prior_network = prior_network
+        self._z_mean = tf.keras.layers.Dense(hidden_dims)
+        self._z_log_var = tf.keras.layers.Dense(hidden_dims)
+
+    def sampling_forward(self, inputs):
+        """Encoding the data into latent space then do sampling.
+
+        Args:
+            inputs (Tensor or Tuple of Tensor): data to be encoded. If it has a prior network, then the argument
+            is a tuple of (prior_input, new_observation).
```
emailweixu: document return value and their shapes

alf/algorithms/vae_test.py
```diff
@@ -0,0 +1,98 @@
...
+fig = plt.figure()
+#import pdb; pdb.set_trace()
+idx = 0
+for i in range(nrows):
+    for j in range(nrows):
+        fig.add_subplot(nrows, nrows, idx + 1)
+        plt.imshow(np.reshape(sampled_outputs[idx], (image_size, image_size)))
+        idx += 1
+
+plt.show()
```
emailweixu: We need unittests which explicitly tell us whether tests are successful or not.

emailweixu: The situation with prior network should also be tested.

alf/algorithms/vae.py
bayesian: not sure about the benefits of deriving from that class, the current way offer more flexibilities of combining with other modules

alf/algorithms/vae_test.py
bayesian: how do we set up the test cases for prior network? we need prior_input and new_obs, we have to do something fancy like memory as prior_input??

bayesian: it is hard to set up a test with prior network, unless we introduce memory as prior_input for the mnist dataset

alf/algorithms/vae.py
emailweixu: In many algorithms, we need the ability to clone a model (e.g. target network for dqn). That class make this possible.

alf/algorithms/vae_test.py
emailweixu: Perhaps use mnist class id as prior information?

alf/algorithms/vae.py
bayesian: 66 - 67 is the choice of merlin paper to combine prior network and encoding network.

69 is kl divergence equation from 
https://en.wikipedia.org/wiki/Kullback%E2%80%93Leibler_divergence
under Multivariate normal distributions

emailweixu: I'm not saying the computations are not correct. What I said it can be simplified by doing this before line 66, so that you don't need to do these subtractions of (z_mean - prior_z_mean) because before line 66 z_mean is the result of the subtraction.

alf/algorithms/vae.py
emailweixu: should be hidden_dim instead of hidden_dims?

bayesian:
added unit tests with prior_network, it works very well.

alf/algorithms/vae_test.py
```diff
@@ -0,0 +1,161 @@
...
+
+        inputs = tf.keras.layers.Input(shape=self.input_shape)
+        prior_inputs = tf.keras.layers.Input(shape=(1,))
+        prior_inputs_one_hot = tf.reshape(tf.one_hot(tf.cast(prior_inputs, tf.int32), 10), shape=(-1, 10))
+        z, kl_loss = encoder.sampling_forward((prior_inputs_one_hot, inputs))
+        outputs = decoding_layers(z)
+
+        loss = tf.reduce_mean(tf.keras.losses.mse(inputs, outputs) * self.original_dim + kl_loss)
+
+        model = tf.keras.Model(inputs=[prior_inputs, inputs], outputs=outputs, name = "vae")
```
emailweixu: no space around '=' for passing arguments: name="vae"

alf/algorithms/vae.py
bayesian: I don't think that it is conditional vae here,  though could be very easily turned into one. cvae does not have prior network, it uses class conditional/classes as both encoding features, and hidden dimension is also concat of conditional  and gaussian mean and var.

here we implemented two vae here.
1) vanilla vae.
2) vae with prior network.

alf/algorithms/vae.py
```diff
@@ -0,0 +1,67 @@
...
+             preprocess_layers: layers to preprocess input data before project into (mean, log_var)
+
+        """
+        super(VariationalAutoEncoder, self).__init__(dim, size)
+        if preprocess_layers is None:
+            # according to appendix 2.4-2.5 in paper: https://arxiv.org/pdf/1803.10760.pdf
+            self._preprocess_layers = tf.keras.Sequential([
+                tf.keras.layers.Dense(2 * hidden_dim, activation='tanh') for i in range(2)])
+        else:
+            self._preprocess_layers = preprocess_layers
```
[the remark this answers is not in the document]
bayesian: why do we need to make a copy of all network argument?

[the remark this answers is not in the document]
emailweixu: An example is to get the target network, like https://github.com/tensorflow/agents/blob/53f732b99f3f368c0aa62dd1efcb3dfbc809840b/tf_agents/agents/dqn/dqn_agent.py#L174

alf/algorithms/vae.py
emailweixu: Anyway, we should clearly document the computation it performs using some equations.

alf/algorithms/vae.py
```diff
@@ -0,0 +1,67 @@
...
+
+        """Create an instance of `VariationalAutoEncoder`
+
+        Args:
+             hidden_dim (int): dimension of latent vector
+             prior_network:     network to compute the priors (mean, log_var)
+             preprocess_layers: layers to preprocess input data before project into (mean, log_var)
+
+        """
+        super(VariationalAutoEncoder, self).__init__(dim, size)
```
emailweixu: `size` is not defined? And this is not a proper way to call Network.__init__

PR #7, opened 2019-04-06
WIP: SimpleAlgorithm and SimpleAgent

emailweixu:
When I am trying implementing MERLIN using tf_agents, I found that I have to write a lot of redundant code. So I tried to encapsulate some typical patterns of RL algorithms so that user can greatly simplify their code by using these two classes.

The overall framework is ready, but I haven't finish test yet. I want to get your feedbacks
simple_algorithm.py and simple_agent.py are good place to start reviewing the code.

### DIFF
--- a/alf/agents/simple_agent.py
+++ b/alf/agents/simple_agent.py
@@ -1,0 +1,150 @@
+# Copyright (c) 2019 Horizon Robotics. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import tensorflow as tf
+from tf_agents.agents import tf_agent
+from tf_agents.environments import trajectory
+from tf_agents.environments.time_step import TimeStep
+from tf_agents.networks.network import Network
+from tf_agents.utils import eager_utils
+
+from alf.algorithms import SimpleAlgorithm, LossStep
+
+
+def run_rnn(rnn_func, inputs, states):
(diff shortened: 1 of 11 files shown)


alf/agents/simple_agent.py
```diff
@@ -0,0 +1,150 @@
...
+    def _train_iter(self, trajectory, next_trajectory, weights):
+        batch_size = trajectory.discount.shape[0]
+
+        if self._is_rnn_policy:
+            critics, _ = run_rnn(self._algorithm.pre_critic, next_trajectory,
+                                 self._algorithm.get_initial_state(batch_size))
+        else:
+            critics, _ = self._algorithm.pre_critic_step(next_trajectory)
+        critics = self._algorithm.critic(trajectory, next_trajectory, critics)
+
```
runjerry: I think this split of pre_critic_step and critic is neat and convenient for value-based approaches and most actor-critic approaches, my minor concern comes from the following two types of approaches:
1. GAE computation seems not that straightforward in this split?
2. Do we want to have another base Agent&Algorithm for model-based approaches?

alf/agents/simple_agent.py
```diff
@@ -0,0 +1,150 @@
...
+        valid_mask = tf.cast(~trajectory.time_steps.is_last(), tf.float32)
+        with tf.GradientTape() as tape:
+            if self._is_rnn_policy:
+                loss, _ = run_rnn(
+                    self._algorithm.loss_step,
+                    LossStep(trajectory=trajectory, critic=critics),
```
runjerry: What is the purpose of this extra layer "LossStep"?

alf/algorithms/simple_algorithm.py
```diff
@@ -0,0 +1,228 @@
...
+    total_loss = self._calc_total_loss(loss_infos)
+    
+    grads = tape.gradient(total_loss, variables_to_train)
+
+    self.update(grads)
+    ```
+    """
+
+    def __init__(self, model, policy=None, collect_policy=None,
+                 optimizer=None):
```
runjerry: How do we enable an algorithm to have more than one network, say an actor network and a critic network (in our planner-controller algorithm there would be 3 or 4 networks), should be wrap them into a single model object or let the algorithm to accept multiple models?

emailweixu:
abandoned.

PR #8, opened 2019-04-10
text encoder and decoder

witwolf:


alf/algorithms/text_codec.py
```diff
@@ -0,0 +1,90 @@
...
+                    where b is batch and seq_len is sequence length
+                Returns:
+                    result Tensor: shape is (b, lstm_size).
+        """
+        model = tf.keras.Sequential()
+        model.add(
+            tf.keras.layers.Embedding(
+                self._vocab_size, self._embed_size, mask_zero=True))
+        model.add(tf.keras.layers.LSTM(self._lstm_size))
+        return model(inputs), network_state
```
emailweixu: model should be constructed at __init__

alf/algorithms/text_codec.py
```diff
@@ -0,0 +1,90 @@
...
+            seq_len (int): sequence length
+            embed_size (int): embedding size
+            lstm_size (int): lstm size for both encoding
+        """
+        self._vocab_size = vocab_size
+        self._seq_len = seq_len
+        self._embed_size = embed_size
+        self._lstm_size = lstm_size
+
+    def __call__(self, inputs, step_type=None, network_state=()):
```
emailweixu: should override  `call` instead

alf/algorithms/text_codec.py
```diff
@@ -0,0 +1,90 @@
...
+        Args:
+            vocab_size (int): vocabulary size
+            seq_len (int): sequence length
+            embed_size (int): embedding size
+            lstm_size (int): lstm size for both encoding
+        """
+        self._vocab_size = vocab_size
+        self._seq_len = seq_len
+        self._embed_size = embed_size
+        self._lstm_size = lstm_size
```
emailweixu: call `super().__init__`

PR #10, opened 2019-04-16
a2c implement

witwolf:


alf/utils/value_op_test.py
```diff
@@ -0,0 +1,66 @@
...
+        rewards = tf.constant([[2.] * 5], tf.float32)
+        is_lasts = tf.constant([[False] * 5], tf.bool)
+        discounts = tf.constant([[0.9] * 5], tf.float32)
+
+        expected = tf.constant([[
+            (((1 * 0.9 + 2) * 0.9 + 2) * 0.9 + 2) * 0.9 + 2,
+            ((1 * 0.9 + 2) * 0.9 + 2) * 0.9 + 2,
+            (1 * 0.9 + 2) * 0.9 + 2,
+            1 * 0.9 + 2,
+            1]], dtype=tf.float32)
```
emailweixu: According to the equation at value_ops.py:34,  Q_t = sum_{t'=t}^T gamma^(t'-t) * r_{t'} + gamma^(T-t+1)*final_value, the expected result should be
```
[
            ((((1 * 0.9 + 2) * 0.9 + 2) * 0.9 + 2) * 0.9 + 2) * 0.9 + 2,,
            (((1 * 0.9 + 2) * 0.9 + 2) * 0.9 + 2) * 0.9 + 2
            ((1 * 0.9 + 2) * 0.9 + 2) * 0.9 + 2,
            (1 * 0.9 + 2) * 0.9 + 2,
            1 * 0.9 + 2
]
```

alf/utils/value_op_test.py
```diff
@@ -48,12 +48,12 @@ def test_discounted_return(self):
             rewards, values, is_lasts, discounts,
             time_major=False), expected)
 
-        # two episodes and last value is zero (discount is zero)
-        discounts = tf.constant([[0.9, 0.9, 0.0, 0.9, 0.9]])
+        # tow episodes, and end normal
```
emailweixu: tow => two

alf/utils/value_ops.py
```diff
@@ -0,0 +1,79 @@
...
+import tensorflow as tf
+
+
+def discounted_return(
+        rewards,
+        values,
+        is_lasts,
+        discounts,
+        time_major=True):
+    if not time_major:
```
emailweixu: If the problem of tf_agents.utils.value_ops.discounted_return is the discount is not 0 when is_last for some environments, we can simply multiply discount with 1-is_last and call the original discounted_return().

Here the is_last is from current time_step. But  is_last should  be from next_step.
And also please comment what's wrong with the original discounted_return() and how discount is specified by tf_agents

alf/agents/a2c/a2c_agent_test.py
```diff
@@ -0,0 +1,175 @@
...
+        tf.compat.v1.enable_resource_variables()
+        self._obs_spec = tensor_spec.TensorSpec([2], tf.float32)
+        self._time_step_spec = ts.time_step_spec(self._obs_spec)
+        self._action_spec = tensor_spec.BoundedTensorSpec([1], tf.float32, -1, 1)
+
+    def test_create_agent(self):
+        a2c_agent.A2CAgent(
+            self._time_step_spec,
+            self._action_spec,
+            actor_net=DummyActorNet(
```
emailweixu: Why do you need to have DummyActorNet and DummyValueNet instead of using ActorDistributionNetwork and ValueNetwork for this test?

alf/agents/a2c/a2c_agent_test.py
```diff
@@ -0,0 +1,175 @@
...
+        actions = agent.policy.action(time_steps).action
+        self.assertEqual(actions.shape.as_list(), [1, 1])
+        self.evaluate(tf.compat.v1.global_variables_initializer())
+        action_values = self.evaluate(actions)
+        tf.nest.map_structure(
+            lambda v, s: self.assertAllInRange(v, s.minimum, s.maximum),
+            action_values, self._action_spec)
+
+    def test_train(self):
+        with tf.compat.v2.summary.record_if(False):
```
emailweixu: tf.summary.record_if()

alf/agents/a2c/a2c_agent.py
```diff
@@ -0,0 +1,188 @@
...
+
+        with tf.GradientTape() as tape:
+            loss_info = self._loss(time_steps, experience.action, weights)
+        grad_variables = self._value_net.variables + self._actor_net.variables
+        grads = tape.gradient(loss_info.loss, grad_variables)
+        self._apply_gradients(grads, grad_variables, self._optimizer)
+        self.train_step_counter.assign_add(1)
+        return tf.nest.map_structure(tf.identity, loss_info)
+
+    def _cal_values_and_returns(self, time_steps):
```
emailweixu: cal_values => calc_values

alf/agents/a2c/example/train_eval.py
```diff
@@ -0,0 +1,194 @@
...
+               eval_interval=100,
+               summary_interval=100,
+               summaries_flush_secs=1,
+               debug_summaries=True,
+               summarize_grads_and_vars=False):
+    root_dir = os.path.expanduser(root_dir)
+    train_dir = os.path.join(root_dir, 'train')
+    eval_dir = os.path.join(root_dir, 'eval')
+
+    train_summary_writer = tf.compat.v2.summary.create_file_writer(
```
emailweixu: Let's not use compat module whenever possible.

alf/agents/a2c/a2c_agent_test.py
witwolf: agent tests in TFAgents using DummyNetworks . i just follow it , i do not know why too .

alf/utils/value_ops.py
witwolf: For TimeLimit Environment , an episode terminate by time exceed or normal end (goal acheived or unable to continue)

1. tfagents set discounts to 1.0 for first case , eg: Pendulum, V(TimeStep; TimeStep.step==StepType.LAST) sould not be zero


2. tfagents set discounts to 0.0 for steps where next_time_step.SteypType==StepType.LAST

alf/agents/a2c/a2c_agent.py
```diff
@@ -0,0 +1,194 @@
...
+            actions_distribution, actions,
+            self.action_spec)
+        valid_mask = tf.cast(~time_steps.is_last(), tf.float32)
+        action_log_prob *= valid_mask
+        advantage = tf.stop_gradient(returns - values)
+        advantage *= valid_mask
+        policy_loss = -tf.reduce_sum(action_log_prob * advantage, axis=-1)
+        policy_loss = tf.reduce_mean(policy_loss)
+
+        td_error = self._td_errors_loss_fn(values, returns)
```
emailweixu: Since returns[-1] - values[-1] == 0 and advantage[-1]==0(because line 126), td_error[-1] , the last time step is not used for calculating loss, which is correct. But I think it's better to explicit them from the loss calculation (by removing actions[-1] from actions) so that it's easier to understand.

alf/agents/a2c/a2c_agent_test.py
```diff
@@ -0,0 +1,172 @@
...
+                time_steps.step_type, time_steps.reward, time_steps.discount)
+
+            agent.policy.variables()
+            if tf.executing_eagerly():
+                loss = lambda: agent.train(experience)
+            else:
+                loss = agent.train(experience)
+            self.assertEqual(self.evaluate(counter), 0)
+            self.evaluate(loss)
+            self.assertEqual(self.evaluate(counter), 1)
```
emailweixu: We need a test to check whether the policy and value are correctly trained.

alf/agents/a2c/a2c_agent.py
```diff
@@ -0,0 +1,194 @@
...
+            time_steps, policy_state).action
+        action_log_prob = common.log_probability(
+            actions_distribution, actions,
+            self.action_spec)
+        valid_mask = tf.cast(~time_steps.is_last(), tf.float32)
+        action_log_prob *= valid_mask
+        advantage = tf.stop_gradient(returns - values)
+        advantage *= valid_mask
+        policy_loss = -tf.reduce_sum(action_log_prob * advantage, axis=-1)
+        policy_loss = tf.reduce_mean(policy_loss)
```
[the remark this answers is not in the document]
witwolf: Policy gradient loss is defined as the sum, over timesteps, of action . reduce_sum is for timesteps and reduce_mean for batch .

PR #14, opened 2019-05-03
ICM algorithm

emailweixu:


alf/algorithms/icm_algorithm.py
```diff
@@ -0,0 +1,163 @@
...
+
+        if tensor_spec.is_discrete(self._action_spec):
+            inverse_loss = tf.nn.softmax_cross_entropy_with_logits(
+                labels=prev_action, logits=action_pred)
+        else:
+            inverse_loss = 0.5 * tf.reduce_mean(
+                tf.square(prev_action - action_pred), axis=-1)
+
+        intrinsic_reward = tf.stop_gradient(forward_loss)
+        intrinsic_reward = self._reward_normalizer.normalize(intrinsic_reward)
```
bayesian: with this normalization, the intrinsic reward could be negative?? likely it could mess up with lots of other stuff.

alf/algorithms/actor_critic_loss.py
```diff
@@ -96,10 +103,12 @@ def __call__(self, training_info: TrainingInfo, value, final_value):
             advantages = returns - value
         else:
             advantages = value_ops.generalized_advantage_estimation(
-                values=value,
-                final_value=final_value,
                 rewards=training_info.reward,
+                values=training_info.value,
```
bayesian: value is in training_info.info.value, not training_info.value

alf/algorithms/icm_algorithm.py
```diff
@@ -0,0 +1,163 @@
...
+
+        action_spec = flat_action_spec[0]
+
+        if tensor_spec.is_discrete(action_spec):
+            self._num_actions = action_spec.maximum - action_spec.minimum + 1
+        else:
+            self._num_actions = action_spec.shape[-1]
+
+        self._action_spec = action_spec
+        self._reward_scale = 1.0
```
bayesian: _reward_scale = reward_scale

alf/algorithms/icm_algorithm.py
emailweixu: Guaranteed positive reward could mess up with other stuff too. For example, for mountain car task, the original goal is to reach the goal as soon as possible. With a positive goal, the agent has less incentive to terminate the episode by reaching the goal.

PR #17, opened 2019-05-13
missing product gamma when calc discount_returns; reduce_mean twice f…

witwolf:
…or entropy loss, not equal with mask of LAST step

### DIFF
--- a/alf/algorithms/actor_critic_loss.py
+++ b/alf/algorithms/actor_critic_loss.py
@@ -22,12 +22,11 @@ from tf_agents.agents.tf_agent import LossInfo
 from tf_agents.utils import common as tfa_common
 
 from alf.algorithms.on_policy_algorithm import TrainingInfo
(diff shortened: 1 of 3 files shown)


alf/algorithms/actor_critic_loss.py
```diff
@@ -121,8 +120,8 @@ def __call__(self, training_info: TrainingInfo, value, final_time_step,
         if self._entropy_regularization is not None:
             entropies = tfa_common.entropy(training_info.action_distribution,
                                            self._action_spec)
-            entropy_loss = -tf.reduce_mean(input_tensor=entropies)
-            loss += self._entropy_regularization * entropy_loss
+            entropy_loss = -self._entropy_regularization * entropies
```
emailweixu: Keep entropy_loss to the original negative entropy: `entropy_loss = - entropies` so that we can see the original entropy in tensorboard. Apply the weight at the next line.

PR #18, opened 2019-05-16
MERLIN algorithm

emailweixu:
Now the merlin_algorithm_test.py can run successfully. It take 10-20 min to reach perfect solution on my machine.

### DIFF
--- a/alf/algorithms/decoding_algorithm.py
+++ b/alf/algorithms/decoding_algorithm.py
@@ -44,9 +44,6 @@ class DecodingAlgorithm(Algorithm):
         self._loss = loss
         self._loss_weight = loss_weight
 
-    def _variables(self):
-        return self._decoder.variables
-
     def train_step(self, inputs, state=None):
         """Train one step.
 
--- a/alf/algorithms/decoding_algorithm.py
+++ b/alf/algorithms/decoding_algorithm.py
@@ -65,4 +65,4 @@ class DecodingAlgorithm(Algorithm):
         return AlgorithmStep(
             outputs=pred,
             state=state,
-            info=LossInfo(self._loss_weight * loss, extra=loss))
+            info=LossInfo(self._loss_weight * loss, extra=()))
--- a/alf/algorithms/decoding_algorithm.py
+++ b/alf/algorithms/decoding_algorithm.py
@@ -1,0 +1,71 @@
+# Copyright (c) 2019 Horizon Robotics. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Decoding algorithm."""
+
+import tensorflow as tf
+
+from tf_agents.agents.tf_agent import LossInfo
+from tf_agents.networks.network import Network
+import tf_agents.specs.tensor_spec as tensor_spec
+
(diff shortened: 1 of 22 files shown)


alf/algorithms/merlin_algorithm.py
```diff
@@ -0,0 +1,441 @@
...
+            activation_fn=tf.keras.activations.tanh,
+            name=name + "/value_net")
+
+        self._rnn = rnn
+        self._actor_net = actor_net
+
+        # TODO: add qvalue_net for predicting Q-value
+
+    def _variables(self):
+        return _collect_variables(self._rnn, self._value_net, self._actor_net)
```
bayesian: _key_net should also be included.

alf/algorithms/merlin_algorithm.py
```diff
@@ -0,0 +1,441 @@
...
+        latent_vector, kld, next_state = self.encode_step(inputs, state)
+
+        # TODO: decoder for action
+        decoder_loss = self.decode_step(latent_vector, observation)
+
+        return AlgorithmStep(
+            outputs=latent_vector,
+            state=next_state,
+            info=LossInfo(
+                loss=decoder_loss.loss + kld,
```
bayesian: same here, should support weighted sum

alf/algorithms/merlin_algorithm.py
```diff
@@ -0,0 +1,441 @@
...
+    else:
+        return tf.keras.layers.StackedRNNCells(cells)
+
+
+def _collect_variables(*modules):
+    modules = tf.nest.flatten(modules)
+    return sum([mod.variables for mod in modules], [])
+
+
+class MemoryBasedPredictor(Algorithm):
```
bayesian: everything should be annotated with @gin.configurable?

alf/algorithms/merlin_algorithm.py
```diff
@@ -0,0 +1,441 @@
...
+            fc_layer_params=(200, ),
+            activation_fn=tf.keras.activations.tanh,
+            name=name + "/actor_net")
+
+        super(MemoryBasedActor, self).__init__(
+            action_spec=action_spec,
+            train_state_spec=get_rnn_cell_state_spec(rnn),
+            action_distribution_spec=actor_net.output_spec)
+
+        self._loss = ActorCriticLoss(action_spec) if loss is None else loss
```
bayesian: make ActorCriticLoss gin.configurable so it could let client enable more powerful features like td/gae

emailweixu: It's already gin.configurable

alf/algorithms/merlin_algorithm.py
emailweixu: good idea

alf/algorithms/decoding_algorithm.py
```diff
@@ -0,0 +1,71 @@
...
+
+from tf_agents.agents.tf_agent import LossInfo
+from tf_agents.networks.network import Network
+import tf_agents.specs.tensor_spec as tensor_spec
+
+from alf.algorithms.algorithm import Algorithm, AlgorithmStep
+from alf.utils.encoding_network import EncodingNetwork
+
+
+class DecodingAlgorithm(Algorithm):
```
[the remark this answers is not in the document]
emailweixu: We'll consider doing that later if necessary.

alf/algorithms/memory.py
```diff
@@ -165,7 +227,6 @@ def read(self, keys, scale=None):
                 assert scale.shape == keys.shape[:len(scale.shape)]
                 k = len(keys.shape) - len(scale.shape)
```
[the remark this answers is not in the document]
emailweixu: Here we allow the possibility of multiple keys sharing one scale

alf/algorithms/merlin_algorithm.py
emailweixu: This weight will be handled by loss_weight in Algorithm

alf/algorithms/merlin_algorithm.py
emailweixu: good catch. But since tf.keras.Layer now derives from tf.Module, we can get rid of _variables() now.

PR #21, opened 2019-05-22
OnPolicyDriver

emailweixu:
OnPolicyDriver runs the eviroment with the algorithm for on-policy training. It's written
in such way that allow using tf.function for run(). So it's much faster than TrainingPolicy.

### DIFF
--- a/alf/algorithms/actor_critic_algorithm.py
+++ b/alf/algorithms/actor_critic_algorithm.py
@@ -27,6 +27,7 @@ from alf.algorithms.on_policy_algorithm import ActionTimeStep
 from alf.algorithms.on_policy_algorithm import OnPolicyAlgorithm
 from alf.algorithms.on_policy_algorithm import TrainingInfo
 from alf.algorithms.algorithm import Algorithm
+from alf.utils.common import scalar_spec
 
 ActorCriticState = namedtuple("ActorCriticPolicyState",
                               ["actor_state", "value_state", "icm_state"])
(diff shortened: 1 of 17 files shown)


alf/drivers/on_policy_driver.py
```diff
@@ -0,0 +1,370 @@
...
+            training_info = training_info._replace(
+                action_distribution=action_distribution)
+
+        if self._final_step_mode == OnPolicyDriver.FINAL_STEP_SKIP:
+            next_time_step, policy_step, _ = self._step(
+                time_step, policy_state)
+            next_state = policy_step.state
+        else:
+            policy_step = self._algorithm.train_step(time_step, policy_state)
+            next_time_step = time_step
```
witwolf: why not perform it with new GradientTape and pass it to next _iter

alf/drivers/on_policy_driver.py
```diff
@@ -0,0 +1,370 @@
...
+                self._algorithm.action_distribution_spec,
+                training_info.action_distribution)
+
+            training_info = training_info._replace(
+                action_distribution=action_distribution)
+
+        if self._final_step_mode == OnPolicyDriver.FINAL_STEP_SKIP:
+            next_time_step, policy_step, _ = self._step(
+                time_step, policy_state)
+            next_state = policy_step.state
```
witwolf: why not perform final step with new GradientType and pass the new tape to next _iter and final_step_mode can be removed ?

witwolf: loop_vars only support numpy array, `Tensor`, and `TensorArray` objects

emailweixu: You said it. GradientType cannot be loop_vars

alf/algorithms/actor_critic_algorithm.py
```diff
@@ -152,6 +151,8 @@ def calc_loss(self, training_info, final_time_step, final_policy_step):
 
             training_info = training_info._replace(
                 reward=training_info.reward + training_info.info.icm_reward)
+            final_time_step = final_time_step._replace(
+                reward=final_policy_step.info.icm_reward)
```
witwolf: why is not 

     final_time_step = final_time_step._replace( 
              reward= final_time_step.reward+final_policy_step.info.icm_reward)

PR #25, opened 2019-05-31
Training example for SocialRobot SimpleNavigation environment

emailweixu:
It takes several hours to reach average episode return of 0.


alf/algorithms/actor_critic_loss.py
```diff
@@ -48,18 +57,21 @@ def __init__(self,
...
+            td_errors_loss_fn (Callable): A function for computing the TD errors
+                loss. This function takes as input the target and the estimated
+                Q values and returns the loss for each element of the batch.
+            use_gae (bool): If True, uses generalized advantage estimation for
+                computing per-timestep advantage. Else, just subtracts value
+                predictions from empirical return.
+            use_td_lambda_return (bool): Only effective if use_gae is True.
+                If True, uses td_lambda_return for training value function.
+                (td_lambda_return = gae_advantage + value_predictions)
+            td_lambda (float): Lambda parameter for TD-lambda computation.
```
Jialn: add description for argument normalize_advantages

alf/examples/actor_critic.py
```diff
@@ -153,49 +157,54 @@ def create_algorithm(env,
 @gin.configurable
 def create_environment(env_name='CartPole-v0',
```
Jialn: The default env name could use 'SocialBot-CartPole-v0' here.

alf/examples/actor_critic.py
```diff
@@ -153,49 +157,54 @@ def create_algorithm(env,
...
                        env_load_fn=suite_gym.load,
-                       random_seed=0,
                        num_parallel_environments=30):
     """Create environment."""
     if num_parallel_environments == 1:
         py_env = env_load_fn(env_name)
     else:
+        if env_load_fn == suite_socialbot.load:
+            logging.info("suite_socialbot environment")
```
witwolf: config `suite_mario.load.wrap_with_process=0` in gin config file instead

emailweixu: Since we always want wrap_with_process=0 in this situation, it's better to fix its behavior in code.

alf/examples/actor_critic.py
emailweixu: I think it's better to use CartPole-v0, as it's more commonly available.

alf/algorithms/actor_critic_loss.py
emailweixu: Added

hnyu:
When I ran the code, it caused error because of 'channel_first' image format. 

`File "/home/haonanyu/venv/lib/python3.6/site-packages/tensorflow/python/eager/execute.py", line 67, in quick_execute
    six.raise_from(core._status_to_exception(e.code, message), None)
  File "<string>", line 3, in raise_from
tensorflow.python.framework.errors_impl.UnimplementedError: Generic conv implementation only supports NHWC tensor format for now. [Op:Conv2D]`

PR #28, opened 2019-06-06
Rendering for suite_socialbot

emailweixu:
tf_agetns ProcessPyEnvironment does not implement render. So we need to render it to show play result.

### DIFF
--- a/alf/environments/suite_socialbot.py
+++ b/alf/environments/suite_socialbot.py
@@ -33,6 +33,29 @@ def is_available():
     return social_bot is not None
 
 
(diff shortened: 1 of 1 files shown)


alf/environments/suite_socialbot.py
```diff
@@ -33,6 +33,29 @@ def is_available():
     return social_bot is not None
 
 
+class ProcessPyEnvironment(parallel_py_environment.ProcessPyEnvironment):
+    """tf_agetns ProcessPyEnvironment with render()."""
```
hnyu: tf_agetns -> tf_agents

PR #31, opened 2019-06-10
icm super mario experiment with only intrinsic reward

witwolf:
it takes about 30 hours to get cumulative reward above 5000 with only  intrinsic reward as training signal
 (30 hours to get cumulative reward above 6000 with both extrinsic and intrinsic reward )


witwolf:
total reward 8042 

super_mario

PR #33, opened 2019-06-11
fix default arg for record

witwolf:


alf/environments/suite_mario.py
```diff
@@ -54,7 +54,9 @@ def load(game,
         frame_stack: Stack k last frames
         data_format：one of `channels_last` (default) or `channels_first`.
                     The ordering of the dimensions in the inputs.
-        record: Record the gameplay to file
+        record: Record the gameplay , see retro.retro_env.RetroEnv.record
+               `False` for not record otherwise record to current working directory or
+               specified director
```
emailweixu: comment how the record can be viewed.

PR #42, opened 2019-06-14
add breakout training example

hnyu:
1. 10 hours to reach 300 on breakout (actions are sampled). Wall clock time needs to be improved a lot
2. add advantage clipping and reward shaping options
3.  use gin files in root_dir if none is provided

### DIFF
--- a/alf/algorithms/actor_critic_algorithm.py
+++ b/alf/algorithms/actor_critic_algorithm.py
@@ -167,10 +167,14 @@ class ActorCriticAlgorithm(OnPolicyAlgorithm):
 
             training_info = training_info._replace(
                 reward=reward_calc_fn(training_info.reward,
-                                   training_info.info.icm_reward))
+                                      training_info.info.icm_reward))
             final_time_step = final_time_step._replace(
                 reward=reward_calc_fn(final_time_step.reward,
-                                   final_policy_step.info.icm_reward))
+                                      final_policy_step.info.icm_reward))
+
+            self.add_reward_summary("training_reward/overall",
+                                    training_info.reward)
+
 
         final_value = final_policy_step.info.value
         ac_loss = self._loss(training_info, training_info.info.value,
--- a/alf/algorithms/actor_critic_algorithm.py
+++ b/alf/algorithms/actor_critic_algorithm.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 from collections import namedtuple
+from typing import Callable
 
 import gin.tf
 
@@ -51,6 +52,7 @@ class ActorCriticAlgorithm(OnPolicyAlgorithm):
                  loss=None,
                  optimizer=None,
                  gradient_clipping=None,
+                 reward_shaping_fn: Callable =None,
                  train_step_counter=None,
                  debug_summaries=False,
                  name="ActorCriticAlgorithm"):
@@ -71,7 +73,10 @@ class ActorCriticAlgorithm(OnPolicyAlgorithm):
           extrinsic_reward_coef: Coefficient for extrinsic reward
           loss (None|ActorCriticLoss): an object for calculating loss. If None,
             a default ActorCriticLoss will be used.
-          optimizer (tf.optimizers.Optimizer): The optimizer for training.
+          optimizer (tf.optimizers.Optimizer): The optimizer for training
+          gradient_clipping (float): positive threshold for clipping gradient norms
+          reward_shaping_fn (Callable): a function that transforms extrinsic
+            immediate rewards
           debug_summaries: True if debug summaries should be created.
         """
 
@@ -89,6 +94,7 @@ class ActorCriticAlgorithm(OnPolicyAlgorithm):
             action_distribution_spec=actor_network.output_spec,
             optimizer=optimizer,
             gradient_clipping=gradient_clipping,
+            reward_shaping_fn=reward_shaping_fn,
             train_step_counter=train_step_counter,
             debug_summaries=debug_summaries,
             name=name)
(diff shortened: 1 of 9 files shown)


alf/algorithms/actor_critic_algorithm.py
```diff
@@ -71,7 +73,10 @@ def __init__(self,
           extrinsic_reward_coef: Coefficient for extrinsic reward
           loss (None|ActorCriticLoss): an object for calculating loss. If None,
             a default ActorCriticLoss will be used.
-          optimizer (tf.optimizers.Optimizer): The optimizer for training.
+          optimizer (tf.optimizers.Optimizer): The optimizer for training
+          gradient_clipping (float): positive threshold for clipping gradient norms
```
emailweixu: If not None,

alf/algorithms/on_policy_algorithm.py
```diff
@@ -265,10 +270,18 @@ def train_complete(
...
             loss_info (LossInfo): loss information
             grads_and_vars (list[tuple]): list of gradient and variable tuples
-            
+
         """
         valid_masks = tf.cast(
             tf.not_equal(training_info.step_type, StepType.LAST), tf.float32)
+
+        # reward shaping
+        if self._reward_shaping_fn is not None:
```
emailweixu: summarize the original reward

alf/examples/ac_breakout.gin
```diff
@@ -0,0 +1,53 @@
...
+# "v0 vs v4: v0 has repeat_action_probability of 0.25
+#  (meaning 25% of the time the previous action will be used instead of the new action),
+#   while v4 has 0 (always follow your issued action)
+# Because we already implements frame_skip in AtariPreprocessing, we should always
+# use 'NoFrameSkip' Atari environments from OpenAI gym
+create_environment.env_name='BreakoutNoFrameskip-v4'
+create_environment.env_load_fn=@suite_atari.load
+create_environment.num_parallel_environments=512
+
+# Do not use this frame_skip! Potential issue with _pool_and_resize()
```
emailweixu: I don't understand this comment. If it should not be used, why it is here.

alf/examples/actor_critic.py
```diff
@@ -222,12 +221,17 @@ def copy_gin_configs(root_dir, gin_files):
...
+
+    if FLAGS.gin_file is None:
+        gin_file = glob.glob(FLAGS.root_dir + "/*.gin")
+        assert gin_file, "No gin files are found! Please provide"
+
     if not FLAGS.play:
         copy_gin_configs(FLAGS.root_dir, FLAGS.gin_file)
-        # TODO: use the gin files under FLAGS.root_dir if FLAGS.gin_file
-        # is not specified
+        gin_file = FLAGS.gin_file
```
emailweixu: if FLAGS.gin_file is None, gin_file will become None

alf/trainers/on_policy_trainer.py
```diff
@@ -88,8 +88,8 @@ def train_():
             driver.run = tf.function(driver.run)
 
         env.reset()
-        time_step = driver.get_initial_time_step()
-        policy_state = driver.get_initial_state()
+        time_step = None
+        policy_state = None
```
emailweixu: I assign the initial value for time_step and policy_state for the reason of keeping the calling signature to driver.run() same. tf.function will recompile the graph based on the calling signature.

alf/utils/common.py
```diff
@@ -259,3 +259,26 @@ def _transform_image(obs):
...
+
+    Can be used as a reward shaping function passed to an algorithm
+    (e.g. ActorCriticAlgorithm).
+    """
+    assert minmax[0] <= minmax[1], "range error"
+    return tf.clip_by_value(r, minmax[0], minmax[1])
+
+
+@gin.configurable
+def reward_scaling(r, scale=1):
```
emailweixu: simply scaling the reward won't have any meaningful effect because the gradient is proportionally scaled.

alf/algorithms/actor_critic_algorithm.py
hnyu: added

alf/algorithms/on_policy_algorithm.py
hnyu: added

alf/examples/ac_breakout.gin
hnyu: updated

alf/examples/actor_critic.py
hnyu: fixed

alf/trainers/on_policy_trainer.py
hnyu: reverted

alf/utils/common.py
hnyu: Depending on the init values of the value functions, the advantages or td errors are not linearly correlated to the reward

alf/algorithms/actor_critic_algorithm.py
```diff
@@ -170,8 +176,6 @@ def calc_loss(self, training_info, final_time_step, final_policy_step):
         ac_loss = self._loss(training_info, training_info.info.value,
                              final_time_step, final_value)
 
-        self.add_reward_summary("reward", training_info.reward)
-
```
emailweixu: We should still keep this one, this is reward for training.
I am also thinking that we probably need to document the meaning of `reward` clearly, otherwise it might be used incorrectly, Or perhaps we should use different name to differentiate different reward: 'environment_reward' and 'reward`

hnyu: It's not gone. It was moved to the parent class on_policy_algorithm.py with the appropriate name

PR #52, opened 2019-06-20
Write gin-config to tensorboard

emailweixu:
So it's a little easier to compare experiments

### DIFF
--- a/alf/trainers/on_policy_trainer.py
+++ b/alf/trainers/on_policy_trainer.py
@@ -25,6 +25,7 @@ from tf_agents.utils import common as tfa_common
 from alf.drivers.on_policy_driver import OnPolicyDriver
 from alf.utils.metric_utils import eager_compute
 from tf_agents.metrics import tf_metrics
+from alf.utils import common
 from alf.utils.common import run_under_record_context, get_global_counter
 
 
@@ -80,7 +81,8 @@ def train(train_dir,
     if eval_env is not None:
         eval_metrics = [
             tf_metrics.AverageReturnMetric(buffer_size=num_eval_episodes),
-            tf_metrics.AverageEpisodeLengthMetric(buffer_size=num_eval_episodes)
+            tf_metrics.AverageEpisodeLengthMetric(
+                buffer_size=num_eval_episodes)
         ]
         eval_summary_writer = tf.summary.create_file_writer(
             eval_dir, flush_millis=summaries_flush_secs * 1000)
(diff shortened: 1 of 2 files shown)


alf/utils/common.py
```diff
@@ -308,3 +309,79 @@ def reward_scaling(r, scale=1):
...
+    """Copy gin config files to root_dir
+
+    Args:
+        root_dir (str): directory path
+        gin_files (None|list[str]): list of file paths
+    """
+    root_dir = os.path.expanduser(root_dir)
+    os.makedirs(root_dir, exist_ok=True)
+    for f in gin_files:
+        shutil.copyfile(f, os.path.join(root_dir, os.path.basename(f)))
```
witwolf: import shutil this module

alf/utils/common.py
```diff
@@ -308,3 +309,79 @@ def reward_scaling(r, scale=1):
...
+    
+    If FLAGS.gin_file is not set, find gin files under FLAGS.root_dir and
+    returns them.
+    Returns:
+        the gin file(s)
+    """
+    gin_file = flags.FLAGS.gin_file
+    if gin_file is None:
+        root_dir = os.path.expanduser(flags.FLAGS.root_dir)
+        gin_file = glob.glob(os.path.join(root_dir, "*.gin"))
```
witwolf: import glob and flags

alf/trainers/on_policy_trainer.py
```diff
@@ -124,15 +126,19 @@ def train_():
...
+                        metrics=eval_metrics,
+                        environment=eval_env,
+                        state_spec=algorithm.predict_state_spec,
+                        action_fn=algorithm.greedy_predict,
+                        num_episodes=num_eval_episodes,
+                        train_step=global_step,
+                        summary_writer=eval_summary_writer,
+                        summary_prefix="Metrics")
                     metric_utils.log_metrics(eval_metrics)
+            if iter == 0:
```
witwolf: it's better to move  line 139-141   before the loop L144 ?

alf/utils/common.py
emailweixu: It is imported in a different commit in my local repo. I'll change it.

alf/trainers/on_policy_trainer.py
emailweixu: The functions need be called in order to get the operative configurations from gin. So we need to do the summary after the major functions

PR #53, opened 2019-06-20
image_scale_transformer as a wrapper to fix the bug in off_policy_driver

hnyu:
Previously the transformation is only applied inside algorithm_step. So time steps inside the replay buffer are not transformed. This could be a potential problem.


alf/environments/wrappers.py
```diff
@@ -102,3 +107,48 @@ def step(self, action):
...
+
+    Note: it treats an observation with len(shape)==4 as image
+    Args:
+        observation (nested Tensor): observations
+        min (float): normalize minimum to this value
+        max (float): normalize maximum to this value
+    Returns:
+        Transfromed observation
+    """
+    def __init__(self, env, minmax=(-1., 1.)):
```
emailweixu: need to update observation_space

alf/environments/wrappers.py
```diff
@@ -102,3 +107,48 @@ def step(self, action):
...
+        observation, reward, done, info = self.env.step(action)
+        return self._transform(observation), reward, done, info
+
+    def _transform(self, observation):
+        def _transform_image(obs):
+            # tf_agent changes all gym.spaces.Box observation to tf.float32.
+            # See _spec_from_gym_space() in tf_agents/environments/gym_wrapper.py
+            min, max = self._minmax
+            if len(obs.shape) == 4:
+                if obs.dtype == tf.uint8:
```
emailweixu: the observation is likely to be a numpy array, can its dtype be compared against tf dtypes?

alf/examples/ac_atari.gin
```diff
@@ -1,15 +1,16 @@
 # You need to install following packages
-# pip3 install atari-py opencv-python 
+# pip3 install atari-py opencv-python
 
 # environment config
 import tf_agents.environments.suite_atari
+import tf.environments.wrappers
```
emailweixu: wrong name. Perhaps we should import alf.environments.wrapper in the python code, because it's used so often.

alf/environments/wrappers.py

alf/environments/wrappers.py
hnyu: it's not correct. fixed

alf/examples/ac_atari.gin
hnyu: moved the import to actor_critic.py

witwolf:
> Previously the transformation is only applied inside algorithm_step. So time steps inside the replay buffer are not transformed. This could be a potential problem.

it cost 4x memory to store transformed time_steps to replay buffer when dtype is transform  
 from `tf.int8` to `tf.float32`

hnyu:
> > Previously the transformation is only applied inside algorithm_step. So time steps inside the replay buffer are not transformed. This could be a potential problem.
> 
> it cost 4x memory to store transformed time_steps to replay buffer when dtype is transform
> from `tf.int8` to `tf.float32`

Oh right. Then we should do transformation again after taking out exps from the replay buffer if we store uint8. Maybe change self._algorithm.train_step to self.algorithm_step and put train_step inside it. 
https://github.com/HorizonRobotics/alf/blob/44fe5fad130ec5152bd4584193699aee79e81d41/alf/drivers/off_policy_driver.py#L188

PR #55, opened 2019-06-21
apply observation transformation to replayed exps

hnyu:
I unified algorithm_step for on-policy and off-policy with a little overloading experience with time_step

### DIFF
--- a/alf/drivers/off_policy_driver.py
+++ b/alf/drivers/off_policy_driver.py
@@ -60,7 +60,6 @@ class OffPolicyDriver(policy_driver.PolicyDriver):
                  algorithm: OffPolicyAlgorithm,
                  observers=[],
                  metrics=[],
-                 training=True,
                  greedy_predict=False,
                  debug_summaries=False,
                  summarize_grads_and_vars=False,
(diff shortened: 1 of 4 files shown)


alf/drivers/policy_driver.py
```diff
@@ -113,12 +113,20 @@ def get_initial_time_step(self):
...
-    def algorithm_step(self, time_step, state):
+    def algorithm_step(self, time_step, state, training=self._training):
         if self._observation_transformer is not None:
             time_step = time_step._replace(
                 observation=self._observation_transformer(time_step.
                                                           observation))
-        return self._algorithm_step(time_step, state)
+        return self._algorithm_step(time_step, state, training)
+
+    def _algorithm_step(self, time_step, state, training):
```
emailweixu: if _algorithm_step() is only used by algorithm_step(), we should merge them into one single function.

alf/drivers/policy_driver.py
```diff
@@ -113,12 +113,20 @@ def get_initial_time_step(self):
                                                      self.env.batch_size)
         return make_action_time_step(time_step, action)
 
-    def algorithm_step(self, time_step, state):
+    def algorithm_step(self, time_step, state, training=self._training):
```
emailweixu: Perhaps we should remove the default value of training to make it a little easier to understand the logic.

alf/drivers/off_policy_driver.py
```diff
@@ -47,14 +47,15 @@ def __init__(self,
                 updated after every step in the environment. Each observer is a
                 callable(time_step.Trajectory).
             metrics (list[TFStepMetric]): An optiotional list of metrics.
-            training (bool): True for training, false for evaluating
+            training (bool): by default False because training info is obtained from
```
emailweixu: Need to comment the meaning of `training`

alf/drivers/policy_driver.py
hnyu: merged

alf/drivers/policy_driver.py
hnyu: removed

alf/drivers/off_policy_driver.py
hnyu: removed training flag and set it to False

witwolf:
keyword argument 'training' should be removed here
https://github.com/HorizonRobotics/alf/blob/6ac650530ac9317f7e8d99a32afd1c53750dc97a/alf/drivers/off_policy_driver_test.py#L137-L138

hnyu:
> keyword argument 'training' should be removed here
> https://github.com/HorizonRobotics/alf/blob/6ac650530ac9317f7e8d99a32afd1c53750dc97a/alf/drivers/off_policy_driver_test.py#L137-L138

Remvoed.

emailweixu:
Need to resolve conflict

Issue #64, opened 2019-06-26
Figures for evaluation metric against the number of environmental steps


emailweixu:
For RL, we need to have figure showing the performance with respect to the number of training environment steps. Currently, we only have the figures of metrics vs. global steps

witwolf:
`Metrics_vs_EnvironmentSteps`  and `Metrics_vs_NumberOfEpisodes` already exist in summary ?

emailweixu:
We need to show the evaluation metrics vs. training environmental steps. Metrics_vs_EnvironmentSteps and Metrics_vs_NumberOfEpisodes are only for training right now

PR #66, opened 2019-06-27
add evaluate metrics vs environment_steps and environment_episodes

witwolf:
#64

### DIFF
--- a/alf/drivers/policy_driver.py
+++ b/alf/drivers/policy_driver.py
@@ -222,10 +222,10 @@ class PolicyDriver(driver.Driver):
                                                    self.env.batch_size)
 
     def get_step_metrics(self):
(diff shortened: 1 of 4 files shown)


alf/drivers/policy_driver.py
```diff
@@ -221,6 +221,14 @@ def get_initial_state(self):
         return common.zero_tensor_from_nested_spec(self._policy_state_spec,
                                                    self.env.batch_size)
 
+    def get_step_metrics(self):
```
emailweixu: It's unclear what metrics are returned.
Make the comment more specific, the two metrics are episodes and environmental steps.

witwolf: updated

PR #60, opened 2019-06-25
PPO algorithm which behaves in the same way as baselines.ppo2

emailweixu:
Also added an option for clip_by_global_norm. This PR is on top of PR #58

PR #69, opened 2019-06-28
Config and result of ppo for SocialBot PR2 grasping

emailweixu:
It took about 4 hours to show reliable grasping behavior. About 5 times faster than pr #20 

This config is on top of PR #60


emailweixu:
It's already merged with PR #60

PR #74, opened 2019-06-28
improve ppo_cart_pole.gin

hnyu:
with reward scaling and mini_batch_size=128, it will stably converge to 200 in 30 seconds

### DIFF
--- a/alf/examples/ppo_cart_pole.gin
+++ b/alf/examples/ppo_cart_pole.gin
@@ -4,6 +4,10 @@ include 'ppo.gin'
 create_environment.env_name="CartPole-v0"
 create_environment.num_parallel_environments=8
 
+# reward scaling
+ActorCriticAlgorithm.reward_shaping_fn = @reward_scaling
+common.reward_scaling.scale = 0.01
+
 # algorithm config
 PPOLoss.entropy_regularization=1e-4
 PPOLoss.gamma=0.98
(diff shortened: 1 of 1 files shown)


emailweixu:
It's worth to do more experiment to understand the reason why the scaling works for this problem.

hnyu:
> It's worth to do more experiment to understand the reason why the scaling works for this problem.

Suppose beta_1=0.9 and beta_2=0.999  (default values), and after the value function converges (gradients are very small for a long time, so are the moments m_t and v_t). When a big noisy gradient comes in, m_t is approximately (1-beta_1)*g = 0.1g and v_t is (1-beta_2)*g=0.001g. The ratio then becomes 0.1g/(0.03g + eps). If g >> eps, then this ratio is 3.33. 

For a small reward scale, g is roughly the same order of magnitude with eps (after convergence).  So the ratio will be roughly 0.1. I think this somehow intuitively explains why a smaller reward scale helps stabilize training. Basically, if a "surprised" gradient g has a similar magnitude with eps, then the gradient variance will be small.

emailweixu:
Did you try increasing eps by 100 times to see if the training is stable without reward scaling?

hnyu:
> Did you try increasing eps by 100 times to see if the training is stable without reward scaling?

I tried eps=1e-4 (default 1e-8). The training is more stable without reward scaling, but the convergence is slower (the gradients tend to be damped by eps).

emailweixu:
What is the result of eps=1e-6, which is 100 times larger than the default?

hnyu:
> What is the result of eps=1e-6, which is 100 times larger than the default?

I observed no significant stability improvement on eps=1e-8 in this setting.

Issue #73, opened 2019-06-28
Unoperative gin config


emailweixu:
Sometime the configuration provided in gin config file may be overridden by python code. One example is alf/examples/ppo_pr2.sh  in PR #20, where the commandline specifies --gin_param='train_eval.num_epochs=10', but unintentionally override by a FLAG '--num_epochs' with default value 10.

It would be nice if we can find a way to write all the unused config from gin_file or gin_params to tensorboard.

PR #75, opened 2019-06-29
write inoperative gin config to tensorboard

witwolf:
#73

### DIFF
--- a/alf/utils/common.py
+++ b/alf/utils/common.py
@@ -25,7 +25,7 @@ from tf_agents.agents.tf_agent import LossInfo
 from tf_agents.specs import tensor_spec
 from tf_agents.utils import common as tfa_common
 from tf_agents.trajectories import trajectory
-from alf.utils import summary_utils
+from alf.utils import summary_utils, gin_utils
 
 
 def zero_tensor_from_nested_spec(nested_spec, batch_size):
@@ -335,8 +335,8 @@ def reward_scaling(r, scale=1):
     return r * scale
 
 
-def _markdownify_operative_config_str(string):
-    """Convert an operative config string to markdown format.
+def _markdownify_gin_config_str(string):
+    """Convert an gin config string to markdown format.
     
     Args:
         string (str): the string from gin.operative_config_str()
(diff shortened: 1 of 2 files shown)


alf/utils/common.py
```diff
@@ -370,16 +370,21 @@ def process(line):
...
     """
-    config_str = gin.operative_config_str()
-    md_config_str = _markdownify_operative_config_str(config_str)
+    operative_config_str = gin.operative_config_str()
+    md_config_str = _markdownify_gin_config_str(operative_config_str)
     tf.summary.text('gin/operative_config', md_config_str)
+    inoperative_config_str = gin_utils.inoperative_config_str()
+    if inoperative_config_str:
+        md_config_str = _markdownify_gin_config_str(inoperative_config_str)
+        tf.summary.text('gin/inoperative_config', md_config_str)
```
hnyu: Maybe write a more descriptive text for `inoperative_config`? For example, "A config string capturing all parameter values configured but not used by the current program (override by explicit call)." So this text is well explained and a person won't get confused even without looking at the code.

witwolf: `in_utils.inoperative_config_str.__doc__` is not appropriate here and explicit str description is too long  

We use gin config with all experiments , and should be familiar with parameters configured at least,   so it should not be a problem ...

hnyu: My point is that people might not be familiar with the concept of "inoperative config" without looking into gin's code. At least we can say several words like "config overridden by program" to make it clear.

witwolf: added at the head  of text summary

alf/utils/common.py
```diff
@@ -379,11 +384,17 @@ def summarize_gin_config():
...
+        operative_config_str,
+        'All parameter values used by configurable functions that are actually called'
+    )
     tf.summary.text('gin/operative_config', md_config_str)
     inoperative_config_str = gin_utils.inoperative_config_str()
     if inoperative_config_str:
-        md_config_str = _markdownify_gin_config_str(inoperative_config_str)
+        md_config_str = _markdownify_gin_config_str(
+            inoperative_config_str,
+            "All parameter values configured but not used by program (override by explicit call)"
```
emailweixu: The configured functions are not called or called by explicit parameter values

PR #76, opened 2019-06-30
breakout 45 mins to 300

hnyu:
Managed to train breakout-v4 to reach the range 250~300 in 45 mins (5M steps) with simple A2C. It basically reproduces the state-of-the-art Atari training configuration. Several things to note:
1. The reported score is by action sampling. If greedy_predict is used, then it should be higher. 
2. Regarding the gin file changes: a) image is scaled to [0,1] instead of [-1,1];  b) elu is changed to relu. For some unknown reasons, these two changes are very important as it prevents crash with lr=1e-3.
3. I rewrote the AtariPreprocessing to include additional preprocessings used by Deepmind and many other implementations. See the doc string of DMAtariPreprocessing. **It is highly suggested that we use this wrapper for any future Atari runs.**


emailweixu:
👍

alf/environments/wrappers.py
```diff
@@ -102,3 +104,200 @@ def step(self, action):
...
+    """
+    def __init__(self,
+                 env,
+                 frame_skip=4,
+                 noop_max=30,
+                 terminal_on_life_loss=False,
+                 screen_size=84):
+        """Constructor for an Atari 2600 preprocessor.
+
+        Args:
```
witwolf: comment should follow the format of other comment in alf.

alf/environments/wrappers.py
```diff
@@ -102,3 +104,200 @@ def step(self, action):
...
+        """Applies the given action in the environment.
+
+        Remarks:
+
+        * If a terminal state (from life loss or episode end) is reached, this may
+            execute fewer than self.frame_skip steps in the environment.
+        * Furthermore, in this case the returned observation may not contain valid
+            image data and should be ignored.
+
+        Args:
```
witwolf: comment should follow the format of other comment in alf.

alf/environments/wrappers.py
```diff
@@ -102,3 +104,200 @@ def step(self, action):
...
+
+        Returns:
+        observation: numpy array, the initial observation emitted by the
+            environment.
+        """
+        if self.game_over:
+            self.reset_with_random_noops()
+            self.lives = self.env.ale.lives()
+            self.game_over = False
+        else:
```
witwolf: is `self.game_over`  always `True` when reset is called ?

hnyu: If terminate_on_life_loss is True, when a life is lost, step() returns `terminal`=True to indicate that reset() should be called. self.game_over is maintained separately to decide if the game has really ended.

alf/environments/wrappers.py
```diff
@@ -102,3 +104,200 @@ def step(self, action):
...
+            if game_over:
+                self.env.reset()
+
+    def fire_on_reset(self):
+        # The following code is from https://github.com/openai/gym/...
+        # ...blob/master/gym/wrappers/atari_preprocessing.py
+        action_meanings = self.env.unwrapped.get_action_meanings()
+        if action_meanings[1] == 'FIRE' and len(action_meanings) >= 3:
+            self.env.step(1)
+            self.env.step(2)
```
emailweixu: what is action 2?

alf/environments/wrappers.py
emailweixu: +1

alf/environments/wrappers.py
hnyu: Can be any other action.  For some reason, people tend to append another action after 'FIRE'. Maybe some games require this.

alf/environments/wrappers.py

alf/environments/wrappers.py

alf/environments/wrappers.py
```diff
@@ -102,3 +104,198 @@ def step(self, action):
...
+    def __init__(self,
+                 env,
+                 frame_skip=4,
+                 noop_max=30,
+                 terminal_on_life_loss=False,
+                 screen_size=84):
+        """Constructor for an Atari 2600 preprocessor.
+
+        Args:
+            env (Gym environment): the environment whose observations are preprocessed.
```
emailweixu: Use the actual type gym.Env

PR #77, opened 2019-07-01
fix bug with ddpg ,and add simple pendulum experiment

witwolf:


alf/examples/ddpg_pendulum.gin
```diff
@@ -0,0 +1,42 @@
...
+create_ddpg_algorithm.actor_fc_layers=(400,300)
+CriticNetwork.observation_fc_layer_params=(400,)
+create_ddpg_algorithm.critic_fc_layers=(300,)
+DdpgAlgorithm.target_update_period=5
+OneStepTDLoss.td_error_loss_fn=@losses.element_wise_huber_loss
+
+# training config
+off_policy_trainer.train.initial_collect_steps=1000
+off_policy_trainer.train.mini_batch_length=2
+off_policy_trainer.train.num_steps_per_iter=%NUM_PARALLEL_ENVIRONMENTS
```
emailweixu: At off_policy_trainer.py:96 num_steps_per_iter is changed to at least mini_batch_length*mini_batch_size

alf/trainers/off_policy_trainer.py
```diff
@@ -124,21 +132,37 @@ def train_():
...
+
+        time_step, policy_state = driver.run(
+            max_num_steps=collect_steps,
+            time_step=time_step,
+            policy_state=policy_state)
+
         for iter in range(num_iterations):
             t0 = time.time()
-            time_step, policy_state = driver.run(
-                max_num_steps=num_steps_per_iter,
```
emailweixu: max_num_steps=num_steps_per_iter if iter > 0 else collect_steps, so that we don't need two driver.run

alf/utils/common.py
```diff
@@ -100,12 +101,15 @@ def get_target_updater(models, target_models, tau=1.0, period=1):
...
 
     def update():
+        update_ops = []
         for model, target_model in zip(models, target_models):
-            tfa_common.soft_variables_update(model.variables,
-                                             target_model.variables, tau)
+            update_op = tfa_common.soft_variables_update(
+                model.variables, target_model.variables, tau)
+            update_ops.append(update_op)
+        return tf.group(*update_ops)
```
emailweixu: does tf.group have any effect in tf 2.0?

alf/examples/ddpg_pendulum.gin
```diff
@@ -0,0 +1,42 @@
+
```
emailweixu: also add a picture of training curve

alf/trainers/off_policy_trainer.py
```diff
@@ -124,21 +132,37 @@ def train_():
...
-                time_step=time_step,
-                policy_state=policy_state)
 
-            experience = replay_buffer.gather_all()
+            if clear_replay_buffer:
+                experience = replay_buffer.gather_all()
+                replay_buffer.clear()
+            else:
+                experience, _ = replay_buffer.get_next(
```
runjerry: Do we might want to make replay_buffer an optional input rather than always initializing by driver.add_replay_buffer()? I am considering adding a sampler to current TF replay buffer, maybe one easy way without changing TF replay buffer is to use as_dataset() of the TF.replay_buffer, then it would be helpful if we could have some flexibility when fetching experience from the replay_buffer.

alf/algorithms/sac_algorithm.py
```diff
@@ -183,6 +183,12 @@ def __init__(self,
             self._target_critic_network2.variables,
             tau=1.0)
 
+    @property
+    def trainable_variables(self):
```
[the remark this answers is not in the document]
witwolf: updated

alf/trainers/off_policy_trainer.py
witwolf: updated

alf/examples/ddpg_pendulum.gin
witwolf: red one represent for evaluate and blue for traning

alf/examples/ddpg_pendulum.gin
witwolf: updated 
 
`num_steps_per_iter` is independent with  `mini_batch_length`  when  training with  sampled transitions  ,  `num_steps_per_iter=num_parallel_envs*per_env_step`

alf/utils/common.py
witwolf: here required  return a tf operator when tf.functions is used

alf/trainers/off_policy_trainer.py
witwolf: perhaps we should do some modification to make it flexible in a separate change .

emailweixu:
For the figure, it seems the episode reward saturate around 200, which is much lower than what is reported here https://gym.openai.com/evaluations/eval_GH4gO123SvukqlMmqhIyeQ/ . Do you have any idea?

PR #79, opened 2019-07-02
finalize atari wrappers

hnyu:
Finalize the Atari wrappers. Previously there was an issue regarding evaluation on Atari games. suite_atari.py hardcodes `auto_reset=False` and thus when a game terminates the environment is not really reset. In this case, if 'Fire' action is taken, then the game will proceeds to the next episode. However, in the early training stage, 'FIRE' might haven't been learned and the evaluation will get stuck. 

Also change terminate_on_life_loss to False so that episodic reward is computed across multiple lives. However, this also makes the training more difficult because of a harder estimation of advantages (longer horizon).

Lastly put the atari configuration in a common header file 'atari.gin'. 

Numbers: 70mins reach 300 (~2000 episode length) with 6M time steps in total, and 130mins reach 400 (~2500 episode length) with 10M time steps in total.


alf/examples/ac_atari.gin
```diff
@@ -1,43 +1,58 @@
...
-suite_atari.load.gym_env_wrappers=(@AtariPreprocessing, @FrameStack4)
+import tf_agents.environments.suite_gym
+import alf.environments.wrappers
+# From OpenAI gym wiki:
+# "v0 vs v4: v0 has repeat_action_probability of 0.25
+#  (meaning 25% of the time the previous action will be used instead of the new action),
+#   while v4 has 0 (always follow your issued action)
+# Because we already implements frame_skip in AtariPreprocessing, we should always
+# use 'NoFrameSkip' Atari environments from OpenAI gym
+create_environment.env_name='BreakoutNoFrameskip-v4'
```
emailweixu: Since the current ac_pong.png is not the result of any of the configs, we should remove it or redo figures.

emailweixu: Perhaps we should still keep a direct correspondence between config and result figure, so there is no confusion how the result is obtained.

alf/environments/wrappers.py
```diff
@@ -170,42 +175,37 @@ def __init__(self,
...
+    def fire(self):
         # The following code is from https://github.com/openai/gym/...
         # ...blob/master/gym/wrappers/atari_preprocessing.py
         action_meanings = self.env.unwrapped.get_action_meanings()
         if action_meanings[1] == 'FIRE' and len(action_meanings) >= 3:
             self.env.step(1)
             self.env.step(2)
 
-    def reset(self):
+    def reset(self, hard_reset=True):
```
emailweixu: comment hard_reset. But instead of adding 'hard_reset', which totally changes the semantic of 'reset' for gym.Env, I'd suggest keeping reset same signature, but adding another function such as start_new_life() which is internally used in this class.

alf/environments/wrappers.py
```diff
@@ -114,32 +114,38 @@ def __getattr__(self, name):
...
-    1. If terminating on losing a life, stepping a NO-OP action instead of resetting
-    2. Random number of NOOPs after reset
-    3. FIRE after step 2
+    1. Random number of NOOPs after reset
+    2. FIRE after a reset or a lost life. This is for the purpose of evaluation with
+       greedy prediction without getting stucked in the early training stage.
+    3. A lost life doesn't result in a terminal state
+
+    NOTE:
+    Some implementations mark a terminal state when a life is lost to help learn value
```
emailweixu: What does "mark a terminal state" mean?

alf/examples/ac_atari.gin
hnyu: removed ac_pong.png; changed ac_atari.gin to be a common header

alf/environments/wrappers.py
hnyu: added a start_new_life() function

alf/environments/wrappers.py

alf/examples/ac_atari.gin
```diff
@@ -1,58 +1,9 @@
...
 import tf_agents.environments.suite_gym
 import alf.environments.wrappers
-# From OpenAI gym wiki:
-# "v0 vs v4: v0 has repeat_action_probability of 0.25
-#  (meaning 25% of the time the previous action will be used instead of the new action),
-#   while v4 has 0 (always follow your issued action)
-# Because we already implements frame_skip in AtariPreprocessing, we should always
-# use 'NoFrameSkip' Atari environments from OpenAI gym
-create_environment.env_name='BreakoutNoFrameskip-v4'
+
```
emailweixu: should be renamed as atari.gin? It has nothing to do with ac

hnyu: renamed

PR #81, opened 2019-07-02
Training curves for ppo_bullet_humanoid

emailweixu:
3 runs, upper 3 curves are evaluation performance, lower 3 curves are training performance.
It took about 10 hours on a machine with i9-7960X CPU and 2080ti GPU to play 50M steps and reach episode reward of 2000-2500 for training and 3000-3500 for evaluation (i.e. using greedy policy)


hnyu:
Can you also provide some important numbers regarding the training time?

emailweixu:
Updated the PR description

PR #86, opened 2019-07-03
fix unittests

emailweixu:


alf/algorithms/vae_test.py
```diff
@@ -20,6 +20,8 @@
 from tensorflow.keras.datasets import mnist
 import matplotlib.pyplot as plt
 
+interactive_mode = False
```
witwolf: constants  on a module level should be capital letters

PR #87, opened 2019-07-03
Add instruction of doing tests to CONTRIBUTING.md

emailweixu:


CONTRIBUTING.md
```diff
@@ -21,12 +21,24 @@ pre-commit install
...
 git push origin PR_change_name
 ```
+6. Change your code based on review comments. The new change should be added
+as NEW commit to your previous commits. Do not use `--amend` option for the 
+commit, because then you will have to use '-f' option to push your change to
+github and review will be more difficult because the new change cannot
+be separated from previous change. For the same reason, if you need to incorporate
+the latest code from master, please use `git pull` instead of `git pull --rebase`.
 
 ## Coding stardard
```
Jialn: coding standard ?

PR #89, opened 2019-07-03
Update ac_cart_pole.gin

emailweixu:
Following PR #74, scale the reward by 0.01. The training is more stable.


emailweixu:
Sorry, the figure is not correct. The x-axis is train steps. Should be environment steps

Issue #83, opened 2019-07-02
Speeding up the loading of SocialRobot environment


emailweixu:
When using 30 or 60 parallel environments, the time for loading the environments become quite long. It seems that the environments are loaded sequentially. We might be able to  speed up the loading by loading the environments parallelly

witwolf:
it's configurable
`ParallelPyEnvironment.start_serially=False`

emailweixu:
It's good to have this option. Unfortunately, it doesn't seem to make SimpleNavigation load faster.

witwolf:
it actually speed up the loading , but  sometimes `gazebo.initialize(port=port)` will failed when loading parallelly, i'll try to fix it

witwolf:
it actually speed up the process,  but it has problem with `social_bot` yet, it seems the spawned `gazebo` process shares some resources and can be Influenced at initialization stage

i tried the test below :

    if __name__ == '__main__':

        gin.parse_config([
            'create_environment.num_parallel_environments=30',
            'ParallelPyEnvironment.start_serially=False'
        ])
        env = create_environment(
            env_name="SocialBot-Pr2Gripper-v0",
            env_load_fn=suite_socialbot.load)

        print('all env created')

        while True:
            time.sleep(5)


and get this error, is show error occurs at `social_bot::Initialize`

    Traceback (most recent call last):
      File "test.py", line 30, in <module>
        env_load_fn=suite_socialbot.load)
      File "/usr/local/lib/python3.5/dist-packages/gin/config.py", line 1032, in wrapper
        utils.augment_exception_message_and_reraise(e, err_str)
      File "/usr/local/lib/python3.5/dist-packages/gin/utils.py", line 49, in augment_exception_message_and_reraise
        six.raise_from(proxy.with_traceback(exception.__traceback__), None)
      File "<string>", line 3, in raise_from
      File "/usr/local/lib/python3.5/dist-packages/gin/config.py", line 1009, in wrapper
        return fn(*new_args, **new_kwargs)
      File "/home/hongyingxiang/FLA/alf/environments/utils.py", line 44, in create_environment
        [lambda: env_load_fn(env_name)] * num_parallel_environments)
      File "/usr/local/lib/python3.5/dist-packages/gin/config.py", line 1032, in wrapper
        utils.augment_exception_message_and_reraise(e, err_str)
      File "/usr/local/lib/python3.5/dist-packages/gin/utils.py", line 49, in augment_exception_message_and_reraise
[...]

PR #92, opened 2019-07-05
fix with suite_socialbot and support parallel load

witwolf:
#83

### DIFF
--- a/alf/environments/suite_socialbot.py
+++ b/alf/environments/suite_socialbot.py
@@ -129,19 +129,22 @@ def _get_unused_port(start, end=65536):
         end (int): port range end
     """
     process_lock = None
+    unused_port = None
     try:
-        unused_port = None
         for port in range(start, end):
+            process_lock = InterProcessLock(
+                path='/tmp/socialbot/{}.lock'.format(port))
+            if not process_lock.acquire(blocking=False):
+                process_lock = None
+                continue
             try:
                 with contextlib.closing(socket.socket()) as sock:
                     sock.bind(('', port))
-                    process_lock = InterProcessLock(
-                        path='/tmp/socialbot/{}.lock'.format(port))
-                    if process_lock.acquire(blocking=False):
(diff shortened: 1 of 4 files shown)


witwolf:
lock handle is  incorrect previous, and now fixed

alf/environments/suite_socialbot.py
```diff
@@ -129,19 +129,22 @@ def _get_unused_port(start, end=65536):
...
     """
     process_lock = None
+    unused_port = None
     try:
-        unused_port = None
         for port in range(start, end):
+            process_lock = InterProcessLock(
+                path='/tmp/socialbot/{}.lock'.format(port))
+            if not process_lock.acquire(blocking=False):
+                process_lock = None
```
hnyu: Line 138 & 147 are probably redundant.

witwolf:
parallel load is not available  with SocialBot due to some unknown problems with gazebo load world file

alf/environments/suite_socialbot.py
witwolf: No redundancy here  .

hnyu:
> parallel load is not available with SocialBot due to some unknown problems with gazebo load world file

If this is the case, is it still necessary for us to change from `BatchedPyEnvironment` to `ParallelPyEnvironment`?

hnyu:
> parallel load is not available with SocialBot due to some unknown problems with gazebo load world file

So the changes in suite_socialbot.py are still necessary if parallel loading is not available? What will be needed to fix the gazebo world file loading issue?

witwolf:
> > parallel load is not available with SocialBot due to some unknown problems with gazebo load world file
> 
> So are the changes in suite_socialbot.py still necessary if parallel loading is not available? What will be needed to fix the gazebo world file loading issue?

changes in suite_socialbot.py is bug fix for lock handle  ,  i'am trying to solve  gazebo world loading issue (#83) in SocialRobot

PR #96, opened 2019-07-08
Update README.md

emailweixu:


README.md
```diff
@@ -1,2 +1,47 @@
...
+git submodule update --init --recursive
+cd tf_agents
+pip install -e .
+cd ..
+pip install -e .
+```
+
+## Examples
+
+### A2C
```
hnyu: Do you think it's necessary that we put game snapshots, reward curves, and statistics for each example?

emailweixu: Yes, we should do that. And we should have some video showing how the agents perform.

Issue #94, opened 2019-07-08
Evaluation is not correct when there is observation_transformer


emailweixu:
eager_compute does not use observation_transformer, which makes the evaluation incorrect. We should use (on/off)_policy_driver to run the algorithm.greedy_predict so that the pre-processing is consistent.

PR #97, opened 2019-07-09
fix evaluation bug

witwolf:
#94


alf/drivers/policy_driver.py
```diff
@@ -109,14 +109,14 @@ def get_initial_time_step(self):
                                                      self.env.batch_size)
         return make_action_time_step(time_step, action)
 
-    def algorithm_step(self, time_step, state, training):
+    def algorithm_step(self, time_step, state, training, greedy_predict):
```
emailweixu: should add some comments because this function is used in many places.

PR #98, opened 2019-07-10
fix rnn test case for on_policy_driver

hnyu:
Before this, on_policy_driver_test.py will report errors about value shapes.


alf/algorithms/actor_critic_algorithm.py
```diff
@@ -142,6 +142,10 @@ def train_step(self, time_step: ActionTimeStep, state=None):
             observation,
             step_type=time_step.step_type,
             network_state=state.value_state)
+        # ValueRnnNetwork will add a time dim to value
+        # See value_rnn_network.py L153
+        if isinstance(self._value_network, ValueRnnNetwork):
```
emailweixu: This is a bug of tf-agents. In fact, I file a PR (https://github.com/tensorflow/agents/pull/151) to tf-agent, but they didn't merge. And the ValueRnnNetworks of tf-agents is not very good because it cannot handle multiple observations. We should replace it with our own in the future.

PR #99, opened 2019-07-10
Fix bug with sac

witwolf:


alf/algorithms/sac_algorithm.py
```diff
@@ -214,7 +214,7 @@ def actor_loss_fn(dqda, action):
             if self._dqda_clipping:
                 dqda = tf.clip_by_value(dqda, -self._dqda_clipping,
                                         self._dqda_clipping)
-            loss = losses.element_wise_squared_loss(
+            loss = 0.5 * losses.element_wise_squared_loss(
```
emailweixu: To  be consistent, we should also have 0.5 for ddpg_algorithm

alf/examples/sac_pendulum.gin
```diff
@@ -0,0 +1,44 @@
+
+import alf.algorithms.sac_algorithm
+import alf.utils.common
+
+# environment config
+NUM_PARALLEL_ENVIRONMENTS=1
```
emailweixu: why only use 1 here?

alf/algorithms/sac_algorithm.py
witwolf: updated

alf/examples/sac_pendulum.gin
witwolf: Its an appropriate value here .  process communication  would be heavy load for this simple gym env when `ParallelPyEnvironment` is used ,it actually cost less training steps when parallel environments are used but it slows down the training it terms of total time

PR #101, opened 2019-07-11
grid search

witwolf:
simple grid search on single machine

TODO: gpu config for each training, run grid search on multi machines


alf/examples/grid_search.py
```diff
@@ -0,0 +1,101 @@
...
+                if isinstance(value, str):
+                    value = eval(value)
+                keys.append(key)
+                values.append(value)
+        return keys, values
+
+    def run(self):
+        processes = []
+        para_keys, para_values = GridSearch.parse_config(self._conf_file)
+        for values in itertools.product(*para_values):
```
emailweixu: should have an option to limit the number of running processes

alf/examples/grid_search.py
```diff
@@ -0,0 +1,101 @@
...
+            parameters = dict(zip(para_keys, values))
+            root_dir = "%s_%d" % (FLAGS.root_dir, len(processes))
```
emailweixu: "%s/%d" is better because tensorboard can get everything under root_dir.

alf/examples/grid_search.py
```diff
@@ -0,0 +1,101 @@
...
+
+    def _worker(self, root_dir, parameters):
+        logging.set_verbosity(logging.INFO)
+        gin_file = common.get_gin_file()
+        if FLAGS.gin_file:
+            common.copy_gin_configs(root_dir, gin_file)
+        gin.parse_config_files_and_bindings(gin_file, FLAGS.gin_param)
+        with gin.unlock_config():
+            gin.parse_config(['%s=%s' % (k, v) for k, v in parameters.items()])
+        train_eval(root_dir + "/train")
```
emailweixu: since train_eval does both train and eval now, we should only pass root_dir to train_eval and let it to add '/train' and '/eval' to root_dir

alf/examples/grid_search.py
```diff
@@ -0,0 +1,101 @@
+# Copyright (c) 2019 Horizon Robotics. All Rights Reserved.
```
emailweixu: we can move this and main.py to another directory since they does not look like examples. 'bin' might be a good choice.

alf/examples/grid_search.py
```diff
@@ -0,0 +1,101 @@
...
+class GridSearch(object):
+    """Grid Search"""
+
+    def __init__(self, conf_file):
+        """
+        Args:
+            conf_file (str): Path to the config file. Config file must be json format,
+                and `parameters` is required. Parameters consist of key and iterable values.
+                see `ddpg_grid_search.json` for a good example
+                for more details.
```
emailweixu: need to also explain "key: string".

emailweixu: remove "for more details"

alf/examples/grid_search.py
```diff
@@ -0,0 +1,101 @@
...
+  --search_file=ddpg_grid_search.json \
+  --gin_file=ddpg_pendulum.gin \
+  --gin_param='create_environment.num_parallel_environments=8' \
+  --alsologtostderr
+```
+"""
+
+
+class GridSearch(object):
+    """Grid Search"""
```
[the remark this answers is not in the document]
witwolf: added

alf/examples/grid_search.py
```diff
@@ -0,0 +1,101 @@
...
+import json
+
+from absl import app
+from absl import flags
+from absl import logging
+import gin
+from alf.examples.main import train_eval
+from alf.utils import common
+
+flags.DEFINE_string('search_file', None,
```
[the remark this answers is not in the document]
witwolf: updated

alf/examples/grid_search.py
witwolf: updated

alf/examples/grid_search.py
witwolf: updated

alf/examples/grid_search.py
witwolf: removed

witwolf: updated

alf/examples/grid_search.py
witwolf: yeah, i'll do this in a seperate change

alf/examples/grid_search.py
witwolf: add `max_worker_num` to limit the number of running processed

alf/examples/main.py
```diff
@@ -92,26 +91,26 @@ def train_eval(train_dir, algorithm_ctor, evaluate=True,
...
-        train_dir (str): directory where checkpoints stores
+        root_dir (str): directory where checkpoints stores
         algorithm_ctor (Callable): callable that create an algorithm
             parameter value is bind with `__main__.train_eval.algorithm_ctor`,
             just config `__main__.train_eval.algorithm_ctor` when using with gin configuration
     """
     env = create_environment(num_parallel_environments=1)
     algorithm = algorithm_ctor(env)
-    on_policy_trainer.play(train_dir, env, algorithm)
+    on_policy_trainer.play(root_dir, env, algorithm)
```
emailweixu: need to update on_policy_trainer.play to accept root_dir instead of train_dir

alf/examples/grid_search.py
```diff
@@ -0,0 +1,156 @@
...
+              ...
+        }
+        ...
+    }
+    `max_worker_num` is max number of parallel search worker processes
+    `parameters` is a dict(param_name=param_value,) of configured search space .
+    param_name is a gin configurable argument str and param_value must be an
+    iterable python object or a str that can be evaluated to an iterable object.
+
+    See `ddpg_grid_search.json` for a example.
```
emailweixu: a example => an example

alf/examples/grid_search.py
```diff
@@ -0,0 +1,156 @@
...
+    """Grid Search"""
+
+    def __init__(self, conf_file):
+        """Create GridSearch instance
+        Args:
+            conf_file (str): Path to the config file.
+        """
+        self._conf = GridSearchConfig(conf_file)
+
+    def run(self):
```
emailweixu: comment

alf/examples/grid_search.py
witwolf: fixed

alf/examples/main.py
witwolf: updated

alf/examples/grid_search.py
witwolf: updated

alf/trainers/on_policy_trainer.py
```diff
@@ -150,6 +151,9 @@ def train_():
                     tf.summary.text('commandline', ' '.join(sys.argv))
 
         checkpointer.save(global_step=global_step.numpy())
+        env.close()
```
hnyu: No such method for TFPyEnvironment created from create_environment

witwolf: Ok, I’ll fix it

alf/bin/grid_search.py
```diff
@@ -0,0 +1,159 @@
...
+        """Run trainings with all possible parameter combinations in configured space
+        """
+
+        processes = []
+        param_keys = self._conf.param_keys
+        param_values = self._conf.param_values
+        max_worker_num = self._conf.max_worker_num
+
+        # `worker_queue` is just a status queue to control the number for working process.
+        # And to simplify, process pool is not used here, just spawn process for new
```
hnyu: Sorry for this late review. Why not just use Pool here? The code will be much shorter and less error-prone in that way.

witwolf: Yeah , u are right , code will be much shorter  and easy to understand here . But some reset  work needs to be done before job start  when multi training taken on the same process , such as reset `global_step`  etc, and it may be a problem to make sure all global states are reset. 
so ,  using  processes to isolate jobs is better here .

hnyu: OK, I see.

hnyu: I found out that you can specify an argument `maxtasksperchild=1` to process.Pool to have the same behavior. 

"maxtasksperchild is the number of tasks a worker process can complete before it will exit and be replaced with a fresh worker process, to enable unused resources to be freed. The default maxtasksperchild is None, which means worker processes will live as long as the pool."

witwolf: so  , Implementation is equivalet

PR #102, opened 2019-07-11
add opencv-python in installation instructions README.md

le-horizon:


le-horizon:
adding to setup.py instead.

PR #104, opened 2019-07-11
add opencv atari etc to dep of alf

le-horizon:


hnyu:
Probably we need also specify the version requirement for each package. Some versions  are not compatible with our code base. See setup.py in tf_agents for how to specify versions.

le-horizon:
Good point Haonan.  Done for the additional packages added.  Skipped "gym" which has been there before.

PR #111, opened 2019-07-13
Support entropy regularization when entropy is not implemented

emailweixu:
I am trying to add some features of SAC to ActorCritic. In SAC, NormalProjectionNetwork.scale_distribution=True is used. This PR makes ActorCritic can also use scale_distrituion=True

Another PR I am working on is the adaptive entropy regularization (alpha in SAC)


emailweixu:
@witwolf for the reason in this PR, the way of handling entropy if sac_algorithm is not correct. Can you check if fixing it can help?

emailweixu:
@witwolf, sorry, ignore my above comments. sac_algorithm does things correctly.

hnyu:
The second half of the note could be simpler. Basically it's a reparameterization trick from x to eps. You don't have to explicitly write out the p.d.f of p_theta(x). You can directly move the \partial\theta out of the expectation and jump to the last step.

alf/utils/entropy_estimator.tex
```diff
@@ -0,0 +1,55 @@
...
+&=& - \int \left(\pian{\log p_\theta(x)}{\theta}p_\theta(x) + \log p_\theta(x) \pian{p_\theta(x)}{\theta}\right) dx \\
+&=& - \int \left(\pian{\log p_\theta(x)}{\theta}p_\theta(x) + \log p_\theta(x) \pian{\log p_\theta(x)}{\theta} p_\theta(x)\right) dx \\
+&=& - \int (1+\log p_\theta(x))\pian{\log p_\theta(x)}{\theta} p_\theta(x) dx \\
+&=& -E_{x\sim p_\theta}\left(\log p_\theta(x)\pian{\log p_\theta(x)}{\theta}\right) -E_{x\sim p_\theta}\left(\pian{\log p_\theta(x)}{\theta}\right) \\
+&=& -\frac{1}{2}E_{x\sim p_\theta}\left(\pian{}{\theta}(\log p_\theta(x))^2\right) \\
+\end{eqnarray*}
+This means that $-\frac{1}{2}\pian{}{\theta}(\log p_\theta(x))^2$ is an unbiased estimator of $\pian{H}{\theta}$. Actually, $-\frac{1}{2}\pian{}{\theta}(c+\log p_\theta(x))^2$ is an unbiased
+estimator for any constant $c$
+
+For some distributions, the sample of $p_\theta$ is generated by transforming $\epsilon \sim q$ by $f_\theta(\epsilon)$, where $q$ is a fixed distribution and $f_\theta$ is a smooth bijective mapping. $p_\theta(x)$ is implicitly defined by $q$ and $f_\theta$ as:
```
hnyu: f_\phi instead of f_\theta

alf/utils/entropy_estimator.tex
```diff
@@ -0,0 +1,55 @@
...
+	\maketitle
+
+For some algorithms, we need to calculate the entropy and its derivative. If there is no analytic formula for the entropy, we can resort to sampling. Given the definition of entropy:
+\begin{equation}
+H(p) = E_{x\sim p_\theta}(-\log p_\theta(x))
+\end{equation}
+we can see that $-\log p(x)$ is an unbiased estimator of $H$ if $x$ is sampled from $p$. It is tempting to use $-\pian{\log p_\theta(x)}{\theta}$ as an estimator of $\pian{H}{\theta}$. However, it is wrong, as shown in the following: 
+\begin{eqnarray*}
+&&	E_{x\sim p_\theta}\left(\pian{\log p_\theta(x)}{\theta}\right) = \int \pian{\log p_\theta(x)}{\theta} p_\theta(x) dx \\
+&=& \int \pian{p_\theta(x)}{\theta} dx = \pian{}{\theta} \int p_\theta(x) dx = \pian{1}{\theta} = 0
```
hnyu: On a side note, this is exactly why we can't directly do SGD to optimize the RL object. Instead we have to do policy gradient or apply a reparameterization trick with SGD (only feasible in this case if the reward function is differentiable).

hnyu: This is exactly why a baseline can be used for policy gradients.

PR #114, opened 2019-07-15
replace grid search  implement with process pool

witwolf:


alf/bin/grid_search.py
```diff
@@ -12,14 +12,16 @@
...
-import multiprocessing
 import itertools
 import json
 
 from absl import app
 from absl import flags
 from absl import logging
 import gin
+
+# serialization overwrite by dill, it's more convenient
```
emailweixu: I don't understand "serialization overwrite by dill"

witwolf: task info was serialized and deserialize by pickle in std lib `multiprocessing` , as pickle can not serialize a member function and lambda ect ., it's  not flexible .  `pathos.multiprocessing` provides a consistent interface with std lib `multiprocessing` and its serialization is overwrite by dill , it's more flexible

PR #117, opened 2019-07-15
Add ppo_icubwalk.gin and result

emailweixu:
This result is trained with the code change in PR #111. In terms of sample efficiency, it is slightly slower than the result reported in [here](https://github.com/HorizonRobotics/SocialRobot/blob/master/examples/train_icub_walk_sac_curve.png). But it reaches higher return eventually.


Jialn:
Sample efficiency is good enough, consider that SAC is an off-policy algorithm and PPO is kind of on-policy.

Jialn:
I found I forgot to update the curve after increasing the max steps per episode from 120 to 200. The figure [here](https://github.com/HorizonRobotics/SocialRobot/blob/master/examples/train_icub_walk_sac_curve.png) is the 120 max step version. We are using 200 steps now.

The figure below is 200 steps version. It seems SAC final return is better. (the ICubWalkDefault one, ignore other configurations) :
image

PR #118, opened 2019-07-16
entropy target algorithm

emailweixu:
This implements the method described in section 5 of "Soft Actor-Critic Algorithms and Applications".
It automatically adjust weight for entropy loss term so that entropy is not smaller than the given target.


emailweixu:
This PR is on top of PR #

Issue #91, opened 2019-07-04
summarize_grads_and_vars and debug_summaries are very slow


hnyu:
In the PPO humanoid example, these two summaries will take roughly 3 seconds per training update while the training update itself is only about 2 seconds.

emailweixu:
We can adjust the summary interval. We need to have something to measure how much time is spent on environment.step, predict, training, summary, and evaluation so that we can make right adjustment.

hnyu:
Is it possible to optimize the summary code?

witwolf:
`summary interval`  do not help , it only reduce the frequency of `write to file` ,  computation for data (particular histogram data ..)  , serialization for metadata etc has no reduction， and this part can be optimized at least

hnyu:
Close this issue for #124

emailweixu:
#120 also helps alleviate this issue

PR #120, opened 2019-07-16
optimize summary functions

witwolf:
this pr has limited contribution to the reduction of summary cost , 
and we should config  parameters below with large values

`_trainer.train.summary_interval`
`_trainer.train.summaries_flush_secs`
`_trainer.train.summary_max_queue`

Reference to #91


emailweixu:
How much summary time does this PR reduce?

alf/utils/summary_utils.py
```diff
@@ -118,3 +143,58 @@ def histogram_continuous(name,
...
+                name='summarize_grads/' + var_name + '_gradient',
+                data=grad_values,
+                step=step)
+            tf.summary.scalar(
+                name='summarize_grads/' + var_name + '_gradient_norm',
+                data=tf.linalg.global_norm([grad_values]),
+                step=step)
+
+
+# tf.summary.scalar = _summary_wrapper(tf.summary.scalar)
```
emailweixu: Should this line be removed?

alf/utils/summary_utils.py
```diff
@@ -118,3 +143,58 @@ def histogram_continuous(name,
...
+tf.summary.histogram = _summary_wrapper(tf.summary.histogram)
```
emailweixu: should we do the same for add_gradients_summaries and add_variables_summaries instead of copying them?

alf/utils/common.py
```diff
@@ -258,20 +261,21 @@ def run_under_record_context(func, summary_dir, summary_interval,
...
-        summary_dir, flush_millis=flush_millis)
+        summary_dir, flush_millis=flush_millis, max_queue=summary_max_queue)
     summary_writer.set_as_default()
     global_step = get_global_counter()
     with tf.summary.record_if(lambda: tf.equal((global_step + 1) %
                                                summary_interval, 0)):
         func()
 
 
-from tensorflow.python.ops.summary_ops_v2 import should_record_summaries
```
emailweixu: ppo_loss.py uses this function

alf/utils/summary_utils.py
witwolf: `add_gradients_summaries` in tfa using `tf.compat.v2.summary` 
so it will be `tf.compat.v2.summary.histogram=_summary_wrapper(tf.compat.v2.summary.histogram)`
we should avoid include `tf.compat` in alf

witwolf:
> How much summary time does this PR reduce?

i did not make sufficient test , i run several group ddpg test with `summary_interval=1000` and default `summary_max_queue` (same for original) ,   it cost about 5% less .


And an appropriate value for  `summary_max_queue` can  reduce flush freq in one training step 
eg: ppo record more than 100 training curves , it flush more than 10 times with the default value in one summary step

PR #121, opened 2019-07-16
entropy target algorithm

emailweixu:
This implements the method described in section 5 of "Soft Actor-Critic Algorithms and Applications".
It automatically adjust weight for entropy loss term so that entropy is not smaller than the given target.


alf/algorithms/entropy_target_algorithm.py
```diff
@@ -0,0 +1,122 @@
...
+
+import gin
+import numpy as np
+import tensorflow as tf
+
+from tf_agents.agents.tf_agent import LossInfo
+from tf_agents.utils import common as tfa_common
+from alf.algorithms.algorithm import Algorithm, AlgorithmStep
+
+from alf.utils import common, dist_utils
```
witwolf: `dist_uitls`  not added in this commit

alf/algorithms/entropy_target_algorithm.py
```diff
@@ -0,0 +1,122 @@
...
+            optimizer=optimizer,
+            debug_summaries=debug_summaries,
+            name="EntropyTargetAlgorithm")
+
+        self._log_alpha = log_alpha
+        self._action_spec = action_spec
+
+        def _calc_default_target_entropy(spec):
+            dims = np.product(spec.shape.as_list())
+            if spec.dtype.is_floating:
```
witwolf: it's more clear to use `tensor_spec.is_continuous`

alf/algorithms/entropy_target_algorithm.py
witwolf: sorry, ignore comment above  `dist_utils` already exists in mater

alf/algorithms/entropy_target_algorithm.py
```diff
@@ -0,0 +1,122 @@
...
+        return AlgorithmStep(
+            outputs=(),
+            state=(),
+            info=LossInfo(
+                loss,
+                extra=EntropyTargetLossInfo(
+                    alpha_loss=alpha_loss, entropy_loss=entropy_loss)))
+
+    def calc_loss(self, training_info):
+        if self._debug_summaries and common.should_record_summaries():
```
witwolf: Is  condition `common.should_record_summaries()`  needed here as it already run under summary context with `run_under_record_context` ?

alf/algorithms/entropy_target_algorithm.py
```diff
@@ -0,0 +1,122 @@
...
+            AlgorithmStep. `info` field is LossInfo, other fields are empty.
+        """
+        entropy, entropy_for_gradient = dist_utils.entropy_with_fallback(
+            distribution, self._action_spec)
+        alpha_loss = self._log_alpha * tf.stop_gradient(entropy -
+                                                        self._target_entropy)
+        alpha = tf.stop_gradient(tf.exp(self._log_alpha))
+        loss = alpha_loss
+        entropy_loss = -entropy
+        loss -= alpha * entropy_for_gradient
```
witwolf: it's better to add comment for this joint loss and  add `target_entropy_loss` for  `EntropyTargetLossInfo`

alf/algorithms/entropy_target_algorithm.py
emailweixu: updated

alf/algorithms/entropy_target_algorithm.py
emailweixu: It shouldn't affect the summary. But it may save a little computation for preparing the value to be summarized.

alf/algorithms/entropy_target_algorithm.py
emailweixu: added

PR #124, opened 2019-07-17
Avoiding multiple summaries in one train step.

emailweixu:
When there are multiple minibatches or num_updates>1, algorithm.train_complete()
is called several times in one global step. We should only do the summary at the last minibatch

### DIFF
--- a/alf/drivers/off_policy_driver.py
+++ b/alf/drivers/off_policy_driver.py
@@ -244,14 +244,12 @@ class OffPolicyDriver(policy_driver.PolicyDriver):
                     lambda x: tf.reshape(x, [mini_batch_size] + list(x.shape)[
                         1:]), batch)
                 batch = _make_time_major(batch)
-                training_info, loss_info, grads_and_vars = self._update(
-                    batch, weight=batch.step_type.shape[1] / mini_batch_size)
-                # somehow tf.function autograph does not work correctly for the
-                # following code:
-                # if u == num_updates - 1 and b + mini_batch_size >= batch_size:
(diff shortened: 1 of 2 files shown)


alf/drivers/off_policy_driver.py
```diff
@@ -244,14 +244,12 @@ def _make_time_major(nest):
...
-                # somehow tf.function autograph does not work correctly for the
-                # following code:
-                # if u == num_updates - 1 and b + mini_batch_size >= batch_size:
                 is_last_mini_batch = tf.logical_and(
                     tf.equal(u, num_updates - 1),
                     tf.greater_equal(b + mini_batch_size, batch_size))
+                common.enable_summary(is_last_mini_batch)
+                training_info, loss_info, grads_and_vars = self._update(
+                    batch, weight=batch.step_type.shape[1] / mini_batch_size)
                 if is_last_mini_batch:
```
hnyu: Do you still need this condition after enable_summary is called?

alf/drivers/off_policy_driver.py
```diff
@@ -244,14 +244,12 @@ def _make_time_major(nest):
...
                     self._training_summary(training_info, loss_info,
                                            grads_and_vars)
```
hnyu: Might be better to always enable summary after training, in case summary is needed.

alf/drivers/off_policy_driver.py
emailweixu: Having this can avoid the extra processing in self._training_summary() if there's any

PR #125, opened 2019-07-17
Extract importance ratio calculation

le-horizon:
simple refactor to help vtrace loss

### DIFF
--- a/alf/algorithms/ppo_loss.py
+++ b/alf/algorithms/ppo_loss.py
@@ -30,6 +30,8 @@ from alf.utils import value_ops
 
 @gin.configurable
 class PPOLoss(ActorCriticLoss):
+    """PPO loss."""
+
     def __init__(self,
                  action_spec,
                  gamma=0.99,
@@ -99,35 +101,34 @@ class PPOLoss(ActorCriticLoss):
         self._check_numerics = check_numerics
 
     def _pg_loss(self, training_info: TrainingInfo, advantages):
-        with tf.name_scope(self.__class__.__name__) as scope:
-            importance_ratio, importance_ratio_clipped = value_ops.action_importance_ratio(
-                action_distribution=training_info.action_distribution,
-                collect_action_distribution=training_info.
-                collect_action_distribution,
-                action=training_info.action,
-                action_spec=self._action_spec,
-                calling_class_name=self.__class__.__name__,
-                scope=scope,
-                importance_ratio_clipping=self._importance_ratio_clipping,
-                log_prob_clipping=self._log_prob_clipping,
-                check_numerics=self._check_numerics,
-                debug_summaries=self._debug_summaries)
-            # Pessimistically choose the maximum objective value for clipped and
-            # unclipped importance ratios.
-            pg_objective = -importance_ratio * advantages
-            pg_objective_clipped = -importance_ratio_clipped * advantages
-            policy_gradient_loss = tf.maximum(pg_objective,
-                                              pg_objective_clipped)
+        scope = tf.name_scope(self.__class__.__name__)
+        importance_ratio, importance_ratio_clipped = value_ops.action_importance_ratio(
+            action_distribution=training_info.action_distribution,
+            collect_action_distribution=training_info.
+            collect_action_distribution,
+            action=training_info.action,
+            action_spec=self._action_spec,
+            clipping_mode=1,
+            scope=scope,
+            importance_ratio_clipping=self._importance_ratio_clipping,
+            log_prob_clipping=self._log_prob_clipping,
+            check_numerics=self._check_numerics,
+            debug_summaries=self._debug_summaries)
+        # Pessimistically choose the maximum objective value for clipped and
+        # unclipped importance ratios.
+        pg_objective = -importance_ratio * advantages
+        pg_objective_clipped = -importance_ratio_clipped * advantages
+        policy_gradient_loss = tf.maximum(pg_objective, pg_objective_clipped)
 
-            if self._debug_summaries and common.should_record_summaries():
-                with tf.name_scope(scope):
-                    tf.summary.histogram('pg_objective', pg_objective)
-                    tf.summary.histogram('pg_objective_clipped',
-                                         pg_objective_clipped)
+        if self._debug_summaries and common.should_record_summaries():
+            with scope:
+                tf.summary.histogram('pg_objective', pg_objective)
+                tf.summary.histogram('pg_objective_clipped',
+                                     pg_objective_clipped)
 
-            if self._check_numerics:
-                policy_gradient_loss = tf.debugging.check_numerics(
-                    policy_gradient_loss, 'policy_gradient_loss')
+        if self._check_numerics:
+            policy_gradient_loss = tf.debugging.check_numerics(
+                policy_gradient_loss, 'policy_gradient_loss')
 
         return policy_gradient_loss
 
--- a/alf/algorithms/ppo_loss.py
+++ b/alf/algorithms/ppo_loss.py
@@ -99,26 +99,35 @@ class PPOLoss(ActorCriticLoss):
         self._check_numerics = check_numerics
 
     def _pg_loss(self, training_info: TrainingInfo, advantages):
-        name_scope = 'PPOLoss'
-        importance_ratio, importance_ratio_clipped = value_ops.action_importance_ratio(
-            training_info, self._action_spec, name_scope,
-            self._importance_ratio_clipping, self._log_prob_clipping,
-            self._check_numerics, self._debug_summaries)
-        # Pessimistically choose the maximum objective value for clipped and
-        # unclipped importance ratios.
-        pg_objective = -importance_ratio * advantages
-        pg_objective_clipped = -importance_ratio_clipped * advantages
-        policy_gradient_loss = tf.maximum(pg_objective, pg_objective_clipped)
+        with tf.name_scope(self.__class__.__name__) as scope:
+            importance_ratio, importance_ratio_clipped = value_ops.action_importance_ratio(
+                action_distribution=training_info.action_distribution,
+                collect_action_distribution=training_info.
+                collect_action_distribution,
+                action=training_info.action,
+                action_spec=self._action_spec,
+                calling_class_name=self.__class__.__name__,
+                scope=scope,
+                importance_ratio_clipping=self._importance_ratio_clipping,
+                log_prob_clipping=self._log_prob_clipping,
+                check_numerics=self._check_numerics,
+                debug_summaries=self._debug_summaries)
+            # Pessimistically choose the maximum objective value for clipped and
+            # unclipped importance ratios.
+            pg_objective = -importance_ratio * advantages
+            pg_objective_clipped = -importance_ratio_clipped * advantages
+            policy_gradient_loss = tf.maximum(pg_objective,
+                                              pg_objective_clipped)
 
-        if self._debug_summaries and common.should_record_summaries():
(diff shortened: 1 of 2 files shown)


alf/algorithms/ppo_loss.py
```diff
@@ -98,58 +99,19 @@ def __init__(self,
...
-                tf.summary.histogram('action_log_prob', action_log_prob)
-                tf.summary.histogram('action_log_prob_sample',
-                                     sample_action_log_probs)
-                tf.summary.histogram('importance_ratio', importance_ratio)
-                tf.summary.scalar(
-                    'importance_ratio_mean',
-                    tf.reduce_mean(input_tensor=importance_ratio))
-                tf.summary.histogram('importance_ratio_clipped',
-                                     importance_ratio_clipped)
+            with tf.name_scope(name_scope):
```
emailweixu: I think using same name_scope at two different places will generate two different name scopes with different suffix by tensorflow.

alf/algorithms/ppo_loss.py
```diff
@@ -98,58 +99,19 @@ def __init__(self,
...
-        importance_ratio = tf.exp(action_log_prob - sample_action_log_probs)
-        if self._check_numerics:
-            importance_ratio = tf.debugging.check_numerics(
-                importance_ratio, 'importance_ratio')
-        importance_ratio_clipped = tf.clip_by_value(
-            importance_ratio, 1 - self._importance_ratio_clipping,
-            1 + self._importance_ratio_clipping)
-
+        name_scope = 'PPOLoss'
+        importance_ratio, importance_ratio_clipped = value_ops.action_importance_ratio(
```
emailweixu: it's better to user key=value form to call the function, it's more robust to future changing of the function signature.

alf/utils/value_ops.py
```diff
@@ -16,6 +16,75 @@
 import tensorflow as tf
 
 from tf_agents.trajectories.time_step import StepType
+from tf_agents.utils import common as tfa_common
+
+from alf.algorithms.rl_algorithm import TrainingInfo
+from alf.utils import common
+
+
+def action_importance_ratio(training_info: TrainingInfo, action_spec,
```
emailweixu: for function here, I is better not to using structure like TrainingInfo, because it depends on how particular structure of TrainingInfo. Just pass current_action_distribution and collect_action_distribution directly.

emailweixu: Also please make sure the clipped importance ratio is same as the one used by vtrace

alf/algorithms/ppo_loss.py
le-horizon: Oops, this sounds scary.  What's the convention here?  Should we leave all debug summaries inside loss classes or is it also Ok to have them in value_ops but need to remember not to do them again in loss classes?

alf/algorithms/ppo_loss.py
le-horizon: done.

alf/utils/value_ops.py
le-horizon: They are indeed different.  Modified to use different clipping for the different losses.

alf/algorithms/ppo_loss.py
le-horizon: Researched a bit, as long as the with tf.name_scope statements are not nested, concatenation of scopes will not happen, see example here:
https://www.tensorflow.org/api_docs/python/tf/Graph#for_example_6

le-horizon: Ok, I tested and this seems to be the only way to make it reuse an existing scope.  It's a bit cumbersome, but hopefully not many callers need to invoke this function..

le-horizon:
Thanks for the comments Wei.  All addressed.  Could you take another look?

alf/utils/value_ops.py
```diff
@@ -18,19 +18,26 @@
...
+
         Args:
-            training_info: passed in directly from loss classes.
+            action_distribution: Probability of actions under target policy.
+            collect_action_distribution: Probability of action from behavior
+                policy, used to sample actions for the rollout.
+            action: action taken during rollout.
             action_spec (nested BoundedTensorSpec): representing the actions.
-            name_scope: the calling class, 'PPOLoss', 'VTraceLoss' etc..
+            calling_class_name: to decide which clipping method to use.
```
emailweixu: classing_class_name is not a good name here, because it's conceivable it's called by other classes. It's better to name as "clipping_mode" and explain the kind of clipping it does.

alf/algorithms/ppo_loss.py
```diff
@@ -99,26 +99,35 @@ def __init__(self,
...
-            with tf.name_scope(name_scope):
-                tf.summary.histogram('pg_objective', pg_objective)
-                tf.summary.histogram('pg_objective_clipped',
-                                     pg_objective_clipped)
-
-        if self._check_numerics:
-            policy_gradient_loss = tf.debugging.check_numerics(
-                policy_gradient_loss, 'policy_gradient_loss')
+        with tf.name_scope(self.__class__.__name__) as scope:
+            importance_ratio, importance_ratio_clipped = value_ops.action_importance_ratio(
```
emailweixu: you don't need to call this under with .... as scope, you are passing scope as an argument. I better way is:
scope = tf.name_scope(...) as scope

alf/algorithms/ppo_loss.py
```diff
@@ -99,26 +99,35 @@ def __init__(self,
...
+                debug_summaries=self._debug_summaries)
+            # Pessimistically choose the maximum objective value for clipped and
+            # unclipped importance ratios.
+            pg_objective = -importance_ratio * advantages
+            pg_objective_clipped = -importance_ratio_clipped * advantages
+            policy_gradient_loss = tf.maximum(pg_objective,
+                                              pg_objective_clipped)
+
+            if self._debug_summaries and common.should_record_summaries():
+                with tf.name_scope(scope):
```
emailweixu: with scope:

alf/utils/value_ops.py
```diff
@@ -61,12 +69,16 @@ def action_importance_ratio(training_info: TrainingInfo, action_spec,
...
+    if calling_class_name == 'PPOLoss':
+        importance_ratio_clipped = tf.clip_by_value(
+            importance_ratio, 1 - importance_ratio_clipping,
+            1 + importance_ratio_clipping)
+    else:  #if name_scope == VTraceLoss.__class__.__name__:
+        raise Exception('Unsupported calling class: ' + calling_class_name)
 
     if debug_summaries and common.should_record_summaries():
-        with tf.name_scope(name_scope):
+        with tf.name_scope(scope):
```
emailweixu: with scope:

le-horizon: Nice, I didn't know this works!  Addressed all comments, please take another look.

alf/utils/value_ops.py
```diff
@@ -22,22 +22,21 @@
...
 
         Args:
             action_distribution: Probability of actions under target policy.
             collect_action_distribution: Probability of action from behavior
                 policy, used to sample actions for the rollout.
             action: action taken during rollout.
             action_spec (nested BoundedTensorSpec): representing the actions.
-            calling_class_name: to decide which clipping method to use.
-            scope: name_scope set outside.
+            clipping_mode: 1: double_sided [-x, +x] clipping for PPOLoss.
```
emailweixu: Using a string instead of a number might have better readability. Also make the comment more specific:
mode for clipping the importance ratio. "double_sided" for clipping the importance ratio into range [1-x, 1+x], which is used by PPO algorithm.

alf/utils/value_ops.py
```diff
@@ -22,22 +22,21 @@
...
                             importance_ratio_clipping, log_prob_clipping,
                             check_numerics, debug_summaries):
     """ ratio for importance sampling, used in PPO loss and vtrace loss.
 
-        Caller has to set tf.name_scope() and nest the call to this function
-            inside the with statement.
+        Caller has to save tf.name_scope() and pass scope to this function.
 
         Args:
             action_distribution: Probability of actions under target policy.
```
emailweixu: include the type in the comment like the following
arg_name (arg_type): description

alf/utils/value_ops.py
```diff
@@ -16,6 +16,86 @@
...
+    importance_ratio = tf.exp(action_log_prob - sample_action_log_probs)
+    if check_numerics:
+        importance_ratio = tf.debugging.check_numerics(importance_ratio,
+                                                       'importance_ratio')
+
+    if clipping_mode == 1:
+        importance_ratio_clipped = tf.clip_by_value(
+            importance_ratio, 1 - importance_ratio_clipping,
+            1 + importance_ratio_clipping)
+    else:  #if name_scope == VTraceLoss.__class__.__name__:
```
emailweixu: not name_scope here

alf/utils/value_ops.py
```diff
@@ -30,13 +30,21 @@ def action_importance_ratio(action_distribution, collect_action_distribution,
...
+                the rollout.
+            action (nested tf.distribution): possibly batched action tuple
+                taken during rollout.
             action_spec (nested BoundedTensorSpec): representing the actions.
-            clipping_mode: 1: double_sided [-x, +x] clipping for PPOLoss.
-            scope: the scope object returned by tf.name_scope(), set outside.
+            clipping_mode (string): mode for clipping the importance ratio.
+                'double_sided': clips the range of importance ratio into
+                    [1-x, 1+x], which is used by PPOLoss.
+                'capping': clips the range of importance ratio into
```
emailweixu: I'd suggest to make importance_ratio_clipping have similar meaning into both cases, so that we can switching between two modes without changing importance_ratio_clipping. This means that for 'capping', we use min(1+importance_ratio_clipping, importance_ratio). And change `x' to importance_ratio_clipping to be clearer.

alf/utils/value_ops.py
```diff
@@ -30,13 +30,21 @@ def action_importance_ratio(action_distribution, collect_action_distribution,
...
+                actions under target policy.
+            collect_action_distribution (nested tf.distribution): distribution
+                over actions from behavior policy, used to sample actions for
+                the rollout.
+            action (nested tf.distribution): possibly batched action tuple
+                taken during rollout.
             action_spec (nested BoundedTensorSpec): representing the actions.
-            clipping_mode: 1: double_sided [-x, +x] clipping for PPOLoss.
-            scope: the scope object returned by tf.name_scope(), set outside.
+            clipping_mode (string): mode for clipping the importance ratio.
```
emailweixu: for python, string is str

alf/utils/value_ops.py
le-horizon: Good point, done.

le-horizon:
Thanks for the careful reviewing Wei.  Please take another look.  Thanks,
Le

alf/utils/value_ops.py
```diff
@@ -82,8 +85,8 @@ def action_importance_ratio(action_distribution, collect_action_distribution,
             importance_ratio, 1 - importance_ratio_clipping,
             1 + importance_ratio_clipping)
     elif clipping_mode == 'capping':
-        importance_ratio_clipped = tf.clip_by_value(importance_ratio, 0,
-                                                    importance_ratio_clipping)
+        importance_ratio_clipped = tf.clip_by_value(
```
emailweixu: tf.minimum(importance_ratio, 1 + importance_ratio_clipping)
This is more efficient.

le-horizon: done.

le-horizon:
Done.  PTAL again.

Thanks,
Le

Issue #126, opened 2019-07-18
Continuous integration


emailweixu:
As the code base become larger, we should have CI test to make sure new changes do not break the tests. https://travis-ci.org/ is a popular site to provide free CI support. This is one example of using travis: https://github.com/PaddlePaddle/PARL/blob/develop/.travis.yml

One CI is available, we can add it as a condition of merge a pull request.

Another thing is to use fixed random seed for tests to make them robuster, otherwise there will be lot of headache.

witwolf:
since job check_stype added in CI ,  code style tools should be unified , tools below are used in our ci job 

    pre-commit==1.17.0 
    cpplint==1.4.4 
    pydocstyle==4.0.0 
    pylint==2.3.1 
    yapf==0.28.0

emailweixu:
Resolved by PR #128

emailweixu:
Can you update CONTRIBUTING.md to include the version of  those code style tools.

PR #128, opened 2019-07-19
add ci

witwolf:
#126 

my Travis account don't have sufficient rights to enable this repo , so this should be enabled on Travis by administrator (i have test it in my forked branch, it cost less than 2 mins total ,  about 30 secs to run unittest , other time for preparing such as pull docker image etc. )
and two things to do:
1. build docker image automatic by `Docker Hub Builder` 
2. long time cost unittests should be optimized and adjust , i just skip these experiments this commit

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,18 +1,26 @@
+# CI job config file used by travis-ci see https://travis-ci.org/ for detail.
+# Jobs are build inside docker container that all dependencies are pre-installed.
+# here we use `horizonrobotics/alf:latest` which built by docker file `.ci-cd/Dockerfile.cpu`
+# and has been pushed to the public docker hub `https://hub.docker.com/`
+# see https://www.docker.com/ for detail about docker and how to build docker image.
+
 language: python
 sudo: required
 dist: trusty
+git:
+  quiet: true
 services:
   - docker
 os:
   - linux
 env:
+  - JOB=check_style
   - JOB=test
 script:
   - docker run
     -v $PWD:/ALF
-    -w /ALF/alf
+    -w /ALF/
     -e PYTHONPATH=/ALF/tf_agents:/ALF
-    -e SKIP_LONG_TESTS=True
     -it horizonrobotics/alf:latest
     /ALF/.ci-cd/build.sh ${JOB}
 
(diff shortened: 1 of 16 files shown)


alf/environments/suite_mario.py
```diff
@@ -30,7 +30,13 @@
 
 
 def is_available():
-    return retro is not None
+    if retro is None:
+        return False
+    try:
+        retro.data.get_romfile_path('SuperMarioBros-Nes')
+    except:
```
emailweixu: In general, we should not use "except:" because it catches all kinds of error include syntax errors. Should use an explicit Exception type.

.travis.yml
```diff
@@ -0,0 +1,22 @@
+language: python
+sudo: required
+dist: trusty
+services:
+  - docker
+os:
+  - linux
+env:
+  - JOB=test
```
emailweixu: let's also add job check_style

alf/utils/value_ops_test.py
```diff
@@ -134,4 +135,7 @@ def test_generalized_advantage_estimation(self):
 
 
 if __name__ == '__main__':
-    tf.test.main()
+    from alf.utils.common import set_per_process_memory_growth
+
+    set_per_process_memory_growth()
+    unittest.main()
```
emailweixu: DiscountedReturnTest is subclass of tf.test.TestCase. Does it work with python unittest?

emailweixu:
@witwolf I added you as administrator of ALF. Can you make the necessary changes?

hnyu:
> 2. long time cost unittests should be optimized and adjust , i just skip these experiments this commit
As long as the total testing time is less than half an hour (or some average reviewing time for a PR), I personally think it's acceptable. Certainly we don't want to have tests running for several hours.

.travis.yml
witwolf: added

alf/utils/value_ops_test.py
witwolf: yes, it works

witwolf:
> > 1. long time cost unittests should be optimized and adjust , i just skip these experiments this commit
> 
> As long as the total testing time is less than half an hour (or our average reviewing time for a PR), I personally think it's acceptable. No need to waste time on optimizing unittest code. Certainly we don't want to have tests running for several hours.

Yeah, i retained these in the test job

Issue #123, opened 2019-07-17
merge code for off_policy_trainer and on_policy_trainer


witwolf:
code implementation  for  `off_policy_trainer` and `on_policy_trainer` have much similarity,
we should merge them

emailweixu:
I think they are different enough to warrant separate implementations. Although it's possible to extract some common parts for re-use or put into main.

emailweixu:
We also need to keep in mind that the training process may vary across different algorithms. We should make it not too difficult to customize training.

PR #130, opened 2019-07-22
merge trainer

witwolf:
#123


alf/trainers/off_policy_trainer.py
```diff
@@ -12,230 +12,116 @@
...
+            replay_buffer.clear()
+            return experience
+
+        def _sample_exp():
+            replay_buffer = self._driver.exp_replayer
+            return replay_buffer.replay(
+                sample_batch_size=self._mini_batch_size,
+                mini_batch_length=self._mini_batch_length)[0]
+
+        self._get_exp = _get_all_exp if clear_replay_buffer else _sample_exp
```
emailweixu: We can let the subclass implement self._get_exp because base class does not use  _get_exp() and different subclass need different _get_exp. And perhaps _get_exp() should just be part of train_iter() to make things simple.

alf/trainers/policy_trainer.py
```diff
@@ -0,0 +1,303 @@
...
+            summary_interval=self._summary_interval,
+            flush_millis=self._summaries_flush_mills,
+            summary_max_queue=self._summary_max_queue)
+        self._save_checkpoint()
+        if self._evaluate:
+            self._eval_env.pyenv.close()
+        self._env.pyenv.close()
+
+    @abc.abstractmethod
+    def train_iter(self, iter_num, policy_state, time_step):
```
emailweixu: Need more comments about what train_iter does because this is major interface for Trainer.

alf/trainers/policy_trainer.py
```diff
@@ -0,0 +1,303 @@
...
+            not self._use_tf_functions)
+        self._env = create_environment()
+        if self._evaluate:
+            self._eval_env = create_environment(num_parallel_environments=1)
+        self._algorithm = self._algorithm_ctor(
+            self._env, debug_summaries=self._debug_summaries)
+        self._driver = self.init_driver()
+
+    @abc.abstractmethod
+    def init_driver(self):
```
emailweixu: Need more comments about what init_driver does because this is major interface for Trainer.
Args and Returns

alf/trainers/policy_trainer.py
```diff
@@ -0,0 +1,303 @@
...
+        """initialize driver"""
+        pass
+
+    def train(self):
```
emailweixu: check initialized

emailweixu:
Because we use gin for configuration, when our code base are mature enough in the future, we should try to make new changes compatible with existing gin configuration files. Otherwise it will be a huge burden to update the gin files.  This is something we should keep in mind in the future.

alf/trainers/off_policy_trainer.py
witwolf: it's common part for sub classes

alf/trainers/policy_trainer.py
witwolf: added

emailweixu:
I am still troubled by the gin config file compatibility issue. What do you think if we add a single class TrainerConfig for the configuration and pass it as argument when constructing Trainer? This way user only needs to write TrainerConfig.name=value in the configuration. It's more robust to future refactorization.

witwolf:
> I am still troubled by the gin config file compatibility issue. What do you think if we add a single class TrainerConfig for the configuration and pass it as argument when constructing Trainer? This way user only needs to write TrainerConfig.name=value in the configuration. It's more robust to future refactorization.

so your suggestion is put configurable args from `Trainer*` to `TrainerConfig`  ?
configurable args set are different for different trainers , it would be confused if put all at one interface ,  can we use `**kwargs`   for different parts and explicit named arg for common parts here ?

emailweixu:
As long as we clearly document the usage of each argument, it should be fine to put them all under TrainerConfig.

alf/trainers/on_policy_trainer.py
```diff
@@ -12,259 +12,41 @@
...
-        else:
-            episode_reward += float(time_step.reward)
-            episode_length += 1
-    if recorder:
-        recorder.close()
-    env.reset()
+class OnPolicyTrainer(Trainer):
+    def __init__(self,
+                 root_dir,
+                 train_interval=20,
```
emailweixu: we can change to 'unroll_length' to be consistent with off_policy_trainer.

alf/trainers/off_policy_trainer.py
```diff
@@ -12,237 +12,116 @@
...
 from alf.drivers.sync_off_policy_driver import SyncOffPolicyDriver
-from alf.utils.metric_utils import eager_compute
-from tf_agents.metrics import tf_metrics
-from alf.utils import common
-from alf.utils.common import run_under_record_context, get_global_counter
+from alf.environments.utils import create_environment
+from alf.trainers.policy_trainer import Trainer
 
 
 @gin.configurable
```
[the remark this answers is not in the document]
witwolf: right reference value should be gin.configruable too ,

alf/trainers/on_policy_trainer.py
witwolf: done

alf/trainers/policy_trainer.py
```diff
@@ -0,0 +1,344 @@
...
+from alf.environments.utils import create_environment
+
+
+@gin.configurable
+class TrainerConfig(object):
+    def __init__(self, trainer_cls, **trainer_kwargs):
+        """
+
+        Args:
+            trainer_cls (class): cls that used for creating a `Trainer` instance,
```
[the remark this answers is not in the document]
witwolf: created an alias name by gin.configurable

alf/trainers/off_policy_trainer.py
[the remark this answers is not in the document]
witwolf: removed , this abstract class is not necessary gin configurable

alf/examples/ppo_icubwalk.gin
```diff
@@ -29,6 +29,7 @@ create_ac_algorithm.value_fc_layers=(256, 128)
 create_ac_algorithm.learning_rate=1e-4
 
 # training config
+TrainerConfig.unroll_length=20
 TrainerConfig.mini_batch_length=1
 TrainerConfig.num_steps_per_iter=16384
```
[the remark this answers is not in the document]
witwolf: fixed .   the original gin file did not config `unroll_length` and use default value  defined in `off_policy_trainer.trian`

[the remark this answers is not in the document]
emailweixu: num_steps_per_iter should removed because it's not used for sync_off_policy_trainer. num_steps_per_iter was used by an old version of OffPolicyTrainer, that's the reason it is here.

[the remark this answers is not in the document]
witwolf: removed

PR #131, opened 2019-07-22
fix with debug_summaries

witwolf:
`common.should_record_summaries` return a `tf.Tensor`  and can not  use it  as a python `bool`


emailweixu:
I sometime also encountered this error, but it always comes with some other error. Once the other errors are fixed, this error is gone.

alf/algorithms/actor_critic_loss.py
```diff
@@ -111,13 +111,18 @@ def __call__(self, training_info: TrainingInfo, value):
...
                 tf.summary.scalar("values", tf.reduce_mean(value))
                 tf.summary.scalar("returns", tf.reduce_mean(returns))
                 tf.summary.scalar("advantages", tf.reduce_mean(advantages))
                 tf.summary.scalar("explained_variance_of_return_by_value",
                                   common.explained_variance(value, returns))
+                return tf.constant(True)
+
+        if self._debug_summaries:
+            tf.cond(common.should_record_summaries(),
+                    _summary, lambda: tf.constant(False))
```
hnyu: Is there any special meaning of returning tf.constant(True) or tf.constant(False)? Can we just return tf.zeros(()) for both cases?

witwolf: return value here has no meaning,  any shape of tensor is ok .  i just follow the source code of  [tf.summary.write](https://github.com/tensorflow/tensorflow/blob/8e423e3d56390671f0d954c90f4fd163ab02a9c1/tensorflow/python/ops/summary_ops_v2.py#L684-L701)

alf/algorithms/entropy_target_algorithm.py
```diff
@@ -122,7 +122,7 @@ def train_step(self, distribution):
                     alpha_loss=alpha_loss, entropy_loss=entropy_loss)))
 
     def calc_loss(self, training_info):
-        if self._debug_summaries and common.should_record_summaries():
```
hnyu: Why was this condition removed?

alf/algorithms/actor_critic_loss.py
```diff
@@ -111,13 +111,18 @@ def __call__(self, training_info: TrainingInfo, value):
         returns, advantages = self._calc_returns_and_advantages(
             training_info, value)
 
-        if self._debug_summaries and common.should_record_summaries():
+        def _summary():
```
emailweixu: can we use something like the following, which is more readable:
```python
if tf.logical_and(self._debug_summaries, common.should_record_summaries()):
```

hnyu: This won't work as I tried.

alf/algorithms/entropy_target_algorithm.py
witwolf: there's only one simple summary `tf.summary.scalar` under this condition , and  it checks `should_record_summaries` in its implementation

alf/algorithms/actor_critic_loss.py
witwolf: > This won't work as I tried.
I've tried it before and it did not work, but it works now , I can't remember what the problem is.

hnyu: If I run ac_cart_pole.gin, it will crash.

witwolf: > If I run ac_cart_pole.gin, it will crash.

yes , it will not work 
i saw it as `tf.cond(tf.logical_and(self._debug_summaries, common.should_record_summaries()) ...`  before ...

PR #134, opened 2019-07-24
run_if for conditional execution

emailweixu:


witwolf:
we should also change code in `./utils/summary_utils.py` too

emailweixu:
> we should also change code in `./utils/summary_utils.py` too

What should be changed?

witwolf:
> > we should also change code in `./utils/summary_utils.py` too
> 
> What should be changed?

`smart_cond.smart_cond` in  `_summary_wrapper` can be replaced with `run_if`

PR #137, opened 2019-07-31
auto mini_batch_length and mini_batch_size in off_policy_trainer

hnyu:
Currently the trainer's mini_batch_size and mini_batch_length will always override off_policy_driver's train().

### DIFF
--- a/alf/drivers/off_policy_driver.py
+++ b/alf/drivers/off_policy_driver.py
@@ -176,11 +176,8 @@ class OffPolicyDriver(policy_driver.PolicyDriver):
             collect_action_distribution=self._action_dist_param_spec)
 
     @tf.function
-    def train(self,
-              experience: Experience,
-              num_updates=1,
-              mini_batch_size=None,
-              mini_batch_length=None):
+    def train(self, experience: Experience, num_updates, mini_batch_size,
+              mini_batch_length):
         """Train using `experience`.
 
         Args:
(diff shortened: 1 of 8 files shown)


alf/experience_replayers/experience_replay.py
```diff
@@ -85,14 +92,20 @@ class OnetimeExperienceReplayer(ExperienceReplayer):
     only once.
 
     Example algorithms: IMPALA, PPO2
+
+    NOTE: this replayer can only be run in the eager mode, because
+    self._experience is updated by python assignment
     """
 
-    def __init__(self, experience_spec, batch_size):
+    def __init__(self, experience_spec, *args, **kwargs):
```
witwolf: `*args, **kwargs` are not used , can we avoid this ?

hnyu: The reason is that the replayer will be created by the base class OffPolicyDriver. For tf_agents buffer, there is an additional arg `batch_size`. This is not used by this class.

alf/trainers/policy_trainer.py
```diff
@@ -258,11 +260,13 @@ def _train(self):
         policy_state = self._driver.get_initial_policy_state()
         for iter_num in range(self._num_iterations):
             t0 = time.time()
-            time_step, policy_state = self.train_iter(
+            time_step, policy_state, steps = self.train_iter(
                 iter_num=iter_num,
                 policy_state=policy_state,
                 time_step=time_step)
-            logging.info('%s time=%.3f' % (iter_num, time.time() - t0))
+            logging.info(
```
witwolf: `Metrics/EnvironmentSteps`  shows the throughput ?

hnyu: yeah, but to get the actual steps per second, we always have to do the math and the elapsed time can be read imprecisely.

alf/experience_replayers/experience_replay.py
witwolf: perhaps  we can  init replay_buffer in sub classes instead of specifing a cls and init it in base class

hnyu: The problem is that each sub-class can potentially use any of the replayers. I've changed the class to a string.

alf/trainers/policy_trainer.py
hnyu: Also what is shown here is the training steps per second, which is different from Metrics/EnvironmentSteps for off-policy methods.

alf/drivers/off_policy_driver.py
```diff
@@ -195,32 +201,26 @@ def train(self,
         experience = self._algorithm.preprocess_experience(experience)
 
         length = experience.step_type.shape[1]
-        if mini_batch_length is None:
-            mini_batch_length = length
-        else:
-            mini_batch_length = min(mini_batch_length, length)
+        mini_batch_length = (mini_batch_length or length)
+        assert length % mini_batch_length == 0, (
+            "length=%s mini_batch_length=%s" % (length, mini_batch_length))
```
le-horizon: be a bit more informative in the error message?  e.g. length=%s not a multiple of mini_batch_length=%s.

alf/drivers/off_policy_driver.py
```diff
@@ -195,32 +201,26 @@ def train(self,
...
 
         batch_size = experience.step_type.shape[0]
-        if mini_batch_size is None:
-            mini_batch_size = batch_size
-        else:
-            mini_batch_size = min(mini_batch_size, batch_size)
-
-        assert length % mini_batch_length == 0
+        mini_batch_size = (mini_batch_size or batch_size)
+        # The reason of this constraint is at L233
```
le-horizon: be a bit more informative, e.g. reason is tf.reshape at L241

hnyu: updated

alf/drivers/off_policy_driver.py
hnyu: updated

PR #142, opened 2019-08-06
fix gridsearch (introduce by pr #114)

witwolf:


witwolf:
> Why did you remove sac_bullet_humanoid.gin?

i noticed that and add it back . it's a misdeletion

PR #143, opened 2019-08-06
add support for discrete action in sac

witwolf:


alf/algorithms/sac_algorithm.py
```diff
@@ -190,53 +194,80 @@ def predict(self, time_step: ActionTimeStep, state=None):
...
-
-        def actor_loss_fn(dqda, action):
-            if self._dqda_clipping:
-                dqda = tf.clip_by_value(dqda, -self._dqda_clipping,
-                                        self._dqda_clipping)
-            loss = 0.5 * losses.element_wise_squared_loss(
-                tf.stop_gradient(dqda + action), action)
-            loss = tf.reduce_sum(loss, axis=list(range(1, len(loss.shape))))
-            return loss
+            action_probs = action_distribution._probs
```
Haichao-Zhang: action_distribution._probs is None when I run the code using tensorflow 2.0.0-beta1. Somehow it is not converted from logits.
Result from print(action_distribution.__dict__):
{'_parameters': {'name': 'Categorical', 'allow_nan_stats': True, 'validate_args': False, 'dtype': tf.int64, 'probs': **None**, 'logits': <tf.Tensor: id=509, shape=(8, 2), dtype=float32, numpy=
array([[ 1.6195892e-03, -4.9218588e-04], ....

alf/examples/sac_cart_pole.gin
```diff
@@ -0,0 +1,44 @@
...
+create_sac_algorithm.actor_fc_layers=(100,)
+create_sac_algorithm.critic_fc_layers=(100,)
+create_sac_algorithm.actor_learning_rate=1e-3
+create_sac_algorithm.critic_learning_rate=1e-3
+create_sac_algorithm.alpha_learning_rate=1e-3
+SacAlgorithm.target_update_tau=0.01
+OneStepTDLoss.td_error_loss_fn=@losses.element_wise_squared_loss
+OneStepTDLoss.gamma=0.98
+
+ActorDistributionNetwork.continuous_projection_net=@NormalProjectionNetwork
```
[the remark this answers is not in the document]
witwolf: removed ,  it's inoperative config, i copied the gin file and didn't notice to remove them

alf/algorithms/sac_algorithm.py
witwolf: distributions it's in package `tfp` not in `tensorflow` , i use `tfp-nightly==0.8.0.dev20190609` in my workstation 

and avoid dependency inconsistency,  i suggest use docker as our dev env
[alf image gpu](https://docker.hobot.cc/harbor/projects/14/repositories/rl%2Ftfagent)
[alf image cpu](https://docker.hobot.cc/harbor/projects/14/repositories/rl%2Ftfagent-cpu)

Haichao-Zhang: Tested with this version of tfp and works good.

emailweixu: _probs seems to be a private member of distribution. We should avoid using it. Can you find some other ways to get it, like the extracting parameters for distribution when it’s stored in experience buffer

PR #145, opened 2019-08-06
add interface for deepmind lab enviroments

witwolf:
many algorithm about exploration using dmlab as test envs, it would be helpful if we want to try these algorithms . and i am try to add merlin examples using dmlab envs mentioned in the paper . 
we do not have to check this pr in if it's not very useful

### DIFF
--- a/alf/environments/suite_dmlab.py
+++ b/alf/environments/suite_dmlab.py
@@ -46,6 +46,31 @@ def action_discretize(action_spec,
 
     TODO: action combinations
 
+    Mapping all valid action values to discrete action
+
+    original deepmind lab environment action_spec:
+
+    `[{'max': 512, 'min': -512, 'name': 'LOOK_LEFT_RIGHT_PIXELS_PER_FRAME'},
+    {'max': 512, 'min': -512, 'name': 'LOOK_DOWN_UP_PIXELS_PER_FRAME'},
+    {'max': 1, 'min': -1, 'name': 'STRAFE_LEFT_RIGHT'},
+    {'max': 1, 'min': -1, 'name': 'MOVE_BACK_FORWARD'},
+    {'max': 1, 'min': 0, 'name': 'FIRE'},
+    {'max': 1, 'min': 0, 'name': 'JUMP'},
+    {'max': 1, 'min': 0, 'name': 'CROUCH'}]`
+
+    and discretized actions:
+
+    0  -> [20,0,0,0,0,0,0] (look left 20 pixels),
+    1  -> [-20,0,0,0,0,0,0] (look right 20 pixels),
+    ...,
+    m  -> [0,0,0,-1,0,0,0] (move back),
+    m+1-> [0,0,0,1,0,0,0] (move forward) ,
+    ...,
+    n  -> [0,0,0,0,1,1,0] (jump and fire),
+    ...
+
+    see `SuiteDMLabTest.test_action_discretize` in `suite_dmlab_test.py` for examples
+
     Args:
         action_spec (list(dict)): action spec
         look_left_right_pixels_per_frame (iterable|str): look left or look right pixels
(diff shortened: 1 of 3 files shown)


alf/environments/suite_dmlab.py
```diff
@@ -0,0 +1,195 @@
...
+def action_discretize(action_spec,
+                      look_left_right_pixels_per_frame=(-20, 20),
+                      look_down_up_pixels_per_frame=(-10, 10),
+                      strafe_left_right=(-1, 1),
+                      move_back_forward=(-1, 1),
+                      fire=(),
+                      jump=(1, ),
+                      crouch=(1, ),
+                      **kwargs):
+    """Discretize action from action_spec
```
hnyu: More details: describe what the discretization does by giving examples, especially what we should expect about the output representation.

alf/environments/suite_dmlab.py
```diff
@@ -0,0 +1,195 @@
...
+                 action_repeat=4,
+                 observation='RGB_INTERLEAVED',
+                 config={},
+                 renderer='software'):
+        """Create an deepmind_lab env
+
+        Args:
+            scene (str): script for the deepmind_lab env. See available script:
+                `https://github.com/deepmind/lab/tree/master/game_scripts/levels`
+            action_repeat (int): the frequency at which the agent experiences the game
```
hnyu: 'frequency' -> 'interval'

alf/environments/suite_dmlab.py
```diff
@@ -0,0 +1,195 @@
...
+        val_max = spec['max']
+        values = config.get(spec['name'], None)
+
+        if values is None:
+            values = list(range(val_min, val_max + 1))
+        elif isinstance(values, str):
+            values = eval(values)
+
+        for value in values:
+            if value < val_min or value > val_max or value == 0:
```
hnyu: why ignoring value == 0?

alf/environments/suite_dmlab.py
```diff
@@ -0,0 +1,195 @@
...
+
+    def seed(self, seed=None):
+        self._lab.reset(seed=seed)
+
+    def close(self):
+        self._lab.close()
+
+    def render(self, mode='rgb_array', close=False):
+        if mode == 'rgb_array':
+            return self._last_obs
```
hnyu: raise errors in other cases

alf/environments/wrappers.py
```diff
@@ -110,6 +110,47 @@ def __getattr__(self, name):
...
+        super().__init__(env)
+        obs_shape = env.observation_space.shape
+        self.observation_space = gym.spaces.Box(
+            low=0, high=255, shape=list(obs_shape[:-1]) + [1], dtype=np.uint8)
+
+    def observation(self, obs):
+        obs = cv2.cvtColor(obs, cv2.COLOR_RGB2GRAY)
+        return np.expand_dims(obs, -1)
+
+    def __getattr__(self, name):
```
hnyu: You don't need this function

alf/environments/wrappers.py
```diff
@@ -110,6 +110,47 @@ def __getattr__(self, name):
...
+        return obs
+
+
+@gin.configurable
+class FrameGrayScale(gym.ObservationWrapper):
+    """Gray scale image observation"""
+
+    def __init__(self, env):
+        super().__init__(env)
+        obs_shape = env.observation_space.shape
```
hnyu: should check the current obs_shape. Make sure it has no more than three dims.

alf/environments/wrappers.py
```diff
@@ -110,6 +110,47 @@ def __getattr__(self, name):
         return getattr(self.env, name)
 
 
+@gin.configurable
+class FrameResize(gym.ObservationWrapper):
+    def __init__(self, env, width=84, height=84):
+        super().__init__(env)
+        self._width = width
+        self._height = height
+        obs_shape = env.observation_space.shape
```
hnyu: should check the current obs_shape. Make sure it has no more than three dims.

alf/environments/suite_dmlab.py
witwolf: valid value for action is   `[min, max]  - {0}`, 0 means idle in this direction

alf/environments/suite_dmlab.py
```diff
@@ -0,0 +1,222 @@
...
+class DeepmindLabEnv(gym.Env):
+    metadata = {'render.modes': ['rgb_array']}
+
+    def __init__(self,
+                 scene,
+                 action_repeat=4,
+                 observation='RGB_INTERLEAVED',
+                 config={},
+                 renderer='hardware'):
+        """Create an deepmind_lab env
```
emailweixu: an => a

PR #148, opened 2019-08-07
PR ac vtrace implementation and unit tests

le-horizon:
This is the new branch with clean commit history (properly rebased from master).

Could you take another look Wei and Haonan?

Thanks,
Le


hnyu:
Did you test additional games with v-trace? Was the performance stable on those games?

le-horizon:
Cart pole was successful.  Simple navigation is still training (painstakingly slow, due to only 8 parallel env possible, otherwise OOM happens, maybe I should restart my machine).

________________________________

le-horizon:
Simple Navigation is gaining return, at similar rate as ac_simple_navigation.  With 10 parallel env and 3 actors after 3 days of training plot is attached.


________________________________

PR #151, opened 2019-08-19
reduce creation of environments in async_off_policy_trainer

witwolf:
`AsyncOffPolicyTrainer`  created   `(N+2) * num_parallel_environments`  env instances  while only `N * num_parallel_environments` were used for training . It takes a lot of time in the start-up phase when using simulated environment, especially for `socail_robot`


alf/drivers/async_off_policy_driver.py
```diff
@@ -108,8 +110,15 @@ def __init__(self,
                 tf.summary.experimental.get_step(). If this is still None, a
                 counter will be created.
         """
+        if isinstance(env_or_env_fn, TFEnvironment):
```
hnyu: Can we just enforce env_or_env_fn to be env? If this is more efficient, then probably there will be no need for an env_fn.

hnyu:
BTW, did you find async training useful in any SocialRobot training case?

alf/drivers/async_off_policy_driver.py
witwolf: total env num depends on `num_parallel_environments`  and  `AsyncOffPolicyDriver.num_envs`  

so  we should move  configurable parameter `num_envs` to `TrainerConfig`, is it appropriate ?

witwolf:
> BTW, did you find async training useful in any SocialRobot training case?

it did only a little help, i just set N=1 to reduce the training time (time for collecting a batch data > training a batch data )

alf/drivers/async_off_policy_driver.py
hnyu: I mean we can directly discard the second branch here and just assume env_or_env_fn is always an environment for efficiency. num_envs stays unchanged.

hnyu:
> > BTW, did you find async training useful in any SocialRobot training case?
> 
> it did only a little help, i just set N=1 to reduce the training time (time for collecting a batch data > training a batch data )

A rule of thumb for this scenario (batch collection time > batch training time) would be to have 2 or 3 actors, each actor with N environments, where N is the number of envs you originally have for your A2C sync training. You are expected to observe a much higher training throughput if there are **enough** resources on the computer.

alf/drivers/async_off_policy_driver.py
```diff
@@ -59,7 +61,7 @@ class AsyncOffPolicyDriver(OffPolicyDriver):
     """
 
     def __init__(self,
-                 env_f: Callable,
+                 env_or_env_fn,
```
emailweixu: how about using a list of environments as argument and remove num_envs. It's a cleaner interface.

alf/drivers/async_off_policy_driver.py
witwolf: env can be created by other functions, `create_environment` is just the default choice 

eg: 
https://github.com/HorizonRobotics/alf/blob/62b3aa1158b7162fac66c7ad568ba59a88266e36/alf/drivers/off_policy_driver_test.py#L112-L118
pass an env_fn

alf/drivers/async_off_policy_driver.py
witwolf: so `AsyncOffPolicyDriver.num_envs` should be moved to upper interface  in that way

alf/drivers/async_off_policy_driver.py
hnyu: OK, never mind.

alf/drivers/async_off_policy_driver.py
emailweixu: It has to be moved out since we are passing a list of environments this way, the caller need to know how many environments. One potential advantage is that we can construct the environments differently if needed.

emailweixu:
@witwolf why not continue with this PR?

witwolf:
new pr #175 with clean commit history

PR #154, opened 2019-08-21
Record training time to tensorboard

emailweixu:
time for each iteration and time for each driver.train

### DIFF
--- a/alf/trainers/off_policy_trainer.py
+++ b/alf/trainers/off_policy_trainer.py
@@ -12,7 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import time
+
 import gin.tf
+import tensorflow as tf
+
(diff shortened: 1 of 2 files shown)


witwolf:
I've implemented a simple decorator to record time cost for function call , but somehow it can not work with `tf.while_loop` in `tf.function` mode 
[commit](https://github.com/HorizonRobotics/alf/commit/38c634950e4186fbed7cb901f4bab9d1e034a65b)

hnyu:
> I've implemented a simple decorator to record time cost for function call , but somehow it can not work with `tf.while_loop` in `tf.function` mode
> [commit](https://github.com/HorizonRobotics/alf/commit/38c634950e4186fbed7cb901f4bab9d1e034a65b)

You need add with tf.control_dependencies to avoid any possible parallelism in the graph mode.

PR #155, opened 2019-08-24
Making normal projection stabler.

emailweixu:
The previous ppo_icubwalk uses state independent standard deviation for action distribution and reaches reward 700 after 10M steps. Using state dependent std can reach higher reward but the training is very unstable. This PR tries to make the training more stable by using a different way of parameterizing Gaussian. In the figure blue and red curves uses new method, orange curve uses old method.
icubwalk-stable-normal-projection

### DIFF
--- a/alf/networks/stable_normal_projection_network.py
+++ b/alf/networks/stable_normal_projection_network.py
@@ -1,0 +1,141 @@
+# Copyright (c) 2019 Horizon Robotics. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Project inputs to a normal distribution object."""
+import math
+
+import gin.tf
+
+import tensorflow as tf
+import tensorflow_probability as tfp
+
+from tf_agents.networks import utils as network_utils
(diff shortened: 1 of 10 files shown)


alf/networks/stable_normal_projection_network.py
```diff
@@ -0,0 +1,141 @@
...
+            c -= 1 / (max_std - min_std)
+
+        if inverse_std_transform == 'exp':
+            std_transform = tf.math.exp
+            std_bias_initializer_value = math.log(c)
+        elif inverse_std_transform == 'softplus':
+            std_transform = tf.nn.softplus
+            std_bias_initializer_value = math.log(math.exp(c) - 1)
+        else:
+            raise TypeError("Unsupported std_transform %s" % std_transform)
```
witwolf: raise TypeError("Unsupported std_transform %s" % inverse_std_transform)

alf/utils/math_ops.py
```diff
@@ -0,0 +1,31 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import gin
+import tensorflow as tf
+"""Various math ops."""
+
+
+@gin.configurable
+def clipped_exp(value, clip_value_min=-20, clip_value_max=2):
```
witwolf: same implementation with `alf.utils.common.std_clip_transform` , should remove one of them

alf/networks/stable_normal_projection_network.py
```diff
@@ -0,0 +1,141 @@
...
+        """
+        self._min_std = min_std
+        self._max_std = max_std
+
+        c = 1 / (init_std - min_std)
+        if max_std is not None:
+            c -= 1 / (max_std - min_std)
+
+        if inverse_std_transform == 'exp':
+            std_transform = tf.math.exp
```
witwolf: numeric overflow  is very likely to happen if `tf.math.exp` was apply without clipping

Jialn:
The final reward seems promising, nearly 200 higher than SAC.

PR #158, opened 2019-08-26
Use commandline flags for play

emailweixu:
It's more convenient to use commandline flags than gin_param

And separate play and train_eval to two files.

Will change README and gin files in social robot after this is approved.

### DIFF
--- a/alf/bin/play.py
+++ b/alf/bin/play.py
@@ -1,0 +1,84 @@
+# Copyright (c) 2019 Horizon Robotics. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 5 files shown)


alf/bin/play.py
```diff
@@ -0,0 +1,84 @@
...
+flags.DEFINE_bool('greedy_predict', True, "use greedy action for evaluation.")
+flags.DEFINE_integer('random_seed', 0, "random seed")
+flags.DEFINE_integer('num_episodes', 10, "number of episodes to play")
+flags.DEFINE_float('sleep_time_per_step', 0.01,
+                   "sleep so many seconds for each"
+                   " step")
+flags.DEFINE_string(
+    'record_file', None, "If provided, video will be recorded"
+    "to a file instead of shown on the screen.")
+flags.DEFINE_multi_string('gin_file', None, 'Paths to the gin-config files.')
```
witwolf: No need for `gin_file` and `gin_param` any more

PR #159, opened 2019-08-27
Support using rollout RNN states for off-policy training

### LINKED ISSUE #70: Options for storing RNN states in experience and using it for training
Reference: https://openreview.net/forum?id=r1lyTjAqYX

emailweixu:
Fix #70 

The correctness of the function is tested using ICM model, which makes the observation as state so that it can be used in the next step. 

Haven't tried its effectiveness on real RNN model yet.

For ICM on supermario, ppo seems to be much more sample efficient. In the figure, orange curve is previous result from A2C, red curve is from PPO.
icm_ppo_vs_ac


alf/examples/ppo_async_icm_super_mario_intrinsic_only.gin
```diff
@@ -0,0 +1,81 @@
+include 'ppo.gin'
+
+# icm on SuperMario using PPO algorithm
+
+CHANNEL_ORDER='channels_first'
```
witwolf: The bug in tf has been fixed, it's not need to change to channel order any more

alf/examples/ppo_async_icm_super_mario_intrinsic_only.gin
```diff
@@ -0,0 +1,81 @@
...
+TrainerConfig.eval_interval = 100
+TrainerConfig.evaluate = True
+TrainerConfig.summarize_grads_and_vars = True
+TrainerConfig.debug_summaries=1
+TrainerConfig.summary_interval = 10
+TrainerConfig.use_rollout_state = True
+TrainerConfig.use_tf_functions = True
+
+PolicyDriver.observation_transformer=@image_scale_transformer
+TFUniformReplayBuffer.max_length=32
```
witwolf: it's inoperative config,  `OnetimeExperienceReplayer` is used by default

alf/drivers/threads.py
```diff
@@ -117,7 +117,15 @@ def repeat_shape_n(nested_spec, n):
         lambda t: tf.TensorSpec([n] + list(t.shape), t.dtype), nested_spec)
 
 
+LearningBatch = namedtuple("LearningBatch", [
+    "time_step", "state", "policy_step", "act_dist_param", "next_time_step",
+    "env_id"
```
witwolf: actually env_id never been used, and is value was set to  no meaning  `tf.ones()`

PR #163, opened 2019-08-30
Refactoring RLAlgorithm interface

emailweixu:
The main goal of the refactoring is to make the code easier to understand, especially for part for the handling of RNN states for off-policy training

Removed OffPolicyAdpater and TrainStepAdapter. Introduced rollout() for the interface, which is used for rollout during training.

### DIFF
--- a/alf/algorithms/ppo_algorithm.py
+++ b/alf/algorithms/ppo_algorithm.py
@@ -19,7 +19,6 @@ import tensorflow as tf
 
 from alf.algorithms.actor_critic_algorithm import ActorCriticAlgorithm
 from alf.algorithms.actor_critic_algorithm import create_ac_algorithm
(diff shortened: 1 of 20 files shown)


alf/algorithms/ppo_algorithm.py
```diff
@@ -63,13 +58,6 @@ def preprocess_experience(self, exp: Experience):
...
-    def predict(self, time_step: ActionTimeStep, state=None):
-        return self._algorithm.train_step(time_step, state)
-
-    @property
-    def predict_state_spec(self):
-        return self._algorithm.train_state_spec
-
 
 @gin.configurable
 def create_ppo_algorithm(env, debug_summaries=False):
```
witwolf: perhaps we should add a parameter `loss_class=PPOLoss` here instead of config it in ppo.gin , it make  code easier to read

PR #165, opened 2019-09-05
allow flag config of logging verbosity

le-horizon:
small change to allow flag controlled verbose messages when playing or training model.

### DIFF
--- a/alf/bin/play.py
+++ b/alf/bin/play.py
@@ -54,10 +54,23 @@ flags.DEFINE_string(
 flags.DEFINE_multi_string('gin_file', None, 'Paths to the gin-config files.')
 flags.DEFINE_multi_string('gin_param', None, 'Gin binding parameters.')
 
+flags.DEFINE_string('log_level', 'i', "Default logging verbosity: d|i|w|e|f.")
+
 FLAGS = flags.FLAGS
 
 
 def main(_):
+    if FLAGS.log_level == 'd':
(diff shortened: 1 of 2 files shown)


emailweixu:
absl.logging already has the following flags, Did you try if they work?

absl.logging:
  --[no]alsologtostderr: also log to stderr?
    (default: 'false')
  --log_dir: directory to write logfiles into
    (default: '')
  --[no]logtostderr: Should only log to stderr?
    (default: 'false')
  --[no]showprefixforinfo: If False, do not prepend prefix to info messages when it's logged to stderr, --verbosity is set to INFO level, and python logging is used.
    (default: 'true')
  --stderrthreshold: log messages at this level, or more severe, to stderr in addition to the logfile.  Possible values are 'debug', 'info', 'warning', 'error', and 'fatal'.  Obsoletes --alsologtostderr. Using --alsologtostderr cancels the effect of
    this flag. Please also note that this flag is subject to --verbosity and requires logfile not be stderr.
    (default: 'fatal')
  -v,--verbosity: Logging verbosity level. Messages logged at this level or lower will be included. Set to 1 for debug logging. If the flag was not set or supplied, the value will be changed from the default of -1 (warning) to 0 (info) after flags are
    parsed.
    (default: '-1')
    (an integer)

le-horizon:
You are right Wei.  Using -v 1 works.  Abandoning this change.

PR #170, opened 2019-09-11
Fix the step counting in run_async()

hnyu:
We still better infer the steps from batch because the actual number of environments is decided by the learning queue capacity but not the total number of environments.

Because observation might contains multiple inputs (not a tensor), so we use reward instead.


le-horizon:
Good to know.  Thanks for the fix Haonan!

PR #173, opened 2019-09-13
fix throughput for on-policy training

hnyu:
@le-horizon  Sorry for your last PR I didn't review carefully. num_updates_per_step is only used in off-policy training. So if you always multiply this, then the throughput number is incorrect for on-policy training 

Edited: There is no easy way to check which trainer is used in policy_trainer.py without introducing circular imports. So I moved the throughput logging into sub-trainer files. Now steps is not required to be returned by train_iter().

### DIFF
--- a/alf/drivers/off_policy_driver.py
+++ b/alf/drivers/off_policy_driver.py
@@ -219,6 +219,10 @@ class OffPolicyDriver(policy_driver.PolicyDriver):
             mini_batch_size (int): number of sequences for each minibatch
             mini_batch_length (int): the length of the sequence for each
                 sample in the minibatch
+
+        Returns:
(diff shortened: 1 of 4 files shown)


le-horizon:
Good point.  Thanks for the fix Haonan!

emailweixu:
It seems that an easier fix is let the off-policy driver return steps*num_updates_per_step

> Edited: There is no easy way to check which trainer is used in policy_trainer.py without introducing circular imports. So I moved the throughput logging into sub-trainer files. Now steps is not required to be returned by train_iter().

emailweixu:
What I really meant is to change the return of train_iter. But I guess this change is also fine.

PR #174, opened 2019-09-13
use tf.shape

emailweixu:
use tf.shape(tensor) instead of tensor.shape so that the operators involving shape can handle dynamic  batch size.
tensor.shape is some python data structure which does not have defined value (i.e. it may contain None) if the shape is dynamically set. We should always use tf.shape() to access the shape of a tensor.

### DIFF
--- a/alf/drivers/off_policy_driver.py
+++ b/alf/drivers/off_policy_driver.py
@@ -234,16 +234,12 @@ class OffPolicyDriver(policy_driver.PolicyDriver):
                     "for off-policy training of RNN.")
 
         experience = tf.nest.map_structure(
-            lambda x: tf.reshape(x, [-1, mini_batch_length] + list(x.shape[2:])
-                                 ), experience)
+            lambda x: tf.reshape(
+                x, common.concat_shape([-1, mini_batch_length],
+                                       tf.shape(x)[2:])), experience)
 
-        batch_size = experience.step_type.shape[0]
+        batch_size = tf.shape(experience.step_type)[0]
         mini_batch_size = (mini_batch_size or batch_size)
-        # The reason of this constraint is tf.reshape at L233
-        # TODO: remove this constraint.
-        assert batch_size % mini_batch_size == 0, (
-            "batch_size=%s not a multiple of mini_batch_size=%s" %
-            (batch_size, mini_batch_size))
 
         def _make_time_major(nest):
             """Put the time dim to axis=0."""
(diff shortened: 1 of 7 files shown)


alf/drivers/off_policy_driver.py
```diff
@@ -292,13 +281,15 @@ def _update(self, experience, weight):
                 lambda state: state[0, ...], experience.state)
         else:
             first_train_state = initial_train_state
-        num_steps = experience.step_type.shape[0]
+        num_steps = tf.shape(experience.step_type)[0]
 
         def create_ta(s):
+            # TensorArray cannot use Tensor (batch_size) as element_shape
+            ta_batch_size = experience.step_type.shape[1]
```
hnyu: Can we directly do int(batch_size) here?

alf/utils/common.py
```diff
@@ -214,23 +214,39 @@ def _tensor_parameters_only(params):
...
     Args:
         x (Tensor): source tensor
         y (Tensor): target tensor. Only its shape will be used.
-    Returns
+    Returns:
         x with extra singular dimensions.
     """
     assert len(x.shape) <= len(y.shape)
-    assert x.shape == y.shape[:len(x.shape)]
+    tf.assert_equal(tf.shape(x), tf.shape(y)[:len(x.shape)])
```
hnyu: x.shape

emailweixu: What’s the problem?

emailweixu: What’s the problem?

alf/drivers/off_policy_driver.py
emailweixu: I didn’t try. But I guess not.

witwolf: > Can we directly do int(batch_size) here?

can not convert a tensor to an integer in graph mode

alf/drivers/off_policy_driver.py
```diff
@@ -253,37 +249,30 @@ def _make_time_major(nest):
         for u in tf.range(num_updates):
             if mini_batch_size < batch_size:
```
witwolf: conditional expression `mini_batch_size < batch_size`  is a tensor  as `batch_size` is a tensor
Peharps we should avoid using a tensor as a bool value

emailweixu: What is the problem of that?

witwolf: Somehow  that using a tensor as bool doesn't work in graph mode sometimes

emailweixu: But this one is not a problem.

PR #175, opened 2019-09-16
reduce env created when using async off policy trainer

witwolf:
checkout from  #151 with clean commit history

### DIFF
--- a/alf/trainers/off_policy_trainer.py
+++ b/alf/trainers/off_policy_trainer.py
@@ -92,18 +92,27 @@ class SyncOffPolicyTrainer(OffPolicyTrainer):
 class AsyncOffPolicyTrainer(OffPolicyTrainer):
     """Perform off-policy training using AsyncOffPolicyDriver"""
 
+    def __init__(self, config):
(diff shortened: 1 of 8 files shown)


alf/trainers/off_policy_trainer.py
```diff
@@ -92,18 +92,27 @@ def train_iter(self, iter_num, policy_state, time_step):
...
+            envs.append(create_environment())
         driver = AsyncOffPolicyDriver(
-            env_f=create_environment,
+            envs=envs,
             algorithm=self._algorithm,
             use_rollout_state=self._config.use_rollout_state,
             unroll_length=self._unroll_length,
             debug_summaries=self._debug_summaries,
             summarize_grads_and_vars=self._summarize_grads_and_vars)
-        driver.start()
```
emailweixu: why do you move driver.start() to train_iter()?

witwolf: we reset the environments and get initial timestep at the begin of training, `EnvThread` should run after that, so `env.*` will not run simultaneously

PR #179, opened 2019-09-17
fix entropy_with_fallback

witwolf:
try ... catch doesn't work in graph mode .  so it can not run success when  SquashToSpecNormal is used  (NormalProjectionNetwork.scale_distribution=True)

### DIFF
--- a/alf/utils/dist_utils.py
+++ b/alf/utils/dist_utils.py
@@ -116,12 +116,13 @@ def entropy_with_fallback(distributions, action_spec, seed=None):
                 action_spec.shape.ndims)
 
     def _compute_entropy(dist: tfp.distributions.Distribution, action_spec):
-        try:
-            entropy = dist.entropy()
-            entropy_for_gradient = entropy
-        except NotImplementedError:
+        if isinstance(dist, SquashToSpecNormal):
             entropy, entropy_for_gradient = estimated_entropy(
                 dist, seed=seed_stream())
+        else:
+            entropy = dist.entropy()
+            entropy_for_gradient = entropy
+
         outer_rank = _calc_outer_rank(dist, action_spec)
         rank = entropy.shape.ndims
         reduce_dims = list(range(outer_rank, rank))


hnyu:
Somehow I feel the current solution is not scalable. Every time we add some distribution then we might need to change this code block.

witwolf:
> Somehow I feel the current solution is not scalable. Every time we add some distribution then we might need to change this code block.

`_calc_outer_rank` use `isinstance()` too . 

 we do not need to change this code  if we implement all necessary interface functions when add a new Distribution.

hnyu:
> > Somehow I feel the current solution is not scalable. Every time we add some distribution then we might need to change this code block.
> 
> `_calc_outer_rank` use `isinstance()` too .
> 
> we do not need to change this code if we implement all necessary interface functions when add a new Distribution.

But for some distribution, it's not possible to calculate the entropy in a closed form.

hnyu:
> > > Somehow I feel the current solution is not scalable. Every time we add some distribution then we might need to change this code block.
> > 
> > 
> > `_calc_outer_rank` use `isinstance()` too .
> > we do not need to change this code if we implement all necessary interface functions when add a new Distribution.
> 
> But for some distribution, it's not possible to calculate the entropy in a closed form and we have to rely on estimating it.

witwolf:
> > > > Somehow I feel the current solution is not scalable. Every time we add some distribution then we might need to change this code block.
> > > 
> > > 
> > > `_calc_outer_rank` use `isinstance()` too .
> > > we do not need to change this code if we implement all necessary interface functions when add a new Distribution.
> 
> But for some distribution, it's not possible to calculate the entropy in a closed form and we have to rely on estimating it.

i see.   what is your recommended solution here ?

hnyu:
> > > > > Somehow I feel the current solution is not scalable. Every time we add some distribution then we might need to change this code block.
> > > > 
> > > > 
> > > > `_calc_outer_rank` use `isinstance()` too .
> > > > we do not need to change this code if we implement all necessary interface functions when add a new Distribution.
> > 
> > 
> > But for some distribution, it's not possible to calculate the entropy in a closed form and we have to rely on estimating it.
> 
> i see. what is your recommended solution here ?

Maybe we should force to implement entropy() with estimated_entropy() if there is no closed form. But for now let's just use isinstance() to handle special cases. No better solution I can think of at this moment.

Issue #178, opened 2019-09-16
Refactoring: remove ceate_???_algorithm


emailweixu:
We should make it possible to directly configure Algorithm from gin instead of relying on create_???_algorithm.
Since we need to pass observation_spec, action_spec, time_spec etc. to many of the the constructors. We can implement several functions: observation_spec(), action_spec(), time_step_spec() to get those specs from gin.

witwolf:
it will make the configurable interface more clearer , i'll do this refactor

Issue #177, opened 2019-09-16
Refactoring: Change MERLIN so that it can be used from ActorCriticAlgorithm


emailweixu:
In fact, MERLIN is a kind of AC algorithm, we should allow the current ActorCriticAlgorithm to use the featue of MERLIN.

PR #181, opened 2019-09-17
remove create_*algorithms and config algorithms directly

witwolf:
#178
Will change gin files in social robot after this and remove create_merlin_algorithm at #177


alf/examples/ppo_icm_super_mario_intrinsic_only.gin
```diff
@@ -15,27 +15,38 @@ suite_mario.load.state="Level1-1"
...
-ActorDistributionNetwork.activation_fn=@tf.nn.elu
-CategoricalProjectionNetwork.logits_init_output_factor=1e-10
-ActorDistributionNetwork.discrete_projection_net=@CategoricalProjectionNetwork
+ac/Adam.learning_rate=3e-5
+
+icm/EncodingNetwork.input_tensor_spec=%observation_spec
+icm/EncodingNetwork.conv_layer_params=%CONV_LAYER_PARAMS
+icm/EncodingNetwork.fc_layer_params=(256,)
+icm/EncodingNetwork.activation_fn=@tf.nn.elu
+icm/to_tensor_spec.shape=(256,)
```
emailweixu: should use a macro for this as it's should be same as  icm/EncodingNetwork.fc_layer_params.
Some other gin files for ICM may also have this problem

witwolf: feature_spec is (fc_layer_params[-1],) for a FC layers encoding network,  
They happen to be equal here as len(fc_layer_params)=1

but gin does not support config below: (Line3 , visit list element)
 
```
icm/encoding_fc_layers=(200,)
icm/EncodingNetwork.fc_layer_params=%icm/encoding_fc_layers
icm/to_tensor_spec.shape=(%icm/encoding_fc_layers[-1],)
ICMAlgorithm.feature_spec=@icm/to_tensor_spec()
```

alf/bin/play.py
```diff
@@ -60,10 +60,11 @@
 def main(_):
     gin_file = common.get_gin_file()
     gin.parse_config_files_and_bindings(gin_file, FLAGS.gin_param)
-    algorithm_ctor = gin.query_parameter(
-        'TrainerConfig.algorithm_ctor').scoped_configurable_fn
+    algorithm_cls = gin.query_parameter(
```
Jialn: Does "_cls" here mean "class"? Maybe change to something like "algorithm_type" is more easier for understanding.

alf/examples/ppo_icm_super_mario_intrinsic_only.gin
Jialn: since basically function "to_tensor_spec" is only used for ICM, maybe we can remove function "to_tensor_spec", use 
`ICMAlgorithm.feature_shape=icm/encoding_fc_layers`, and add
`feature_spec=tf.TensorSpec(shape=feature_shape, dtype=tf.float32)` in icm_alg.py,

alf/bin/play.py
witwolf: Personly think cls is better as we configured a class not  str type here

alf/examples/ac_cart_pole.gin
```diff
@@ -10,20 +11,34 @@ ActorCriticAlgorithm.reward_shaping_fn = @reward_scaling
 common.reward_scaling.scale = 0.01
 
 # algorithm config
+observation_spec=@get_observation_spec()
+action_spec=@get_action_spec()
+
+actor/ActorDistributionNetwork.input_tensor_spec=%observation_spec
```
Jialn: Seems we are writing the same input_tensor and output_tensor spec in every gin file... Can we fix input_tensor to observation_spec and output_tensor to action_sepc in python?

alf/examples/ppo_icm_super_mario_intrinsic_only.gin
witwolf: algorithms constructor using `tf.TensorSpec`  to specifiy input or output spec, we should keep interface consistency

alf/examples/ac_cart_pole.gin
witwolf: we remove create_algorithm* and configure algorithm directly, this  increase flexibility and as consquence  it may lost some reusability

Jialn: I know. I mean can we still give a default tensor spec configuration for algorithm class?

alf/bin/play.py
emailweixu: the original algorithm_ctor might be better, because it's still conceivable that we use a function to create the algorithm in the future.

alf/examples/ac_cart_pole.gin
emailweixu: It's possible to do that for some of alf classes.  But not possible for tf_agents classes.

alf/examples/ppo_icm_super_mario_intrinsic_only.gin
emailweixu: I would be much better If gin can accept function calling with arguments

alf/bin/play.py
witwolf: i have renamed it back to `algorithm_ctor`

alf/utils/common.py
```diff
@@ -685,3 +685,48 @@ def _if_true():
         return tf.constant(True)
 
     tf.cond(cond, _if_true, lambda: tf.constant(False))
+
+
+@gin.configurable
+def to_tensor_spec(shape, dtype=tf.float32):
```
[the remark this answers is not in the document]
witwolf: added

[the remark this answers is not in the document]
witwolf: i added it before , somehow it did not work  and throw an Exception , maybe i added it in a wrong way ...

alf/examples/ac_cart_pole.gin
le-horizon: Or we can have a base.gin file that has all the default setup, and later gin files can override them only if needed?

alf/examples/ppo_icm_super_mario_intrinsic_only.gin
emailweixu: Since we know it's same as the 200 used in icm/encoding_fc_layers=(200,),  we should use a macro so that the error of changing only one of them won't occur.

> feature_spec.shape is (fc_layer_params[-1],) for a FC layers encoding network,
> They happen to be equal here as len(fc_layer_params)=1
> 
> but gin does not support config below: (Line3 , visit list element)
> 
> ```
> icm/encoding_fc_layers=(200,)
> icm/EncodingNetwork.fc_layer_params=%icm/encoding_fc_layers
> icm/to_tensor_spec.shape=(%icm/encoding_fc_layers[-1],)
> ICMAlgorithm.feature_spec=@icm/to_tensor_spec()
> ```

alf/examples/icm_mountain_car.gin
```diff
@@ -10,23 +10,41 @@ load_with_random_max_episode_steps.max_steps=1200
...
+ac/Adam.learning_rate=5e-5
 
-create_ac_algorithm.use_icm=1
-create_ac_algorithm.actor_fc_layers=(256,)
-create_ac_algorithm.value_fc_layers=(256,)
-create_ac_algorithm.encoding_fc_layers=(200,)
+icm/EncodingNetwork.input_tensor_spec=%observation_spec
+icm/EncodingNetwork.fc_layer_params=(200,)
+icm/EncodingNetwork.activation_fn=@tf.nn.elu
+icm/tf.TensorSpec.shape=(200,)
```
emailweixu: Can we just use icm/TensorSpec here?

witwolf: yes,  there is no other class or function named `TensorSpec` , so  a dotted module name prefix is not necessary required

alf/examples/ac_cart_pole.gin
witwolf: One consideration is that we copy gin files to output dir (tensorboard events and checkpoints) but missing included gin files,  so may be keep all configuration in just one gin file is better choice current now

alf/examples/ppo_icm_super_mario_intrinsic_only.gin
witwolf: > I would be much better If gin can accept function calling with arguments

it would be more flexible and easier to use , i'll  see if it is easy to extend.

PR #189, opened 2019-09-23
add NoisyArray for testing intrinsic rewards

hnyu:
@emailweixu Added a NoisyArray environment for testing the robustness to stochastic events. A comparison between KME and ICM is shown in the figure below. 

noisy_array_comp

where BLUE is KME+MI_loss_embedding, GRAY is KME+inverse_loss_embedding, RED is ICM+inverse_loss_embedding, and ORANGE is ICM+MI_loss_embedding. So basically ICM gets stuck by always changing the noise vector. See code comments for environment description. The array length is 11 (plus a noise vector of length 100) and the max steps is 50.

This environment might help test your goal generation method for sanity check.

@witwolf  Also change `auto_reset` form False to True in suite_mario.py. The reason is that if max_episode_steps=0, then the TimeLimit wrapper will not be applied and the tf_agents env wrapper is not reset after game_over. So the metric won't know which steps are StepType.FIRST and the AverageReturn metric will become broken because it resets accumulated rewards according to StepType.FIRST flags.

### DIFF
--- a/alf/environments/simple/noisy_array.py
+++ b/alf/environments/simple/noisy_array.py
@@ -18,23 +18,20 @@ import numpy as np
 
 import cv2
 
-import unittest
-from absl.testing import parameterized
-
 
 class NoisyArray(gym.Env):
     """
     A synthetic noisy array to test the agent's robustness to random noises. The
-    binary array has a length of (2K+1+M), where the subarray of length (2K+1)
-    is a onehot vector with 1 representing the agent's current location, and the
-    remaining M bits constitute a noise vector in {0,1}^M. For example (K=2,
+    binary array has a length of (K+M), where the subarray of length K is a
+    onehot vector with 1 representing the agent's current location, and the
+    remaining M bits constitute a noise vector in {0,1}^M. For example (K=5,
     M=3):
 
         0 0 1 0 0 | 0 1 1
 
     and the agent is at i==2 now.
 
-    The agent always starts from i==0. The goal is to reach i==2K (it cannot
+    The agent always starts from i==0. The goal is to reach i==K-1 (it cannot
     step on the noise vector). It has three actions: LEFT, RIGHT, and FIRE. The
     FIRE action changes the noise vector into some random M bits, without
     changing the agent's position. Both LEFT and RIGHT won't change the noise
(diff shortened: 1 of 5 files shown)


alf/environments/simple/noisy_array.py
```diff
@@ -0,0 +1,173 @@
...
+                game_ends += int(done)
+
+        self.assertEqual(game_ends, 2)
+        self.assertEqual(total_rewards, 2)
+        self.assertEqual(r, 1.0)
+        self.assertTrue(done)
+
+
+if __name__ == "__main__":
+    unittest.main()
```
emailweixu: Should put the test to a separate file ending with _test.py so that CI will run it automatically.

alf/environments/suite_simple.py
```diff
@@ -0,0 +1,67 @@
...
+
+@gin.configurable
+def load(game,
+         env_args=dict(),
+         discount=1.0,
+         wrap_with_process=True,
+         frame_skip=None,
+         frame_stack=None,
+         max_episode_steps=0,
+         spec_dtype_map=None):
```
emailweixu: comments for these arguments.

alf/environments/simple/noisy_array.py
```diff
@@ -0,0 +1,173 @@
...
+    def _gen_observation(self, act):
+        movement = act - 1
+        # If the current position is beyond the right boundary, put the agent
+        # back to the left
+        self._position = max(self._position + movement, 0)
+        self._position %= self._K
+
+        self._game_over = (self._position == self._K - 1)
+
+        # step type
```
emailweixu: what does "step type" mean here?

alf/environments/simple/noisy_array.py
hnyu: Done.

alf/environments/suite_simple.py
hnyu: Done.

alf/environments/simple/noisy_array.py
hnyu: Removed.

hnyu:
I also tested ICM on the noisy array with auto noise that changes it self at every time step, regardless of the agent's action. As expected, ICM can work well in this case. The reason is that because the noise vector is always changing independently, the inverse loss has to focus on the semantically meaningful array to predict the agent's actions. So the learned embedding is robust to the white noise. This is essentially the noise experiment the original ICM paper showed to demonstrate the robustness of inverse loss.

icm_5_100_auto_noise


What the ICM paper didn't show is that if the noise/stochasticity is dependent on the actions. Then the inverse loss doesn't have the pressure of focusing on only the meaningful semantics, because the noise vector can also help predict the action of FIRE. So ICM fails in this case.

PR #191, opened 2019-09-25
add RND algorithm for intrinsic motivation

hnyu:
Added another intrinsic motivation baseline method: 

`Exploration by Random Network Distillation, Burda et al. 2019.`

An arbitrary trial on SuperMario with intrinsic reward only produced the curve attached after 17 hours. Could be better if hyperparameters are tuned carefully.


alf/algorithms/rnd_algorithm.py
```diff
@@ -0,0 +1,77 @@
...
+    Suppose we have a fixed randomly initialized target network g: s -> e_t and
+    a trainable predictor network h: s -> e_p, then the intrinsic reward is
+
+    r = |e_t - e_p|^2
+
+    The reward is expected to be higher for novel states.
+    """
+
+    def __init__(self,
+                 target_net: Network,
```
witwolf: comments for these arguments.

hnyu: added

PR #197, opened 2019-10-02
Add scalar_loss for LossInfo

emailweixu:
Also
1. updated ICMAlorithm to support multiple hidden layers.
2. minor refactoring of ActorCriticAlgorithm so that it's easier to add new sub-algorithm in the future.
3. improved icm_mountain_car.gin so it can train within an hour

### DIFF
--- a/alf/algorithms/icm_algorithm.py
+++ b/alf/algorithms/icm_algorithm.py
@@ -174,4 +174,5 @@ class ICMAlgorithm(Algorithm):
                         inverse_loss=inverse_loss))))
 
     def calc_loss(self, info: ICMInfo):
-        return info.loss
+        loss = tf.nest.map_structure(tf.reduce_mean, info.loss)
+        return LossInfo(scalar_loss=loss.loss, extra=loss.extra)
--- a/alf/algorithms/icm_algorithm.py
+++ b/alf/algorithms/icm_algorithm.py
@@ -11,17 +11,19 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from collections import namedtuple
 
 import gin.tf
 import tensorflow as tf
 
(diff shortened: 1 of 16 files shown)


alf/algorithms/icm_algorithm.py
```diff
@@ -134,9 +163,15 @@ def train_step(self, inputs, state, calc_intrinsic_reward=True):
...
-                    forward_loss=forward_loss, inverse_loss=inverse_loss)))
+            info=ICMInfo(
+                reward=intrinsic_reward,
+                loss=LossInfo(
+                    loss=forward_loss + inverse_loss,
+                    extra=dict(
+                        forward_loss=forward_loss,
+                        inverse_loss=inverse_loss))))
+
+    def calc_loss(self, info: ICMInfo):
```
hnyu: should avoid masking out across-episode transitions by converting the loss to scalar. Maybe define a wrapper called mask_invariant_loss()

alf/algorithms/rnd_algorithm.py
```diff
@@ -102,4 +104,7 @@ def train_step(self, inputs, state):
         return AlgorithmStep(
             outputs=intrinsic_reward,
```
hnyu: outputs should be ()

alf/utils/summary_utils.py
```diff
@@ -221,3 +221,32 @@ def summarize_action_dist(action_distributions,
...
+def add_mean_hist_summary(name, value):
+    """Generate mean and histogram summary of value.
+
+    Args:
+        name (str): name of the summary
+        value (Tensor): tensor to be summarized
+    Returns:
+        None
+    """
+    tf.summary.histogram(name + "/value", value)
```
hnyu: This function can call the one below.

alf/algorithms/icm_algorithm.py
emailweixu: added reduce_mean

PR #198, opened 2019-10-03
Pr update tf_agents as submodule to allow encoding networks in actor and value networks

le-horizon:
This change updates our dependency to tensorflow-gpu 2.0.0b1, which is the highest version so far.

Our docker image on gpu cluster has the following setup: Py3.6  cudnn7.4  cuda10.0  Nvidia Driver 410.93.  I've tested on that and training works.  Make sure you have a compatible env.  I've also tested on cudnn7.6, nvidia 418, which also works.

Motivation:
The reason for this change is we want encoding network in actor and value networks so we can have mixed inputs (e.g. language + image) to a & c networks, but that needs newer version of tf2. So this change just upgrades to tf2 and updates submodule pointer of tf_agents to the version after the encoding network enhancement.

### DIFF
--- a/.ci-cd/requirements.txt
+++ b/.ci-cd/requirements.txt
@@ -1,5 +1,6 @@
-tfp-nightly==0.8.0.dev20190609
 tf-agents-nightly==0.2.0.dev20190415
+tensorflow==2.0.0rc1
+tfp-nightly==0.8.0.dev20190609
 numpy
 opencv-python>=3.4.1.15
 gym
--- a/.ci-cd/requirements.txt
+++ b/.ci-cd/requirements.txt
@@ -1,6 +1,5 @@
-tf-agents-nightly==0.2.0.dev20190415
-tensorflow==2.0.0rc1
-tfp-nightly==0.8.0.dev20190609
+tensorflow==2.0.0
+tensorflow-probability==0.8.0
 numpy
 opencv-python>=3.4.1.15
 gym
--- a/alf/drivers/on_policy_driver.py
+++ b/alf/drivers/on_policy_driver.py
@@ -168,7 +168,7 @@ class OnPolicyDriver(policy_driver.PolicyDriver):
             loop_vars=[time_step, policy_state],
             maximum_iterations=maximum_iterations,
             back_prop=False,
-            name="driver_loop")
+            name="")
         return time_step, policy_state
 
     def _train_loop_body(self, counter, time_step, policy_state,
--- a/alf/drivers/on_policy_driver.py
+++ b/alf/drivers/on_policy_driver.py
@@ -272,4 +272,4 @@ class OnPolicyDriver(policy_driver.PolicyDriver):
 
         self._train_step_counter.assign_add(1)
 
-        return next_time_step, next_state
+        return [next_time_step, next_state]
--- a/alf/drivers/policy_driver.py
+++ b/alf/drivers/policy_driver.py
@@ -75,7 +75,7 @@ class PolicyDriver(driver.Driver):
             tf_metrics.EnvironmentSteps(),
         ]
         # This is a HACK.
-        # Due to tensorflow metric API change:
+        # Due to tf_agents metric API change:
         # https://github.com/tensorflow/agents/commit/b08a142edf180325b63441ec1b71119c393c4a64,
         # after tensorflow v20190807, these two metrics will cause error during
         # playing of the trained model, when num_parallel_envs > 1.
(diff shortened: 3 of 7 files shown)


alf/drivers/policy_driver.py
```diff
@@ -73,9 +73,21 @@ def __init__(self,
...
             tf_metrics.NumberOfEpisodes(),
             tf_metrics.EnvironmentSteps(),
-            tf_metrics.AverageReturnMetric(buffer_size=metric_buf_size),
-            tf_metrics.AverageEpisodeLengthMetric(buffer_size=metric_buf_size),
         ]
+        # This is a HACK.
+        # Due to tensorflow metric API change:
+        # https://github.com/tensorflow/agents/commit/b08a142edf180325b63441ec1b71119c393c4a64,
+        # after tensorflow v20190807, these two metrics will cause error during
+        # playing of the trained model, when num_parallel_envs > 1.
```
hnyu: For playing with the trained model, should the batch_size be always one?

hnyu:
I don't quite understand what the title of this PR means. Also the code change and comment don't explain well. Can you elaborate on this?

alf/drivers/on_policy_driver.py
```diff
@@ -168,7 +168,7 @@ def train(self, max_num_steps, time_step, policy_state):
             loop_vars=[time_step, policy_state],
             maximum_iterations=maximum_iterations,
             back_prop=False,
-            name="driver_loop")
+            name="")
```
hnyu: why remove the name?

alf/drivers/policy_driver.py
```diff
@@ -82,12 +82,16 @@ def __init__(self,
...
         if training:
-            standard_metrics += [
-                tf_metrics.AverageReturnMetric(
-                    batch_size=env.batch_size, buffer_size=metric_buf_size),
-                tf_metrics.AverageEpisodeLengthMetric(
-                    batch_size=env.batch_size, buffer_size=metric_buf_size),
-            ]
+            batch_size = env.batch_size
+        else:
+            batch_size = 1
```
emailweixu: I think this won't work. The checkpoint need the same batch size for play and train.

alf/drivers/policy_driver.py
```diff
@@ -73,9 +73,21 @@ def __init__(self,
         standard_metrics = [
             tf_metrics.NumberOfEpisodes(),
             tf_metrics.EnvironmentSteps(),
-            tf_metrics.AverageReturnMetric(buffer_size=metric_buf_size),
-            tf_metrics.AverageEpisodeLengthMetric(buffer_size=metric_buf_size),
         ]
+        # This is a HACK.
+        # Due to tensorflow metric API change:
```
emailweixu: tensorflow => tf-agents

alf/drivers/policy_driver.py
```diff
@@ -73,9 +73,21 @@ def __init__(self,
...
+        # https://github.com/tensorflow/agents/commit/b08a142edf180325b63441ec1b71119c393c4a64,
+        # after tensorflow v20190807, these two metrics will cause error during
```
emailweixu: tensorflow tf-agents

alf/drivers/policy_driver.py
le-horizon: Done.

alf/drivers/policy_driver.py
le-horizon: Good point.  Tried and true.

.ci-cd/requirements.txt
```diff
@@ -1,5 +1,6 @@
-tfp-nightly==0.8.0.dev20190609
 tf-agents-nightly==0.2.0.dev20190415
```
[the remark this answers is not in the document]
le-horizon: Removed.  As we are using submodule.

alf/drivers/policy_driver.py
hnyu: OK, I thought num_parallel_envs > 1 is for playing. What you really mean is that if during training we have num_parallel_envs > 1 and we have num_parallel_envs==1 for playing, after this update, then these two metrics will not work.

hnyu:
> This change updates our dependency to tensorflow-gpu 2.0.0b1, which is the highest version so far.
> 
> Our docker image on gpu cluster has the following setup: Py3.6 cudnn7.4 cuda10.0 Nvidia Driver 410.93. I've tested on that and training works. Make sure you have a compatible env. I've also tested on cudnn7.6, nvidia 418, which also works.
> 
> Motivation:
> The reason for this change is we want encoding network in actor and value networks so we can have mixed inputs (e.g. language + image) to a & c networks, but that needs newer version of tf2. So this change just upgrades to tf2 and updates submodule pointer of tf_agents to the version after the encoding network enhancement.

Should the version be "tensorflow-gpu==2.0.0" or "tensorflow-gpu==2.0.0b1". Seems like that the stable release is the former.

PR #200, opened 2019-10-04
ActorCriticAlgorithm refactoring: Agent

emailweixu:
The purpose of this PR is to make it easier to apply sub-algorithms for different RL algorithms.
To do this, I make ActorCriticAglrotihm a pure actor critic algorithm and add an Agent class for integrating different algorithms.
Agent is responsible to use other algorithms to get intrinsic reward, learn internal representation, generate goals, etc. It will use a rl algorithm to learn policy.

### DIFF
--- a/alf/algorithms/actor_critic_algorithm.py
+++ b/alf/algorithms/actor_critic_algorithm.py
@@ -13,59 +13,37 @@
 # limitations under the License.
 """Actor critic algorithm."""
 
-import functools
-from typing import Callable
-
 import gin.tf
-
 import tensorflow as tf
 
 from tf_agents.networks.actor_distribution_network import ActorDistributionNetwork
 from tf_agents.networks.actor_distribution_rnn_network import ActorDistributionRnnNetwork
-from tf_agents.networks.encoding_network import EncodingNetwork
 from tf_agents.networks.network import Network, DistributionNetwork
 from tf_agents.networks.value_network import ValueNetwork
 from tf_agents.networks.value_rnn_network import ValueRnnNetwork
 from tf_agents.trajectories.policy_step import PolicyStep
 
 from alf.algorithms.actor_critic_loss import ActorCriticLoss
-from alf.algorithms.algorithm import Algorithm
-from alf.algorithms.entropy_target_algorithm import EntropyTargetAlgorithm
-from alf.algorithms.icm_algorithm import ICMAlgorithm
 from alf.algorithms.on_policy_algorithm import OnPolicyAlgorithm
-from alf.algorithms.rl_algorithm import ActionTimeStep, TrainingInfo, LossInfo, namedtuple
-
-import sys
+from alf.algorithms.rl_algorithm import ActionTimeStep, TrainingInfo, namedtuple
 
 ActorCriticState = namedtuple(
-    "ActorCriticPolicyState", ["actor", "value", "icm"], default_value=())
+    "ActorCriticState", ["actor", "value"], default_value=())
 
-ActorCriticInfo = namedtuple(
-    "ActorCriticInfo", ["value", "icm", "entropy_target"], default_value=())
-
-ActorCriticAlgorithmLossInfo = namedtuple(
-    "ActorCriticAlgorithmLossInfo", ["ac", "icm", "entropy_target"],
-    default_value=())
+ActorCriticInfo = namedtuple("ActorCriticInfo", ["value"])
 
 
 @gin.configurable
 class ActorCriticAlgorithm(OnPolicyAlgorithm):
+    """Actor critic algorithm"""
+
     def __init__(self,
                  action_spec,
                  actor_network: DistributionNetwork,
                  value_network: Network,
-                 encoding_network: Network = None,
-                 intrinsic_curiosity_module=None,
-                 intrinsic_reward_coef=1.0,
-                 extrinsic_reward_coef=1.0,
-                 enforce_entropy_target=False,
                  loss=None,
                  loss_class=ActorCriticLoss,
                  optimizer=None,
-                 gradient_clipping=None,
-                 clip_by_global_norm=False,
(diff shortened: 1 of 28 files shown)


alf/drivers/on_policy_driver.py
```diff
@@ -174,7 +174,7 @@ def train(self, max_num_steps, time_step, policy_state):
     def _train_loop_body(self, counter, time_step, policy_state,
                          training_info_ta):
 
-        next_time_step, policy_step, action = self._step(
+        next_time_step, policy_step, action, time_step = self._step(
```
hnyu: Should better name `time_step` here as `transformed_time_step`

alf/drivers/on_policy_driver.py
```diff
@@ -227,12 +227,12 @@ def create_ta(s):
...
+            next_time_step, policy_step, action, time_step = self._step(
                 time_step, policy_state)
             next_state = policy_step.state
         else:
+            time_step = self._algorithm.transform_timestep(time_step)
             policy_step = common.algorithm_step(self._algorithm.rollout,
-                                                self._observation_transformer,
                                                 time_step, policy_state)
             action = common.sample_action_distribution(policy_step.action)
             next_time_step = time_step
```
hnyu: For the two different branches, `next_time_step` is not consistently transformed.  I believe that nontransformed `next_time_step` should be returned.

alf/drivers/policy_driver.py
```diff
@@ -206,11 +206,9 @@ def _step(self, time_step, policy_state):
         step_func = self._algorithm.rollout if self._training else (
             self._algorithm.greedy_predict
             if self._greedy_predict else self._algorithm.predict)
-        policy_step = common.algorithm_step(
-            step_func,
-            self._observation_transformer,
-            time_step,
-            state=policy_state)
+        transformed_time_step = self._algorithm.transform_timestep(time_step)
```
hnyu: It seems that here returning transformed_time_step is only for keeping the shaped rewards info for on_policy_driver. It this is True, I think this additional return adds unnecessary complication to the code structure.

alf/algorithms/rl_algorithm.py
```diff
@@ -347,13 +379,6 @@ def train_complete(self,
         valid_masks = tf.cast(
             tf.not_equal(training_info.step_type, StepType.LAST), tf.float32)
 
-        # reward shaping
-        if self._reward_shaping_fn is not None:
-            # record unshaped extrinsic rewards given by the environment
-            self.add_reward_summary("reward/raw", training_info.reward)
-            training_info = training_info._replace(
-                reward=self._reward_shaping_fn(training_info.reward))
-
```
hnyu: This "raw" reward is no longer summarized anywhere.

alf/algorithms/actor_critic_algorithm.py
```diff
@@ -78,213 +56,68 @@ def __init__(self,
...
+        """Rollout for one step."""
         value, value_state = self._value_network(
-            observation,
+            time_step.observation,
             step_type=time_step.step_type,
             network_state=state.value)
         # ValueRnnNetwork will add a time dim to value
         # See value_rnn_network.py L153
         if isinstance(self._value_network, ValueRnnNetwork):
             value = tf.squeeze(value, axis=1)
```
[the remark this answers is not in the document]
emailweixu: I don't change this. Since github does report conflict, I think it will be automatically removed after merge.

alf/algorithms/agent.py
```diff
@@ -0,0 +1,266 @@
...
+
+AgentLossInfo = namedtuple(
+    "AgentLossInfo", ["rl", "icm", "entropy_target"], default_value=())
+
+
+@gin.configurable
+class Agent(OnPolicyAlgorithm):
+    """Agent
+
+    Agent is a master algorithm that integrate different different algorithms
```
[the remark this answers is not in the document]
emailweixu: fixed

alf/algorithms/agent.py
```diff
@@ -0,0 +1,266 @@
...
+        """
+        reward = external_reward
+        if self._extrinsic_reward_coef != 1.0:
+            reward *= self._extrinsic_reward_coef
+
+        if self._icm is not None:
+            self.add_reward_summary("reward/icm", info.icm.reward)
+            reward += self._intrinsic_reward_coef * info.icm.reward
+
+        if id(reward) != id(external_reward):
```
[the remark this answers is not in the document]
emailweixu: I think we don't need to consider that case. It does not hurt to display "overall" in the case. And it might be helpful by reminding that reward is scaled.

alf/drivers/on_policy_driver.py
emailweixu: Changed.

alf/drivers/on_policy_driver.py
emailweixu: good catch

alf/drivers/policy_driver.py
emailweixu: It's possible that tranform_timestep may do something else to the time_step in the future. And I cannot think of better way of doing it.

alf/algorithms/rl_algorithm.py
emailweixu: Part of the reason for removing it is that I don't find a good place to do it. And part of reason is that it is not that useful.

alf/drivers/on_policy_driver.py
```diff
@@ -231,9 +231,10 @@ def create_ta(s):
                 time_step, policy_state)
```
hnyu: The returned time_step should also be transformed_time_step. And when final_training_info is created below, the time_step should be transformed_time_step to make the two branches consistent.

emailweixu: fixed

PR #202, opened 2019-10-04
improve grid_search to support gpu allocation

hnyu:
The original search script ignores the allocation of gpus; all the jobs would be put on the same device. After this change, we're basically done for single-machine job scheduling and grid search

### DIFF
--- a/alf/bin/grid_search.py
+++ b/alf/bin/grid_search.py
@@ -14,6 +14,10 @@
 
 import itertools
 import json
+import os
+import time
+import random
+from collections import Iterable
 
 from absl import app
 from absl import flags
@@ -21,13 +25,15 @@ from absl import logging
 import gin
 
 # `pathos.multiprocessing` provides a consistent interface with std lib `multiprocessing`
-# and it's more  flexible
+# and it's more flexible
 from pathos import multiprocessing
+from multiprocessing import Queue, Manager
 from alf.bin.train import train_eval
 from alf.utils import common
 
 flags.DEFINE_string('search_config', None,
                     'Path to the grid search config file.')
+
 FLAGS = flags.FLAGS
 r"""Grid search.
 
@@ -51,6 +57,8 @@ class GridSearchConfig(object):
     For example:
     {
         "desc": "desc text",
+        "use_gpu": true,
+        "gpus": [0, 1],
         "max_worker_num": 8,
         "parameters": {
             "ac/Adam.learning_rate": [1e-3, 8e-4],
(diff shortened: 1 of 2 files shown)


alf/bin/grid_search.py
```diff
@@ -77,16 +92,33 @@ def __init__(self, conf_file):
...
+        # check conf keys
+        for k in conf.keys():
+            assert k in self._all_keys_, "Invalid conf key: %s" % k
+
+        self._desc = conf.get('desc', "Grid search")
         self._param_keys = param_keys
         self._param_values = param_values
-        self._max_worker_num = conf['max_worker_num']
+        self._max_worker_num = conf.get('max_worker_num', 1)
+        self._use_gpu = conf.get("use_gpu", False)
```
emailweixu: should default to whether gpu is available.

alf/bin/grid_search.py
```diff
@@ -125,21 +174,34 @@ def run(self):
...
         process_pool.join()
 
-    def _worker(self, root_dir, parameters):
+    def _worker(self, root_dir, parameters, device_queue):
+        # sleep for random seconds to avoid crowded launching
+        time.sleep(random.uniform(0, 3))
+
+        device = device_queue.get()
+        if self._conf.use_gpu:
+            os.environ["CUDA_VISIBLE_DEVICES"] = str(device)
```
emailweixu: Did you verify that changing CUDA_VISIBLE_DEVICES has effect after the process is started?

alf/bin/grid_search.py
```diff
@@ -125,21 +174,34 @@ def run(self):
...
+        device_queue = self._init_device_queue(max_worker_num)
 
         task_count = 0
         for values in itertools.product(*param_values):
             parameters = dict(zip(param_keys, values))
             root_dir = "%s/%d" % (FLAGS.root_dir, task_count)
             process_pool.apply_async(
                 func=self._worker,
-                args=[root_dir, parameters],
+                args=[root_dir, parameters, device_queue],
```
emailweixu: it is simpler to directly pass device_id or job_id instead of device_queue here.

alf/bin/grid_search.py
hnyu: Yes, as long as there hasn't been no long tf message output (e.g., the long message after you create a tensor for the first time)

alf/bin/grid_search.py
hnyu: The problem with that is that the total number of jobs might be bigger than number of workers. So the scheduling is actually dynamic instead of static. I only know which device is available after one worker finishes and put the device back into the queue.

alf/bin/grid_search.py
hnyu: Checking whether gpu is available or not will bring in additional dependency (other lib) here, because if we check gpu with TF here in the beginning, then it will trigger a process so that we cannot set CUDA_VISIBLE_DEVICES later for a subprocess any more.

Issue #193, opened 2019-09-26
inoperative config info seems incorrect


hnyu:
I included "atari.gin" in my gin file to run a job. In TensorBoard "text", I see the following inoperative info:

![Screenshot from 2019-09-26 16-54-34](https://user-images.githubusercontent.com/51248379/65732359-754d1d80-e07e-11e9-9fa3-e9a4cbf0b686.png)

However, as I check these confs are indeed used by my job. Is there any bug here?

witwolf:
<img width=320 src=https://user-images.githubusercontent.com/1267359/65733108-b46f8300-e0ff-11e9-808b-28e9b6547fac.png>

configurations in red rect are using in sub processes (env process ), operative config can not capture them, only record configurations for the main process

do u check  `CategoricalProjectionNetwork.logits_init_output_factor=1e-10` ?

witwolf:
create an eval_env in current process (make sure not wrap env in a new process ) with 
`TrainerConfig.evaluate=True`
`TrainerConfig.eval_interval=INT_MAX`  would be help

hnyu:
> <img alt="" width="320" src="https://user-images.githubusercontent.com/1267359/65733108-b46f8300-e0ff-11e9-808b-28e9b6547fac.png">
> 
> configurations in red rect are used in sub processes (env process ), operative config can not capture them, only record configurations for the main process
> 
> do u check `CategoricalProjectionNetwork.logits_init_output_factor=1e-10` ?

Yes I checked CategoricalProjectionNetwork.logits_init_output_factor. It's 1e-10. For some reason, it's recognized as inoperative.

hnyu:
I found out the reason is that `CategoricalProjectionNetwork` is used in the namescope `alf/`. So actually there is an operative config `actor/CategoricalProjectionNetwork.logits_init_output_factor = 1e-10`. However, the global scope conf is thought to be inoperative.

PR #204, opened 2019-10-05
write operative gin confs to disk

hnyu:
Write a gin configration to a file instead of just copying. Because the user can
    1) manually change the gin confs after loading a conf file into the code, or
    2) include a gin file in another gin file while only the latter might be
       copied to `root_dir`.
So here we just dump the actual used gin conf string to a file. 

We should always write both operative and inoperative confs because for some confs that are indeed used by subprocesses, they are always shown as "inoperative".

### DIFF
--- a/alf/bin/train.py
+++ b/alf/bin/train.py
@@ -76,8 +76,6 @@ def train_eval(root_dir):
 
 def main(_):
     gin_file = common.get_gin_file()
-    if FLAGS.gin_file:
-        common.copy_gin_configs(FLAGS.root_dir, gin_file)
     gin.parse_config_files_and_bindings(gin_file, FLAGS.gin_param)
     train_eval(FLAGS.root_dir)
 
--- a/alf/utils/adaptive_normalizer.py
+++ b/alf/utils/adaptive_normalizer.py
@@ -49,11 +49,13 @@ class AdaptiveNormalizer(EMATensorNormalizer):
         super(AdaptiveNormalizer, self).__init__(
             tensor_spec, norm_update_rate=self._update_ema_rate)
 
-    def normalize(self, tensors):
+    def normalize(self, tensors, clip_value=1.0):
         """Normalized the reward
 
         Args:
             tensors (nested Tensor): tensors to be normalized
+            clip_value (float): the normalized reward will be clipped to +/-
+                this value
         Returns:
             normalized reward. with mean equal to 0 and variance equal to 1
               over the time.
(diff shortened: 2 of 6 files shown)


alf/utils/common.py
```diff
@@ -721,3 +721,67 @@ def get_action_spec():
     """
     assert _env, "set a global env by `set_global_env` before using the function"
     return _env.action_spec()
+
+
+def extract_spec(nest):
```
emailweixu: extract_spec in off_policy_driver should be changed to use this one.

alf/utils/common.py
```diff
@@ -721,3 +721,67 @@ def get_action_spec():
...
+    """
+    Extract tensor spec for each element of a nested structure.
+    It assumes that the first dimension of each element is the batch size.
```
emailweixu: Arguments: and Returns:

alf/bin/train.py
```diff
@@ -76,8 +76,6 @@ def train_eval(root_dir):
 
 def main(_):
     gin_file = common.get_gin_file()
-    if FLAGS.gin_file:
-        common.copy_gin_configs(FLAGS.root_dir, gin_file)
```
emailweixu: This change will make continuing training or playing without gin_file fail.

witwolf: should copy gin files to root_dir , `operative.gin`  do not include all operative config actually (sub processes #193 )

hnyu: > should copy gin files to root_dir , `operative.gin` do not include all operative config actually (sub processes #193 )

I've updated the PR description for the motivation. For the subprocess issue, I think we should create a dummy environment inside the main process.

hnyu: > This change will make continuing training or playing without gin_file fail.

Why? If FLAGS.gin_file is None, then nothing happens. `gin_file` will be globbed inside `root_dir`

emailweixu: You need to copy inoperative configs

alf/utils/adaptive_normalizer.py
```diff
@@ -49,19 +49,24 @@ def __init__(self,
         super(AdaptiveNormalizer, self).__init__(
             tensor_spec, norm_update_rate=self._update_ema_rate)
 
-    def normalize(self, tensors):
+    def normalize(self, tensors, clip_value=1.0):
```
emailweixu: This averager is used in different places. I think default should be no clip.

hnyu: updated. But note that tf_agents does have a default clip value of 5.0 originally. Not it's set to -1.0 (ignored)

PR #205, opened 2019-10-08
Pr target navigation with language and image input and rnn agent

le-horizon:
Hey Jiangtao, Yingxiang and Wei,

This change utilizes encoding network in tf_agents' actor and value networks to process input image and language.

Only tested with ball as input.  Since other people may also need to process mixed input, pushing this change into master to make it easier for folks to experiment.

Could you take a look?

Thanks,
Le


alf/environments/wrappers.py
```diff
@@ -27,24 +27,43 @@
...
         """
         super().__init__(env)
         self._frames = collections.deque(maxlen=stack_size)
         self._channel_order = channel_order
         self._stack_size = stack_size
+        self._field_to_stack = field_to_stack
+
+        raw_space = self.env.observation_space
+        if field_to_stack:
+            assert hasattr(raw_space, 'spaces'), 'observation is not dict'
```
emailweixu: assert isinstance(raw_space, gym.space.Dict)

alf/environments/wrappers.py
```diff
@@ -27,24 +27,43 @@
...
+            assert raw_space.spaces[field_to_stack], (
```
emailweixu: If field_to_stack is not in raw_space, I think this will raise exception.
You can use `raw_space.contains(field_to_stack)`

alf/environments/wrappers.py
```diff
@@ -27,24 +27,43 @@
...
+                '{} not in observation.spaces'.format(field_to_stack))
+            space = raw_space.spaces[field_to_stack]
```
emailweixu: space =raw_space[field_to_stack]

alf/environments/wrappers.py
```diff
@@ -53,18 +72,40 @@ def __init__(self, env, stack_size=4, channel_order='channels_last'):
             shape = [
                 stack_size * space.shape[0],
             ] + list(space.shape[1:])
-        self.observation_space = gym.spaces.Box(
+        box_space = gym.spaces.Box(
```
emailweixu: should use the 'high' and 'low' from the original space

alf/environments/wrappers.py
```diff
@@ -53,18 +72,40 @@ def __init__(self, env, stack_size=4, channel_order='channels_last'):
...
             low=0, high=255, shape=shape, dtype=np.uint8)
+        if field_to_stack:
+            self.observation_space.spaces[field_to_stack] = box_space
```
emailweixu: self.observation_space does not exist before this line.

alf/environments/wrappers.py
```diff
@@ -27,24 +27,43 @@
...
+                space corresponding to the key in FrameStack.
         """
         super().__init__(env)
         self._frames = collections.deque(maxlen=stack_size)
         self._channel_order = channel_order
         self._stack_size = stack_size
+        self._field_to_stack = field_to_stack
+
+        raw_space = self.env.observation_space
+        if field_to_stack:
```
emailweixu: if field_to_stack is not specified, we should stack all items.

alf/environments/wrappers.py
```diff
@@ -27,24 +27,43 @@
...
+            field_to_stack (str): optional name of the key to the
+                Dict env.observation_space.  If specified, only use the
+                space corresponding to the key in FrameStack.
         """
         super().__init__(env)
         self._frames = collections.deque(maxlen=stack_size)
         self._channel_order = channel_order
         self._stack_size = stack_size
+        self._field_to_stack = field_to_stack
+
```
emailweixu: I feel that the code can be simplified by using a list of indices of the spaces that require stacking.

alf/examples/ac_target_navigation.gin
```diff
@@ -0,0 +1,87 @@
...
+observation_spec=@get_observation_spec()
+action_spec=@get_action_spec()
+vocab_size=@get_vocab_size()  # must have sentence as part of observation
+
+num_embedding_dims=16
+fc_layer_params=(256,)
+get_conv_layers.conv_layer_params=((16, 3, 2), (32, 3, 2))
+conv_layers = @get_conv_layers()
+sentence/generate_network.structure=(
+    'tf.keras.Sequential',
```
emailweixu: Do you think @layers.Sequential, @layers.Embedding are better than string?

alf/utils/common.py
```diff
@@ -714,3 +719,79 @@ def get_action_spec():
...
+
+@gin.configurable
+def get_vocab_size():
+    """Get the vocabulary size of observations provided by the global environment.
+
+    Returns:
+      vocab_size (int): size of the environment's/teacher's vocabulary
+    """
+    assert _env, "set a global env by `set_global_env` before using the function"
+    return _env.observation_spec()['sentence'].shape[0]
```
emailweixu: report error if observation_spec is not correct.

alf/utils/common.py
```diff
@@ -714,3 +719,79 @@ def get_action_spec():
...
+
+
+@gin.configurable
+def get_conv_layers(conv_layer_params):
```
emailweixu: comments

le-horizon:
Dear reviewers,

Tests are passing.  Please take a look and see if there is anything you want me addressed in this PR.

Thanks,
Le

alf/environments/wrappers.py
le-horizon: changed as recommended.

alf/environments/wrappers.py
le-horizon: Using a list of the keys now.

alf/environments/wrappers.py
le-horizon: Done.

alf/utils/common.py
```diff
@@ -878,6 +893,24 @@ def get_conv_layers(conv_layer_params):
...
+    Args:
+      input_shape (list of ints): input tensor shape.
+      batch_size (int): batch size.
+
+    Returns:
+      A network that passes input directly to output.
+    """
+    x = tf.keras.Sequential()
+    x.add(
+        tf.keras.layers.Lambda(
```
emailweixu: Why not just return layers.Lambda(...)

alf/utils/common.py
```diff
@@ -878,6 +893,24 @@ def get_conv_layers(conv_layer_params):
...
+            lambda x: x + 0, input_shape=input_shape, batch_size=batch_size))
```
emailweixu: why do you need to pass input_shape and batch_size? They are not the required arguments for Lambda()

alf/environments/wrappers.py
```diff
@@ -27,44 +29,144 @@
...
+            result = collections.OrderedDict()
+            for frame in self._frames:
+                for key, field in frame.items():
+                    if key not in result:
+                        result[key] = collections.deque(
+                            maxlen=self._stack_size)
+                    result[key].append(field)
+            for key in result:
+                if key in self._fields_to_stack:
+                    if (key == 'image'
```
emailweixu: This can be a problem if the key is not "image" but still a image. It's probably better to always respect _channel order

alf/environments/wrappers.py
```diff
@@ -27,44 +29,144 @@
...
+                        stacked_space = gym.spaces.Box(
+                            low=np.array(low),
+                            high=np.array(high),
+                            dtype=sp.dtype)
+
+                    elif name == 'sentence':
+                        assert isinstance(sp, gym.spaces.MultiDiscrete)
+                        nvec = [
+                            stack_size * sp.nvec[0],
+                        ] + list(sp.nvec[1:])
```
emailweixu: should be: nvec = sp.nvec * stack_size

alf/environments/wrappers.py
```diff
@@ -27,44 +29,144 @@
...
+                        stacked_space = gym.spaces.MultiDiscrete(nvec)
+
+                    else:
+                        assert name not in self._white_list_fields, (
```
emailweixu: why do we need to have white_list_fields?

alf/environments/wrappers.py
```diff
@@ -27,44 +29,144 @@
 class FrameStack(gym.Wrapper):
     """Stack previous `stack_size` frames, applied to Gym env."""
 
-    def __init__(self, env, stack_size=4, channel_order='channels_last'):
+    def __init__(
+            self,
+            env,
+            stack_size=4,
+            channel_order='channels_last',
+            fields_to_stack=None,
```
emailweixu: Perhaps you can use the similar way as common.image_scale_transformer to handle the field name to make it more general in the case of nested observation.
https://github.com/HorizonRobotics/alf/blob/d8464e3d16d62309f3e3b64d8a2f57fc21382cbc/alf/utils/common.py#L438

alf/environments/utils.py
```diff
@@ -155,8 +155,8 @@ def create_environment(env_name='CartPole-v0',
     """
     if nonparallel:
         # Each time we can only create one unwrapped env at most
-        py_env = env_load_fn(env_name)
-        py_env.seed(np.random.randint(0, np.iinfo(np.int32).max))
+        py_env = env_load_fn(env_name, wrap_with_process=True)
```
emailweixu: It's fixed by #244

alf/drivers/policy_driver.py
```diff
@@ -125,17 +125,17 @@ def _training_summary(self, training_info, loss_info, grads_and_vars):
                                                   self._train_step_counter)
         if self._debug_summaries:
             common.add_action_summaries(training_info.action,
-                                        self.env.action_spec())
+                                        self._env.action_spec())
```
emailweixu: why make this change?

le-horizon: You are right, no need for this.  I misread the code.

alf/environments/utils.py
le-horizon: Cool

alf/drivers/policy_driver.py
le-horizon: This file is a bit confusing, sometimes it uses .env sometimes ._env.  I was trying to make it consistent.

le-horizon:
Hey Wei,

Thanks for the suggestions and discussions.

I made several refactors: moved network architecture out of gin file, refactored FrameStack to support nested fields specified by paths.

Please take another look.

Thanks,
Le

alf/utils/common.py
emailweixu: You can simply use tf_agents.networks.utils.mlp_layers

alf/utils/common.py
```diff
@@ -864,3 +890,42 @@ def write_gin_configs(root_dir, gin_file):
...
+                glorot_uniform(),
+                dtype=tf.float32,
+                name='ConvLayer'))
+    layers.append(tf.keras.layers.Flatten())
+    image_processing_layers = tf.keras.Sequential(layers)
+    return image_processing_layers
+
+
+@gin.configurable
+def get_identity_layer():
```
emailweixu: Put it in alf/layers.py. And we make it have same usage as other keras layers.
```python
@gin.configurable
class IdentityLayer(tf.keras.layers.Lambda):
    def __init__(self):
         super().__init__(lambda x: x)
```

alf/networks/target_navigation.py
```diff
@@ -0,0 +1,115 @@
...
+from alf.utils import common
+
+
+def get_ac_networks(conv_layer_params=None,
+                    num_embedding_dims=None,
+                    fc_layer_params=None,
+                    num_state_tiles=None,
+                    num_sentence_tiles=None):
+    """
+    Generate the actor and value networks
```
emailweixu: comment arguments. explain why using tiles

alf/environments/wrappers.py
```diff
@@ -27,44 +29,118 @@
 class FrameStack(gym.Wrapper):
     """Stack previous `stack_size` frames, applied to Gym env."""
 
-    def __init__(self, env, stack_size=4, channel_order='channels_last'):
+    def __init__(
```
emailweixu: As this class is so simple anymore, we need some unittest

alf/environments/wrappers_test.py
```diff
@@ -0,0 +1,75 @@
...
+        assert (
+            obs['image'].shape,
+            obs['states'].shape,
+        ) == (
+            (2, 2, 3 * 4),  # 3 channels * 4
+            (4 * 4, ),  # 4 dimensions * 4 == 16
+        )
+        env.close()
+
+    def test_framestack_some_fields(self):
```
emailweixu: should not use socialbot to test this. It's better to use a fake environment so that it does not depend on socialbot

alf/networks/target_navigation.py
```diff
@@ -0,0 +1,142 @@
...
+            inputs like image observation.
+    """
+    observation_spec = common.get_observation_spec()
+    action_spec = common.get_action_spec()
+
+    conv_layers = tf.keras.Sequential(
+        tf_agents.networks.utils.mlp_layers(
+            conv_layer_params=conv_layer_params,
+            activation_fn=tf.keras.activations.softsign,
+            kernel_initializer=tf.compat.v1.keras.initializers.
```
emailweixu: make these two configurable.

alf/environments/wrappers.py
```diff
@@ -27,44 +30,161 @@
...
+
+            elif isinstance(sp, gym.spaces.MultiDiscrete):
+                nvec = [sp.nvec] * stack_size
+                stacked_space = gym.spaces.MultiDiscrete(nvec)
+
+            else:
+                stacked_space = sp
+            return stacked_space
+
+        def _traverse(d, fields_to_stack=None, prefix=""):
```
emailweixu: It seems that this function will change the original space of wrapped the environment. We should make a copy so that the space of the wrapped environment will not be changed.

alf/environments/wrappers.py
```diff
@@ -27,44 +30,161 @@
...
+                    transformed, fields_to_stack = _traverse(
+                        sp, fields_to_stack, path)
+                else:
+                    if fields_to_stack is None or path in fields_to_stack:
+                        transformed = _stack_space(sp)
+                        if fields_to_stack:
+                            fields_to_stack = [
+                                item for item in fields_to_stack
+                                if item != path
+                            ]
```
emailweixu: why need L105-L109?

alf/environments/wrappers.py
le-horizon: Good catch!  Done, and this will break the new test added.

alf/environments/wrappers.py
le-horizon: This is to compute the remaining fields that we haven't stacked, for verification purposes, so that all fields specified in fields_to_stack have to appear in input obs.

alf/networks/target_navigation.py
le-horizon: Done.

alf/examples/ac_target_navigation.gin
```diff
@@ -36,6 +36,9 @@ fc_layer_params=(256,)
 num_state_tiles=800
 num_sentence_tiles=800
 
+mlp_layers.activation_fn=@tf.keras.activations.softsign
+mlp_layers.kernel_initializer=@tf.compat.v1.keras.initializers.glorot_uniform()
```
emailweixu: It it possible not to use compat.v1 version?

le-horizon: Done.

alf/examples/ac_target_navigation.gin
```diff
@@ -37,7 +37,7 @@ num_state_tiles=800
 num_sentence_tiles=800
 
 mlp_layers.activation_fn=@tf.keras.activations.softsign
-mlp_layers.kernel_initializer=@tf.compat.v1.keras.initializers.glorot_uniform()
+mlp_layers.kernel_initializer=@tf.keras.initializers.GlorotUniform
```
emailweixu: Missing ()?: GlorotUniform()

le-horizon: You are right!  Fixed.  This is scary, the code can still run without complaint..

Issue #206, opened 2019-10-08
ParallelPyEnvironment will cause sys.exit() to hang after main() finishes


hnyu:
I've tested for quite some time to pin down an issue (reason unknown at this moment) that once we created tf_agents' ParallelPyEnvironment in main(), after the code finishes, sys.exit() will hang forever without releasing the GPU memory (running in CPU mode also has this hanging issue). This makes grid_search.py fail. The minimal re-producible example is to replace the `train_eval(root_dir)` function with the following:

```
@gin.configurable
def train_eval(root_dir):
    env = create_environment()
    env.pyenv.close()
``` 
and run any training job. The script will never exit. 

And I have located the issue exactly at sys.exit(main(argv)) in absl's app.py. It's more like an issue happened inside sys.exit() rather than in absl's code. 

@witwolf  When you ran grid_search.py, did you ever have such a problem?

witwolf:
what are the parameters (gin configurations )  used for create_environment ?
There is indeed a problem with `wrap_with_process=True` , but it only happen with `suite_socialbot.load`  and  `suite_mario.load` when `num_parallel_environments=1`

witwolf:
> what are the parameters (gin configurations ) used for create_environment ?
> There is indeed a problem with `wrap_with_process=True` , but it only happen with `suite_socialbot.load` and `suite_mario.load` when `num_parallel_environments=1`

The innermost process is not joined ,i will fixed this

hnyu:
> what are the parameters (gin configurations ) used for create_environment ?
> There is indeed a problem with `wrap_with_process=True` , but it only happen with `suite_socialbot.load` and `suite_mario.load` when `num_parallel_environments=1`

Yes, the issue only happens if wrap_with_process=True (I also set it to True for num_parallel_environments>1). After I set it to False, then it works. Thanks for finding out the reason.

PR #207, opened 2019-10-08
close env in ProcessPyEnvironment when receive _CLOSE

witwolf:
#206 

`ParallelPyEnvironment` do not close envs when it is closed,    it may become a problem when envs are wrapped in process, those processed will not be terminated 

This pr overwrites `_worker`  , close env when receive `_CLOSE` (just add one line code)

### DIFF
--- a/alf/environments/suite_socialbot.py
+++ b/alf/environments/suite_socialbot.py
@@ -45,7 +45,7 @@ class ProcessPyEnvironment(parallel_py_environment.ProcessPyEnvironment):
             env_constructor, flatten=flatten)
 
     def _worker(self, conn, env_constructor, flatten=False):
-        """It's a little different with `super()._worker`, it close environment when
+        """It's a little different with `super()._worker`, it closes environment when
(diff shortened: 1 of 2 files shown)


emailweixu:
I also suggest to send a PR to tf_agent

alf/environments/suite_socialbot.py
```diff
@@ -40,6 +44,60 @@ def __init__(self, env_constructor, flatten=False):
         super(ProcessPyEnvironment, self).__init__(
             env_constructor, flatten=flatten)
 
+    def _worker(self, conn, env_constructor, flatten=False):
+        """It's a little different with `super()._worker`, it close environment when
```
emailweixu: "it closs" => it closes

witwolf: fixed

witwolf:
> I also suggest to send a PR to tf_agent

Ok, i'll send a PR to `tf_agent`

PR #208, opened 2019-10-08
RLAlgorithm is Algorithm

emailweixu:
The purpose for this is to subclass RLAlgorithm from Algorithm so that we have one implementation of train_complete. With the refactoring, each algorithm can have zero, one or multiple optimizers. Each optimizer is responsible for optimizing a set of modules. If the optimizer for an algorithm is not defined, it is to be optimized by the optimizer of its parent algorithm.

### DIFF
--- a/alf/algorithms/agent.py
+++ b/alf/algorithms/agent.py
@@ -13,10 +13,9 @@
 # limitations under the License.
 """Agent for integrating multiple algorithms."""
 
-import functools
 from typing import Callable
 
-import gin.tf
+import gin
 import tensorflow as tf
 
 from tf_agents.networks.network import Network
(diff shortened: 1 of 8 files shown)


alf/utils/common.py
```diff
@@ -13,23 +13,88 @@
...
+        discount=time_step.discount,
+        observation=time_step.observation,
+        prev_action=prev_action)
+
+
+LossInfo = namedtuple(
+    "LossInfo",
+    [
+        "loss",  # batch loss shape should be (T, B)
+        "scalar_loss",  # shape is ()
```
witwolf: scalar_loss is () or (B,)  [scalar_loss for  icm , rnd .. are (B,) ]

alf/algorithms/agent.py
```diff
@@ -85,58 +84,30 @@ def __init__(self,
                 immediate rewards
             observation_transformer (Callable): transformation applied to
                 `time_step.observation`
```
witwolf: L69 typo  "intrinsid=>intrinsic"

alf/algorithms/agent.py
```diff
@@ -13,10 +13,9 @@
 # limitations under the License.
 """Agent for integrating multiple algorithms."""
 
-import functools
 from typing import Callable
 
-import gin.tf
+import gin
 import tensorflow as tf
```
witwolf: L41  Maybe its more clear that `Agent ` inherit from  `RLAlgorithm` instead of `OnPolicyAlgorithm`

alf/algorithms/algorithm.py
```diff
@@ -154,25 +290,51 @@ def train_complete(self,
...
             loss_info (LossInfo): loss information
             grads_and_vars (list[tuple]): list of gradient and variable tuples
         """
-
         with tape:
             loss_info = self.calc_loss(training_info)
             if valid_masks is not None:
                 loss_info = tf.nest.map_structure(
-                    lambda l: tf.reduce_mean(l * valid_masks), loss_info)
+                    lambda l: tf.reduce_mean(l * valid_masks)
```
witwolf: when is the shape not equal to 2 ,  and why is the `else` branch doing the same thing

alf/algorithms/algorithm.py
```diff
@@ -78,13 +115,111 @@ def __init__(self,
...
+        self._gradient_clipping = gradient_clipping
+        self._clip_by_global_norm = clip_by_global_norm
         self._debug_summaries = debug_summaries
 
     @property
-    def optimizer(self):
-        """Return the optimizer for this algorithm."""
-        return self._optimizer
+    def trainable_variables(self):
+        return sum([var_set for _, var_set in self._get_opt_and_var_sets()],
```
witwolf: `_get_opt_and_var_sets` can be replaced with  `_get_cached_opt_and_var_sets`

emailweixu: It is intentionally this to have the same behavior as tf.Module.trainable_variables. It is expected to always give the correct result even with after new submodules are added.

alf/algorithms/agent.py
emailweixu: It's indeed OnPolicyAlgorithm because it can be trained using OnPolicyDriver. You can put comments related to Agent to this PR #200

alf/algorithms/algorithm.py
emailweixu: When it's not 2 it's expected to be a scalar.

alf/algorithms/agent.py
emailweixu: Fixed

Issue #209, opened 2019-10-08
Unittest for end-to-end training using alf.bin.train


emailweixu:
Currently we don't have any unittests for this. We should add one. ac_cart_pole.gin might be a good one since it's can be trained within a couple of minutes.

emailweixu:
Fixed by #210

PR #210, opened 2019-10-09
unit test for alf.bin.train

witwolf:
#209

### DIFF
--- a/.ci-cd/Dockerfile.cpu
+++ b/.ci-cd/Dockerfile.cpu
@@ -14,15 +14,7 @@
 
 FROM tensorflow/tensorflow:2.0.0-py3
 
-#
-RUN apt install -y \
-        libsm6  \
-        libxext-dev \
-        libxrender1
-
-# python libs requirement by alf
-COPY requirements.txt  /tmp/requirements.txt
-RUN pip install -i https://mirrors.aliyun.com/pypi/simple/ -r /tmp/requirements.txt
+COPY ./sources.list /etc/apt/sources.list
 
 # install code style tools
 RUN pip3 install -i https://mirrors.aliyun.com/pypi/simple/ \
@@ -32,4 +24,19 @@ RUN pip3 install -i https://mirrors.aliyun.com/pypi/simple/ \
     pylint==2.3.1 \
     yapf==0.28.0
 
-RUN apt install -y  git clang-format
+RUN apt update && apt install -y  git clang-format
+
+RUN apt install -y \
+        libsm6  \
+        libxext-dev \
+        libxrender1 \
+        xvfb
+
+# clean
+RUN  rm -rf /var/lib/apt/lists/*
+
+# python libs requirement by alf
+COPY requirements.txt  /tmp/requirements.txt
+RUN pip install -i https://mirrors.aliyun.com/pypi/simple/ -r /tmp/requirements.txt
+
+
--- a/.ci-cd/Dockerfile.cpu
+++ b/.ci-cd/Dockerfile.cpu
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM tensorflow/tensorflow:2.0.0b1-py3
+FROM tensorflow/tensorflow:2.0.0-py3
 
 #
 RUN apt install -y \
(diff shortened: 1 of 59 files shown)


witwolf:
As we need to get evaluate metrics in some way, i found it's  a easy way that getting it from evaluate tfevents . I'll change this if any better way you suggest.

.ci-cd/Dockerfile.cpu
```diff
@@ -32,4 +24,19 @@ RUN pip3 install -i https://mirrors.aliyun.com/pypi/simple/ \
...
+        libxext-dev \
+        libxrender1 \
+        xvfb
+
+# clean
+RUN  rm -rf /var/lib/apt/lists/*
+
+# python libs requirement by alf
+COPY requirements.txt  /tmp/requirements.txt
+RUN pip install -i https://mirrors.aliyun.com/pypi/simple/ -r /tmp/requirements.txt
```
emailweixu: We do we need to use aliyun?

.ci-cd/sources.list
```diff
@@ -0,0 +1,10 @@
+deb http://mirrors.aliyun.com/ubuntu/ bionic main restricted universe multiverse
+deb http://mirrors.aliyun.com/ubuntu/ bionic-security main restricted universe multiverse
+deb http://mirrors.aliyun.com/ubuntu/ bionic-updates main restricted universe multiverse
+deb http://mirrors.aliyun.com/ubuntu/ bionic-proposed main restricted universe multiverse
+deb http://mirrors.aliyun.com/ubuntu/ bionic-backports main restricted universe multiverse
+deb-src http://mirrors.aliyun.com/ubuntu/ bionic main restricted universe multiverse
+deb-src http://mirrors.aliyun.com/ubuntu/ bionic-security main restricted universe multiverse
+deb-src http://mirrors.aliyun.com/ubuntu/ bionic-updates main restricted universe multiverse
+deb-src http://mirrors.aliyun.com/ubuntu/ bionic-proposed main restricted universe multiverse
+deb-src http://mirrors.aliyun.com/ubuntu/ bionic-backports main restricted universe multiverse
```
emailweixu: what are these for?

alf/bin/train_test.py
```diff
@@ -0,0 +1,131 @@
...
+import subprocess
+from pathlib import Path
+import unittest
+import numpy as np
+
+from absl import logging
+from absl.testing import parameterized
+
+from tensorboard.backend.event_processing import event_file_loader
+from tensorboard.util import tensor_util
```
emailweixu: tb_util might be a better name

alf/bin/train_test.py
```diff
@@ -0,0 +1,131 @@
...
+                    tensor = tensor_util.make_ndarray(item.tensor)
+                    episode_lengths.append(int(tensor))
+
+    logging.info("Episode returns, %s, episode lengths: %s", episode_returns,
+                 episode_lengths)
+    return episode_returns, episode_lengths
+
+
+class TrainTest(parameterized.TestCase, unittest.TestCase):
+    def test_ac_cart_pole(self):
```
le-horizon: It would be useful to have one test that verifies that training can run, e.g. set it to run only one or two iterations.

alf/bin/train_test.py
```diff
@@ -0,0 +1,131 @@
...
+                elif item.tag == 'Metrics/AverageEpisodeLength':
+                    tensor = tensor_util.make_ndarray(item.tensor)
+                    episode_lengths.append(int(tensor))
+
+    logging.info("Episode returns, %s, episode lengths: %s", episode_returns,
+                 episode_lengths)
+    return episode_returns, episode_lengths
+
+
+class TrainTest(parameterized.TestCase, unittest.TestCase):
```
le-horizon: It would be nice to be able to run these tests in parallel.

witwolf:
> What are the time needed for these tests?

less than 20 minutes at ci server

.ci-cd/sources.list
witwolf: apt-get source , it's very very slow to download package from the default mirror source in China

.ci-cd/Dockerfile.cpu
witwolf: it's faster than the default mirror source when install python packages by pip

alf/bin/train_test.py
witwolf: virtual buffer is used for gym environments, it may have problems to run these tests parallelly

alf/bin/train_test.py
witwolf: tensor_util is used for deserialize a tensor from bytes , the name is ok

.ci-cd/sources.list
emailweixu: But the CI server is not in China. Isn't this only used for CI?

witwolf: this is only used to build the image (i build it at k8s server in cn ) , CI server uses pre build image,

PR #211, opened 2019-10-09
fix game inoperative confs

hnyu:
If games are created in subprocesses, then the related gin confs will be always "inoperative", which is quite confusing. This PR fixes this issue by creating an unwrapped env in the main process, and if eval=True, use this env for evaluation.

### DIFF
--- a/alf/environments/suite_dmlab.py
+++ b/alf/environments/suite_dmlab.py
@@ -18,7 +18,7 @@ import gin.tf
 from tf_agents.environments import suite_gym
 from tf_agents.environments import wrappers
 from alf.environments.suite_socialbot import ProcessPyEnvironment
-from alf.environments.utils import UnwrappedEnvChecker
+from alf.environments.suite_socialbot import UnwrappedEnvChecker
 
 _unwrapped_env_checker_ = UnwrappedEnvChecker()
 
--- a/alf/environments/suite_dmlab.py
+++ b/alf/environments/suite_dmlab.py
@@ -19,7 +19,6 @@ from tf_agents.environments import suite_gym
 from tf_agents.environments import wrappers
 from alf.environments.suite_socialbot import ProcessPyEnvironment
 
-
 # This flag indicates whether there has been an unwrapped env in the process
 _unwrapped_env_in_process_ = False
 
--- a/alf/environments/suite_dmlab.py
+++ b/alf/environments/suite_dmlab.py
@@ -18,9 +18,9 @@ import gin.tf
 from tf_agents.environments import suite_gym
 from tf_agents.environments import wrappers
 from alf.environments.suite_socialbot import ProcessPyEnvironment
+from alf.environments.utils import UnwrappedEnvChecker
 
-# This flag indicates whether there has been an unwrapped env in the process
-_unwrapped_env_in_process_ = False
+_unwrapped_env_checker_ = UnwrappedEnvChecker()
 
 # `DeepmindLab` are required,
 #   see `https://github.com/deepmind/lab` to build `DeepmindLab`
@@ -209,10 +209,8 @@ def load(scene,
     Returns:
         A PyEnvironmentBase instance.
     """
-    global _unwrapped_env_in_process_
-    assert not _unwrapped_env_in_process_, \
-        "You cannot create more envs once there has been an env in the main process!"
-    _unwrapped_env_in_process_ |= not wrap_with_process
+    global _unwrapped_env_checker_
+    _unwrapped_env_checker_.check_and_update(wrap_with_process)
 
     if max_episode_steps is None:
         max_episode_steps = 0
--- a/alf/environments/suite_dmlab.py
+++ b/alf/environments/suite_dmlab.py
@@ -22,6 +22,9 @@ from alf.environments.suite_socialbot import ProcessPyEnvironment
 # `DeepmindLab` are required,
 #   see `https://github.com/deepmind/lab` to build `DeepmindLab`
 
+# This flag indicates whether there has been an unwrapped env in the main proc
+_unwrapped_env_in_main_process_ = False
+
 try:
     import deepmind_lab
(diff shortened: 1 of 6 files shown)


emailweixu:
This can not always solve the problem. SocialRobot is always in a subprocess.

alf/environments/utils.py
```diff
@@ -19,33 +19,48 @@
...
-    xarg = {}
-    if env_load_fn in (suite_socialbot.load, env_load_fn == suite_mario.load):
-        # No need to wrap with process if num_parallel_environments > 1,
-        # because ParallelPyEnvironment will do that.
-        xarg = dict(wrap_with_process=num_parallel_environments == 1)
+    global _unwrapped_env_in_main_process_
+
+    wrappable = (env_load_fn in (suite_socialbot.load, suite_mario.load,
+                                 suite_dmlab.load))
+    if wrappable:
```
emailweixu: This logic should belong to load().

alf/trainers/policy_trainer.py
```diff
@@ -217,8 +217,12 @@ def initialize(self):
             not self._use_tf_functions)
         self._env = create_environment()
         common.set_global_env(self._env)
+        # Create an unwrapped env to expose subprocess gin confs which otherwise
+        # will be marked as "inoperative"
+        unwrapped_env = create_environment(force_unwrapped=True)
```
emailweixu: num_parallel_environments=1

alf/environments/utils.py
hnyu: updated

alf/trainers/policy_trainer.py
hnyu: No need to specify that because if num_parallel_environments > 1, then it's definitely wrapped. force_unwrapped=True will only create a single env

alf/environments/utils.py
```diff
@@ -19,33 +19,40 @@
...
 
-    if num_parallel_environments == 1:
+    xarg = dict()
+    if force_unwrapped or num_parallel_environments == 1:
+        if wrappable:
+            xarg = dict(wrap_with_process=not force_unwrapped)
+        # for unwrappable envs, there is no arg called "wrap_with_process"
         py_env = env_load_fn(env_name, **xarg)
     else:
         py_env = parallel_py_environment.ParallelPyEnvironment(
```
emailweixu: should not wrap for num_parallel_environments > 1

hnyu: fixed

hnyu: I have reverted the change of moving the global flag inside load(). The reason is that load might be called by ParallelPyEnvironment with wrap_with_process=False. In this case, there is no env created in the main process but the global flag will be set to True. The flag should really be set only when load() is called in the first branch in create_environment()

hnyu:
Turns out that I don't have to check the flag inside the env_ctor() function. The logic would still hold for checking outside in load(), because ParallelPyEnvironment will create a lambda wrapping load()

witwolf:
can we make `OPERATIVE_CONFIG` , `IMPORTED_MODULES` shared and modifiable between processes  by 
```python
gin.config._OPERATIVE_CONFIG=multiprocessing.Manager.dict()
gin.config._IMPORTED_MODULES=multiprocessing.Manager.dict()
```
and rewrite 
`ProcessPyEnvironment.start`  (pass shared variables to sub processes)

Just an idea, not verified

alf/environments/suite_socialbot.py
```diff
@@ -91,6 +94,10 @@ def load(environment_name,
     Returns:
         A PyEnvironmentBase instance.
     """
+    global _unwrapped_env_in_process_
+    assert not _unwrapped_env_in_process_, \
+        "You cannot create more envs once there has been an env in the main process!"
+    _unwrapped_env_in_process_ |= not wrap_with_process
```
emailweixu: One way to reduce code duplication is to make a class for this. We declare a global object g_wrap_checker for each load module. And calling g_wrap_checker.check() inside the load function.

hnyu:
> can we make `OPERATIVE_CONFIG` , `IMPORTED_MODULES` shared and modifiable between processes by
> 
> ```python
> gin.config._OPERATIVE_CONFIG=multiprocessing.Manager.dict()
> gin.config._IMPORTED_MODULES=multiprocessing.Manager.dict()
> ```
> 
> and rewrite
> `ProcessPyEnvironment.start` (pass shared variables to sub processes)
> 
> Just an idea, not verified

Seems like more complicated than just creating an unwrapped env in the main process.

alf/environments/suite_socialbot.py
hnyu: Changed.

alf/environments/suite_dmlab.py
```diff
@@ -206,6 +209,9 @@ def load(scene,
     Returns:
         A PyEnvironmentBase instance.
     """
+    global _unwrapped_env_checker_
```
emailweixu: no need to have this?

alf/environments/suite_socialbot.py
```diff
@@ -33,6 +33,43 @@
 DEFAULT_SOCIALBOT_PORT = 11345
 
 
+class UnwrappedEnvChecker(object):
```
emailweixu: Should be in environments/utils.py

hnyu: That will create a circular dependency: suite_socialbot -> utils -> suite_socialbot

alf/environments/suite_dmlab.py
hnyu: confirmed with @witwolf , dmlab can only create one env, too

hnyu: removed

alf/environments/suite_socialbot.py
hnyu: changed

PR #213, opened 2019-10-10
improve run names for grid search

hnyu:
Improve the naming of grid search runs


hnyu:
> Did not notice the latest update commit, ignore the comment above

I added truncation of the entire name to some max length.

PR #214, opened 2019-10-10
use tf.test.TestCase for unittests

witwolf:
`tf.test.TestCase` is being [setup](https://github.com/tensorflow/tensorflow/blob/1cf0898dd4331baf93fe77205550f2c2e6c90ee5/tensorflow/python/framework/test_util.py#L1771) to use fixed seed,  make all test cases in ALF inherit from it 

```python
  def setUp(self):
    self._ClearCachedSession()
    random.seed(random_seed.DEFAULT_GRAPH_SEED)
    np.random.seed(random_seed.DEFAULT_GRAPH_SEED)
    # Note: The following line is necessary because some test methods may error
    # out from within nested graph contexts (e.g., via assertRaises and
    # assertRaisesRegexp), which may leave ops._default_graph_stack non-empty
    # under certain versions of Python. That would cause
    # ops.reset_default_graph() to throw an exception if the stack were not
    # cleared first.
    ops._default_graph_stack.reset()  # pylint: disable=protected-access
    ops.reset_default_graph()
    random_seed.set_random_seed(random_seed.DEFAULT_GRAPH_SEED)
    # Reset summary writer in case another test used set_as_default() with their
    # summary writer.
    context.context().summary_writer = None
```

### DIFF
--- a/alf/algorithms/generator_test.py
+++ b/alf/algorithms/generator_test.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 
 import math
-import unittest
 
 from absl import logging
 from absl.testing import parameterized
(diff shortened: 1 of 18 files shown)


emailweixu:
👏

PR #216, opened 2019-10-11
unittest for alf.bin.train

witwolf:
this Pr is from #210 with clean commits


alf/trainers/policy_trainer.py
```diff
@@ -81,7 +84,8 @@ def __init__(self,
                 `async_off_policy_trainer`]
             algorithm_ctor (Callable): callable that create an
                 `OffPolicyAlgorithm` or `OnPolicyAlgorithm` instance
-            random_seed (int): random seed
+            random_seed (None|int): random seed, if None DEFAULT_GRAPH_SEED will
```
emailweixu: Previously, it's expected that different run have different random seed, the the variance of the performance among runs is meaningful. This change causes that different training uses same random seed. I suggest we still default to the old behavior if seed is None. And set a fixed seed for the purpose of unittest.

witwolf: fixed

PR #217, opened 2019-10-12
fix run name to be handled by tf checkpoint

hnyu:
In PR #213 , I added some special chars in the grid search run names. Those chars can be handled correctly by Tensorboard, however, loading checkpoint will have some troubles. So I have to remove special chars.

### DIFF
--- a/alf/bin/grid_search.py
+++ b/alf/bin/grid_search.py
@@ -195,10 +195,10 @@ class GridSearch(object):
                 for key in x:
                     try:
                         val = x.get(key)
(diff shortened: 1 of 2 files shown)


hnyu:
The second commit is because that now regardless of num_envs, it's always ParallelPyEnvironment unless nonparallel=True is specified. So when we play a game, we should set nonparallel=True because the render() will call env.pyenv.envs[0] where envs is only owned by BatchedPyEnvironment but not ParallelPyEnvironment.

Issue #220, opened 2019-10-14
async training currently has gpu memory releasing issue


hnyu:
For grid search, async training cannot effectively release GPU memory after each search. The reason is still under investigation. Generally, whether it's sync or async shouldn't matter, after the fix #112.

hnyu:
Closed due to #221

PR #221, opened 2019-10-14
fix the env resource issue for async training

hnyu:
This PR fixes the issue #220


alf/trainers/policy_trainer.py
```diff
@@ -261,7 +261,12 @@ def train(self):
         self._save_checkpoint()
         if self._evaluate:
             self._eval_env.pyenv.close()
-        self._env.pyenv.close()
+        try:
+            # async off-policy driver has several envs, including self._env
+            for env in self._envs:
+                env.pyenv.close()
+        except:
```
emailweixu: Never use the `except` without specifying what kind of exception you want to capture. This can even capture spelling errors

alf/trainers/policy_trainer.py
```diff
@@ -261,7 +261,12 @@ def train(self):
         self._save_checkpoint()
         if self._evaluate:
             self._eval_env.pyenv.close()
-        self._env.pyenv.close()
+        try:
```
emailweixu: It's better to explicitly test wether self._envs exists

alf/trainers/policy_trainer.py
```diff
@@ -261,7 +261,12 @@ def train(self):
...
+            # async off-policy driver has several envs, including self._env
+            for env in self._envs:
+                env.pyenv.close()
+        except:
+            self._env.pyenv.close()
```
emailweixu: I feel the logic here is highly depending on the logic at off_policy_trainer, which is not good from the point of view of software engineering. We should have a cleaner expectation about self._env and self._envs in the base class  and the subclass follows those expectations.

hnyu:
The third commit is just to remove old observation transformers, which is probably not worth another PR.

PR #224, opened 2019-10-15
set random seed for environments

### LINKED ISSUE #222: TrainTest fails sometimes
======================================================================
FAIL: test_ppo_cart_pole (bin.train_test.TrainTest)
test_ppo_cart_pole (bin.train_test.TrainTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/ALF/alf/bin/train_test.py", line 100, in test_ppo_cart_pole
    self._test_train('ppo_cart_pole.gin', _test_func)
  File "/ALF/alf/bin/train_test.py", line 126, in _test_train
    assert_func(episode_returns, episode_lengths)
  File "/ALF/alf/bin/train_test.py", line 97, in _test_func
    self.assertGreater(np.mean(returns[-2:]), 198)
AssertionError: 197.8499984741211 not greater than 198

@witwolf It seems that the determinism isn't working as expected? There are some other cases failed sometimes, and I have to manually restart the testing job each time.

witwolf:
Fixes #222 

set seeds for environments , these seeds are generated according to  a random seed  `TrainerConfig.random_seed`

### DIFF
--- a/alf/environments/utils.py
+++ b/alf/environments/utils.py
@@ -17,6 +17,7 @@ import traceback
 import tensorflow as tf
 import gin.tf
 from absl import logging
+import numpy as np
 
 from tf_agents.environments import suite_gym, parallel_py_environment, tf_py_environment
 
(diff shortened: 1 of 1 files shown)


alf/environments/utils.py
```diff
@@ -155,10 +156,17 @@ def create_environment(env_name='CartPole-v0',
     if nonparallel:
         # Each time we can only create one unwrapped env at most
         py_env = env_load_fn(env_name)
+        py_env.seed(np.random.randint(0, np.iinfo(np.int32).max))
```
hnyu: So according to my understanding, if np is deterministic, then the seeds for py_env will also be deterministic. The entire training/playing will have deterministic results, right? In issue #222, you mentioned that there are some other possible reasons for non-deterministic unittest results. Should we address those factors?

witwolf: envs  (gym) will behave deterministically when np is deterministic.  This pr does not solve the issue #222 . It just set seed for environment, and make settings as deterministic as possible when seed is set .

PR #225, opened 2019-10-15
entropy regularization for Generator

emailweixu:


alf/algorithms/generator.py
```diff
@@ -36,20 +36,41 @@ class Generator(Algorithm):
...
+
+    * entropy_regularization = 0: the minimization is achieved by simply
+      minimizing loss_func(net([noise, inputs]))
+
+    * entropy_regularization > 0: the minimization is achieved using amortized
+      Stein variational gradient descent (SVGD). See the following paper for
+      reference:
+
+      Feng et al "Learning to Draw Samples with Amortized Stein Variational
+      Gradient Descent" https://arxiv.org/pdf/1707.06626.pdf
```
runjerry: It is a clean way to support both regularized and non-regularized training, and entropy-based regularization does include a large variations of regularization methods. However, I was wondering if we might possibly go beyond entropy regularization in the future? If so, might we want a more abstract generator class?

emailweixu: We can do that when there is need. Otherwise if writing a very general generator now, it's probably won't satisfy the new need when it actually comes.

PR #226, opened 2019-10-15
Conditional ops

emailweixu:
conditional_update and select_from_mask are used for goal generator.

### DIFF
--- a/alf/utils/conditional_ops.py
+++ b/alf/utils/conditional_ops.py
@@ -89,7 +89,7 @@ def run_if(cond, func):
     Args:
         cond (tf.Tensor): scalar bool Tensor
         func (Callable): function to be run
-    Reutrns:
+    Returns:
         None
     """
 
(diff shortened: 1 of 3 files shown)


alf/utils/conditional_ops.py
```diff
@@ -0,0 +1,100 @@
...
+    ```
+    However, when `cond` is tf bool scalar tensor, the above code does not
+    always do what we want because tensorflow does not allow bool scalar tensor
+    to be used in the same way as python bool. So we have to use tf.cond to do
+    the job.
+
+    Args:
+        cond (tf.Tensor): scalar bool Tensor
+        func (Callable): function to be run
+    Reutrns:
```
Haichao-Zhang: Reutrns  -> Returns

alf/utils/conditional_ops.py
```diff
@@ -0,0 +1,100 @@
...
+                tgt, scatter_indices, updt), target, updates)
+
+    total = tf.shape(cond)[0]
+    n = tf.shape(scatter_indices)[0]
+    return tf.case(((n == 0, lambda: target),
+                    (n == total, lambda: func(*args, **kwargs))),
+                   default=_update_subset)
+
+
+def run_if(cond, func):
```
Haichao-Zhang: since this "run_if" function is moved from common.py to here. do we need to update the related import where this function is used across alf?

emailweixu: It's imported in common.py. So it won't cause any issue.

PR #227, opened 2019-10-15
smart_add

emailweixu:
add() which handles addend as None or ()

### DIFF
--- a/alf/utils/math_ops.py
+++ b/alf/utils/math_ops.py
@@ -29,3 +29,25 @@ def clipped_exp(value, clip_value_min=-20, clip_value_max=2):
     """
     value = tf.clip_by_value(value, clip_value_min, clip_value_max)
     return tf.exp(value)
+
+
+def smart_add(x, y):
+    """Add two Tensors which may be None or ().
+
+     If x or y is None, the other tensor is returned.
+     Args:
+          x (Tensor|None|()):
+          y (Tensor(|None|())):
+     Returns:
+          x + y
+     """
+
+    def _ignore(t):
+        return t is None or (isinstance(t, tuple) and len(t) == 0)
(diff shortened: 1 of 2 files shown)


alf/utils/math_ops.py
```diff
@@ -29,3 +29,25 @@ def clipped_exp(value, clip_value_min=-20, clip_value_max=2):
     """
     value = tf.clip_by_value(value, clip_value_min, clip_value_max)
     return tf.exp(value)
+
+
+def smart_add(x, y):
```
le-horizon: Maybe call it add_ignore_empty()

alf/utils/math_ops.py
```diff
@@ -29,3 +29,25 @@ def clipped_exp(value, clip_value_min=-20, clip_value_max=2):
...
+
+    def _ignore(t):
+        return t is None or (isinstance(t, tuple) and len(t) == 0)
+
+    if _ignore(y):
+        return x
+    elif _ignore(x):
+        return y
+    else:
+        return x + y
```
le-horizon: What if both are empty?

Probably better to do this:
```
x_0 = x
y_0 = y
if _ignore(x):
  x_0 = 0
if _ignore(y):
  y_0 = 0

return x_0 + y_0
```

emailweixu: If they are not tensor, probably this function will raise some exception. And it's reasonable to do so.

emailweixu: If both are empty, the returned value is also empty. And there is no extra computation in tf, which is good.

alf/utils/math_ops.py
emailweixu: changed.

PR #228, opened 2019-10-15
improve grid_search for multi-restarts

hnyu:
Improve grid_search.py to include the option of multiple repeats for each configuration. When no parameters is specified, this can also be used to issue multiple repeats given the original gin file.

### DIFF
--- a/alf/bin/grid_search.py
+++ b/alf/bin/grid_search.py
@@ -60,6 +60,7 @@ class GridSearchConfig(object):
         "use_gpu": true,
         "gpus": [0, 1],
         "max_worker_num": 8,
+        "repeats": 3,
         "parameters": {
             "ac/Adam.learning_rate": [1e-3, 8e-4],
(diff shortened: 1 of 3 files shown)


witwolf:
This can be done with a trick 
```json
{
  "desc": "Simple repeate experiment",
  "use_gpu": true,
  "gpus": [
    0,
    1
  ],
  "max_worker_num": 6,
  "parameters": {
    "_UNUSED_": "range(10)"    # repeat 10 times 
  }
}
```

hnyu:
> This can be done with a trick
> 
> ```json
> {
>   "desc": "Simple repeate experiment",
>   "use_gpu": true,
>   "gpus": [
>     0,
>     1
>   ],
>   "max_worker_num": 6,
>   "parameters": {
>     "_UNUSED_": "range(10)"    # repeat 10 times 
>   }
> }
> ```

Yes, indeed. I just want to make the multi-restarts (and also the name generation) more friendly for use without a hack like this.

PR #229, opened 2019-10-15
PR refactor off_policy_algorithm train

runjerry:
Move the following functionalities from driver to algorithm
- off_policy train
- exp_observers and exp_replayer

### DIFF
--- a/alf/algorithms/actor_critic_algorithm.py
+++ b/alf/algorithms/actor_critic_algorithm.py
@@ -45,6 +45,8 @@ class ActorCriticAlgorithm(OnPolicyAlgorithm):
                  loss_class=ActorCriticLoss,
                  optimizer=None,
                  debug_summaries=False,
+                 summarize_grads_and_vars=False,
+                 summarize_action_distributions=False,
                  name="ActorCriticAlgorithm"):
         """Create an ActorCriticAlgorithm
 
@@ -63,6 +65,10 @@ class ActorCriticAlgorithm(OnPolicyAlgorithm):
             optimizer (tf.optimizers.Optimizer): The optimizer for training
             debug_summaries (bool): True if debug summaries should be created.
             name (str): Name of this algorithm.
+            summarize_grads_and_vars (bool): If True, gradient and network
+                variable summaries will be written during training.
+            summarize_action_distributions (bool): If True, generate summaris
+                for the action distributions.
             """
         super(ActorCriticAlgorithm, self).__init__(
             action_spec=action_spec,
@@ -74,6 +80,8 @@ class ActorCriticAlgorithm(OnPolicyAlgorithm):
             action_distribution_spec=actor_network.output_spec,
             optimizer=optimizer,
             debug_summaries=debug_summaries,
+            summarize_grads_and_vars=summarize_grads_and_vars,
+            summarize_action_distributions=summarize_action_distributions,
             name=name)
 
         self._actor_network = actor_network
--- a/alf/algorithms/actor_critic_algorithm.py
+++ b/alf/algorithms/actor_critic_algorithm.py
@@ -45,8 +45,6 @@ class ActorCriticAlgorithm(OnPolicyAlgorithm):
                  loss_class=ActorCriticLoss,
                  optimizer=None,
                  debug_summaries=False,
-                 summarize_grads_and_vars=False,
-                 summarize_action_distributions=False,
                  name="ActorCriticAlgorithm"):
         """Create an ActorCriticAlgorithm
 
@@ -65,10 +63,6 @@ class ActorCriticAlgorithm(OnPolicyAlgorithm):
             optimizer (tf.optimizers.Optimizer): The optimizer for training
             debug_summaries (bool): True if debug summaries should be created.
             name (str): Name of this algorithm.
-            summarize_grads_and_vars (bool): If True, gradient and network
-                variable summaries will be written during training.
-            summarize_action_distributions (bool): If True, generate summaris
-                for the action distributions.
             """
         super(ActorCriticAlgorithm, self).__init__(
             action_spec=action_spec,
@@ -80,8 +74,6 @@ class ActorCriticAlgorithm(OnPolicyAlgorithm):
             action_distribution_spec=actor_network.output_spec,
             optimizer=optimizer,
             debug_summaries=debug_summaries,
-            summarize_grads_and_vars=summarize_grads_and_vars,
-            summarize_action_distributions=summarize_action_distributions,
             name=name)
 
         self._actor_network = actor_network
--- a/alf/algorithms/off_policy_algorithm.py
+++ b/alf/algorithms/off_policy_algorithm.py
@@ -220,11 +220,21 @@ class OffPolicyAlgorithm(RLAlgorithm):
         """
         return experience
 
-    def prepare_off_policy_specs(self, env_batch_size,
-                                 time_step: ActionTimeStep, exp_replayer: str,
-                                 metrics):
+    def prepare_exp_replayer(self, exp_replayer: str):
+        """Prepare experience replayer."""
+
+        if exp_replayer == "one_time":
+            self._exp_replayer = OnetimeExperienceReplayer()
+        elif exp_replayer == "uniform":
+            self._exp_replayer = SyncUniformExperienceReplayer(
+                self._experience_spec, self._env_batch_size)
+        else:
+            raise ValueError("invalid experience replayer name")
+        self.add_experience_observer(self._exp_replayer.observe)
+
+    def prepare_off_policy_specs(self, time_step: ActionTimeStep):
         """Prepare various tensor specs for off_policy training.
-        
+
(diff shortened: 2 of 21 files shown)


alf/algorithms/actor_critic_algorithm.py
```diff
@@ -45,6 +45,8 @@ def __init__(self,
                  loss_class=ActorCriticLoss,
                  optimizer=None,
                  debug_summaries=False,
+                 summarize_grads_and_vars=False,
+                 summarize_action_distributions=False,
```
emailweixu: Right now, summarize_grads_and_vars and summarize_action_distributions are only used by the root algorithm. Perhaps it's better to omit this for all the subclasses to reduce clutter. We can add a function RLAlgorithm.set_summary_setting() to change these flags.

alf/algorithms/rl_algorithm.py
```diff
@@ -14,6 +14,9 @@
 """Base class for RL algorithms."""
 
 from abc import abstractmethod
+import itertools
```
emailweixu: not used

alf/algorithms/rl_algorithm.py
```diff
@@ -77,7 +84,10 @@ def __init__(self,
                  clip_by_global_norm=False,
                  reward_shaping_fn: Callable = None,
                  observation_transformer: Callable = None,
+                 train_step_counter=None,
```
emailweixu: this is no longer needed.

alf/algorithms/rl_algorithm.py
```diff
@@ -132,11 +157,60 @@ def action_distribution_spec(self):
         """Return the action distribution spec for the action distributions."""
         return self._action_distribution_spec
 
+    @property
```
emailweixu: Just add an observe() function, so the algorithm can do whatever it needs with the experience.

alf/drivers/off_policy_driver.py
```diff
@@ -210,155 +145,8 @@ def train(self,
...
-                action_distribution=action_distribution,
-                collect_action_distribution=collect_action_distribution)
-
-        loss_info, grads_and_vars = self._algorithm.train_complete(
-            tape=tape, training_info=training_info, weight=weight)
-
-        del tape
-
-        return training_info, loss_info, grads_and_vars
+        return self._algorithm.train(
```
emailweixu: off_policy_driver.train() can be removed. Call off_policy_algorithm.train instead

alf/drivers/on_policy_driver.py
```diff
@@ -84,17 +82,14 @@ def __init__(self,
             observers (list[Callable]): An optional list of observers that are
                 updated after every step in the environment. Each observer is a
                 callable(time_step.Trajectory).
-            metrics (list[TFStepMetric]): An optional list of metrics.
+            # metrics (list[TFStepMetric]): An optional list of metrics.
```
emailweixu: why it's commented out?

alf/drivers/policy_driver.py
```diff
@@ -16,7 +16,7 @@
 import os
 from typing import Callable
 
-import psutil
+# import psutil
```
emailweixu: remove

alf/algorithms/off_policy_algorithm.py
```diff
@@ -206,3 +219,266 @@ def preprocess_experience(self, experience: Experience):
...
+            action_distribution=self._action_dist_param_spec,
+            step_type=self._time_step_spec.step_type,
+            reward=self._time_step_spec.reward,
+            discount=self._time_step_spec.discount,
+            info=info_spec,
+            collect_info=self._processed_experience_spec.info,
+            collect_action_distribution=self._action_dist_param_spec)
+
+    @tf.function
+    def train(self,
```
emailweixu: To make things more flexible, I feel obtaining data from exp_player should also be a part of job of train()

alf/algorithms/actor_critic_algorithm.py
emailweixu: How about this?

runjerry: So should we make summarize_grads_and_vars and summarize_action_distribution global flags and maybe set in gin file, or keep the (new commit) PR way to use RLAlgorithm.set_sumary_setting() to set both in Trainer.initialize()?

alf/algorithms/off_policy_algorithm.py
```diff
@@ -206,3 +219,283 @@ def preprocess_experience(self, experience: Experience):
             processed experience
         """
         return experience
+
+    def prepare_off_policy_specs(self, env_batch_size,
+                                 time_step: ActionTimeStep, exp_replayer: str,
```
emailweixu: It's better to add metrics and exp_replayer using a different function since it's not about preparing specs.
And for metrics, since the Algorithm.training_summary() assumes a particular order of the metrics. It makes more sense to let algorithm to create the standard metrics and add another function in algorithm to let user to add additional metrics if they want.  Then PolicyDriver can get the metrics from algorithm.

alf/algorithms/on_policy_algorithm.py
```diff
@@ -69,3 +69,8 @@ def train_step(self, exp: Experience, state):
             observation=exp.observation,
             prev_action=exp.prev_action)
         return self.rollout(time_step, state, with_experience=True)
+
+    def prepare_on_policy_specs(self, metrics=[]):
```
emailweixu: See comments for prepare_off_policy_specs()

alf/drivers/off_policy_driver.py
```diff
@@ -21,13 +21,10 @@
 from tf_agents.trajectories.time_step import StepType
 from tf_agents.environments.tf_environment import TFEnvironment
 from tf_agents.specs.distribution_spec import DistributionSpec
-from tf_agents.specs.distribution_spec import nested_distributions_from_specs
+from tf_agents.specs.tensor_spec import TensorSpec
```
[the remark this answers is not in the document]
runjerry: There was story behind this change (witnessed by @le-horizon). Before my updating the tf_agent to current version, there was a strange bug. The action_distribution_spec (which was initialized as a BoundedTensorSpec) as not recognized as a tf.TensorSpec. So I felt that maybe staying with tf_agents' TensorSpec will have a slightly better robustness regarding the match between tf2.0 version and tf_agents version.

alf/algorithms/rl_algorithm.py
```diff
@@ -132,11 +149,72 @@ def action_distribution_spec(self):
...
+            alf.utils.summary_utils.summarize_action_dist(
+                training_info.action_distribution, self._action_spec)
+            if training_info.collect_action_distribution:
+                alf.utils.summary_utils.summarize_action_dist(
+                    action_distributions=training_info.
+                    collect_action_distribution,
+                    action_specs=self._action_spec,
+                    name="collect_action_dist")
+
+        for metric in self._metrics:
```
emailweixu: This depends on self._metrics being set. Should skip if _metrics is not set

alf/algorithms/rl_algorithm.py
```diff
@@ -132,11 +149,72 @@ def action_distribution_spec(self):
...
+    def exp_observers(self):
+        """Return experience observers."""
+        return self._exp_observers
+
+    @abstractmethod
+    def observe(self):
+        """An algorithm can decide what to do with experience."""
+        pass
+
+    def prepare_metrics(self, metrics=[]):
```
emailweixu: set_metrics might be a better name. Also should comment that len(metrics) should be >= 2 as required by the code at L208

alf/algorithms/off_policy_algorithm.py
```diff
@@ -206,3 +219,284 @@ def preprocess_experience(self, experience: Experience):
             processed experience
         """
         return experience
+
+    def prepare_exp_replayer(self, exp_replayer: str):
```
emailweixu: set_exp_replayer seems to be a better name

alf/algorithms/rl_algorithm.py
```diff
@@ -132,11 +149,72 @@ def action_distribution_spec(self):
         """Return the action distribution spec for the action distributions."""
         return self._action_distribution_spec
 
+    @property
+    def exp_observers(self):
+        """Return experience observers."""
+        return self._exp_observers
+
+    @abstractmethod
+    def observe(self):
```
emailweixu: this function is not used anywhere.

alf/drivers/policy_driver.py
```diff
@@ -223,15 +160,15 @@ def _step(self, time_step, policy_state):
...
+        if self._algorithm.exp_observers and self._training:
             action_distribution_param = common.get_distribution_params(
                 policy_step.action)
             exp = make_experience(
                 time_step,
                 policy_step._replace(action=action),
                 action_distribution=action_distribution_param,
                 state=policy_state if self._use_rollout_state else ())
-            for observer in self._exp_observers:
+            for observer in self._algorithm.exp_observers:
```
emailweixu: Can be moved to rl_algorithm.observe()

PR #233, opened 2019-10-16
safe_mean_summary

emailweixu:
Skip summary if data is empty

### DIFF
--- a/alf/utils/summary_utils.py
+++ b/alf/utils/summary_utils.py
@@ -245,7 +245,7 @@ def get_current_scope():
 
 
 def add_mean_hist_summary(name, value):
-    """Generate mean and histogram summary of value.
+    """Generate mean and histogram summary of `value`.
 
     Args:
         name (str): name of the summary
@@ -258,9 +258,9 @@ def add_mean_hist_summary(name, value):
 
 
 def safe_mean_hist_summary(name, value):
-    """Generate mean and histogram summary of value.
+    """Generate mean and histogram summary of `value`.
 
-    It skip the summary if value is empty.
+    It skips the summary if `value` is empty.
 
     Args:
         name (str): name of the summary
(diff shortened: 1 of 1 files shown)


alf/utils/summary_utils.py
```diff
@@ -236,6 +257,25 @@ def add_mean_hist_summary(name, value):
     add_mean_summary(name + "/mean", value)
 
 
+def safe_mean_hist_summary(name, value):
+    """Generate mean and histogram summary of value.
+
+    It skip the summary if value is empty.
```
Haichao-Zhang: skip -> skips

alf/utils/summary_utils.py
```diff
@@ -248,3 +288,22 @@ def add_mean_summary(name, value):
     if not value.dtype.is_floating:
         value = tf.cast(value, tf.float32)
     tf.summary.scalar(name, tf.reduce_mean(value))
+
+
+def safe_mean_summary(name, value):
+    """Generate mean summary of value.
+
+    It skip the summary if value is empty.
```
Haichao-Zhang: skip -> skips

alf/utils/summary_utils.py
```diff
@@ -16,6 +16,7 @@
 import tensorflow as tf
 from tensorboard.plugins.histogram import metadata
 from tensorflow.python.ops import summary_ops_v2
+from alf.utils.conditional_ops import run_if
```
Haichao-Zhang: shall we import from common.py for consistency? not sure what's the best practice here

emailweixu: Perhaps we should import every thing under alf.util from alf.utils in the future so that you need not concern which module a symbol is from. But even with that, for modules under util, you have to import from the actual source of a symbol to prevent circular import.

PR #235, opened 2019-10-16
Fix Algorithm.trainable_variables

emailweixu:


alf/algorithms/algorithm.py
```diff
@@ -157,8 +157,9 @@ def _is_alg(obj):
             return isinstance(obj, Algorithm)
 
         def _is_module_or_var(obj):
-            return (isinstance(obj, (tf.Variable, tf.Module))
-                    and not isinstance(obj, Algorithm))
+            return ((isinstance(obj, tf.Variable) and obj.trainable)
```
hnyu: if obj.trainable is False, then obj.trainable_variables will be [], which is fine in that case

emailweixu: Variable does not have property trainable_variables

Haichao-Zhang: Should the function name be updated to something like "_is_module_or_trainable_var" in order to reflect its functionality?

emailweixu: It's fine as it's a local function which only used here.

Issue #237, opened 2019-10-17
images are not stored as tf.uint8 in replay buffers


hnyu:
Due to tf_agents implementations in `gym_wrapper.py _spec_from_gym_space()`, all inputs of spec `gym.spaces.Box` will be mapped to the same dtype specified in `dtype_map`. This will have issues when the inputs are a mixture of uint8 and float32, or when the action is float32. So the default type for now is always float32. We are spending x4 memory storing image inputs right now.

hnyu:
PR #239

PR #238, opened 2019-10-17
Fix several bugs

hnyu:
1. image_scale_transformer is not applied in off-policy driver's train() because the shape check fails (with one additional "length" dimension). This bug completely will completely destroy any off-policy training such as PPO.
2. throughput computation is incorrect when num_steps_per_iter is small for on-policy driver due to the ceil() operation
3. Write optimizers info to Tensorboard
4. Image inputs are not stored as uint8 in replay buffers, see #237. There is not a trivial solution right now; leave it to future work.

### DIFF
--- a/alf/algorithms/algorithm.py
+++ b/alf/algorithms/algorithm.py
@@ -204,11 +204,7 @@ class Algorithm(tf.Module):
     def _get_opt_and_var_sets(self):
         opt_and_var_sets = []
         optimizer_and_module_sets = self.get_optimizer_and_module_sets()
-        optimizer_info = []
         for opt, module_set in optimizer_and_module_sets:
-            optimizer_info.append(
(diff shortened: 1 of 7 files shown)


alf/algorithms/algorithm.py
```diff
@@ -220,6 +221,7 @@ def _get_opt_and_var_sets(self):
                 else:
                     raise ValueError("Unsupported module type %s" % module)
             opt_and_var_sets.append((opt, vars))
+        self._optimizer_info = '\n\n'.join(optimizer_info)
```
emailweixu: This is quite hacky. I suggest call algorithm.get_optimizer_and_module_sets() outside and generate the info there.

hnyu: Moved the optimizer info to a member function.

PR #239, opened 2019-10-18
support image inputs as uint8

hnyu:
Address the issue #237. The idea is to write our own `_spec_from_gym_specs` in our own `suite_gym.py`. Now a tensor spec's dtype will be read from Gym's space dtype. So the input tensor types are defined by the environment specs. The replay buffer also store data that comply with the original input specs without being transformed. For off-policy training with uint8 image inputs, this will save CPU memory required by replay buffers.

### DIFF
--- a/alf/drivers/off_policy_driver.py
+++ b/alf/drivers/off_policy_driver.py
@@ -123,7 +123,8 @@ class OffPolicyDriver(policy_driver.PolicyDriver):
         self._time_step_spec = common.extract_spec(time_step)
         self._action_spec = self._env.action_spec()
 
-        policy_step = algorithm.rollout(time_step, self._initial_state)
+        policy_step = algorithm.rollout(
+            algorithm.transform_timestep(time_step), self._initial_state)
         info_spec = common.extract_spec(policy_step.info)
         self._policy_step_spec = PolicyStep(
             action=self._action_spec,
@@ -173,12 +174,14 @@ class OffPolicyDriver(policy_driver.PolicyDriver):
             action_distribution=action_dist,
             state=initial_state if self._use_rollout_state else ())
 
-        processed_exp = algorithm.preprocess_experience(exp)
+        processed_exp = algorithm.preprocess_experience(
+            algorithm.transform_timestep(exp))
         self._processed_experience_spec = self._experience_spec._replace(
-            info=common.extract_spec(processed_exp.info))
+            info=common.extract_spec(processed_exp.info))._replace(
+                observation=common.extract_spec(processed_exp.observation))
 
-        policy_step = common.algorithm_step(algorithm.train_step, exp,
-                                            initial_state)
+        policy_step = common.algorithm_step(algorithm.train_step,
+                                            processed_exp, initial_state)
         info_spec = common.extract_spec(policy_step.info)
         self._training_info_spec = make_training_info(
             action=self._action_spec,
--- a/alf/drivers/off_policy_driver.py
+++ b/alf/drivers/off_policy_driver.py
@@ -177,8 +177,8 @@ class OffPolicyDriver(policy_driver.PolicyDriver):
         processed_exp = algorithm.preprocess_experience(
             algorithm.transform_timestep(exp))
         self._processed_experience_spec = self._experience_spec._replace(
-            info=common.extract_spec(processed_exp.info))._replace(
-                observation=common.extract_spec(processed_exp.observation))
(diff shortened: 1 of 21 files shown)


alf/examples/atari.gin
```diff
@@ -1,5 +1,5 @@
 # environment config
-import tf_agents.environments.suite_gym
+import alf.environments.suite_gym
```
witwolf: perhaps it's better to `import alf.environments.suite_*` in `alf/environments/utils`  , so avoid to import suite_* for every gin configuration file.  

[L19-L21 modification ](https://github.com/HorizonRobotics/alf/commit/22142289d9c9d609c82fa655f1703864c6e50af7#diff-0d12de508ba2ef2d70e9ec2274a9c09cL19-L21) in  #211    make some   examples for super mario and social bot not 
 available

hnyu: > And [L19-L21 modification ](https://github.com/HorizonRobotics/alf/commit/22142289d9c9d609c82fa655f1703864c6e50af7#diff-0d12de508ba2ef2d70e9ec2274a9c09cL19-L21) in #211 make some examples for super mario and social bot not
> available

What exact modifications are you referring to?

hnyu:
Also, this PR will make `observation_spec` for image games have uint8 in it. And `observation_spec` will be used by tf_agents (only) to create networks. As I tested, those model parameters will have their own dtype defined, uncorrelated with the input tensor spec. However, as suggested by @emailweixu, for now I will convert all uint8 in `observation_spec` to float32 before returning.

alf/examples/atari.gin
hnyu: If we import all suite_* in utils, then there is a circular import dependency which we should always avoid. I think it makes more sense to move some common utils like `ProcessPyEnvironment` and `UnwrappedEnvChecker` to `utils.py` and import suite_* in individual gin files.

hnyu: What I can do at most is to import suite_* in `utils.py` for suites that don't rely on `utils.py`: `suite_gym` and `suite_simple`.

witwolf: ok , i see

alf/utils/common.py
```diff
@@ -414,30 +414,41 @@ def get_global_counter(default_counter=None):
...
+        assert obs.dtype == tf.uint8, "Image must have dtype uint8!"
+        obs = tf.cast(obs, tf.float32)
+        return ((max - min) / 255.) * obs + min
+
+    def _traverse_path(obs, path):
+        """Traverse `path` and transform the image at the path end"""
+        if not path:
+            return _transform_image(obs)
+        step = path[0]
+        return obs._replace(
```
emailweixu: obs is not always a namedtuple. It can be dictionary, list, unnamed tuple, namedtuple. I think @Haichao-Zhang might use unnamed tuple in his task.

alf/drivers/off_policy_driver.py
```diff
@@ -173,12 +174,14 @@ def _to_distribution_spec(spec):
             action_distribution=action_dist,
             state=initial_state if self._use_rollout_state else ())
 
-        processed_exp = algorithm.preprocess_experience(exp)
+        processed_exp = algorithm.preprocess_experience(
+            algorithm.transform_timestep(exp))
         self._processed_experience_spec = self._experience_spec._replace(
-            info=common.extract_spec(processed_exp.info))
+            info=common.extract_spec(processed_exp.info))._replace(
+                observation=common.extract_spec(processed_exp.observation))
```
emailweixu: _replace() can accept multiple key-value pairs:
```python
self._experience_spec._replace(
     info=common.extract_spec(processed_exp.info),                 
     observation=common.extract_spec(processed_exp.observation))
```

alf/utils/common.py
hnyu: If someone is going to this transformer and obs is a nest structure, it has to be a namedtuple; otherwise there is simply no way to specify which field to transform. Requirement has been written in the comment.

alf/drivers/off_policy_driver.py
hnyu: changed

alf/utils/common.py
hnyu: Added support for dictionary

Haichao-Zhang: I'm using named tuple as the observation structure. Unnamed tuple is not allowed by TFPyEnvironment.

Issue #241, opened 2019-10-20
_cached_opt_and_var_sets inconsistent between initial and sequential calls


hnyu:
I tried printing out how many variables each optimizer is responsible for optimization. I printed in two places. The first place is inside `algorithm._get_opt_and_var_sets()`, which is before `self._cached_opt_and_var_sets` is set, and I got something like below

```
{'amsgrad': False,
 'beta_1': 0.9,
 'beta_2': 0.999,
 'decay': 0.0,
 'epsilon': 1e-07,
 'learning_rate': 3e-05,
 'name': 'Adam'} 24
{'amsgrad': False,
 'beta_1': 0.9,
 'beta_2': 0.999,
 'decay': 0.0,
 'epsilon': 1e-07,
 'learning_rate': 1e-05,
 'name': 'Adam'} 26
{'amsgrad': False,
 'beta_1': 0.9,
 'beta_2': 0.999,
 'decay': 0.0,
 'epsilon': 1e-07,
 'learning_rate': 0.001,
 'name': 'Adam'} 1
```
For each row, the first is the optimizer config and the second is `len(vars)`, which makes sense given my job. However, if I printed inside `train_complete`, after the results of calling `self._get_cached_opt_and_var_sets()`, the output is 

```
{'amsgrad': False,
 'beta_1': 0.9,
 'beta_2': 0.999,
 'decay': 0.0,
 'epsilon': 1e-07,
 'learning_rate': 3e-05,
 'name': 'Adam'} 75
{'amsgrad': False,
 'beta_1': 0.9,
 'beta_2': 0.999,
 'decay': 0.0,
 'epsilon': 1e-07,
 'learning_rate': 1e-05,
 'name': 'Adam'} 26
{'amsgrad': False,
 'beta_1': 0.9,
 'beta_2': 0.999,
 'decay': 0.0,
 'epsilon': 1e-07,
 'learning_rate': 0.001,
 'name': 'Adam'} 1
``` 
A simple calculation suggests that somehow the parent algorithm (holds the first optimizer above) also takes into account all leaf trainable variables somewhere in the code (24 + 24 + 26 + 1 = 75). So some variables will be also optimized by the parent optimizer even though I have specified child optimizers for them (our tape is "persistent" and grads can be computed multiple times). 

Verified that this also happened in the eager mode.

hnyu:
[...]

PR #242, opened 2019-10-21
fix opt_and_var_sets

hnyu:
The change is to check duplications based on variable ids among all trainable (module) variables in the current or child algorithms. And added a check to make sure that each variable will only be optimized by one optimizer in the end.  

This PR should fix issue #241. I suspect that the issue is due to that somewhere in the parent algorithm TF sets a class member pointing to all trainable variables, before `train_complete` being called.

### DIFF
--- a/alf/algorithms/algorithm.py
+++ b/alf/algorithms/algorithm.py
@@ -219,7 +219,7 @@ class Algorithm(tf.Module):
             new_var_ids = set(map(id, new_vars))
             dup_ids = var_ids & new_var_ids
             assert not dup_ids, \
-                (("Variables %s might have multiple optimizers! Consider "
+                (("Modules/variables %s might have multiple optimizers! Consider "
                  + "specifying attributes in _trainable_attributes_to_ignore()")
                  % new_module_or_var.name)
             var_ids.update(new_var_ids)
@@ -288,15 +288,6 @@ class Algorithm(tf.Module):
                 else:
                     raise ValueError("Unsupported module type %s" % module)
             opt_and_var_sets.append((opt, vars))
-        # Check that each trainable variable is handled by one and only one
-        # optimizer (although it might not have any gradient).
-        var_ids = sum([list(map(id, vars)) for _, vars in opt_and_var_sets],
-                      [])
-        for opt, vars in opt_and_var_sets:
-            for v in vars:
-                assert var_ids.count(id(v)) == 1, \
-                    ("Variable '%s' is optimized by %d optimizers!" %
-                        (v.name, var_ids.count(id(v))))
         return opt_and_var_sets
 
     def get_optimizer_info(self):
--- a/alf/algorithms/algorithm.py
+++ b/alf/algorithms/algorithm.py
@@ -266,7 +266,8 @@ class Algorithm(tf.Module):
                 else:
                     raise ValueError("Unsupported module type %s" % module)
             opt_and_var_sets.append((opt, vars))
-        # Check that each variable is optimized by at one and only one optimizer
+        # Check that each trainable variable is handled by one and only one
+        # optimizer (although it might not have any gradient).
         var_ids = sum([list(map(id, vars)) for _, vars in opt_and_var_sets],
                       [])
         for opt, vars in opt_and_var_sets:
--- a/alf/algorithms/algorithm.py
+++ b/alf/algorithms/algorithm.py
@@ -201,9 +201,8 @@ class Algorithm(tf.Module):
                     module_sets.append(module_set)
                 else:
                     module_sets[0].extend(module_set)
-                # The reason for updating module_ids set is that in the child
-                # alg modules, we might have members that point to their network
-                # variables
+            # The reason for updating var_ids set is that we might have members
+            # referring to network variables in children
             var_ids.update(child_var_ids)
 
         # Module has a higher priority than Variable
--- a/alf/algorithms/algorithm.py
+++ b/alf/algorithms/algorithm.py
@@ -151,9 +151,9 @@ class Algorithm(tf.Module):
         """Get the optimizers and the corresponding module sets.
 
         Returns:
-            list[tuple(Opimizer, list[Module])]: optimizer can be None, which
-                means that no optimizer is specified for the corresponding
-                modules.
+            list[tuple(Algorithm_name, Opimizer, list[Module])]: optimizer
+                can be None, which means that no optimizer is specified for the
+                corresponding modules.
             var_ids (set[int]): a set of variable ids including all distinct
                 trainable variables from the current algorithm towards the
                 hierarchy below; used to check duplicates
@@ -186,7 +186,7 @@ class Algorithm(tf.Module):
 
         module_sets = [copy.copy(s) for s in self._init_module_sets]
         optimizers = copy.copy(self._init_optimizers)
-        algorithm_names = [self.__class__.__name__] * len(optimizers)
+        algorithm_names = [self.name] * len(optimizers)
 
         # This set stores all the seen distinct variables so far in this alg
         var_ids = set(
@@ -216,9 +216,15 @@ class Algorithm(tf.Module):
             if new_ids == ids:
                 module_sets[0].append(module)
             else:  # some variables have been added; the rest are added individually
-                module_sets[0].extend([
+                rest_vars = [
                     v for v in module.trainable_variables if id(v) in new_ids
-                ])
+                ]
+                rest_var_names = [v.name for v in rest_vars]
+                logging.warning(
+                    ("Variables %s might have an optimizer different from" +
+                     " what their parent module %s does") % (rest_var_names,
+                                                             module.name))
+                module_sets[0].extend(rest_vars)
             var_ids.update(new_ids)
 
         for var in self._get_children(_is_trainable_var):
@@ -285,8 +291,9 @@ class Algorithm(tf.Module):
         for alg_name, opt, module_set in optimizer_and_module_sets:
             optimizer_info.append(
                 "Algorithm: \"%s\" Optimizer: \"%s\" Modules: \"%s\"" \
-                    % (alg_name, opt.get_config(), '; '.join(
-                        sorted([m.name for m in module_set if m is not None]))))
+                    % (alg_name,
+                       opt.get_config() if opt is not None else None,
+                       '; '.join(sorted([m.name for m in module_set if m is not None]))))
         return '\n\n'.join(optimizer_info)
 
     @property
--- a/alf/algorithms/algorithm.py
+++ b/alf/algorithms/algorithm.py
@@ -172,9 +172,6 @@ class Algorithm(tf.Module):
             list[tuple(Algorithm_name, Opimizer, list[Module])]: optimizer
                 can be None, which means that no optimizer is specified for the
                 corresponding modules.
-            vars (list[tf.Variable]): a list of variables including all distinct
-                trainable variables from the current algorithm towards the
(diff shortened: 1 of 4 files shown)


alf/algorithms/algorithm.py
```diff
@@ -156,28 +159,71 @@ def get_optimizer_and_module_sets(self):
...
+        # This set stores all the seen distinct variables so far in this alg
+        var_ids = set(
+            map(
+                id,
+                tf.nest.flatten(
+                    tf.nest.map_structure(_get_trainable_vars, module_sets))))
 
         for alg in self._get_children(_is_alg):
-            for opt, module_set in alg.get_optimizer_and_module_sets():
+            opt_and_module, child_var_ids = alg.get_optimizer_and_module_sets()
```
Haichao-Zhang: This function name as well as its documentation might need to be updated: https://github.com/HorizonRobotics/alf/blob/d58103f299558f5fdeae218bd23515aed5c9b557/alf/algorithms/algorithm.py#L147-L153

hnyu: The name is fine since var_ids is just an auxiliary variable to assist getting optimizers and modules.

alf/algorithms/algorithm.py
```diff
@@ -263,12 +266,14 @@ def _get_opt_and_var_sets(self):
                 else:
                     raise ValueError("Unsupported module type %s" % module)
             opt_and_var_sets.append((opt, vars))
-        # Check that each variable is optimized by only one optimizer
-        var_ids = [i for _, vars in opt_and_var_sets for i in map(id, vars)]
-        for _, vars in opt_and_var_sets:
+        # Check that each variable is optimized by at one and only one optimizer
```
Haichao-Zhang: at one -> at least one

alf/algorithms/algorithm.py
```diff
@@ -217,17 +264,24 @@ def _get_opt_and_var_sets(self):
...
         return opt_and_var_sets
 
     def get_optimizer_info(self):
         """Return the optimizer info for all the modules in a string.
         """
-        optimizer_and_module_sets = self.get_optimizer_and_module_sets()
+        optimizer_and_module_sets, _ = self.get_optimizer_and_module_sets()
         optimizer_info = []
-        for opt, module_set in optimizer_and_module_sets:
+        for alg_name, opt, module_set in optimizer_and_module_sets:
```
witwolf: opt may be None  when there is  no optimizer for root algorithm , should filter out `opt == None`

hnyu: The assertion inside `train_complete()` will make sure that each opt is not None: `assert optimizer is not None, "optimizer needs to be provides at __init__()"`

alf/algorithms/algorithm.py
hnyu: fixed

hnyu:
One flaw of the current PR design is that if the parent algorithm passes a module to several subalgorithms which have their own optimizers, then this module's variables will always be optimized by the first (in the DFS sense) optimizer that sees them. Before this PR, there is an undefined behavior because those variables are simultaneously optimized by multiple optimizers that see them.

alf/algorithms/algorithm.py
hnyu: changed

alf/algorithms/algorithm.py
```diff
@@ -151,33 +154,79 @@ def get_optimizer_and_module_sets(self):
...
+            var_ids.update(child_var_ids)
+
+        # Module has a higher priority than Variable
+        for module in self._get_children(_is_trainable_module):
+            ids = set(map(id, module.trainable_variables))
+            new_ids = ids - var_ids
+            # all ids not appeared in var_ids; add the entire module (better info)
+            if new_ids == ids:
                 module_sets[0].append(module)
+            else:  # some variables have been added; the rest are added individually
```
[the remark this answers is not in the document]
hnyu: My KME code actually has this because I want to compute Hessian with an inner tape, and I have a reference to trainable variables of a module.

alf/algorithms/algorithm.py
```diff
@@ -151,33 +154,79 @@ def get_optimizer_and_module_sets(self):
...
+                return obj.trainable_variables
+            elif _is_trainable_var(obj):
+                return [obj]
+            else:
+                assert "%s is not a trainble module or variable!" % obj.name
 
         module_sets = [copy.copy(s) for s in self._init_module_sets]
         optimizers = copy.copy(self._init_optimizers)
-        module_ids = set(map(id, sum(module_sets, [])))
+        algorithm_names = [self.__class__.__name__] * len(optimizers)
```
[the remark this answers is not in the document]
hnyu: changed

alf/algorithms/algorithm.py
```diff
@@ -151,33 +154,79 @@ def get_optimizer_and_module_sets(self):
             list[tuple(Opimizer, list[Module])]: optimizer can be None, which
```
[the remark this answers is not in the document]
hnyu: updated

alf/algorithms/algorithm.py
```diff
@@ -217,17 +266,27 @@ def _get_opt_and_var_sets(self):
...
         """
-        optimizer_and_module_sets = self.get_optimizer_and_module_sets()
+        optimizer_and_module_sets, _ = self.get_optimizer_and_module_sets()
         optimizer_info = []
-        for opt, module_set in optimizer_and_module_sets:
+        for alg_name, opt, module_set in optimizer_and_module_sets:
             optimizer_info.append(
-                "optimizer %s: modules %s" % (opt.get_config(), ' '.join(
-                    [m.name for m in module_set if m is not None])))
+                "Algorithm: \"%s\" Optimizer: \"%s\" Modules: \"%s\"" \
```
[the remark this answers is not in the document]
hnyu: The optimizer config has single quote inside, which makes the format less readable.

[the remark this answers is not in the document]
hnyu: The optimizer config str has single quotes inside, which makes the resulting format less readable if I do that.

emailweixu:
> One flaw of the current PR design is that if the parent algorithm passes a module to several subalgorithms which have their own optimizers, then this module's variables will always be optimized by the first (in the sense of DFS tree) optimizer that sees them. Before this PR, there is an undefined behavior because those variables are simultaneously optimized by multiple optimizers that see them in the tree. @emailweixu

In this case, we should report an error and exit. We can add a member function to algorithm to indicate the modules/algorithms/variables which it does not want to optimize.

hnyu:
> > One flaw of the current PR design is that if the parent algorithm passes a module to several subalgorithms which have their own optimizers, then this module's variables will always be optimized by the first (in the sense of DFS tree) optimizer that sees them. Before this PR, there is an undefined behavior because those variables are simultaneously optimized by multiple optimizers that see them in the tree. @emailweixu
> 
> In this case, we should report an error and exit. We can add a member function to algorithm to indicate the modules/algorithms/variables which it does not want to optimize.

As mentioned in issue #241 , somehow the parent algorithm will hold all trainable variables automatically before `train_complete`. There will always been overlap of variables in that case. It's tedious to specify which variables the parent algorithm doesn't want to optimize.

emailweixu:
> > > One flaw of the current PR design is that if the parent algorithm passes a module to several subalgorithms which have their own optimizers, then this module's variables will always be optimized by the first (in the sense of DFS tree) optimizer that sees them. Before this PR, there is an undefined behavior because those variables are simultaneously optimized by multiple optimizers that see them in the tree. @emailweixu
> > 
> > 
> > In this case, we should report an error and exit. We can add a member function to algorithm to indicate the modules/algorithms/variables which it does not want to optimize.
> 
> As mentioned in issue #241 , somehow the parent algorithm will hold all trainable variables automatically before `train_complete`. There will always been overlap of variables in that case. It's tedious to specify which variables the parent algorithm doesn't want to optimize.

Perhaps we can do something like this to avoid the issue?https://github.com/tensorflow/tensorflow/blob/1cf0898dd4331baf93fe77205550f2c2e6c90ee5/tensorflow/python/module/module.py#L256

emailweixu:
> > > > One flaw of the current PR design is that if the parent algorithm passes a module to several subalgorithms which have their own optimizers, then this module's variables will always be optimized by the first (in the sense of DFS tree) optimizer that sees them. Before this PR, there is an undefined behavior because those variables are simultaneously optimized by multiple optimizers that see them in the tree. @emailweixu
> > > 
> > > 
> > > In this case, we should report an error and exit. We can add a member function to algorithm to indicate the modules/algorithms/variables which it does not want to optimize.
> > 
> > 
> > As mentioned in issue #241 , somehow the parent algorithm will hold all trainable variables automatically before `train_complete`. There will always been overlap of variables in that case. It's tedious to specify which variables the parent algorithm doesn't want to optimize.
> 
> Perhaps we can do something like this to avoid the issue?https://github.com/tensorflow/tensorflow/blob/1cf0898dd4331baf93fe77205550f2c2e6c90ee5/tensorflow/python/module/module.py#L256

I tested this by ignoring the attributes in _TF_MODULE_IGNORED_PROPERTIES at Algorith._get_children. It does work.

hnyu:
> > > > > One flaw of the current PR design is that if the parent algorithm passes a module to several subalgorithms which have their own optimizers, then this module's variables will always be optimized by the first (in the sense of DFS tree) optimizer that sees them. Before this PR, there is an undefined behavior because those variables are simultaneously optimized by multiple optimizers that see them in the tree. @emailweixu
> > > > 
> > > > 
> > > > In this case, we should report an error and exit. We can add a member function to algorithm to indicate the modules/algorithms/variables which it does not want to optimize.
> > > 
> > > 
> > > As mentioned in issue #241 , somehow the parent algorithm will hold all trainable variables automatically before `train_complete`. There will always been overlap of variables in that case. It's tedious to specify which variables the parent algorithm doesn't want to optimize.
> > 
> > 
> > Perhaps we can do something like this to avoid the issue?https://github.com/tensorflow/tensorflow/blob/1cf0898dd4331baf93fe77205550f2c2e6c90ee5/tensorflow/python/module/module.py#L256
> 
> I tested this by ignoring the attributes in _TF_MODULE_IGNORED_PROPERTIES at Algorith._get_children. It does work.

Revised

alf/algorithms/algorithm.py
```diff
@@ -187,52 +205,46 @@ def _get_trainable_vars(obj):
...
-                    tf.nest.map_structure(_get_trainable_vars, module_sets))))
+        vars = tf.nest.flatten(
+            tf.nest.map_structure(_get_trainable_vars, module_sets))
+        var_ids = set(map(id, vars))
+
+        def _check_vars(new_vars):
+            new_var_ids = set(map(id, new_vars))
+            dup_ids = var_ids & new_var_ids
+            dup_vars = [v for v in new_vars if id(v) in dup_ids]
+            assert not dup_vars, \
```
emailweixu: It's better to report the error on module level if it is available: def _check_trainable(var_or_module)

alf/algorithms/algorithm.py
```diff
@@ -141,54 +141,125 @@ def __init__(self,
...
+
+        def _check_module_or_var(new_module_or_var):
+            if isinstance(new_module_or_var, tf.Module):
+                new_vars = new_module_or_var.trainable_variables
+            else:
+                new_vars = [new_module_or_var]
+            new_var_ids = set(map(id, new_vars))
+            dup_ids = var_ids & new_var_ids
+            assert not dup_ids, \
+                (("Variables %s might have multiple optimizers! Consider "
```
emailweixu: Module/Variable

PR #243, opened 2019-10-23
make train test traceable when error occurs

witwolf:
logging training stdout & stderr output to stderr (stdout is suppressed at ci) to make it traceable when error occurs. 

and this pr can  solve  the issue:  [build-times-out-because-no-output-was-received](https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received) (`travis_wait` not support for a docker run command )

### DIFF
--- a/alf/bin/train_test.py
+++ b/alf/bin/train_test.py
@@ -19,6 +19,7 @@ import subprocess
 from pathlib import Path
 import numpy as np
 import sys
+import time
 
 import logging as sys_logging
 from absl import logging
@@ -37,23 +38,11 @@ def run_and_stream(cmd, cwd):
     """
     logging.info("Running %s", " ".join(cmd))
 
-    # create a logger for sub process outputs
-    # 1. logging all outputs of sub process to sys.stderr to make it traceable when an error
-    #   occurs (ci suppresses stdout output to prevent producing a big log file than 4 MB)
-    # 2. set a simple formatter without prefix for the logger, because a log_prefix
-    #   already exists for sub process log
-    logger = logging.ABSLLogger('')
-    handler = sys_logging.StreamHandler(sys.stderr)
-    handler.setFormatter(sys_logging.Formatter('%(message)s'))
-    logger.addHandler(handler)
-
     process = subprocess.Popen(
-        cmd, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, cwd=cwd)
+        cmd, stdout=sys.stdout, stderr=sys.stderr, cwd=cwd)
 
     while process.poll() is None:
-        with io.TextIOWrapper(process.stdout, encoding="utf-8") as text_io:
-            for line in text_io:
-                logger.info(line.strip())
+        time.sleep(0.1)
 
     assert process.returncode == 0, ("cmd: {0} exit abnormally".format(
         " ".join(cmd)))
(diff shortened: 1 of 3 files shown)


alf/bin/train_test.py
```diff
@@ -34,17 +36,22 @@ def run_and_stream(cmd, cwd):
         cwd (str): working directory for the process
     """
     logging.info("Running %s", " ".join(cmd))
+
+    logger = logging.ABSLLogger('')
```
emailweixu: instead of changing the code here, can we simply redirect stdout and stderr to stderr at build.sh?
```sh
python3 -m unittest discover -p "*_test.py" -v 1>&2
```
If not, we need some comment in the code to explain the reason of doing all these stuff so that future readers can understand it.

witwolf: yes, it can be done with`python3 -m unittest discover -p "*_test.py" -v 1>&2`  , 
but there is a  potential problem  [Log length exceeded 4 MB](https://docs.travis-ci.com/user/common-build-problems/#log_length_exceeded) if we log for all stdout  (now the log file is about 3.2MB)

emailweixu: I see. Then please add some comments here.

witwolf: added

witwolf:
> how about other tests, don’t they have the same issue?

other tests are not in sub process like train test  (python -m  unittest ... -> python  .._test_py-> python -m alf.bin.train ... )

alf/bin/train_test.py
```diff
@@ -34,17 +36,27 @@ def run_and_stream(cmd, cwd):
...
+    logger.addHandler(handler)
+
     process = subprocess.Popen(
         cmd, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, cwd=cwd)
 
     while process.poll() is None:
         with io.TextIOWrapper(process.stdout, encoding="utf-8") as text_io:
             for line in text_io:
-                logging.info(line.strip())
+                logger.info(line.strip())
```
emailweixu: Now the original stdout goes to stderr, and the original stderr goes to stdout (because stderr=subprocess.STDOUT at line 50). So I am still confused. And why not simply use stdout=subprocess.STDERR at line 50?

emailweixu:
> other tests are not in sub process like train test (python -m unittest ... -> python .._test_py-> python -m alf.bin.train ... )

I am still confused. If the stdout of train_test is suppressed, aren't the stdout of other tests also suppressed?

witwolf:
> > other tests are not in sub process like train test (python -m unittest ... -> python .._test_py-> python -m alf.bin.train ... )
> 
> I am still confused. If the stdout of train_test is suppressed, aren't the stdout of other tests also suppressed?

train test cost much time,  may encounter such a problem  [build-times-out-because-no-output-was-received](https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received) when no stdout output  .  

solve this with  `travis_wait`  in the new commit   (https://github.com/travis-ci/travis-ci/issues/6934 )

alf/bin/train_test.py
```diff
@@ -18,7 +18,10 @@
 import subprocess
 from pathlib import Path
 import numpy as np
+import sys
+import time
 
+import logging as sys_logging
```
emailweixu: Not used

witwolf: removed

Issue #240, opened 2019-10-18
segmentation fault for play with socialbot environments


emailweixu:
The problem can be fixed if wrap_process=True is used in the following line (and remove the next line)
https://github.com/HorizonRobotics/alf/blob/d58103f299558f5fdeae218bd23515aed5c9b557/alf/environments/utils.py#L158-L159

hnyu:
I have no idea what's going here. I tested self._eval_env without being wrapped, and it worked well. It seems that for playing, the innermost Python environment crashes for either reset() or step(), as I traced the error back to the FrameStack wrapper.

witwolf:
I  tested it on my machine and did not encounter such a problem.   what is the details ?

hnyu:
> I tested it on my machine and did not encounter such a problem. what is the details ?

Make sure to pull the newest version where play.py creates an environment with `nonparallel=True`. For example, after you train a checkpoint for ac_simple_navigation.gin and load it to play, the segfault will occur.

hnyu:
Or perhaps for now we can just add a hack of always wrapping SocialBot environments. But this will also show incorrect gin information for them.

witwolf:
found out the reason  is that gazebo operations on an instance (creation, reset, step ...)  can not run in multiple thread .   I'll fix this issue

PR #244, opened 2019-10-24
Create and step env in a separate thread

witwolf:
#240 

social_bot envs (gazebo) requires `step` and `reset` run in the same thread which it created in ,
so add  a wrapper `ThreadPyEnvironment` that create and step it in a separate thread .

### DIFF
--- a/alf/environments/utils.py
+++ b/alf/environments/utils.py
@@ -11,6 +11,12 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+# multiprocessing.dummy provides a pure *multithreaded* threadpool that works
+# in both python2 and python3 (concurrent.futures isn't available in python2).
+#   https://docs.python.org/2/library/multiprocessing.html#module-multiprocessing.dummy
+from multiprocessing import dummy as mp_threads
+
 import random
 import sys
 import traceback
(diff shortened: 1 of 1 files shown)


hnyu:
> social_bot envs (gazebo) requires `step` and `reset` run in the same thread which it created in ,

Looking at the code, I suppose what you really mean is that `step` and `reset` have to be in a different thread than the main thread that creates the environment, right?

emailweixu:
> #240
> 
> social_bot envs (gazebo) requires `step` and `reset` run in the same thread which it created in ,
> so add a wrapper `ThreadPyEnvironment` that create and step it in a separate thread .

Why?

witwolf:
> > social_bot envs (gazebo) requires `step` and `reset` run in the same thread which it created in ,
> 
> Looking at the code, I suppose what you really mean is that `create`, `step` and `reset` have to be in a different thread than the main thread, right?

they can be in the main thread , only need to keep they are in the same thread

witwolf:
> > #240
> > social_bot envs (gazebo) requires `step` and `reset` run in the same thread which it created in ,
> > so add a wrapper `ThreadPyEnvironment` that create and step it in a separate thread .
> 
> Why?

have not dig deep into gazebo yet,  just run a simple test with code below: 

```python
import social_bot
import gym
import threading

if __name__ == '__main__':
    env = gym.make('SocialBot-Pr2Gripper-v0')

    def _reset():
        env.reset()

    t = threading.Thread(target=_reset)
    t.start()
    t.join()
```

or 
```python
if __name__ == '__main__':
    env = gym.make('SocialBot-Pr2Gripper-v0')
    env.reset()

    def _step():
        env.step(env.action_space.sample())

    t = threading.Thread(target=_step)
    t.start()
    t.join()
```

it crashes at `reset`  (second code fragment crashes at `step`) with segmentation fault, and bt show it's related to `Ogre`

image

alf/environments/utils.py
```diff
@@ -11,6 +11,12 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+# multiprocessing.dummy provides a pure *multithreaded* threadpool that works
+# in both python2 and python3 (concurrent.futures isn't available in python2).
```
emailweixu: Can our code probably won't run using python2? I don't think so.

witwolf:
> But it seems that all calls to the environment in the current play() are from the same thread (i.e. the main thread)

paly() is run in graph mode  , `step` is  called in thread managed by tf

alf/environments/utils.py
witwolf: removed (i copied these lines from `batched_py_environment.py`)

emailweixu:
> > But it seems that all calls to the environment in the current play() are from the same thread (i.e. the main thread)
> 
> paly() is run in graph mode , `step` is called in thread managed by tf

I see.

PR #246, opened 2019-10-28
make agent support for ddpg & sac

witwolf:
1.  Modify `Agent` to support  for ddpg, sac .

2. remove `with_experience` from func `rollout`, make it easier to understand 
This only affect `_reward_normalizer` in ICM (may calculate intrinsic  reward multiple times ) , and i have tested on `ppo_icm_super_mario_intrinsic_only` ,it does not affect performance (right:master, left:PR_agent1)

TODO:
since we can train  an `OnPolicyAlgorithm`  with an `OffPolicyTrainer` , it's no necessary to calculate all things (loss, info ...) when rollout , it can save much computation . what to calculate is actually determined by what trainer is used .

### DIFF
--- a/alf/algorithms/agent.py
+++ b/alf/algorithms/agent.py
@@ -157,9 +157,6 @@ class Agent(OnPolicyAlgorithm):
         new_state = AgentState()
         info = AgentInfo()
         observation = self._encode(time_step)
-
-        # TODO
-        # avoid computing this when rl_algorithm do not need preprocess experience
         if self._icm is not None:
             icm_step = self._icm.train_step(
                 (observation, time_step.prev_action), state=state.icm)
@@ -173,7 +170,7 @@ class Agent(OnPolicyAlgorithm):
         info = info._replace(rl=rl_step.info)
 
         # TODO
-        # avoid computing this when collecting exps (off policy train)
+        # avoid computing this when rollout (off policy train)
         if self._entropy_target_algorithm:
             et_step = self._entropy_target_algorithm.train_step(rl_step.action)
             info = info._replace(entropy_target=et_step.info)
--- a/alf/algorithms/agent.py
+++ b/alf/algorithms/agent.py
@@ -141,20 +141,28 @@ class Agent(OnPolicyAlgorithm):
 
         return PolicyStep(action=rl_step.action, state=new_state, info=())
 
-    def rollout(self,
-                time_step: ActionTimeStep,
-                state: AgentState,
-                with_experience=False):
+    def greedy_predict(self, time_step: ActionTimeStep, state=None, eps=0.1):
+        observation = self._encode(time_step)
+
+        new_state = AgentState()
+
+        rl_step = self._rl_algorithm.greedy_predict(
+            time_step._replace(observation=observation), state.rl)
+        new_state = new_state._replace(rl=rl_step.state)
+
+        return PolicyStep(action=rl_step.action, state=new_state, info=())
+
+    def rollout(self, time_step: ActionTimeStep, state: AgentState):
         """Rollout for one step."""
         new_state = AgentState()
         info = AgentInfo()
         observation = self._encode(time_step)
 
+        # TODO
+        # avoid computing this when rl_algorithm do not need preprocess experience
         if self._icm is not None:
             icm_step = self._icm.train_step(
-                (observation, time_step.prev_action),
-                state=state.icm,
-                calc_intrinsic_reward=not with_experience)
+                (observation, time_step.prev_action), state=state.icm)
             info = info._replace(icm=icm_step.info)
             new_state = new_state._replace(icm=icm_step.state)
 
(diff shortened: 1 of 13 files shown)


emailweixu:
Sac/ddpg cannot work with icm even with this change, because icm need to have use_rollout_state=False when using off_policy_driver. Perhaps we can leave that as the next PR. As to removing with_experience, @hnyu may not agree because he added that to save the computation for his own intrinsic reward algorithm.

hnyu:
> 1. Modify `Agent` to support  for ddpg, sac .
> 2. remove `with_experience` from func `rollout`, make it easier to understand
>    This only affect `_reward_normalizer` in ICM (may calculate intrinsic  reward multiple times ) , and i have tested on `ppo_icm_super_mario_intrinsic_only` ,it does not affect performance (right:master, left:PR_agent1)
> 
> 
> 
> TODO:
> since we can train an `OnPolicyAlgorithm` with an `OffPolicyTrainer` , it's no necessary to calculate all things (loss, info ...) when rollout , it can save much computation . what to calculate is actually determined by what trainer is used .



> Sac/ddpg cannot work with icm even with this change, because icm need to have use_rollout_state=False when using off_policy_driver. Perhaps we can leave that as the next PR. As to removing with_experience, @hnyu may not agree because he added that to save the computation for his own intrinsic reward algorithm.

Right, with_experience is mainly for saving computational burden. And I think it will serve a general purpose of differentiating quantities that should be computed once during exp collection from quantities that can be computed&optimized multiples times (e.g., costs) during one training iteration.

hnyu:
> Sac/ddpg cannot work with icm even with this change, because icm need to have use_rollout_state=False when using off_policy_driver. Perhaps we can leave that as the next PR. As to removing with_experience, @hnyu may not agree because he added that to save the computation for his own intrinsic reward algorithm.

Why do you say that "icm need to have use_rollout_state=False when using off_policy_driver" ?

hnyu:
> Sac/ddpg cannot work with icm even with this change, because icm need to have use_rollout_state=False when using off_policy_driver. Perhaps we can leave that as the next PR. As to removing with_experience, @hnyu may not agree because he added that to save the computation for his own intrinsic reward algorithm.

Why do you say that "icm need to have use_rollout_state=False when using off_policy_driver" ?

hnyu:
>    This only affect `_reward_normalizer` in ICM (may calculate intrinsic  reward multiple times ) , and i have tested on `ppo_icm_super_mario_intrinsic_only` ,it does not affect performance (right:master, left:PR_agent1)
> 
> 


Seems to me that the left figure has a better sample efficiency than the right one. Is that just because of initialization randomness?

alf/algorithms/sac_algorithm.py
```diff
@@ -349,14 +349,8 @@ def train_step(self, exp: Experience, state: SacState):
         info = SacInfo(actor=actor_info, critic=critic_info, alpha=alpha_info)
         return PolicyStep(action_distribution, state, info)
 
-    def train_complete(self,
-                       tape: tf.GradientTape,
-                       training_info: TrainingInfo,
-                       weight=1.0):
-        ret = super().train_complete(
-            tape=tape, training_info=training_info, weight=weight)
+    def after_train(self):
```
emailweixu: need to implement the following (see Algorithm._trainable_attributes_to_ignore)
```python
def _trainable_attributes_to_ignore(self):
    return ['_target_critic_network1', '_target_critic_network2']
```

witwolf: added

witwolf:
> 
> Right, `with_experience` is mainly for saving computational burdens and for normalizers&averagers to have correct statistics. And I think it will serve a general purpose of differentiating quantities that should be computed once during exp collection from quantities that can be computed&optimized multiples times (e.g., costs) during one training iteration.

set `calc_intrinsic_reward=False` when collecting experiences in the new commit

alf/algorithms/agent.py
```diff
@@ -141,20 +141,28 @@ def predict(self, time_step: ActionTimeStep, state: AgentState):
...
+        return PolicyStep(action=rl_step.action, state=new_state, info=())
+
+    def rollout(self, time_step: ActionTimeStep, state: AgentState):
         """Rollout for one step."""
         new_state = AgentState()
         info = AgentInfo()
         observation = self._encode(time_step)
 
+        # TODO
+        # avoid computing this when rl_algorithm do not need preprocess experience
```
hnyu: What do you mean by this comment? I believe that for any algorithm, icm rewards should always be computed during rollout.

alf/algorithms/agent.py
```diff
@@ -164,7 +172,32 @@ def rollout(self,
...
+        # TODO
+        # avoid computing this when collecting exps (off policy train)
+        if self._entropy_target_algorithm:
+            et_step = self._entropy_target_algorithm.train_step(rl_step.action)
+            info = info._replace(entropy_target=et_step.info)
+
+        return PolicyStep(action=rl_step.action, state=new_state, info=info)
+
+    def train_step(self, exp: Experience, state):
+        new_state = AgentState()
```
hnyu: Do you need to first convert Experience to ActionTimeStep here?

alf/algorithms/rnd_algorithm.py
```diff
@@ -98,8 +100,11 @@ def train_step(self, inputs, state):
             tf.square(pred_embedding - tf.stop_gradient(target_embedding)),
             axis=-1)
 
-        intrinsic_reward = tf.stop_gradient(loss)
-        intrinsic_reward = self._reward_normalizer.normalize(intrinsic_reward)
+        intrinsic_reward = ()
+        if calc_intrinsic_reward:
+            intrinsic_reward = tf.stop_gradient(loss)
+            intrinsic_reward = self._reward_normalizer.normalize(
+                intrinsic_reward)
```
hnyu: Please also add this check in ICMAlgorithm

alf/algorithms/sac_algorithm.py
```diff
@@ -380,3 +380,6 @@ def _calc_critic_loss(self, training_info):
 
         critic_loss = critic_loss1.loss + critic_loss2.loss
         return LossInfo(loss=critic_loss, extra=critic_loss)
+
+    def _trainable_attributes_to_ignore(self):
+        return ['target_critic_network1', 'target_critic_network2']
```
emailweixu: should be ['_target_critic_network1", '_target_critic_network2']

witwolf: fixed

alf/algorithms/agent.py
witwolf: It's not necessary , and _rl_algorithm.train_step require Experience as input

alf/algorithms/agent.py
witwolf: Removed,  I thought it should be  calculated  at  `train_step` for some off policy algorithms that do not need preprocessing experiences

PR #247, opened 2019-10-28
add a default 'cast_transformer' for 'RLAlgorithm'

witwolf:
algorithm networks usually need  tensors with type tf.float32 as input and observations often have mixed dtypes , so  add a default `cast_transformer` for `RLAlgorithm`  .  And fix a bug with obs_transformer  in `OffPolicyAlgorithm`

### DIFF
--- a/alf/utils/common.py
+++ b/alf/utils/common.py
@@ -414,6 +414,19 @@ def get_global_counter(default_counter=None):
 
 
 @gin.configurable
+def cast_transformer(observation, dtype=tf.float32):
+    """Cast observation
+
+    Args:
+         observation (nested Tensor): observation
+         dtype (Dtype): The destination type.
+    Returns:
+        casted observation
+    """
+    return tf.nest.map_structure(lambda o: tf.cast(o, dtype), observation)
+
+
+@gin.configurable
 def image_scale_transformer(observation, fields=None, min=-1.0, max=1.0):
     """Scale image to min and max (0->min, 255->max)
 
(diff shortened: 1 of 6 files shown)


alf/utils/common.py
```diff
@@ -440,23 +453,28 @@ def _traverse_path(obs, path):
...
-            new_val = _traverse_path(getattr(obs, step), path[1:])
-            return obs._replace(**{step: new_val})
+        if isinstance(obs, tuple):
+            if hasattr(obs, '_fields'):
+                new_val = _traverse_path(getattr(obs, step), path[1:])
+                return obs._replace(**{step: new_val})
+            else:
+                step = int(step)
+                transformed = _traverse_path(obs[step], path[1:])
+                return obs[:step] + (transformed, ) + obs[step + 1:]
```
emailweixu: This will fail if obs is a list

hnyu: Honestly I don't think supporting unnamed tuple is a good choice. Basically the user has to provide a sequence of integer indices which might be error-prone.

alf/utils/common.py
```diff
@@ -440,23 +453,28 @@ def _traverse_path(obs, path):
...
-                             "a dict or namedtuple!"))
+                             "a dict or tuple!"))
 
     fields = fields or [""]
     for field in fields:
         # remove '' in the path
         path = [step for step in field.split(".") if step]
         observation = _traverse_path(observation, path)
-    return observation
+    return cast_transformer(observation, tf.float32)
```
hnyu: This will fail if observation has language input

witwolf: removed

alf/utils/common.py
witwolf: there  is a problem in tf_agent EncodingNetwork that can not take  namedtuple or OrderDict observation as input , so add support for unnamed tuple for now

alf/utils/common_test.py
```diff
@@ -23,15 +23,16 @@ def test_transform_image(self):
         observation = tf.zeros(shape, dtype=tf.uint8)
         common.image_scale_transformer(observation)
 
-        T1 = common.namedtuple('T1', ['x', 'y'])
+        T1 = common.namedtuple('T1', ['x'])
         T2 = common.namedtuple('T2', ['a', 'b', 'c'])
-        T3 = common.namedtuple('T3', ['l', 'm'])
+        T3 = common.namedtuple('T3', ['l'])
```
hnyu: Should not remove 'm' because it's used below: 
```
with self.assertRaises(Exception) as _:	    
            common.image_scale_transformer(
                observation, fields=["x.b.m"])  # empty ()
```

alf/utils/common.py
hnyu: OK, remember to update the function comment

emailweixu: > there is a problem in tf_agent EncodingNetwork that can not take namedtuple or OrderDict observation as input , so add support for unnamed tuple for now

EncodingNetwork does support dict observation:
https://github.com/HorizonRobotics/alf/pull/205/files#diff-4f8caad82255bf76df66e98c7f055370

witwolf: Yes, it supports, i made a mistake before (with a wrong preprocessing_layers) . Revert to the previous implementation in the new commit

alf/utils/common_test.py
witwolf: reverted

PR #253, opened 2019-10-30
Better EntropyTargetAlgorithm and WindowAverager

emailweixu:
Instead of using an optimizer for EntropyTargetAlgorithm, I change to use manual rule for updating log_alpha which is more responsive to the changing dynamics.

### DIFF
--- a/alf/algorithms/entropy_target_algorithm.py
+++ b/alf/algorithms/entropy_target_algorithm.py
@@ -65,12 +65,12 @@ class EntropyTargetAlgorithm(Algorithm):
 
         Args:
             action_spec (nested BoundedTensorSpec): representing the actions.
-            initial_log_alpha (float): initial value for alpha.
+            initial_alpha (float): initial value for alpha.
             target_entropy (float): the lower bound of the entropy. If not
                 provided, a default value proportional to the action dimension
                 is used.
-            slow_update_rate (float):
-            fast_update_rate (float):
+            slow_update_rate (float): minimal update rate for log_alpha
+            fast_update_rate (float): maximum update rate for log_alpha
             min_alpha (float): the minimal value of alpha. If <=0, exp(-100) is
                 used.
             optimizer (tf.optimizers.Optimizer): The optimizer for training. If
--- a/alf/algorithms/entropy_target_algorithm.py
+++ b/alf/algorithms/entropy_target_algorithm.py
@@ -12,20 +12,21 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from collections import namedtuple
-
+from absl import logging
 import gin
 import numpy as np
 import tensorflow as tf
 
-from tf_agents.utils import common as tfa_common
-from alf.algorithms.algorithm import Algorithm, AlgorithmStep, LossInfo
+from tf_agents.trajectories.time_step import StepType
 
+from alf.algorithms.algorithm import Algorithm, AlgorithmStep, LossInfo
 from alf.utils import dist_utils
+from alf.utils.averager import ScalarWindowAverager
+from alf.utils.common import namedtuple, run_if, should_record_summaries
 from alf.utils.dist_utils import calc_default_target_entropy
 
-EntropyTargetLossInfo = namedtuple("EntropyTargetLossInfo",
-                                   ["alpha_loss", "entropy_loss"])
+EntropyTargetLossInfo = namedtuple("EntropyTargetLossInfo", ["entropy_loss"])
+EntropyTargetInfo = namedtuple("EntropyTargetInfo", ["step_type", "loss"])
 
 
 @gin.configurable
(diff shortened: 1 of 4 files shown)


alf/algorithms/entropy_target_algorithm.py
```diff
@@ -35,42 +36,64 @@ class EntropyTargetAlgorithm(Algorithm):
...
     the entropy is not smaller than `target_entropy`.
 
-    It's described in section 5 of the following paper:
-    Haarnoja et al "Soft Actor-Critic Algorithms and Applications" arXiv:1812.05905v2
+    The algorithm has two stages:
+    1. init stage. During this stage, the alpha is not changed. It transitions
+       to fast_stage once entropy drops below `target_entropy`.
+    2. adjust stage. During this stage, log_alpha is adjusted using this formula:
+       ((above + 0.5 * below) * increasing - (below + 0.5 * above) * decreasing) * update_rate.
+       Note that log_alpha will always be increased if entropy is increasing
```
hnyu: Explanation inconsistent with the formula. From the formula, log_alpha will always be increased if the entropy is increasing (whether it's above or below the target), otherwise it will always be decreased.

alf/algorithms/entropy_target_algorithm.py
```diff
@@ -35,42 +36,64 @@ class EntropyTargetAlgorithm(Algorithm):
     It tries to adjust the entropy regularization (i.e. alpha) so that the
     the entropy is not smaller than `target_entropy`.
 
-    It's described in section 5 of the following paper:
-    Haarnoja et al "Soft Actor-Critic Algorithms and Applications" arXiv:1812.05905v2
+    The algorithm has two stages:
+    1. init stage. During this stage, the alpha is not changed. It transitions
+       to fast_stage once entropy drops below `target_entropy`.
```
hnyu: fast_stage -> adjust_stage

alf/algorithms/entropy_target_algorithm.py
emailweixu: The comment does not contradict with the formula. It says log_alpha will increase even in the situation of the entropy being below the target as long as the entropy is increasing.

hnyu: I mean for the sentence "Same reason for always decreasing log_alpha when the entropy is above the target entropy."

hnyu: And "even when the entropy is below the target entropy" should be "even when the entropy is above the target" because usually people would expect that if the entropy is above the target, alpha should be decreased. Here you are using the word "even when" to represent a different situation from the common sense.

hnyu: You mean "even in the situation of the entropy being above the target"? Because if it's below, log_alpha is expected to increase (so no need to use the word "even"). Or are you talking about the "entropy loss" (negative) instead of "entropy" (positive)?

alf/algorithms/entropy_target_algorithm.py
```diff
@@ -35,42 +36,64 @@ class EntropyTargetAlgorithm(Algorithm):
...
+    2. adjust stage. During this stage, log_alpha is adjusted using this formula:
+       ((above + 0.5 * below) * increasing - (below + 0.5 * above) * decreasing) * update_rate.
+       Note that log_alpha will always be increased if entropy is increasing
+       even when the entropy is below the target entropy. This is to prevent
+       overshooting log_alpha to a too big value. Same reason for always
+       decreasing log_alpha when the entropy is above the target entropy.
```
hnyu: "always decreasing ... when the entropy is decreasing"

alf/algorithms/entropy_target_algorithm.py
```diff
@@ -35,42 +36,64 @@ class EntropyTargetAlgorithm(Algorithm):
...
 
         Args:
             action_spec (nested BoundedTensorSpec): representing the actions.
-            initial_log_alpha (float): initial value for log(alpha). The default
-                value corresponds to alpha ~ 0.05.
+            initial_log_alpha (float): initial value for alpha.
             target_entropy (float): the lower bound of the entropy. If not
                 provided, a default value proportional to the action dimension
                 is used.
+            slow_update_rate (float):
```
hnyu: Comment on slow_update_rate (the min possible update rate)

alf/algorithms/entropy_target_algorithm.py
```diff
@@ -35,42 +36,64 @@ class EntropyTargetAlgorithm(Algorithm):
...
+                 fast_update_rate=np.log(2),
                  min_alpha=1e-4,
                  debug_summaries=False):
         """Create an EntropyTargetAlgorithm
 
         Args:
             action_spec (nested BoundedTensorSpec): representing the actions.
-            initial_log_alpha (float): initial value for log(alpha). The default
-                value corresponds to alpha ~ 0.05.
+            initial_log_alpha (float): initial value for alpha.
```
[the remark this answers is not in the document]
emailweixu: fixed

alf/algorithms/entropy_target_algorithm.py
emailweixu: This means alpha is either doubled or halved.

alf/algorithms/entropy_target_algorithm.py
```diff
@@ -35,42 +36,64 @@ class EntropyTargetAlgorithm(Algorithm):
...
     """
 
     def __init__(self,
                  action_spec,
-                 initial_log_alpha=-3.0,
+                 initial_alpha=0.01,
                  target_entropy=None,
-                 optimizer=None,
+                 slow_update_rate=0.01,
+                 fast_update_rate=np.log(2),
```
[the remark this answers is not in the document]
emailweixu: This means alpha is either doubled or halved.

alf/algorithms/entropy_target_algorithm.py
emailweixu: done

PR #265, opened 2019-11-13
simplify trac

emailweixu:
Using PolicyStep.info to store necessary information instead of DataBuffer


alf/algorithms/trac_algorithm.py
```diff
@@ -61,9 +63,6 @@ class TracAlgorithm(ActorCriticAlgorithm):
 
     def __init__(self,
                  action_spec,
-                 observation_spec,
```
witwolf: It seems that other parameters can also be removed

emailweixu: ActorCriticAlgorithm needs other parameters.

witwolf: i see

PR #269, opened 2019-11-15
modify trac to support other ac algorithms

witwolf:


alf/algorithms/trac_algorithm.py
```diff
@@ -59,44 +60,33 @@ class TracAlgorithm(ActorCriticAlgorithm):
...
-                and network state.
-            value_network (Network): A function that returns value tensor from neural
-                net predictions for each observation given observation and nwtwork
-                state.
-            loss (None|ActorCriticLoss): an object for calculating loss. If
-                None, a default loss of class loss_class will be used.
-            loss_class (type): the class of the loss. The signature of its
-                constructor: loss_class(action_spec, debug_summaries)
-            optimizer (tf.optimizers.Optimizer): The optimizer for training
+            ac_algorithm_cls (type): Actor Critic Algorithm cls (AC, PPO, SAC...)
```
emailweixu: DDPG can also be supported. We only need to support the distance between two deterministic action distribution.

alf/algorithms/trac_algorithm.py
```diff
@@ -170,21 +175,32 @@ def _update_total_dists(new_action, exp, total_dists):
...
         num_steps = exp_array.step_type.size()
+        # element_shape for `TensorArray` can be (None, ...)
+        batch_size = tf.shape(exp_array.step_type.read(0))[0]
         state = tf.nest.map_structure(lambda x: x.read(0), exp_array.state)
-        batch_size = exp_array.step_type.element_shape[0]
         # exp_array.state is no longer needed
         exp_array = exp_array._replace(state=())
         initial_state = common.zero_tensor_from_nested_spec(
-            self.train_state_spec.actor, batch_size)
+            self.train_state_spec, batch_size)
```
emailweixu: The structure of initial_state is train_state_spec. But the state from predict is predict_state_spec. They might be different and cause line 188 to fail.

alf/algorithms/trac_algorithm.py
```diff
@@ -157,8 +154,16 @@ def _calc_change(self, exp_array):
         def _dist(d1, d2):
             if isinstance(d1, tfp.distributions.Categorical):
                 return tf.reduce_sum(tf.square(d1.logits - d2.logits), axis=-1)
+            elif isinstance(d1, SquashToSpecNormal):
```
emailweixu: Put the following at line 155 instead:
```python
if isinstance(d1, SquashToSpecNormal):
    d1 = d1.input_distribution
    d2 = d2.input_distrubution
```

alf/algorithms/trac_algorithm.py
```diff
@@ -170,21 +175,32 @@ def _update_total_dists(new_action, exp, total_dists):
...
             exp = tf.nest.map_structure(lambda x: x.read(t), exp_array)
             state = common.reset_state_if_necessary(
                 state, initial_state, exp.step_type == StepType.FIRST)
-            new_action, state = self._actor_network(
-                exp.observation, step_type=exp.step_type, network_state=state)
+            time_step = ActionTimeStep()
+            time_step = time_step._replace(
+                observation=exp.observation,
+                step_type=exp.step_type,
+            )
```
emailweixu: can be simplified:
```python
            time_step = ActionTimeStep(
                observation=exp.observation,
                step_type=exp.step_type,
            )
```

witwolf: simplified

alf/algorithms/trac_algorithm.py
witwolf: done

alf/algorithms/trac_algorithm.py
witwolf: changed to  `rollout` instead of `predict`

alf/algorithms/trac_algorithm.py
witwolf: added

witwolf: `kl_divergence` implemented in `SquashToSpecNormal` checks action mean and magnitude for two distributions, and it fails in graph mode,  have not found the reason yet.

alf/algorithms/trac_algorithm.py
```diff
@@ -68,15 +69,15 @@ def __init__(self,
...
+            ac_algorithm_cls (type): Actor Critic Algorithm cls.
             action_dist_clip_per_dim (float): action dist clip per dimension
             debug_summaries (bool): True if debug summaries should be created.
             name (str): Name of this algorithm.
         """
         ac_algorithm = ac_algorithm_cls(
             action_spec=action_spec, debug_summaries=debug_summaries)
 
-        assert isinstance(ac_algorithm._actor_network, DistributionNetwork)
+        assert hasattr(ac_algorithm, '_actor_network')
```
emailweixu: TracAlgorithm does not use ac_algorithm._actor_network.

alf/algorithms/trac_algorithm.py
```diff
@@ -187,14 +192,12 @@ def _update_total_dists(new_action, exp, total_dists):
...
                 state, initial_state, exp.step_type == StepType.FIRST)
-            time_step = ActionTimeStep()
-            time_step = time_step._replace(
-                observation=exp.observation,
-                step_type=exp.step_type,
-            )
-            policy_step = self._ac_algorithm.predict(
+            time_step = ActionTimeStep(
+                observation=exp.observation, step_type=exp.step_type)
+            policy_step = self._ac_algorithm.rollout(
```
emailweixu: Only action_distribution is needed here. rollout() has extra computations such as value.

alf/algorithms/trac_algorithm.py
witwolf: `TrustedUpdater` update parameters for  `ac_algorithm._actor_network`  (L106)

alf/algorithms/trac_algorithm.py
witwolf: previous methods
1.  `_ac_algorithm_actor_network()`  assume `state` has a filed named `actor`
2. `_ac_algorithm.predict` use predict_state, but  state stored in `exp_array`

emailweixu: Yes, I know this. Or do you think tensorflow can optimized away the unused computation? If so, using rollout is the cleanest way. If not, I think we still should avoid the unnecessary computation. Perhaps we can convert the train_state to predict_state before putting it to TracInfo at L110 and L117?

witwolf: converted `train_state` to `predict_state` .

alf/algorithms/algorithm.py
```diff
@@ -313,6 +313,10 @@ def train_state_spec(self):
         """Returns the RNN state spec for train_step()."""
         return self._train_state_spec
 
+    def train_state_2_predict_state(self, state):
+        """Convert RNN state for train_step() to RNN state for predict()"""
```
emailweixu: Use a better name: convert_train_state_to_predict_state
And assert:
tf.nest.assert_same_structure(train_state_spec, predict_state_spec)

witwolf: renamed

PR #272, opened 2019-11-20
increment global step every update

witwolf:


hnyu:
The original definition of an "iteration" is once every rollout, I believe. This will change that definition. But from the perspective of summary, I think this makes more sense. @emailweixu

hnyu:
If there are multiple summaries between two rollouts, will the reward curve look weird? For example, duplicate points that cause flat curve segments.

witwolf:
> If there are multiple summaries between two rollouts, will the reward curve look weird? For example, duplicate points that cause flat curve segments.

Yes,  it will happen for metrics curve （AverageReturn, AverageEpisodeLength,EvnStep .. Update multiple times without collecting new data) . Modifying this is mainly because it’s helpful to summary for every update (summay_interval=1) when debugging

emailweixu:
Perhaps separate metric summary fro training summary?

witwolf:
> Perhaps separate metric summary fro training summary?

separated from training summary, only summary metric after rollout

PR #273, opened 2019-11-22
Conditional mi estimator

emailweixu:
Add support for estimating conditional mutual information in MIEstimator


alf/algorithms/mi_estimator.py
```diff
@@ -72,6 +86,11 @@ class MIEstimator(Algorithm):
     assumption that y samples from one batch are independent. If the additional
     memory is not a concern, we recommend 'buffer' sampler so that there is no
     need to worry about the assumption of independence.
+
+    MIEstimator can be also used to estimate conditional mutual information
+    MI(X,Y|Z). In this case, you should let `x` to represent X and Z, and `y` to
```
hnyu: This statement seems incorrect, as MI(X, Y | Z) is not equal to MI([X,Z], Y). Because 'x' and 'y' in the former both condition on 'z' in the denominator while only one of them conditions on 'z' in the latter.  This is same for both 'JSD' and 'ML'

alf/algorithms/mi_estimator_test.py
```diff
@@ -17,10 +17,55 @@
 from absl import logging
 from absl.testing import parameterized
 import tensorflow as tf
+import tensorflow_probability as tfp
 
+from tf_agents.networks.network import Network
```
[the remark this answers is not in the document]
emailweixu: EncodingNetwork is Network, but not keras.Model

alf/algorithms/mi_estimator.py
emailweixu: The way it is used make it estimating MI(X, Y|Z). For MI([X,Z], Y), y_distritution should sample the marginal distribution P(Y). For MI(X,Y|Z), y_distribution should sample P(Y|z)

hnyu: ok I see.

PR #274, opened 2019-11-25
training terminates based on steps

hnyu:
Sometimes when running comparison experiments, it's better to run a fixed number of env steps (frames) regardless of the training configuration.


alf/trainers/policy_trainer.py
```diff
@@ -86,7 +87,10 @@ def __init__(self,
             algorithm_ctor (Callable): callable that create an
                 `OffPolicyAlgorithm` or `OnPolicyAlgorithm` instance
             random_seed (None|int): random seed, a random seed is used if None
-            num_iterations (int): number of update iterations
+            num_iterations (int): number of update iterations (ignored if 0)
+            num_env_steps (int): number of environment steps (ignored if 0). The
+                total number of FRAMES will be (`num_env_steps`*`frame_skip`) for
```
emailweixu: need to explain what frame_skip is.

PR #275, opened 2019-11-26
add max entropy for EntropyTargetAlgorithm

hnyu:
For different games, it's impractical to have a fixed init_alpha because the value might be too large for having meaningful exploration in some cases, depending on the action spec and rewards. The current target entropy algorithm only ensures the entropy won't be too low to deviate from a target. I added a max entropy limit to encourage exploitation even when the init alpha is big. 

Alpha:
EntropyTargetAlgorithm_alpha
(final stable alpha: ~3e-3)

Stage:
EntropyTargetAlgorithm_stage

Entropy:
EntropyTargetAlgorithm_avg_entropy

### DIFF
--- a/alf/algorithms/entropy_target_algorithm.py
+++ b/alf/algorithms/entropy_target_algorithm.py
@@ -24,6 +24,7 @@ from alf.utils import dist_utils
 from alf.utils.averager import ScalarWindowAverager
 from alf.utils.common import namedtuple, run_if, should_record_summaries
 from alf.utils.dist_utils import calc_default_target_entropy
+from alf.utils.dist_utils import calc_default_max_entropy
 
 EntropyTargetLossInfo = namedtuple("EntropyTargetLossInfo", ["entropy_loss"])
 EntropyTargetInfo = namedtuple("EntropyTargetInfo", ["step_type", "loss"])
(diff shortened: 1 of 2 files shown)


alf/algorithms/entropy_target_algorithm.py
```diff
@@ -144,6 +168,19 @@ def calc_loss(self, training_info: EntropyTargetInfo):
         avg_entropy = self._avg_entropy.average(entropy)
 
         def _init():
+            below = avg_entropy < self._max_entropy
+            increasing = tf.cast(avg_entropy > prev_avg_entropy, tf.float32)
+            # -1 * increasing + 0.5 * (1 - increasing)
+            update_rate = (
+                0.5 - 1.5 * increasing) * self._very_slow_update_rate
+            # cannot use tf.cond because the two return types are different
+            run_if(below, lambda: self._stage.assign(-1))
```
emailweixu: We can avoid `run_if` or `cond`:
```
self._stage.assign_sub(below)
```

alf/algorithms/entropy_target_algorithm.py
```diff
@@ -144,6 +168,19 @@ def calc_loss(self, training_info: EntropyTargetInfo):
...
+            run_if(
```
emailweixu: We can simply the code by removing run_if. An extra ajust of log_alpha does not matter.

alf/utils/dist_utils.py
```diff
@@ -157,3 +157,23 @@ def calc_default_target_entropy(spec):
...
+    """Calc default max entropy
+    Args:
+        spec (TensorSpec): action spec
+    Returns:
+        A default max entropy for adjusting the entropy weight
+    """
+    if tensor_spec.is_continuous(spec):
+        # pick a very conservative high value in the continuous case
+        # (upper bound for free; better than nothing)
+        e = 100
```
emailweixu: This is too  big.  The entropy for Gaussian is 0.5*log(2*pi*e*sigma^2). Even if sigma=10, it's just 3.72. So I would say 2 per dimension might be reasonable.

alf/algorithms/entropy_target_algorithm.py
hnyu: updated; should be assign_add(below) (from -2 to -1)

PR #278, opened 2019-11-27
add logo

hnyu:
I made some changes based on the original one using GIMP


hnyu:
> Nice design! One minor comment: visually, the current design seems a bit unbalanced and leans towards left.

yeah I prefer left-aligning the letters. Are you suggesting center alignment?

hnyu:
> Nice design! One minor comment: visually, the current design seems a bit unbalanced and leans towards left.

Or maybe I can decrease the size of the subject line

PR #279, opened 2019-12-03
merlin for dmlab env collect good objects

witwolf:
merlin example on simple dmlab env `room_collect_good_objects` , it cost about 10M env (about 15 hours ) steps to get episode returns greater than 9 

i also tested it on other complex dmlab environment, such as `water_maze` and `explore_goal_locations`, have not get good results yet, they need much more env steps (up to 10^9 total env steps for 192 parallel runner mentioned in the paper) 

It should be noted that to simplify implementation , ours is a little different from  the paper
1. we only encode/decode for image 
2. we do not add log p(a_i) as input to value net and has no q network
 
Main purpose of this pr: Verify that the implementation is correct and it works on some task


alf/utils/resnet50.py
```diff
@@ -0,0 +1,91 @@
+# Copyright (c) 2019 Horizon Robotics. All Rights Reserved.
```
hnyu: The filename would be better called "resnet50_block.py"

alf/utils/resnet50.py
```diff
@@ -0,0 +1,91 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import tensorflow as tf
+from tensorflow.keras import layers
+
+
+def conv_block(input_tensor,
```
hnyu: Will it be better if this function is a class that derives from keras.Model?

alf/utils/resnet50.py
```diff
@@ -0,0 +1,91 @@
...
+    bn_name_base = 'bn' + str(stage) + block + '_branch'
+
+    conv_fn = layers.Conv2DTranspose if transpose else layers.Conv2D
+
+    x = conv_fn(
+        filters1, (1, 1),
+        strides=strides,
+        kernel_initializer='he_normal',
+        name=conv_name_base + '2a')(input_tensor)
+    x = layers.BatchNormalization(axis=bn_axis, name=bn_name_base + '2a')(x)
```
emailweixu: BatchNormalization need to differentiate training or inference mode. Need to have training=False for inference.

alf/algorithms/merlin_algorithm.py
```diff
@@ -440,50 +443,89 @@ def calc_loss(self, training_info: TrainingInfo):
...
+    2.1.1 of "Unsupervised Predictive Memory in a Goal-Directed Agent"
+    """
+
+    def __init__(self, input_tensor_spec, name='ResnetEncodingNetwork'):
+        """Create a `ResnetEncodingNetwork` instance
+
+        Args:
+            input_tensor_spec (TensorSpec|nested TensorSpec): input observations spec.
+        """
+        super().__init__(input_tensor_spec, (), name)
```
emailweixu: specify the argument name for `()` for readability

alf/utils/resnet50.py
emailweixu: Right, let's have an uniform way to construct model: create the layers or Variables in `__init__` (of keras.Model or tf.Module) and use them later in call() or other functions. It's easier to reason about the owner of the Variables.

alf/examples/merlin_dmlab_collect_good_objects.gin
```diff
@@ -0,0 +1,77 @@
...
+MerlinAlgorithm.decoders=@img_dec/DecodingAlgorithm()
+MerlinAlgorithm.latent_dim=%LATENT_DIM
+MerlinAlgorithm.lstm_size=(256, 256)
+MerlinAlgorithm.memory_size=1350
+
+# ActorCriticLoss.gamma = 0.96
+# ActorCriticLoss.use_gae=True
+# ActorCriticLoss.use_td_lambda_return=True
+# ActorCriticLoss.td_lambda=0.9
+# ActorCriticLoss.entropy_regularization=0.01
```
emailweixu: do you need these parameters?

alf/examples/merlin_dmlab_collect_good_objects.gin
```diff
@@ -0,0 +1,77 @@
...
+# `DeepmindLab` are required,
+#   see `https://github.com/deepmind/lab` to build `DeepmindLab`
+
+# environment config
+import alf.environments.suite_dmlab
+import alf.environments.wrappers
+
+# environment config
+create_environment.env_load_fn=@suite_dmlab.load
+create_environment.env_name="contributed/dmlab30/rooms_collect_good_objects_train"
```
emailweixu: What is the performance on  rooms_collect_good_objects_test?
For some tasks, we need to be able to specify different env for evaluation.

alf/examples/merlin_dmlab_collect_good_objects.gin
witwolf: these are default parameters mentioned in the paper , i did not use them in this experiment (it has better performance when not use them)

alf/utils/resnet50.py
witwolf: okay, i'll modify it

alf/utils/resnet50.py
witwolf: should we add addition parameter `training` flag for `Network.call` ?

alf/examples/merlin_dmlab_collect_good_objects.gin
```diff
@@ -0,0 +1,77 @@
+# MERLIN on deepmindlab rooms_collect_good_objects
+
+import alf.algorithms.merlin_algorithm
+import alf.algorithms.decoding_algorithm
+import alf.trainers.on_policy_trainer
+
+# `DeepmindLab` are required,
+#   see `https://github.com/deepmind/lab` to build `DeepmindLab`
```
emailweixu: adding more detail for build dmlab data package:
https://github.com/deepmind/lab/blob/master/data/brady_konkle_oliva2008/README.md
python package:
https://github.com/deepmind/lab/blob/master/python/pip_package/README.md

alf/examples/merlin_dmlab_collect_good_objects.gin
witwolf: evaluate for a intermediate checkpoint ( evaluate episode returns 7-8 ),
it seems good 

rooms_collect_good_objects_train

rooms_collect_good_objects_test

alf/utils/resnet50.py
witwolf: renamed

alf/utils/resnet50.py
witwolf: add a class `BottleneckBlock` inherited from `tf.keras.models.Model`

alf/examples/merlin_dmlab_collect_good_objects.gin
witwolf: > For some tasks, we need to be able to specify different env for evaluation.

Let's do this at another pr

alf/utils/resnet50.py
witwolf: to differentiate training or inference,  it's better to add add a parameter `training` flag for `Network.call` to handle all layers (Dropout also needs though we haven't used it yet)

alf/examples/merlin_dmlab_collect_good_objects.gin
```diff
@@ -5,7 +5,11 @@ import alf.algorithms.decoding_algorithm
 import alf.trainers.on_policy_trainer
 
 # `DeepmindLab` are required,
-#   see `https://github.com/deepmind/lab` to build `DeepmindLab`
+#   see `https://github.com/deepmind/lab/blob/master/python/pip_package/README.md`
+#   to build `DeepmindLab` Python module
+#  Stimuli dataset should downloaded
```
emailweixu: should be downloaded before building the python module.

alf/examples/merlin_dmlab_collect_good_objects.gin
emailweixu: > evaluate for a intermediate checkpoint ( evaluate episode returns 7-8 ),
> it seems good
Do you use play to do this. But for play, the batch size is 1, without training=False, BatchNormalization might be bad?

alf/examples/merlin_dmlab_collect_good_objects.gin
emailweixu: Then please make a comment about this.

alf/utils/resnet50_block.py
```diff
@@ -0,0 +1,105 @@
...
+import tensorflow as tf
+from tensorflow.keras import layers
+
+
+class BottleneckBlock(tf.keras.models.Model):
+    def __init__(self,
+                 kernel_size,
+                 filters,
+                 stage,
+                 block,
```
emailweixu: No need to have `stage` and `block`, just `name` is enough.

alf/algorithms/merlin_algorithm.py
```diff
@@ -440,50 +443,97 @@ def calc_loss(self, training_info: TrainingInfo):
...
+
+        for i, stride in enumerate([2, 1, 2, 1, 2, 1]):
+            enc_layers.append(
+                resnet50_block.BottleneckBlock(
+                    kernel_size=(3, 3),
+                    filters=(64, 32, 64),
+                    stage=i,
+                    block='block',
+                    strides=stride,
+                    name='block%d' % i))
```
emailweixu: name='%s/block%d' % (name, i)

alf/algorithms/merlin_algorithm.py
```diff
@@ -440,50 +443,97 @@ def calc_loss(self, training_info: TrainingInfo):
...
+
+        enc_layers.extend([
+            tf.keras.layers.Flatten(),
+            tf.keras.layers.Dense(500, activation='tanh')
```
emailweixu: name=...

alf/algorithms/merlin_algorithm.py
```diff
@@ -440,50 +443,97 @@ def calc_loss(self, training_info: TrainingInfo):
...
+        for i, stride in enumerate(reversed([2, 1, 2, 1, 2, 1])):
+            dec_layers.append(
+                resnet50_block.BottleneckBlock(
+                    kernel_size=(3, 3),
+                    filters=(64, 32, 64),
+                    strides=stride,
+                    stage=i,
+                    block='deconv',
+                    transpose=True,
+                    name='block%d' % i))
```
emailweixu: name='%s/block%d' % (name, i)

alf/algorithms/merlin_algorithm.py
```diff
@@ -440,50 +443,97 @@ def calc_loss(self, training_info: TrainingInfo):
...
+
+        dec_layers.append(
+            tf.keras.layers.Conv2DTranspose(
+                filters=3, kernel_size=1, activation='sigmoid'))
```
emailweixu: name

alf/algorithms/merlin_algorithm.py
```diff
@@ -440,50 +443,97 @@ def calc_loss(self, training_info: TrainingInfo):
...
+
+        Args:
+             input_tensor_spec (TensorSpec): input latent spec.
+        """
+        super().__init__(input_tensor_spec, state_spec=(), name=name)
+
+        dec_layers = []
+        dec_layers.extend([
+            tf.keras.layers.Dense(500, activation='relu'),
+            tf.keras.layers.Dense(8 * 8 * 64, activation='relu'),
```
emailweixu: need `name` so that names in variable summaries are understandable..

emailweixu:
Let's also add the replay video to the readme

alf/examples/merlin_dmlab_collect_good_objects.gin
witwolf: > Do you use play to do this.

Yes, and the checkpoint is on about 5M env step, performance is not good yet , 
 also played after converged (15M), result is much better, average episode returns about 9.4   

> But for play, the batch size is 1, without training=False, BatchNormalization might be bad?

i am not sure about this

alf/utils/resnet50_block.py
witwolf: removed

alf/algorithms/merlin_algorithm.py
```diff
@@ -440,50 +443,94 @@ def calc_loss(self, training_info: TrainingInfo):
...
+
+        enc_layers = []
+
+        for i, stride in enumerate([2, 1, 2, 1, 2, 1]):
+            enc_layers.append(
+                resnet50_block.BottleneckBlock(
+                    kernel_size=(3, 3),
+                    filters=(64, 32, 64),
+                    strides=stride,
+                    name='block%d' % i))
```
emailweixu: should `name` of this Network be included?

witwolf: it's under the scope of `ResnetEncodingNetwork`,  L453

README.md
```diff
@@ -84,3 +84,11 @@ python -m alf.bin.play --root_dir=LOG_DIR
   Python package gym-retro>=0.7.0 is required for this experiment and also a suitable `SuperMarioBros-Nes` rom should be obtained and imported (roms are not included in gym-retro). See [this doc](https://retro.readthedocs.io/en/latest/getting_started.html#importing-roms) on how to import roms.
 
   <img src="alf/examples/icm_super_mario_intrinsic_only.png" width = "300" height ="200" alt="super-mario-training-curve"/> <img src="alf/examples/icm_super_mario_intrinsic_only.gif" width = "300" height ="200" alt="super-mario-video"/>
+  
+
+### Merlin
+* [Room collect good objects](alf/examples/merlin_dmlab_collect_good_objects.gin).
```
hnyu: This bullet should be called "Collect good objects in rooms"? (current syntax seems strange). Also please add a sentence briefly summarizing what algorithm (MERLINE in this case) solves what kind of problem, just like in the ICM case above.

README.md
```diff
@@ -84,3 +84,11 @@ python -m alf.bin.play --root_dir=LOG_DIR
   Python package gym-retro>=0.7.0 is required for this experiment and also a suitable `SuperMarioBros-Nes` rom should be obtained and imported (roms are not included in gym-retro). See [this doc](https://retro.readthedocs.io/en/latest/getting_started.html#importing-roms) on how to import roms.
 
   <img src="alf/examples/icm_super_mario_intrinsic_only.png" width = "300" height ="200" alt="super-mario-training-curve"/> <img src="alf/examples/icm_super_mario_intrinsic_only.gif" width = "300" height ="200" alt="super-mario-video"/>
+  
+
+### Merlin
+* [Room collect good objects](alf/examples/merlin_dmlab_collect_good_objects.gin). 
+  `DeepmindLab` are required,  Follow the instruction at [DeepmindLab](https://github.com/deepmind/lab/blob/master/python/pip_package/README.md) to install the environment.
```
hnyu: "are" -> "is"

alf/algorithms/merlin_algorithm.py
emailweixu: How is the scope used? Don’t you need to use ‘with self.name_scope’ in order to use the scope?

witwolf: `ResnetEncodingNetwork` is a subclass of `keras_network.Network` ,  these variables are created in `ResnetEncodingNetwork.__init__`, they are sub models

witwolf: looked into source code,  variables (weights) of a Layer or Model are created at build phase (or first call),  and `Layer.call` or `Model.call` is  called under its scope, 

```python
with backend.name_scope(self._name_scope()):
  ...
  self.call(...)
  ...
```

emailweixu:
I checked [Collect Good Objects](https://github.com/deepmind/lab/tree/master/game_scripts/levels/contributed/dmlab30#collect-good-objects) task. It seems that this task doesn't not need much memory ability. It only requires that the agent can differentiate good and bad objects. And the types of good objects are fixed across episodes. We should test on a task where memory is really necessary.

witwolf:
> I checked [Collect Good Objects](https://github.com/deepmind/lab/tree/master/game_scripts/levels/contributed/dmlab30#collect-good-objects) task. It seems that this task doesn't not need much memory ability. It only requires that the agent can differentiate good and bad objects. And the types of good objects are fixed across episodes. We should test on a task where memory is really necessary.

ok,  i am still trying it on  [Water Maze](https://github.com/deepmind/lab/tree/master/game_scripts/levels/contributed/dmlab30#watermaze) ,performance was not good,  got episode return about 25~30 after 30M env steps (> 200 mentioned in the paper)

emailweixu:
> ok, i am still trying it on [Water Maze](https://github.com/deepmind/lab/tree/master/game_scripts/levels/contributed/dmlab30#watermaze) ,performance was not good, got episode return about 25~30 after 30M env steps (> 200 mentioned in the paper)

I think in order to solve this task, reward should be part of the observation so that the agent can possibly associate reward with the target object.

witwolf:
> I think in order to solve this task, reward should be part of the observation so that the agent can possibly associate reward with the target object.

I'll complete the unfinished part of the algorithm

PR #282, opened 2019-12-07
put data buffer on CPU

hnyu:
tested that this will only consume CPU memory by default

### DIFF
--- a/alf/utils/data_buffer.py
+++ b/alf/utils/data_buffer.py
@@ -50,20 +50,19 @@ class DataBuffer(tf.Module):
                 shape=shape,
                 trainable=False)
 
-        self._current_size = tfa_common.create_variable(
-            name=name + "/size",
-            initializer=0,
-            dtype=tf.int32,
-            shape=(),
-            trainable=False)
-        self._current_pos = tfa_common.create_variable(
-            name=name + "/pos",
-            initializer=0,
-            dtype=tf.int32,
-            shape=(),
-            trainable=False)
-
         with tf.device(self._device):
+            self._current_size = tfa_common.create_variable(
+                name=name + "/size",
+                initializer=0,
+                dtype=tf.int32,
+                shape=(),
+                trainable=False)
+            self._current_pos = tfa_common.create_variable(
+                name=name + "/pos",
+                initializer=0,
+                dtype=tf.int32,
+                shape=(),
+                trainable=False)
             self._buffer = tf.nest.map_structure(_create_buffer, data_spec)
             # TF 2.0 checkpoint does not handle tuple. We have to convert
             # _buffer to a flattened list in order to make the checkpointer save the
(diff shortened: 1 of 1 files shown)


hnyu:
> Also need to change the places where DataBuffer is used to allow the choice of device or use the original device.

In most cases there is no need to store data on GPU because there won't be an efficiency issue of storing on CPU. CPU is also the default option in tf_agents replay buffers.

emailweixu:
> > Also need to change the places where DataBuffer is used to allow the choice of device or use the original device.
> 
> In most cases there is no need to store data on GPU because there won't be an efficiency issue of storing on CPU. CPU is also the default option of tf_agents replay buffers and we didn't bother changing the device when using their buffers.
> 
> I think that unless the replay buffer is frequently called (e.g., every rollout step, which is not the case for our current usage), storing on CPU is good enough. If someone wants to have a high efficiency, then he can provide the device when writing new code.

alf/utils/data_buffer.py
```diff
@@ -74,15 +82,17 @@ def add_batch(self, batch):
...
         n = tf.minimum(batch_size, self._capacity)
         indices = tf.range(self._current_pos,
                            self._current_pos + n) % self._capacity
         indices = tf.expand_dims(indices, axis=-1)
-        tf.nest.map_structure(
-            lambda buf, bat: buf.scatter_nd_update(indices,
-                                                   tf.stop_gradient(bat[-n:])),
-            self._buffer, batch)
+
+        with tf.device(self._device):
```
emailweixu: It's better to put this at line 86

alf/utils/data_buffer.py
```diff
@@ -115,11 +125,14 @@ def get_batch_by_indices(self, indices):
         """
         indices = (indices +
                    (self._current_pos - self._current_size)) % self._capacity
-        return tf.nest.map_structure(
-            lambda buffer: tf.gather(buffer, indices, axis=0), self._buffer)
+        with tf.device(self._device):
```
emailweixu: move to line 126

alf/utils/data_buffer.py
```diff
@@ -57,12 +62,15 @@ def _create_buffer(tensor_spec):
...
             shape=(),
             trainable=False)
-        # TF 2.0 checkpoint does not handle tuple. We have to convert
-        # _buffer to a flattened list in order to make the checkpointer save the
-        # content in self._buffer. This seems to be fixed in
-        # tf-nightly-gpu 2.1.0.dev20191119
-        # TODO: remove this after upgrading tensorflow
-        self._flattened_buffer = tf.nest.flatten(self._buffer)
+
+        with tf.device(self._device):
```
emailweixu: put this before line 53

hnyu: The reason why I didn't include self._current_size and self._current_pos in tf.device is that when calling current_size(), clear() and pop(), they don't have to be transferred from one device to another. basically, only using device when the buffer content is accessed.

emailweixu: At line 94, if indices is on GPU, but buf on CPU,  the system need to switch and transfer data between GPU and CPU, which  is slow.  And for the scalar like _current_size, they are mostly used  in CPU code. GPU can only do parallel computations like vector and matrix math.

alf/utils/data_buffer.py
```diff
@@ -113,13 +122,16 @@ def get_batch_by_indices(self, indices):
...
+            indices = (indices + (self._current_pos - self._current_size)
+                       ) % self._capacity
+            return tf.nest.map_structure(
+                lambda buffer: tf.gather(buffer, indices, axis=0),
+                self._buffer)
 
     def get_all(self):
-        return self._buffer
+        with tf.device(self._device):
```
emailweixu: no computation or allocation here. no need to specify device

PR #283, opened 2019-12-09
add BaseObservationWrapper that support nested transformation

witwolf:


alf/environments/wrappers.py
```diff
@@ -12,187 +12,166 @@
...
-from absl import logging
-
+import gin
 from tf_agents.environments import wrappers
 from tf_agents.trajectories.time_step import StepType
+from alf.utils import common
+
+
+def traverse_transform_space(space, levels, field, func):
+    """Transform space
```
emailweixu: Need more explanation on what this function does: given the argument `space`, `levels`, `field` and `func` what kind of output it generates?

alf/environments/wrappers.py
```diff
@@ -12,187 +12,166 @@
...
+
+    Args:
+         space (gym.Space): space to be transformed
+         levels (list[str]): levels
```
emailweixu: comment what the valid values for levels

alf/environments/wrappers.py
```diff
@@ -12,187 +12,166 @@
...
+         field (str): filed name, a multi-step path denoted by "A.B.C"
+         func (Callable): transform func
```
emailweixu: comment its expected arguments and outputs

alf/environments/wrappers.py
```diff
@@ -12,187 +12,166 @@
...
+        return observation
+
+    def transform_space(self, observation_space, field=None):
+        """Transform space
+
+        Subclass should override this to perform transformation
+
+        Args:
+             observation_space (gym.Space): space to be transformed
+             field (str): field to be transformed
```
emailweixu: comment the format of `field`

alf/environments/wrappers.py
```diff
@@ -12,187 +12,166 @@
...
+        return observation_space
+
+    def transform_observation(self, observation, field=None):
+        """Transform observation
+
+        Subclass should override this to perform transformation
+
+        Args:
+             observation (ndarray): observation to be transformed
+             field (str): field to be transformed
```
emailweixu: Returns:

alf/environments/wrappers.py
```diff
@@ -12,187 +12,166 @@
...
+        """
+        return observation_space
+
+    def transform_observation(self, observation, field=None):
+        """Transform observation
+
+        Subclass should override this to perform transformation
+
+        Args:
+             observation (ndarray): observation to be transformed
```
emailweixu: The type should be ndarray or dict?

alf/environments/wrappers.py
```diff
@@ -12,187 +12,166 @@
...
+    The key interface functions are:
+    1. transform_space(): transform space.
+    2. transform_observation(): transform observation.
+    """
+
+    def __init__(self, env, fields=None):
+        """
+        Args:
+            env (gym.Env): the gym environment
+            fields (list[str]): fields to be applied transformation
```
emailweixu: comment the format of fields

alf/environments/wrappers.py
```diff
@@ -12,187 +12,166 @@
...
+        """
+        super().__init__(env)
+        paths = []
+        for field in fields or [""]:
+            # remove '' in the path
```
emailweixu: how does the following line perform "remove '' in the path" ?

alf/environments/wrappers.py
emailweixu: Returns:

alf/environments/wrappers.py
emailweixu: format of `field`

alf/environments/wrappers.py
```diff
@@ -12,187 +12,166 @@
...
-            channel_order (str): one of `channels_last` or `channels_first`.
-                The ordering of the dimensions in the images.
-                `channels_last` corresponds to images with shape
-                `(height, width, channels)` (Atari's default) while
-                `channels_first` corresponds to images with shape
-                `(channels, height, width)`.
-            fields_to_stack (list of str): optional paths to the fields of the
+            channel_order (str): The ordering of the dimensions in the images.
+                should be one of `channels_last` or `channels_first`.
+            fields (list[str]): optional paths to the fields of the
```
emailweixu: comment the format of fields

alf/environments/wrappers.py
```diff
@@ -12,187 +12,166 @@
...
-                spaces corresponding to the keys in FrameStack.  If is None,
-                stack all recognized spaces.
-                When input space is not a Dict, we just try to stack if we can.
-                Fields_to_stack doesn't apply anymore.
+                spaces corresponding to the keys in FrameStack.
         """
-        super().__init__(env)
-        self._frames = deque(maxlen=stack_size)
         self._channel_order = channel_order
+        if self._channel_order == 'channels_last':
```
emailweixu: assert channel_order in [...]

alf/utils/common.py
```diff
@@ -433,6 +433,31 @@ def _cast(obs):
     return tf.nest.map_structure(_cast, observation)
 
 
+def traverse_transform_observation(obs, levels, field, func):
+    """
```
emailweixu: Explain what this function does.

alf/utils/common.py
```diff
@@ -433,6 +433,31 @@ def _cast(obs):
...
+    Args:
+        obs (nested Tensor|nested ndarray): observations to be applied the transformation
+        levels (list[str]): levels
```
emailweixu: explain what are the valid values of `levels` and its relationship with `field`.

alf/utils/common.py
```diff
@@ -433,6 +433,31 @@ def _cast(obs):
...
+        field (str): full field path
```
emailweixu: format of field?

alf/utils/common.py
```diff
@@ -433,6 +433,31 @@ def _cast(obs):
...
+        func (Callable): transform func
+    """
+    if not levels:
+        return func(obs, field)
+    level = levels[0]
+    if isinstance(obs, tuple) and hasattr(obs, '_fields'):
```
emailweixu: Perhaps add a utility function `is_namedtuple()`?

alf/utils/common.py
```diff
@@ -449,33 +474,42 @@ def image_scale_transformer(observation, fields=None, min=-1.0, max=1.0):
...
+
+@gin.configurable
+def scale_transformer(observation, scale, dtype=tf.float32, fields=None):
+    """Scale observation
+
+    Args:
+         observation (nested Tensor): observation to be scaled
+         scale (float): scale factor
+         dtype (Dtype): The destination type.
+         fields (list[str]): fields to be scaled
```
emailweixu: format of `fields`

alf/utils/common.py
```diff
@@ -803,16 +837,23 @@ def get_observation_spec():
...
     """
     assert _env, "set a global env by `set_global_env` before using the function"
     specs = _env.observation_spec()
-    return tf.nest.map_structure(
+    specs = tf.nest.map_structure(
         lambda spec: (tf.TensorSpec(spec.shape, tf.float32)
                       if spec.dtype == tf.uint8 else spec), specs)
 
+    if field:
+        for f in field.split('.'):
```
hnyu: namedtuple cannot be indexed by a string

alf/utils/common.py
```diff
@@ -803,16 +837,23 @@ def get_observation_spec():
 
     https://github.com/HorizonRobotics/alf/pull/239#issuecomment-544644558
 
+    Args:
+        field (str): a multi-step path denoted by "A.B.C".
```
hnyu: assert field is a string

alf/utils/common.py
```diff
@@ -433,6 +433,31 @@ def _cast(obs):
     return tf.nest.map_structure(_cast, observation)
 
 
+def traverse_transform_observation(obs, levels, field, func):
```
hnyu: `field` seems unnecessary here? `levels` contains the split tokens of `field`

alf/utils/common.py
```diff
@@ -449,33 +474,42 @@ def image_scale_transformer(observation, fields=None, min=-1.0, max=1.0):
         Transfromed observation
     """
 
-    def _transform_image(obs):
+    def _transform_image(obs, field):
```
hnyu: `field` not used in this local function

alf/utils/common.py
```diff
@@ -449,33 +474,42 @@ def image_scale_transformer(observation, fields=None, min=-1.0, max=1.0):
...
+    """Scale observation
+
+    Args:
+         observation (nested Tensor): observation to be scaled
+         scale (float): scale factor
+         dtype (Dtype): The destination type.
+         fields (list[str]): fields to be scaled
+    """
+
+    def _scale_obs(obs, field):
```
hnyu: `field` not used in this local function

alf/environments/wrappers.py
```diff
@@ -12,187 +12,166 @@
...
+    def observation(self, observation):
+        for levels, field in self._paths:
+            observation = common.traverse_transform_observation(
+                obs=observation,
+                levels=levels,
+                field=field,
+                func=self.transform_observation)
+        return observation
+
+    def transform_space(self, observation_space, field=None):
```
hnyu: `field` seems not necessary if this function is always called on a leaf

alf/environments/wrappers.py
```diff
@@ -12,187 +12,166 @@
...
+
+        Subclass should override this to perform transformation
+
+        Args:
+             observation_space (gym.Space): space to be transformed
+             field (str): field to be transformed
+        """
+        return observation_space
+
+    def transform_observation(self, observation, field=None):
```
hnyu: `field` seems not necessary if this function is always called on a leaf

alf/environments/wrappers.py
```diff
@@ -12,187 +12,166 @@
...
-                return np.concatenate(self._frames, axis=0)
+            raise ValueError("Unsupported space:%s" % observation_space)
+
+    def transform_observation(self, observation, fields=None):
+        queue = self._frames.get(fields, None)
+        if not queue:
+            queue = deque(maxlen=self._stack_size)
+            for _ in range(self._stack_size):
+                queue.append(observation)
+            self._frames[fields] = queue
```
le-horizon: can fields be None here?  If so, index to _frames will be None?

alf/environments/wrappers.py
```diff
@@ -12,187 +12,166 @@
...
-            fields_to_stack (list of str): optional paths to the fields of the
+            channel_order (str): The ordering of the dimensions in the images.
+                should be one of `channels_last` or `channels_first`.
+            fields (list[str]): optional paths to the fields of the
                 Dict env.observation_space.  If specified, only use the
-                spaces corresponding to the keys in FrameStack.  If is None,
-                stack all recognized spaces.
-                When input space is not a Dict, we just try to stack if we can.
-                Fields_to_stack doesn't apply anymore.
+                spaces corresponding to the keys in FrameStack.
```
le-horizon: If fields is None, then no field is transformed.  This is different from the current assumption of FrameStack (all fields are stacked, if fields is None).  Please add a comment about what happens if fields is None.  Also please double check to make sure current gin files have the right fields specified.  I think ac_target_nav is the only one, but I may be wrong.

alf/environments/wrappers.py
witwolf: field is necessary in some case , from `FrameStack` , field represents a unique key for a queue (many queues ,each for a field ) that handle the stacking of this field

alf/utils/common.py
witwolf: actually , the spec will never be a namedtuple instance

see [`_spec_from_gym_space`  ](https://github.com/HorizonRobotics/alf/blob/5c648814b44e3a32e51a1f494964d14354976881/alf/environments/suite_gym.py#L65-L71),  it maps `gym.spaces.Tuple` to a tuple and we do not support unamed tuple observation in alf , and there is no space  `gym.spaces.UnamedTuple`

alf/environments/wrappers.py
witwolf: `None` can be  a key value for `dict`,  and the field will never be None here (empty str when observation is not nested )

alf/environments/wrappers.py
emailweixu: A more concrete comment: Transform the child space in `observation_space` indicated by `field` using `func`

alf/utils/common.py
emailweixu: More detailed comment: Transform the child observation in `observation` indicated by `field` using `func`

alf/utils/common.py
hnyu: I think the user can optionally convert a dict spec to namedtuple spec (for easier field access)? This way seems more consistent with what we have for other specs defined in ALF (they are always namedtuples).

alf/environments/wrappers.py
hnyu: Oh I see

alf/utils/common.py
witwolf: i think it's no necessary to convert a dict spec to namedtuple,  dict is  enough easy  for field accessing in  `observation` and `observation_space`

alf/environments/wrappers.py
witwolf: updated

alf/environments/wrappers.py
emailweixu: Agree. Need to comment the behavior of "fields=None", which is to transform the non-dict observation. In this case observation cannot be dictionary.
In general, the comment of a function needs to contain enough information for the user to use the function correctly.

witwolf: comment updated

alf/utils/common.py
witwolf: added `is_namedtuple` in `nest_utils.py`

PR #285, opened 2019-12-17
Replay buffer allowing different number of samples for each environment.

emailweixu:


alf/experience_replayers/replay_buffer.py
```diff
@@ -0,0 +1,163 @@
...
+                tf.Assert(remaining == 0, [
+                    "batch_size=", batch_size, "cannot be divided by "
+                    "num_environments=", self._num_envs
+                ])
+
+            r = tf.random.uniform([self._num_envs, batch_size_per_env])
+            num_positions = self._current_size - batch_length + 1  # [num_envs]
+            num_positions = tf.expand_dims(num_positions, -1)  # [num_envs, 1]
+            pos = tf.cast(r * tf.cast(num_positions, tf.float32), tf.int32)
+            pos = tf.minimum(pos, num_positions - 1)  # [num_envs, K]
```
hnyu: Is this min necessary if tf.cast always rounds down?

emailweixu: Because of floating multiplication at line 148, I am worried that it might be larger than num_positions - 1

alf/experience_replayers/replay_buffer.py
```diff
@@ -0,0 +1,163 @@
...
+                (current_pos + 1) % self._max_length)
+            current_size = tf.gather(self._current_size, env_ids, axis=0)
+            self._current_size.scatter_nd_update(
+                tf.expand_dims(env_ids, -1),
+                tf.minimum(current_size + 1, self._max_length))
+
+    def get_batch(self, batch_size, batch_length):
+        """Randomly get `batch_size` tranjectories from the buffer.
+
+        Note: The environment where the sampels are from are ordered in the
```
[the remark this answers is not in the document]
witwolf: typo,  tranjectories=> trajectories  , sampels=>samples

witwolf:
Where will this `ReplayBuffer`  used ?  Multi envs are only used for async training, and for async off policy training (not OneTimeExp ), one env seems  fine

emailweixu:
> Where will this `ReplayBuffer` used ? Multi envs are only used for async training, and for async off policy training (not OneTimeExp ), one env seems fine

I am thinking of using it when different environments generate steps at different pace. For example, in hierarchical RL, the higher level policy can generate steps at different intervals.

alf/experience_replayers/replay_buffer.py
hnyu: I think int(x*r) <= x-1 for any integer x and 0<r<1

hnyu:
> Where will this `ReplayBuffer` used ? Multi envs are only used for async training, and for async off policy training (not OneTimeExp ), one env seems fine

Should be useful for async training for SAC and DDPG

PR #286, opened 2019-12-17
SarsaAlgorithm

emailweixu:
Tested on pendulum. In the figure, sac and ddpg is from sac_pendulum.gin and ddpg_pendulum.gin.
sarsa-legend
sarsa

### DIFF
--- a/alf/algorithms/entropy_target_algorithm.py
+++ b/alf/algorithms/entropy_target_algorithm.py
@@ -11,7 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
+"""An algorithm for adjusting entropy regularization strength."""
 from absl import logging
 import gin
 import numpy as np
@@ -26,7 +26,7 @@ from alf.utils.common import namedtuple, run_if, should_record_summaries
 from alf.utils.dist_utils import calc_default_target_entropy
 from alf.utils.dist_utils import calc_default_max_entropy
 
-EntropyTargetLossInfo = namedtuple("EntropyTargetLossInfo", ["entropy_loss"])
+EntropyTargetLossInfo = namedtuple("EntropyTargetLossInfo", ["neg_entropy"])
 EntropyTargetInfo = namedtuple("EntropyTargetInfo", ["step_type", "loss"])
 
 
@@ -68,6 +68,7 @@ class EntropyTargetAlgorithm(Algorithm):
                  slow_update_rate=0.01,
                  fast_update_rate=np.log(2),
                  min_alpha=1e-4,
+                 average_window=2,
                  debug_summaries=False):
         """Create an EntropyTargetAlgorithm
 
@@ -88,6 +89,7 @@ class EntropyTargetAlgorithm(Algorithm):
                 state 2
             min_alpha (float): the minimal value of alpha. If <=0, exp(-100) is
                 used.
+            average_window (int): window size for averaging past entropies.
             optimizer (tf.optimizers.Optimizer): The optimizer for training. If
                 not provided, will use the same optimizer of the parent
                 algorithm.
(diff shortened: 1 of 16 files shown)


alf/algorithms/sarsa_algorithm.py
```diff
@@ -0,0 +1,293 @@
...
+            (time_step.observation, target_action),
+            step_type=time_step.step_type,
+            network_state=state.target_critic)
+
+        prev_return = tf.stop_gradient(time_step.reward +
+                                       time_step.discount * target_critic)
+        info = SarsaInfo(
+            actor_loss=actor_loss, critic=prev_critic, returns=prev_return)
+
+        if not is_train_step and not isinstance(self._actor_network,
```
witwolf: use `not isinstance(action_distribution, tf.Tensor)` like `common.to_distribution` seems better

alf/utils/value_ops.py
```diff
@@ -194,10 +193,8 @@ def one_step_discounted_return(rewards, values, step_types, discounts):
     discounts = discounts[1:]
     rewards = rewards[1:]
     values = values[1:]
-    step_types = step_types[:-1]
 
-    is_lasts = tf.cast(tf.equal(step_types, StepType.LAST), tf.float32)
```
witwolf: Why change this ? For `LAST step` , the value is  not  `reward + discount * V(FIRST step of new_episode)`,  it's `reward`. 
Thought it's masked out when calculate loss, it's better to use the correct calculation

alf/utils/metric_utils.py
```diff
@@ -37,14 +37,17 @@ def __init__(self, time_step_spec, action_spec, policy_state_spec,
         super(Policy, self).__init__(
             time_step_spec=time_step_spec,
             action_spec=action_spec,
-            policy_state_spec=policy_state_spec)
+            policy_state_spec=(policy_state_spec, action_spec))
         # attribute named `_action_fn` already exist in parent
         self._action_fn1 = action_fn
 
     def _action(self, time_step, policy_state=(), seed=None):
-        policy_step = self._action_fn1(time_step, policy_state)
```
witwolf: `reset_state_if_necessary` for new episode .  And is there better way to store `prev_action` ?  it's very strange to store in `state`

alf/algorithms/sarsa_algorithm.py
emailweixu: It's more trouble because action_distribution might be a nest.

alf/utils/value_ops.py
emailweixu: My bad. Will restore.

alf/utils/metric_utils.py
emailweixu: tfa_eager_compute() handles that.

emailweixu: I agree. But this is the easiest way to add prev_action using tfa_eager_compute().
Perhaps we should use OnPolicyDriver for eval too.

alf/algorithms/sarsa_algorithm.py
witwolf: i see

alf/algorithms/entropy_target_algorithm.py
```diff
@@ -154,16 +157,43 @@ def train_step(self, distribution, step_type):
                 step_type=step_type,
                 loss=LossInfo(
                     loss=-entropy_for_gradient,
-                    extra=EntropyTargetLossInfo(entropy_loss=-entropy))))
+                    extra=EntropyTargetLossInfo(neg_entropy=-entropy))))
 
-    def calc_loss(self, training_info: EntropyTargetInfo):
+    def calc_loss(self, training_info: EntropyTargetInfo, valid_mask=None):
```
witwolf: never pass  argument value for `valid_mask` when `calc_loss` is called,  and is it same with `mask` in L164 ?

emailweixu: valid_mask will be used in some other places.

PR #287, opened 2019-12-18
update adaptive_normalizer

hnyu:


alf/utils/adaptive_normalizer.py
```diff
@@ -14,73 +14,97 @@
...
+        self._m2_averager = AdaptiveAverager(tensor_spec, speed=speed)
         self._auto_update = auto_update
         self._variance_epsilon = variance_epsilon
-        super(AdaptiveNormalizer, self).__init__(
-            tensor_spec, norm_update_rate=self._update_ema_rate)
 
-    def normalize(self, tensors, clip_value=-1.0):
-        """Normalized the reward
+    def update(self, tensor):
+        self._mean_averager.update(tf.reduce_mean(tensor))
```
emailweixu: I think this not correct when tensor is not scalar.

hnyu: Right, fixed

hnyu: Right, fixed

alf/utils/adaptive_normalizer.py
```diff
@@ -14,73 +14,97 @@
...
+        self._mean_averager = AdaptiveAverager(tensor_spec, speed=speed)
+        self._m2_averager = AdaptiveAverager(tensor_spec, speed=speed)
         self._auto_update = auto_update
         self._variance_epsilon = variance_epsilon
-        super(AdaptiveNormalizer, self).__init__(
-            tensor_spec, norm_update_rate=self._update_ema_rate)
 
-    def normalize(self, tensors, clip_value=-1.0):
-        """Normalized the reward
+    def update(self, tensor):
```
emailweixu: you need to first get the outer rank, and reduce over the outer ranks. tensor may or may not be batched.
```python
tf_agent.utils.nest_utils.get_outer_rank(tensor, tensor_spec)
```

hnyu:
Regarding the numerical issue after training for a long time: we increment the steps number based on env batches, so after M env steps, the steps counter is actually M/B. Suppose M is 1 billion and B is 50, then the counter is 2x10^7 which can be handled by float32

alf/utils/adaptive_normalizer.py
```diff
@@ -48,52 +51,76 @@ def __init__(self,
...
-            scale=None,
-            variance_epsilon=self._variance_epsilon)
-        if clip_value > 0:
-            normalized_tensor = tf.clip_by_value(normalized_tensor,
-                                                 -clip_value, clip_value)
-        return normalized_tensor
+
+        def _normalize(m, m2, spec, t):
+            var = m2.get() - tf.square(m.get())
+            outer_dims = get_outer_rank(t, spec)
```
emailweixu: what if outer_dims==0?

emailweixu:
> Regarding the numerical issue after training for a long time: we increment the steps number based on env batches, so after M env steps, the steps counter is actually M/B. Suppose M is 1 billion and B is 50, then the counter is 2x10^7 which can be handled by float32

np.float32(1+5e-8) - 1 == 0

alf/utils/adaptive_normalizer.py
hnyu: BatchSquash will handle that correctly
```
x = [0 0]
batch_squash = BatchSquash(0)
batch_squash.flatten(x) -> [[0 0]]
batch_squash.unflatten(batch_squash.flatten(x)) -> [0 0]
```

alf/utils/adaptive_normalizer.py
```diff
@@ -14,73 +14,124 @@
...
-    def normalize(self, tensors, clip_value=-1.0):
-        """Normalized the reward
+    def update(self, tensor):
+        def _update(averager, spec, t):
+            outer_dims = get_outer_rank(t, spec)
+            batch_squash = BatchSquash(outer_dims)
+            t = batch_squash.flatten(t)
+            averager.update(tf.reduce_mean(t, axis=0))
+
+        tf.nest.map_structure(_update, self._mean_averager, self._tensor_spec,
```
emailweixu: If we want to handle nested structure, perhaps it's better to do it in averager.

hnyu: updated

alf/utils/adaptive_normalizer.py
```diff
@@ -14,73 +14,110 @@
...
+                t = tf.clip_by_value(t, -clip_value, clip_value)
+            t = batch_squash.unflatten(t)
+            return t
 
-    def update(self, tensors):
-        self._update_ema_rate.assign(
-            self._speed / tf.cast(self._total_env_steps, tf.float32))
-        self._total_env_steps.assign_add(1)
-        super(AdaptiveNormalizer, self).update(tensors)
+        return tf.nest.map_structure(_normalize, self._mean_averager,
```
emailweixu: self._mean_averager.get(), m2_averager.get()

PR #289, opened 2019-12-20
Simplify data structures and spec preparations

emailweixu:
Now passed off_policy_driver_test.py and on_policy_driver_test. Perhaps some other tests might fail.
Just send it out to get your opinion.

Changed the following:
* Removes action_distribution from TrainingInfo and Experience
* PolicyStep.action from RLAlgorithm.rollout and train_step is always action Tensor
* Automatically handles Distribution in PolicyStep.info
* Spec preparations are moved to RLAlgorithm fro OffPolicyAlgorithn
* Remove final steps. It complicated code, may not have much benefit, and can cause bug if rollout() has side effects.


alf/algorithms/actor_critic_algorithm.py
```diff
@@ -93,10 +98,8 @@ def predict(self, time_step: ActionTimeStep, state: ActorCriticState):
             state=ActorCriticState(actor=actor_state),
             info=())
 
-    def rollout(self,
-                time_step: ActionTimeStep,
-                state: ActorCriticState,
-                is_train_step=False):
+    def rollout_action_distribution(self, time_step: ActionTimeStep,
+                                    state: ActorCriticState, mode):
```
witwolf: 1. since this algorithm can work with on policy and off policy trainers, the mode can be one of [`ON_POLICY_TRAINING`,`OFF_POLICY_TRAINING`] , and for `OFF_POLICY_TRAINING` it's no necessary to compute value info
2. `OnPolicyAlgorithm.train_step` calls this function, and it needs value info, i think it's better to explicit implement `train_step` for on policy algorithm which support off policy training
3.  Its subclass `PPOAlgorithm` is an off policy algorithm, the mode is `ROLLOUT`, and it needs value info ,it's better to overwrite this function

alf/algorithms/rl_algorithm.py
```diff
@@ -265,11 +366,47 @@ def predict(self, time_step: ActionTimeStep, state=None):
...
+              policy_step.state should be consistent with `predict_state_spec`
+        """
+        policy_step = self.rollout_action_distribution(time_step, state)
         return policy_step._replace(info=())
 
+    ON_POLICY_TRAINING = 0
+    OFF_POLICY_TRAINING = 1
+    ROLLOUT = 2
+
     @abstractmethod
```
witwolf: can we make deterministic policy algorithm override `predict_action_distribution` or `rollout_action_distribution` and return deterministic distribution if necessary , and all subclasses do not override `rollout` and `predict`, otherwise, it will make the code very confusing

alf/algorithms/rl_algorithm.py
```diff
@@ -265,11 +366,47 @@ def predict(self, time_step: ActionTimeStep, state=None):
                 `action_distribution_spec`
               policy_step.state should be consistent with `predict_state_spec`
         """
-        policy_step = self.rollout(time_step, state)
+        if self.__class__.predict_action_distribution != RLAlgorithm.predict_action_distribution:
```
witwolf: perhaps we should find other way to do this

alf/drivers/policy_driver.py
```diff
@@ -47,18 +46,21 @@ def __init__(self,
...
                 callable(time_step.Trajectory).
-            use_rollout_state (bool): Include the RNN state for the experiences
-                used for off-policy training
             metrics (list[TFStepMetric]): An optional list of metrics.
             training (bool): True for training, false for evaluating
             greedy_predict (bool): use greedy action for evaluation (i.e.
                 training==False).
         """
+        assert mode in [
+            "on_policy_training", "predict", "greedy_predict", "rollout"
```
witwolf: `rollout => off_policy_training` seems easier to understand ?

alf/algorithms/rl_algorithm.py
emailweixu: As an interface, `rollout` and `predict` are better because they are more general - the policy only output an action. `predict_action_distribution` and `rollout_action_distribution` are only useful for `greedy_predict`. Perhaps we can remove `predict_action_distribution`, `rollout_distribution` and `greedy_predict`, and change `predict` to include an additional argument `epsilon_greedy`?
```
@abstractmethod
def predict(self, time_step, state, epsilon_greedy)
```
And do not provide default implementation for predict. Subclass has to implement  it.  We can provide an utility function for generating epsilon greedy action from a distribution. This way the interface will be much cleaner, but at the expense of more code in the subclass.

I was thinking about this way but eventually changed to the current form. See https://github.com/HorizonRobotics/alf/blob/f18d45ed271bca4ba65353d9f297605d07259394/alf/utils/common.py#L691. 

What do you think @witwolf @hnyu?

emailweixu: Now I removed rollout_action_distribution and predict_action_distribution. The subclasses do not seem to need much more lines.

alf/algorithms/rl_algorithm.py
emailweixu: removed

alf/drivers/policy_driver.py
emailweixu: updated to constant integer OFF_POLICY_TRAINING

alf/algorithms/actor_critic_algorithm.py
emailweixu: We can do that later. Right now we don't train ActorCriticAlgorithm using off-policy.

alf/drivers/policy_driver.py
```diff
@@ -48,19 +53,19 @@ def __init__(self,
...
+            epsilon_greedy (float):  a floating value in [0,1], representing the
+                chance of action sampling instead of taking argmax. This can
+                help prevent a dead loop in some deterministic environment like
+                Breakout. Only used for mode=PREDICT.
         """
-        assert mode in [
-            "on_policy_training", "predict", "greedy_predict", "rollout"
-        ]
         self._mode = mode
+        self._epsilon_greedy = 0.1
```
witwolf: `self._epsilon_greedy = epsilon_greedy`

alf/trainers/policy_trainer.py
```diff
@@ -417,7 +423,10 @@ def play(root_dir,
         algorithm (OnPolicyAlgorithm): the training algorithm
         checkpoint_name (str): name of the checkpoint (e.g. 'ckpt-12800`).
             If None, the latest checkpoint under train_dir will be used.
-        greedy_predict (bool): use greedy action for evaluation.
+        epsilon_greedy (bool): a floating value in [0,1], representing the
```
witwolf: bool => float

alf/drivers/policy_driver.py
```diff
@@ -32,12 +32,17 @@
...
+    PREDICT = 2
+
     def __init__(self,
                  env,
                  algorithm,
                  observers=[],
                  metrics=[],
-                 mode="on_policy_training"):
+                 epsilon_greedy=0.1,
+                 mode=PREDICT):
```
witwolf: missing comment for modes

alf/drivers/policy_driver.py
```diff
@@ -48,19 +53,19 @@ def __init__(self,
                 callable(time_step.Trajectory).
             metrics (list[TFStepMetric]): An optional list of metrics.
             training (bool): True for training, false for evaluating
```
witwolf: no argument `training` in function signature now

alf/drivers/on_policy_driver.py
```diff
@@ -66,17 +66,18 @@ def __init__(self,
...
             train_interval (int):
         """
         super(OnPolicyDriver, self).__init__(
             env=env,
             algorithm=algorithm,
             observers=observers,
             metrics=metrics,
-            mode="on_policy_training" if training else
-            ("greedy_predict" if greedy_predict else "predict"))
+            mode=self.ON_POLICY_TRAINING if training else self.PREDICT)
```
witwolf: `epsilon_greedy=epsilon_greedy`  for `__init__`

alf/drivers/policy_driver.py
emailweixu: fixed

alf/trainers/policy_trainer.py
emailweixu: fixed

alf/drivers/policy_driver.py
emailweixu: added

alf/drivers/policy_driver.py
emailweixu: removed

alf/drivers/on_policy_driver.py
emailweixu: fixed

alf/algorithms/sarsa_algorithm.py
```diff
@@ -156,58 +156,29 @@ def __init__(self,
...
-
-    def greedy_predict(self, time_step: ActionTimeStep, state=None, eps=0.1):
-        if isinstance(self._actor_network, DistributionNetwork):
-            return super().greedy_predict(time_step, state, eps=eps)
-        action, actor_state = self._actor_network(
-            time_step.observation,
-            step_type=time_step.step_type,
-            network_state=state.actor)
+    def predict(self, time_step: ActionTimeStep, state, epsilon_greedy):
+        _, action, actor_state = self._get_action(
```
witwolf: Wrong order of parameters call `_get_action` 
```python
_, action, actor_state = self._get_action(
            actor_network=self._actor_network, 
            time_step=time_step, 
            state=state.actor, 
            epsilon_greedy=epsilon_greedy)
```

alf/bin/play.py
```diff
@@ -42,7 +42,7 @@
     'checkpoint_name', None, "name of the checkpoint "
     "(e.g. 'ckpt-12800`). If None, the latest checkpoint under train_dir will "
     "be used.")
-flags.DEFINE_bool('greedy_predict', True, "use greedy action for evaluation.")
+flags.DEFINE_bool('epsilon_greedy', 0.1, "probability of sampling action.")
```
hnyu: DEFINE_bool -> DEFINE_float, and the default value should be 0?

alf/drivers/async_off_policy_driver.py
```diff
@@ -159,11 +159,7 @@ def get_training_exps(self):
         """
         batch = self._tfq.learn_queue.dequeue_all()
         # convert the batch to the experience format
-        exp = make_experience(
-            batch.time_step,
-            batch.policy_step,
-            batch.act_dist_param,
-            state=batch.state)
```
hnyu: make_experience should store the "current" state instead of the new state in `policy_step`. See the original call of `make_experience` in policy_driver.py when `make_experience` still has 4 arguments.

alf/drivers/threads.py
```diff
@@ -118,9 +120,8 @@ def repeat_shape_n(nested_spec, n):
         lambda t: tf.TensorSpec([n] + list(t.shape), t.dtype), nested_spec)
 
 
-LearningBatch = namedtuple(
-    "LearningBatch",
-    ["time_step", "state", "policy_step", "act_dist_param", "next_time_step"])
+LearningBatch = namedtuple("LearningBatch",
+                           ["time_step", "policy_step", "next_time_step"])
```
hnyu: Might still need "state". See the comment above about `make_experience`

alf/drivers/threads.py
```diff
@@ -170,39 +169,33 @@ def __init__(self,
...
-                                     unroll_length) if store_state else (),
-                policy_step=repeat_shape_n(self._policy_step_spec,
-                                           unroll_length),
-                act_dist_param=repeat_shape_n(self._act_dist_param_spec,
-                                              unroll_length),
-                next_time_step=repeat_shape_n(self._time_step_spec,
-                                              unroll_length)))
+                time_step=unrolled_time_step_spec,
+                policy_step=unrolled_policy_step_spec,
+                next_time_step=unrolled_time_step_spec))
```
hnyu: next_time_step seems no longer useful

alf/algorithms/actor_critic_algorithm.py
```diff
@@ -17,23 +17,25 @@
...
 from alf.algorithms.on_policy_algorithm import OnPolicyAlgorithm
-from alf.algorithms.rl_algorithm import ActionTimeStep, TrainingInfo, namedtuple
+from alf.data_structures import ActionTimeStep, namedtuple, PolicyStep
+from alf.utils import common
 
 ActorCriticState = namedtuple(
     "ActorCriticState", ["actor", "value"], default_value=())
 
-ActorCriticInfo = namedtuple("ActorCriticInfo", ["value"])
+ActorCriticInfo = namedtuple("ActorCriticInfo",
```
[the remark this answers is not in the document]
emailweixu: ActorCriticInfo is only used for certain algorithms. That's why I didn't put to data_structures.py

alf/algorithms/actor_critic_algorithm.py
```diff
@@ -81,22 +83,22 @@ def __init__(self,
...
             state=ActorCriticState(actor=actor_state),
-            info=())
+            info=ActorCriticInfo(action_distribution=action_dist))
 
-    def rollout(self,
-                time_step: ActionTimeStep,
-                state: ActorCriticState,
-                is_train_step=False):
+    def rollout(self, time_step: ActionTimeStep, state: ActorCriticState,
+                mode):
```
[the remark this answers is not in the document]
emailweixu: This function is commented in the base class.

alf/algorithms/rl_algorithm.py
```diff
@@ -123,14 +111,103 @@ def __init__(self,
...
+    def observation_spec(self):
+        """Return the observation spec."""
+        return self._observation_spec
+
+    @property
+    def train_step_info_spec(self):
+        """The spec for the PolicyInfo.info returned from train_step()."""
+        if self._train_step_info_spec is not None:
+            return self._train_step_info_spec
+        batch_size = 4
```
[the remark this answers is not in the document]
emailweixu: It's arbitrary. batch_size=1 can work too.

alf/bin/play.py
emailweixu: Good catch. Changed to DEFINE_float. The greedy_predict=True means epsilon_greedy=0.1. greedy_predict=1 means espsilon_greedy=1.0

alf/drivers/async_off_policy_driver.py
emailweixu: This is a serious bug.

alf/drivers/threads.py
emailweixu: It's still used at line 426.

alf/algorithms/on_policy_algorithm.py
```diff
@@ -45,62 +46,10 @@ class OnPolicyAlgorithm(OffPolicyAlgorithm):
...
-              state (nested Tensor): should be consistent with `train_state_spec`
-              info (nested Tensor): everything necessary for training. Note that
-                ("action_distribution", "action", "reward", "discount",
-                "is_last") are automatically collected by OnPolicyDriver. So
-                the user only need to put other stuff (e.g. value estimation)
-                into `policy_step.info`
-        """
-        pass
-
     # Implement train_step() to allow off-policy training for an
```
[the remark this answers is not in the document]
emailweixu: Actually, this is to allow "off-policy training for an OnPolicyAlgorithm".

PR #291, opened 2020-01-06
pretty print optimizer info in text summary

witwolf:


alf/algorithms/algorithm.py
```diff
@@ -297,11 +312,24 @@ def get_optimizer_info(self):
...
+                    Algorithm=alg_name,
+                    Optimizer=opt.get_config() if opt is not None else None,
+                    Modules=[
+                        dict(
+                            ModuleName=m.name,
+                            TrainableVariables=[
+                                v.name for v in _get_trainable_vars(m)
+                            ]) for m in module_set if m is not None
+                    ]))
+        json_pretty_str_info = json.dumps(
```
emailweixu: what's the difference with pprint.pprint()? Why get_json_type?

alf/algorithms/algorithm.py
```diff
@@ -297,11 +312,24 @@ def get_optimizer_info(self):
...
+            obj=optimizer_info, indent=2, default=get_json_type)
+
+        def _markdownify(paragraph):
+            return "    ".join(
+                (os.linesep + paragraph).splitlines(keepends=True))
+
+        return _markdownify(json_pretty_str_info)
```
emailweixu: why _markdownify is needed?

witwolf: with _markdownify, tensorboard text summary will treat text content with  html `pre` label  and keep original format



without _markdownify

alf/algorithms/algorithm.py
witwolf: optimizer config contain non primitive type attrs

PR #294, opened 2020-01-06
Do not record before actual training

hnyu:
Sometimes at global_step==0  we have undefined behaviors, e.g., calling preprocess_experience on some random time_step. This will log intrinsic rewards before global_step is changed to 1 in _train() and the initial outlier point will destroy the whole curve. Originally we have 

```
tf.equal((global_step + 1) % summary_interval, 0)
```
But this will have an issue for metric_summary in off_policy_algorithm if we log every minibatch and num_minibatch and summary_interval are both even numbers.


emailweixu:
This might also happen at other global_step if training is resumed from checkpoint. Perhaps we should instead calling util.enable_summary(False) initially and util.enable_summary(True) right before real training starts?

PR #295, opened 2020-01-06
Knowledge base for introducing key concepts in ALF

emailweixu:
Also fixed some minor bugs.

Everybody is welcome to contribute contents to this article for things that you think are important for using and developing ALF.

### DIFF
--- a/alf/trainers/policy_trainer.py
+++ b/alf/trainers/policy_trainer.py
@@ -273,6 +273,7 @@ class Trainer(object):
         if self._evaluate:
             self._eval_env = unwrapped_env
 
+    @gin.configurable('alf.trainers.PolicyTrainer._create_environment')
     def _create_environment(self, nonparallel=False):
         """Create and register an env."""
         env = create_environment(nonparallel=nonparallel)
--- a/alf/trainers/policy_trainer.py
+++ b/alf/trainers/policy_trainer.py
@@ -273,7 +273,7 @@ class Trainer(object):
         if self._evaluate:
             self._eval_env = unwrapped_env
 
-    @gin.configurable('alf.trainers.PolicyTrainer._create_environment')
+    @gin.configurable('alf.trainers.Trainer._create_environment')
     def _create_environment(self, nonparallel=False):
         """Create and register an env."""
         env = create_environment(nonparallel=nonparallel)
--- a/docs/kb.md
+++ b/docs/kb.md
@@ -35,7 +35,7 @@ Algorithm is the most important concept in ALF. TODO: more description about the
 
 `ActionTimeStep` is a data structure extended from `TimeStep` of [TF-Agents](https://github.com/tensorflow/agents)
 for storing the information from the result of each environment step. It contains
-four fields:
+five fields:
 * `observation`: observation from the environment. It can be a nest of Tensors.
 * `step_type`: type of this step. It has three possible values:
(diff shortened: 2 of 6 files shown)


docs/kb.md
```diff
@@ -0,0 +1,91 @@
...
+
+## Algorithm
+
+Algorithm is the most important concept in ALF. TODO: more description about the design.
+
+## ActionTimeStep
+
+`ActionTimeStep` is a data structure extended from `TimeStep` of [TF-Agents](https://github.com/tensorflow/agents)
+for storing the information from the result of each environment step. It contains
+four fields:
```
Haichao-Zhang: four fields -> five fields?

docs/kb.md
```diff
@@ -0,0 +1,91 @@
...
+creates an environemnt in the following way to avoid
+`gym.wrappers.time_limit.TimeLimit`:
+```python
+  gym_spec = gym.spec(environment_name)
+  gym_env = gym_spec.make()
+```
+
+Then TF-Agents uses its own wrapper `tf_agents.environments.wrappers.TimeLimit`
+to wrap the environment to limit the steps so that it does not change the
+discount when `max_episode_steps` is reached.
```
hnyu: Could add a subsection 

### Value learning for different time step types
| Step type       | Discount           | Value used for bootstrapping the previous value  | Value learned | Note|
| ------------- |:-------------:| -----:|-----:|------:|
| FIRST      | 1 | No | Yes | First step of an episode|
| MID     | 1   |Yes  |  Yes | Any step other than FIRST and LAST |
| LAST | 0      |   No | No | Last step because of a normal game end |
| LAST | 1 | Yes | No | Last step because of time limit |

docs/kb.md
```diff
@@ -0,0 +1,100 @@
...
+## Algorithm
+
+Algorithm is the most important concept in ALF. TODO: more description about the design.
+
+## ActionTimeStep
+
+`ActionTimeStep` is a data structure extended from `TimeStep` of [TF-Agents](https://github.com/tensorflow/agents)
+for storing the information from the result of each environment step. It contains
+five fields:
+* `observation`: observation from the environment. It can be a nest of Tensors.
```
le-horizon: It would be better to include when exactly is each field taken, e.g. observation is taken after the previous action and a world step; step_type is whether the current state (after previous action and environment step) is a first or middle or last state; discount is how much next step's value would contribute to this state.  (Not sure if this is accurate.)

This should clarify the sequence of things quite a bit.

alf/trainers/policy_trainer.py
```diff
@@ -273,6 +273,7 @@ def initialize(self):
         if self._evaluate:
             self._eval_env = unwrapped_env
 
+    @gin.configurable('alf.trainers.PolicyTrainer._create_environment')
```
hnyu: Why do you name "PolicyTrainer" instead of just "Trainer"?

emailweixu: Should be Trainer. I thought the class name is PolicyTrainer.

emailweixu:
> Also, file name kb is a bit cryptic, maybe definitions or tips_for_dev?

changed to knowledge_base.md

docs/kb.md
emailweixu: Added more detail

PR #300, opened 2020-01-09
Pr attention target nav

le-horizon:
Hey Wei and Haonan,

Adding attention to target navigation plus other supporting changes.  Attention isn't successful yet, but the change is getting large, so pushing it to avoid all the rebasing.

Could you take a look?

Thanks,
Le

### DIFF
--- a/alf/environments/suite_socialbot.py
+++ b/alf/environments/suite_socialbot.py
@@ -74,6 +74,10 @@ def load(environment_name,
         A PyEnvironmentBase instance.
     """
     _unwrapped_env_checker_.check_and_update(wrap_with_process)
+    if gym_env_wrappers is None:
+        gym_env_wrappers = ()
+    if env_wrappers is None:
+        env_wrappers = ()
 
     gym_spec = gym.spec(environment_name)
     if max_episode_steps is None:
--- a/alf/utils/common.py
+++ b/alf/utils/common.py
@@ -708,8 +708,8 @@ def get_states_shape():
         Returns 0 if internal states is not part of observation.
     """
     assert _env, "set a global env by `set_global_env` before using the function"
-    assert isinstance(_env.observation_spec(), dict), "observation not a dict"
-    if 'states' in _env.observation_spec():
+    if isinstance(_env.observation_spec(),
+                  dict) and ('states' in _env.observation_spec()):
         return _env.observation_spec()['states'].shape
     else:
         return 0
@@ -728,6 +728,11 @@ def get_action_spec():
     return _env.action_spec()
 
 
+def get_env():
+    assert _env, "set a global env by `set_global_env` before using the function"
+    return _env
+
+
 @gin.configurable
 def get_vocab_size():
     """Get the vocabulary size of observations provided by the global environment.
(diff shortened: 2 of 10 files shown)


alf/utils/common.py
```diff
@@ -205,7 +205,7 @@ def _get_summary_enabled_var():
     global _summary_enabled_var
     if _summary_enabled_var is None:
         _summary_enabled_var = tf.Variable(
-            False, dtype=tf.bool, name="summary_enabled")
+            False, dtype=tf.bool, trainable=False, name="summary_enabled")
```
le-horizon: Making it non-trainable, as otherwise model checkpointing would fail.

alf/utils/summary_utils.py
```diff
@@ -136,7 +136,8 @@ def histogram_continuous(name,
             bucket_counts = tf.cast(
                 tf.reduce_sum(input_tensor=one_hots, axis=0), dtype=tf.float64)
             edges = tf.linspace(bucket_min, bucket_max, bucket_count + 1)
-            edges = tf.concat([edges[:-1], [float(bucket_max)]], 0)
+            edges = tf.concat([edges[:-1], [tf.cast(bucket_max, tf.float64)]],
```
le-horizon: summarizing in a tf.function, float doesn't work, has to use tf.cast explicitly.

alf/environments/suite_socialbot.py
```diff
@@ -74,6 +74,10 @@ def load(environment_name,
         A PyEnvironmentBase instance.
     """
     _unwrapped_env_checker_.check_and_update(wrap_with_process)
+    if gym_env_wrappers is None:
+        gym_env_wrappers = ()
+    if env_wrappers is None:
+        env_wrappers = ()
```
emailweixu: Why do you need this? The default is already `()`

alf/utils/common.py
```diff
@@ -708,8 +708,8 @@ def get_states_shape():
         Returns 0 if internal states is not part of observation.
     """
     assert _env, "set a global env by `set_global_env` before using the function"
-    assert isinstance(_env.observation_spec(), dict), "observation not a dict"
-    if 'states' in _env.observation_spec():
+    if isinstance(_env.observation_spec(),
```
emailweixu: should raise if states is not part of the observation?

alf/utils/common.py
```diff
@@ -737,8 +742,11 @@ def get_vocab_size():
         Returns 0 if language is not part of observation.
     """
     assert _env, "set a global env by `set_global_env` before using the function"
-    if 'sentence' in _env.observation_spec():
-        return _env.observation_spec()['sentence'].shape[0]
+    if isinstance(_env.observation_spec(),
```
emailweixu: should raise if there is not vocab_size

alf/environments/suite_socialbot.py
le-horizon: This is for command line argument override for gin configs.
--gin_param="xxx=()"
Parens can be mis-interpreted by the linux shell, passing in None is easier.

alf/utils/common.py
le-horizon: Good point.  Done.

alf/utils/common.py
le-horizon: Done.

PR #304, opened 2020-01-11
Fix autograph failure

emailweixu:
TF autograph has bug handling `\` (https://github.com/tensorflow/tensorflow/issues/35765)


emailweixu:
> LG. Is it because tf functions have stricter standards?

It's a TF bug. The transformed code is incorrect with `\`.

PR #305, opened 2020-01-16
Fix entropy_target for Agent off policy training

emailweixu:
Cannot correctly handle mode PREPARE_SPEC for off policy training


alf/algorithms/rl_algorithm.py
```diff
@@ -124,6 +129,7 @@ def use_rollout_state(self):
     def use_rollout_state(self, flag):
         # TODO: need to set the flag for the child algorithms.
```
hnyu: remove TODO

PR #306, opened 2020-01-16
Change the order of runs in grid_search

emailweixu:
So that runs with different configs can run together. This is useful
if we want to get a peek about the performance difference between different settings.

### DIFF
--- a/alf/bin/grid_search.py
+++ b/alf/bin/grid_search.py
@@ -250,9 +250,9 @@ class GridSearch(object):
         device_queue = self._init_device_queue(max_worker_num)
 
         task_count = 0
-        for values in itertools.product(*param_values):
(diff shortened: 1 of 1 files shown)


alf/bin/grid_search.py
```diff
@@ -250,9 +250,9 @@ def run(self):
         device_queue = self._init_device_queue(max_worker_num)
 
         task_count = 0
-        for values in itertools.product(*param_values):
-            parameters = dict(zip(param_keys, values))
-            for repeat in range(self._conf.repeats):
+        for repeat in range(self._conf.repeats):
+            for values in itertools.product(*param_values):
+                parameters = dict(zip(param_keys, values))
```
hnyu: task_count should be re-init to 0 in the inner loop for each parameters

PR #309, opened 2020-01-21
add image_encoding_network

hnyu:
Added an implementation of ImageEncodingNetwork. The main difference with EncodingNetwork from tf_agents is that this can output the unflattened conv result and calculate its shape automatically. This implementation focuses only on image inputs with the "padding" option introduced (for ImageDecodingNetwork as well). Also the interface is simplified and no input_tensor_spec is needed. 

Can be paired with the existing ImageDecodingNetwork and VAE to create a variational image autoencoder.


alf/utils/image_networks.py
```diff
@@ -0,0 +1,216 @@
...
+# limitations under the License.
+
+import gin.tf
+import tensorflow as tf
+
+
+@gin.configurable
+class ImageEncodingNetwork(tf.keras.models.Model):
+    """
+    A general template class for create convolutional encoding networks.
```
Haichao-Zhang: creating

alf/utils/image_networks.py
```diff
@@ -0,0 +1,216 @@
+# Copyright (c) 2019 Horizon Robotics. All Rights Reserved.
```
Haichao-Zhang: can update the year to 2020

alf/utils/image_networks.py
```diff
@@ -0,0 +1,216 @@
...
+                None, then the output would preserve the shape of an image (will
+                not be flattened).
+            activation_fn (tf.nn.activation): activation for conv layers
+            output_activation_fn (tf.nn.activation): only used when there are
+                postprocessing FC layers
+            name (str): network name
+        """
+        super().__init__(name=name)
+
+        assert isinstance(conv_layer_params, list)
```
Haichao-Zhang: This assertion seems to be inconsistent with the comment on support tuple as well ("conv_layer_params (list[tuple])").

alf/utils/image_networks.py
```diff
@@ -0,0 +1,216 @@
...
+
+        Args:
+            input_size (int): the input image size
+            conv (bool): whether calculate the intermediate conv output size
+                or the final FC output size
+
+        Returns:
+            a tuple representing the output shape
+        """
+        if self._postprocess_fc_layers and not conv:
```
Haichao-Zhang: What is the corresponding scenario when self._postprocess_fc_layer_params is not None and conv is True?

alf/utils/image_networks.py
```diff
@@ -0,0 +1,216 @@
...
+                 name="image_decoding_network"):
+        """
+        Initialize the layers for decoding a latent vector into an image.
+
+        Args:
+            deconv_layer_params (list[tuple]): a non-empty list of elements
+                (num_filters, kernel_size, strides).
+            start_decoding_size (int): the initial size we'd like to have for
+                the feature map
+            start_decoding_filters (int): the initial number of fitlers we'd like
```
Haichao-Zhang: fitlers -> filters

alf/utils/image_networks.py
```diff
@@ -0,0 +1,216 @@
...
+                transposed convolutions.
+            activation_fn (tf.nn.activation): activation for hidden layers
+            output_activation_fn (tf.nn.activation): activation for the output
+                layer. Usually our image inputs are normalized to [0, 1] or [-1, 1],
+                so this function should be tf.nn.sigmoid or tf.nn.tanh.
+            name (str): network name
+        """
+        super().__init__(name=name)
+
+        assert isinstance(deconv_layer_params, list)
```
Haichao-Zhang: This assertion seems to be inconsistent with the comment on support tuple as well ("deconv_layer_params (list[tuple])").

alf/utils/image_networks.py
```diff
@@ -0,0 +1,216 @@
...
+            z = tf.reshape(z, [tf.shape(z)[0], -1])
+            for fc_l in self._postprocess_fc_layers:
+                z = fc_l(z)
+        return z
+
+
+@gin.configurable
+class ImageDecodingNetwork(tf.keras.models.Model):
+    """
+    A general template class for create transposed convolutional decoding networks.
```
Haichao-Zhang: create -> creating

alf/utils/image_networks.py
hnyu: list[tuple] means "a list of tuples", not "list of tuple"

alf/utils/image_networks.py
hnyu: It will return the last conv's shape, regardless of postprocess FC layers

PR #311, opened 2020-01-22
Speed up summary

emailweixu:
Currently TrainerConfig.summary_interval controls how often summary is written.
It was expected that running speed with a very large value of summary_interval should be almost
same as debug_summaries=False and summarize_grads_and_vars=False.
But somehow it was not true. In one of my experiments, the difference is as large as 30%.
This PR partially solves this problem by forcing TF to compile two versions of _train().

TODO: Check whether on policy training also has this problem.

### DIFF
--- a/alf/utils/summary_utils.py
+++ b/alf/utils/summary_utils.py
@@ -418,7 +418,7 @@ class record_time(object):
         self._tag = tag
         caller = logging.get_absl_logger().findCaller()
         # token is a string of filename:lineno:tag
-        token = caller[0] + ':' + str(caller[0]) + ':' + tag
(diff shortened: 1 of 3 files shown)


alf/utils/summary_utils.py
```diff
@@ -406,10 +416,21 @@ def __init__(self, tag):
             tag (str): the summary tag for the the time.
         """
         self._tag = tag
+        caller = logging.get_absl_logger().findCaller()
+        # token is a string of filename:lineno:tag
+        token = caller[0] + ':' + str(caller[0]) + ':' + tag
```
hnyu: caller[0] -> caller[1]

PR #317, opened 2020-01-27
more normalizers in the correct way

hnyu:
Also correct EMNormalizer and WindowNormalizer, in addition to AdaptiveNormalizer


emailweixu:
Some of the normalizers are not used anywhere in the code. Need test to make sure they can run.

Issue #315, opened 2020-01-23
trac_ddpg_pendulum failed


witwolf:
when  test with trac_ddpg_pendulum
```bash
python -m alf.bin.train --root_dir=tdp --gin_file=trac_ddpg_pendulum
```
get error msg below, still investagte on it 

    Traceback (most recent call last):
      File "/usr/lib/python3.6/runpy.py", line 193, in _run_module_as_main
        "__main__", mod_spec)
      File "/usr/lib/python3.6/runpy.py", line 85, in _run_code
        exec(code, run_globals)
      File "/home/hongyingxiang/FLA/alf/bin/train.py", line 88, in <module>
        app.run(main)
      File "/usr/local/lib/python3.6/dist-packages/absl/app.py", line 300, in run
        _run_main(main, args)
      File "/usr/local/lib/python3.6/dist-packages/absl/app.py", line 251, in _run_main
        sys.exit(main(argv))
      File "/home/hongyingxiang/FLA/alf/bin/train.py", line 79, in main
        train_eval(FLAGS.root_dir)
      File "/usr/local/lib/python3.6/dist-packages/gin/config.py", line 1032, in wrapper
        utils.augment_exception_message_and_reraise(e, err_str)
      File "/usr/local/lib/python3.6/dist-packages/gin/utils.py", line 49, in augment_exception_message_and_reraise
        six.raise_from(proxy.with_traceback(exception.__traceback__), None)
      File "<string>", line 3, in raise_from
      File "/usr/local/lib/python3.6/dist-packages/gin/config.py", line 1009, in wrapper
        return fn(*new_args, **new_kwargs)
      File "/home/hongyingxiang/FLA/alf/bin/train.py", line 73, in train_eval
        trainer.train()
      File "/home/hongyingxiang/FLA/alf/trainers/policy_trainer.py", line 315, in train
        summary_max_queue=self._summary_max_queue)
      File "/home/hongyingxiang/FLA/alf/utils/common.py", line 265, in run_under_record_context
        func()
      File "/home/hongyingxiang/FLA/alf/trainers/policy_trainer.py", line 345, in _train
        time_step=time_step)
      File "/home/hongyingxiang/FLA/alf/trainers/off_policy_trainer.py", line 74, in _train_iter
        update_counter_every_mini_batch=self._config.
      File "/home/hongyingxiang/FLA/alf/algorithms/off_policy_algorithm.py", line 105, in train
        mini_batch_length, update_counter_every_mini_batch)
[...]

PR #319, opened 2020-01-29
fix with trac ddpg

witwolf:
#315

### DIFF
--- a/alf/algorithms/trac_algorithm.py
+++ b/alf/algorithms/trac_algorithm.py
@@ -35,7 +35,8 @@ nest_map = tf.nest.map_structure
 TracExperience = namedtuple(
     "TracExperience", ["observation", "step_type", "state", "action_param"])
 
-TracInfo = namedtuple("TracInfo", ["observation", "state", "ac"])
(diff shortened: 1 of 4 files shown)


alf/algorithms/trac_algorithm.py
```diff
@@ -112,20 +113,29 @@ def predict(self, time_step: ActionTimeStep, state, epsilon_greedy):
...
-                state=self._ac_algorithm.convert_train_state_to_predict_state(
-                    state),
-                ac=policy_step.info))
+        return self._replace_info(policy_step, time_step, state)
 
     def train_step(self, exp: Experience, state):
         policy_step = self._ac_algorithm.train_step(exp, state)
+        return self._replace_info(policy_step, exp, state)
+
+    def _replace_info(self, policy_step, time_step, state):
```
emailweixu: We can make ddpg also include action_distribution in info structure so we can avoid doing this here.

emailweixu: > We can make ddpg also include action_distribution in info structure so we can avoid doing this here.

I did this in a different PR #321

PR #321, opened 2020-01-30
Fix TracAlgorithm with DDPG

emailweixu:
Also makes it work with EntropyTargetAlgorithm

### DIFF
--- a/alf/algorithms/ddpg_algorithm.py
+++ b/alf/algorithms/ddpg_algorithm.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 """Deep Deterministic Policy Gradient (DDPG)."""
 
-from collections import namedtuple
 import gin.tf
 import tensorflow as tf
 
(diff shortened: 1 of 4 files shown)


alf/algorithms/trac_algorithm.py
```diff
@@ -239,16 +227,13 @@ def _update_total_dists(new_action, exp, total_dists):
...
-                new_action = policy_step.action
-            else:
-                assert (
-                    common.is_namedtuple(policy_step.info)
-                    and "action_distribution" in policy_step.info._fields
-                ), ("PolicyStep.info from ac_algorithm.predict() should be "
-                    "a namedtuple containing `action_distribution` in order to "
-                    "use TracAlgorithm.")
-                new_action = policy_step.info.action_distribution
+            assert (
```
witwolf: redundant assertion, already checked in `_make_policy_step`

alf/algorithms/ddpg_algorithm.py
```diff
@@ -157,10 +159,13 @@ def _sample(a, ou):
                 tf.less(tf.random.uniform((), 0, 1),
                         epsilon_greedy), lambda: a + ou(), lambda: a)
 
-        action = tf.nest.map_structure(_sample, action, self._ou_process)
+        noisy_action = tf.nest.map_structure(_sample, action, self._ou_process)
```
witwolf: should clip noise action  to match action spec

alf/algorithms/trac_algorithm.py
emailweixu: That only asserts for train_step() and rollout(). Here is predict().

alf/algorithms/ddpg_algorithm.py
emailweixu: fixed.

PR #323, opened 2020-02-03
deterministic train and play

hnyu:
Makes training and playing truly deterministic under a fixed seed, using tf2.1 newly enabled env variable "TF_DETERMINISTIC_OPS=1". Tested on GPU:
1. `train_play_test.py` all cases will now definitely pass when `test_perf=True` is set
2. `grid_search.py` will generate identical runs if a fixed seed is shared between runs

To be further investigated: when `random_seed=None`, `grid_search.py` with CPU training tends to generate two identical runs (probably they are launched too closely in time and time is used as seeds), while GPU training won't (before tf2.1).

### DIFF
--- a/alf/bin/train_play_test.py
+++ b/alf/bin/train_play_test.py
@@ -37,8 +37,11 @@ def run_cmd(cmd, cwd=None):
     """
     logging.info("Running %s", " ".join(cmd))
 
+    new_env = os.environ.copy()
+    new_env["CUDA_VISIBLE_DEVICES"] = ''  # force to use CPU for testing
+
     process = subprocess.Popen(
-        cmd, stdout=sys.stderr, stderr=sys.stderr, cwd=cwd)
+        cmd, stdout=sys.stderr, stderr=sys.stderr, cwd=cwd, env=new_env)
 
     process.communicate()
 
@@ -141,6 +144,14 @@ PPO_TRAIN_CONF = OFF_POLICY_TRAIN_CONF + [
     'TrainerConfig.num_updates_per_train_step=2'
 ]
 PPO_TRAIN_PARAMS = _to_gin_params(PPO_TRAIN_CONF)
+
+TEST_PERF_CONF = [
+    # train long enough (but not too long) to test the performance
+    'TrainerConfig.num_iterations=10',
+    'TrainerConfig.evaluate=True',
+    'TrainerConfig.eval_interval=10'
+]
+TEST_PERF_PARAMS = _to_gin_params(TEST_PERF_CONF)
 
 # Run COMMAND in a virtual X server environment
 XVFB_RUN = ['xvfb-run', '-a', '-e', '/dev/stderr']
@@ -272,7 +283,8 @@ class TrainPlayTest(tf.test.TestCase):
             'python3', '-m', 'alf.bin.train',
             '--root_dir=%s' % root_dir,
             '--gin_file=%s' % gin_file,
-            '--gin_param=TrainerConfig.random_seed=1'
+            '--gin_param=TrainerConfig.random_seed=1',
+            '--gin_param=TrainerConfig.use_tf_functions=False'
         ]
         if 'DISPLAY' not in os.environ:
             cmd = XVFB_RUN + cmd
@@ -289,7 +301,7 @@ class TrainPlayTest(tf.test.TestCase):
         """
         cmd = [
             'python3', '-m', 'alf.bin.play',
-            '--root_dir=%s' % root_dir, '--num_episodes=1', '--random_seed=0'
+            '--root_dir=%s' % root_dir, '--num_episodes=1'
         ]
         if 'DISPLAY' not in os.environ:
             cmd = XVFB_RUN + cmd
@@ -318,10 +330,13 @@ class TrainPlayTest(tf.test.TestCase):
 
     def test_ac_cart_pole(self):
         def _test_func(returns, lengths):
-            self.assertGreater(np.mean(returns[-2:]), 198)
-            self.assertGreater(np.mean(lengths[-2:]), 198)
+            self.assertEqual(returns[-1], 200)
+            self.assertEqual(lengths[-1], 200)
 
-        self._test(gin_file='ac_cart_pole.gin', test_perf_func=_test_func)
+        self._test(
+            gin_file='ac_cart_pole.gin',
+            test_perf_func=_test_func,
+            extra_train_params=TEST_PERF_PARAMS)
 
     def test_ac_simple_navigation(self):
         self._test(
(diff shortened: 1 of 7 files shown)


alf/trainers/policy_trainer.py
```diff
@@ -248,6 +248,7 @@ def __init__(self, config: TrainerConfig):
     def initialize(self):
         """Initializes the Trainer."""
         if self._random_seed is not None:
+            os.environ["TF_DETERMINISTIC_OPS"] = str(1)
             random.seed(self._random_seed)
             np.random.seed(self._random_seed)
             tf.random.set_seed(self._random_seed)
```
le-horizon: Let's extract the two blocks and move into a e.g. common.set_random_seed(seed) function?

hnyu: Moved

alf/trainers/policy_trainer.py
```diff
@@ -455,11 +448,6 @@ def play(root_dir,
     root_dir = os.path.expanduser(root_dir)
     train_dir = os.path.join(root_dir, 'train')
 
-    if random_seed is not None:
-        random.seed(random_seed)
-        np.random.seed(random_seed)
-        tf.random.set_seed(random_seed)
-
```
le-horizon: Don't forget to common.set_random_seed here?

hnyu: This has been moved to play.py because env is initialized before this.

hnyu:
Some issues still exist; close for now

le-horizon:
Do you know if pendulum's failure is because seed is bad or because behavior is still random?

hnyu:
> Do you know if pendulum's failure is because seed is bad or because behavior is still random?

That seed worked on my desktop. I think CPU mode is still nondeterministic.

hnyu:
OK, here is the current (tmp) solution.
1. On GPU, all behaviors are deterministic if a seed is given. 
2. On CPU, behaviors are deterministic only in eager mode but not in graph mode if a seed is given. Combining the previous observation 

> To be further investigated: when random_seed=None, grid_search.py with CPU training tends to generate two identical runs (probably they are launched too closely in time and time is used as seeds), while GPU training won't (before tf2.1).

I suspect that the real issue is that part of TF (or some third-party libs) uses the current time as the seed no matter what has been provided by the user, in graph mode.

3. To make the CI testing results deterministic, I change the _test_train to enable eager mode. In order to speed up the training, I now only train for 10 iterations for `test_ac_cart_pole`, `test_ppo_cart_pole`, `test_ddpg_pendulum`, `test_sac_pendulum`, and check their performance before convergence. 

- Disadvantage: every time someone changes something related, for example, even improving the DDPG algorithm, the test will fail because that number won't match. So that person is responsible for updating the number.
- Advantage: if someone writes some code which doesn't mean to affect DDPG but somehow introduces a bug in DDPG, and this bug won't affect the final performance in a substantial way, then the current way can still detect it.

4. To make the behaviors consistent between desktop and github CI, I also force every training in `train_play_test.py` to run on CPU.

alf/bin/train_play_test.py
```diff
@@ -318,8 +322,8 @@ def test_ac_breakout(self):
 
     def test_ac_cart_pole(self):
         def _test_func(returns, lengths):
-            self.assertGreater(np.mean(returns[-2:]), 198)
-            self.assertGreater(np.mean(lengths[-2:]), 198)
+            self.assertGreater(returns[-1], 195)
```
le-horizon: instead of assertGreater, just assert last 5 values == (....)?

This way, any unintended change in training behavior will be detected by the test.

alf/bin/train_play_test.py
```diff
@@ -331,7 +335,7 @@ def test_ac_simple_navigation(self):
 
     def test_ddpg_pendulum(self):
         def _test_func(returns, lengths):
-            self.assertGreater(np.mean(returns[-5:]), -200)
+            self.assertGreater(returns[-1], -200)
```
le-horizon: same here, and the other places as well?

alf/trainers/off_policy_trainer.py
```diff
@@ -85,8 +85,11 @@ def __init__(self, config):
         self._driver_started = False
 
     def _init_driver(self):
-        for _ in range(1, self._config.num_envs):
-            self._create_environment()
+        for i in range(1, self._config.num_envs):
+            # [self._random_seed, self._random_seed + batch_size) has been used
+            # in policy_trainer.py
+            self._create_environment(
+                random_seed=self._random_seed + i * common._env.batch_size)
```
le-horizon: Assuming batch_size == self._config.num_envs, maybe just use + i * self._config.num_envs, so code is more readable?

alf/trainers/off_policy_trainer.py
```diff
@@ -85,8 +85,11 @@ def __init__(self, config):
         self._driver_started = False
 
     def _init_driver(self):
-        for _ in range(1, self._config.num_envs):
-            self._create_environment()
+        for i in range(1, self._config.num_envs):
+            # [self._random_seed, self._random_seed + batch_size) has been used
```
le-horizon: where is self._random_seed + batch_size used in policy_trainer?

alf/bin/train_play_test.py
hnyu: Discussed with Wei about this, we probably don't want a very strict test case.

alf/trainers/off_policy_trainer.py
hnyu: num_envs is how many batched environments we have (only for async training). batch_size is the size of a batched environment

alf/trainers/off_policy_trainer.py
hnyu: in initialize() we first create an environment as the global one

le-horizon: Oh, the comment says from random_seed to random_seed + batch have been used?  Misunderstood it.

hnyu:
@emailweixu  Do you have additional comments?

alf/bin/train_play_test.py
```diff
@@ -224,7 +227,7 @@ def _test(self,
               extra_train_params=None,
               test_play=True,
               extra_play_params=None,
-              test_perf=False,
+              test_perf=True,
```
emailweixu: Is the argument necessary? Seems having `test_perf_func` is enough

alf/trainers/off_policy_trainer.py
emailweixu: Should handle the case where self._random_seed is None.

alf/bin/train_play_test.py
hnyu: I think we can just leave it there in case that we want to turn off performance checking to save some CI time.

alf/trainers/off_policy_trainer.py
```diff
@@ -85,8 +85,14 @@ def __init__(self, config):
...
+            # in policy_trainer.py
+            if self._random_seed is None:
+                self._create_environment()
+            else:
+                self._create_environment(
+                    random_seed=self._random_seed + i * common._env.batch_size)
```
emailweixu: Why was this bug not caught by CI?

hnyu: Maybe because this is only async training with multiple envs?

emailweixu: Perhaps we should add a test for it. We can do it in another PR.

hnyu:
Also made some changes to address the issue of multiple CPU runs having identical results if they are launched in a short time interval (e.g., by `grid_search.py`). See #296

PR #327, opened 2020-02-07
rollout summary for off-policy training

emailweixu:
Previously, for off-policy training, some of the summaries from rollout_info are from replay buffer, whose statistics can be significant from the current on-going rolllout. This PR generate summaries for the current rollout instead of from replay buffer.
See this colab notebook for example:
https://colab.research.google.com/drive/1YctzlIdgdya-JPkdzQnp0pqWR875ANY7#scrollTo=_vtbpjfnKYMR

### DIFF
--- a/alf/algorithms/agent.py
+++ b/alf/algorithms/agent.py
@@ -233,7 +233,9 @@ class Agent(OnPolicyAlgorithm):
             new_state = new_state._replace(icm=icm_step.state)
 
         rl_step = self._rl_algorithm.train_step(
-            exp._replace(observation=observation), state.rl)
+            exp._replace(
+                observation=observation, rollout_info=exp.rollout_info.rl),
+            state.rl)
 
         new_state = new_state._replace(rl=rl_step.state)
         info = info._replace(rl=rl_step.info)
@@ -294,8 +296,16 @@ class Agent(OnPolicyAlgorithm):
                                              new_loss_info.scalar_loss),
                 extra=loss_info.extra._replace(**{name: new_loss_info.extra}))
 
+        def _make_training_info(training_info, name):
+            if training_info.rollout_info == ():
+                rollout_info = ()
+            else:
+                rollout_info = getattr(training_info.rollout_info, name)
+            info = getattr(training_info.info, name)
+            return training_info._replace(info=info, rollout_info=rollout_info)
+
         rl_loss_info = self._rl_algorithm.calc_loss(
-            training_info._replace(info=training_info.info.rl))
+            _make_training_info(training_info, 'rl'))
         loss_info = rl_loss_info._replace(
             extra=AgentLossInfo(rl=rl_loss_info.extra))
         loss_info = _update_loss(loss_info, training_info, 'icm', self._icm)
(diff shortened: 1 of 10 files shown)


alf/algorithms/agent.py
```diff
@@ -233,7 +233,9 @@ def train_step(self, exp: Experience, state):
             new_state = new_state._replace(icm=icm_step.state)
 
         rl_step = self._rl_algorithm.train_step(
-            exp._replace(observation=observation), state.rl)
+            exp._replace(
+                observation=observation, rollout_info=exp.rollout_info.rl),
```
hnyu: Why do we still need to put rollout_info in train_step if now summarize_rollout is separated from summarize_train? From my understanding, if we want to summarize rollout statistics, we can only do it once when collecting experiences.

emailweixu: Some algorithms need rollout_info (e.g. PPO)

alf/algorithms/agent.py
```diff
@@ -294,8 +296,16 @@ def _update_loss(loss_info, training_info, name, algorithm):
                                              new_loss_info.scalar_loss),
                 extra=loss_info.extra._replace(**{name: new_loss_info.extra}))
 
+        def _make_training_info(training_info, name):
```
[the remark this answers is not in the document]
emailweixu: right. I make it function since we may need to use it multiple times later.

alf/algorithms/agent.py
hnyu: I mean whether updating `rollout_info=exp.rollout_info.rl` here won't matter as this field will no longer be used later in the code (not for summary anymore), right?

hnyu: I mean this update `rollout_info=exp.rollout_info.rl` here is meaningless because it won't be used later in `OnPolicyAlgorithm train_step`

alf/algorithms/rl_algorithm.py
```diff
@@ -293,7 +293,23 @@ def observe(self, exp: Experience):
...
-    def training_summary(self, training_info, loss_info, grads_and_vars):
+    def summarize_rollout(self, training_info):
+        if self._debug_summaries:
+            summary_utils.add_action_summaries(
+                training_info.action, self._action_spec, "rollout_action")
+            self.add_reward_summary("rollout_reward/extrinsic",
+                                    training_info.reward)
+
+        if self._summarize_action_distributions:
+            field = nest_utils.find_field(training_info.info,
```
Haichao-Zhang: Why do we perform ```summarize_rollout``` using ```training_info.info``` instead of ```training_info.rollout_info```? This part is the same as ```summarize_train```.
Maybe useful to document what are the fields of ```training_info```  are expected to be used in this function (and ```summarize_train```).

alf/algorithms/agent.py
emailweixu: An algorithm can choose what ever it does with rollout_info. There is no guarantee it will not be used.

alf/algorithms/rl_algorithm.py
emailweixu: policy_step.info is saved to training_info.info. I'll change it to rollout_info.

PR #331, opened 2020-02-10
add mrevenge with RND

hnyu:
Update `rnd_algorithm.py` and add the example of training MRevenge.


alf/utils/common.py
```diff
@@ -1078,3 +1078,18 @@ def set_random_seed(seed, eager_mode):
     random.seed(seed)
     np.random.seed(seed)
     tf.random.set_seed(seed)
+
+
+@gin.configurable
```
emailweixu: TensorSpec is a gin configurable defined in utils/external_configurables.py

alf/algorithms/rnd_algorithm.py
```diff
@@ -91,29 +124,38 @@ def train_step(self,
...
             observation = self._observation_normalizer.normalize(observation)
 
+        if self._encoder_net is not None:
+            observation = tf.stop_gradient(self._encoder_net(observation)[0])
+
         pred_embedding, _ = self._predictor_net(observation)
         target_embedding, _ = self._target_net(observation)
 
-        loss = 0.5 * tf.reduce_mean(
+        loss = tf.reduce_sum(
```
emailweixu: why change to reduce_sum?

alf/examples/ppo_rnd_mrevenge.gin
```diff
@@ -0,0 +1,78 @@
+# RND on MontezumaRevenge
```
emailweixu: add to train_play_test.py

alf/algorithms/rnd_algorithm.py
```diff
@@ -44,30 +43,64 @@ class RNDAlgorithm(Algorithm):
...
+                 reward_adapt_speed=None,
                  observation_adapt_speed=None,
                  observation_spec=None,
+                 learning_rate=None,
+                 clip_value=-1.0,
+                 stacked_frames=True,
                  name="RNDAlgorithm"):
         """
         Args:
+            encoder_net (Network): a shared network that encodes imgs to
```
Haichao-Zhang: ```encodes imgs ...```, is there an explicit assumption that the inputs have to be images? It seems not.

hnyu: If it's not, then this can be None

alf/algorithms/rnd_algorithm.py
hnyu: That's what I usually have for experiments

Haichao-Zhang: Is this related to why a small learning rate can be used for training with good results in your experiments?

hnyu: I never used a small learning rate? This is not learning improvements, just RND

alf/examples/rnd_super_mario.gin
```diff
@@ -65,6 +67,7 @@ TrainerConfig.use_tf_functions=True
 TrainerConfig.debug_summaries=True
 TrainerConfig.summarize_grads_and_vars=False
 TrainerConfig.summary_interval=10
+TrainerConfig.update_counter_every_mini_batch=True
```
emailweixu: Is this just for debug?

hnyu: For updating the curve more frequently, which is what I always use

emailweixu: But this does not have effect for on_policy_trainer.

hnyu: I thought this example uses PPO

PR #332, opened 2020-02-13
Fix diayn and derive GoalGenerator from RLAlgorithm

emailweixu:
Some new changes are not tested against diayn. Added it to train_play_test.

The generator need to drive from RLAlgorithm so that we use rl algorithm to train a policy to generate the goal.

### DIFF
--- a/alf/algorithms/goal_generator.py
+++ b/alf/algorithms/goal_generator.py
@@ -20,8 +20,9 @@ import functools
 import tf_agents.specs.tensor_spec as tensor_spec
 from tf_agents.trajectories.time_step import StepType
 
-from alf.algorithms.algorithm import Algorithm, AlgorithmStep, LossInfo
-from alf.data_structures import TrainingInfo, namedtuple, ActionTimeStep
+from alf.algorithms.rl_algorithm import RLAlgorithm
+from alf.data_structures import (ActionTimeStep, Experience, LossInfo,
+                                 namedtuple, PolicyStep, TrainingInfo)
 
 import alf.utils.common as common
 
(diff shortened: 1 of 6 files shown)


alf/algorithms/goal_generator.py
```diff
@@ -30,28 +31,40 @@
...
+class RandomCategoricalGoalGenerator(RLAlgorithm):
     """Random Goal Generation Module
 
     This module generates a random categorical goal for the agent
     in the beginning of every episode.
     """
 
-    def __init__(self, num_of_goals, name="RandomCategoricalGoalGenerator"):
+    def __init__(self,
+                 observation_spec,
```
Haichao-Zhang: For the ```observation_spec```here, I understand this is due to the change of the base class to ```RLAlgorithm```, but still it feels a bit unnatural to take something unused as input.

alf/algorithms/goal_generator.py
```diff
@@ -86,6 +96,7 @@ def train_step(self,
             calc_goal_update (bool): perform goal alculattion the goal if true;
```
Haichao-Zhang: comments for ```calc_goal_update``` can be removed

alf/algorithms/goal_generator.py
hnyu: #335

hnyu: > For the `observation_spec`here, I understand this is due to the change of the base class to `RLAlgorithm`, but still it feels a bit unnatural to take something unused as input.

Check out this issue #335

emailweixu: You actually use it at line 66.

PR #334, opened 2020-02-13
shuffle with gradient

emailweixu:
tf.random.shuffle() has no gradient defined for `value`.  This `shuffle`
can propagate gradient of `value` by using `gather`.

Also fix _shuffle_sampler for MIEstimator


alf/utils/math_ops.py
```diff
@@ -82,3 +82,22 @@ def max_n(inputs):
...
+    can propagate gradient of `value` by using `gather`.
+
+    Args:
+        values (nested Tensor): nested Tensor to be shuffled. All the tensor
+            need to have the same batch size (i.e. shape[0]).
+        seed (int): Used to create a random seed for the distribution.
+    Returns:
+        shuffled value along dimension 0.
+    """
+    batch_size = tf.shape(tf.nest.flatten(values)[0])[0]
```
hnyu: can use `get_nest_batch_size` in `nest_utils.py`

PR #355, opened 2020-02-15
Update setup.py and CI

emailweixu:
alf/environments/simple/noisy_array_test is the only test which can pass for now.
We need to add successful tests to build.sh one by one


emailweixu:
> Nice starting point! LG. Would be good to add a TODO to add instructions on how to build and update alf-* docker images and tags, in our wiki, if it's not already there.
> 
> Thanks!

It's in docs/update.md

PR #366, opened 2020-02-19
encoding networks

hnyu:
Encoding network assumes unnested tensor specs. I don't think it's necessary to handle nested cases (like tfa's way) which will depend on various scenarios and it's not an appropriate job that an encoding network should do. For nest_concatenate or other preprocessing combiners, we should do that before calling an encoding network (maybe write a new wrapper class).


hnyu:
Close this PR and refer to #369 (duplicate) .

PR #367, opened 2020-02-19
Pr pytorch data structure

runjerry:
Keep most of the data structures in the original data_structures, unifying ActionTimeStep and TimeStep.


alf/data_structures.py
```diff
@@ -124,30 +150,6 @@ def make_action_time_step(time_step: TimeStep, prev_action, first_env_id=0):
         'state'  # state passed to rollout() to generate `action`
     ])
 
-
```
emailweixu: why do you remove this function? Seems we still need this.

alf/data_structures.py
```diff
@@ -42,55 +40,83 @@ def namedtuple(typename, field_names, default_value=None, default_values=()):
...
-            return tf.equal(self.step_type, StepType.LAST)
-        return np.equal(self.step_type, StepType.LAST)
-
+        if torch.is_tensor(self.step_type):
+            return torch.eq(self.step_type, StepType.LAST)
+        raise ValueError('step_type is not a Torch Tensor')
 
-def make_action_time_step(time_step: TimeStep, prev_action, first_env_id=0):
-    """Make ActionTimeStep from TimeStep.
+    # def __hash__(self):
```
emailweixu: what is the purpose of this function?

alf/data_structures.py
```diff
@@ -42,55 +40,83 @@ def namedtuple(typename, field_names, default_value=None, default_values=()):
...
+    The first `TimeStep` in a sequence will equal `StepType.FIRST`. The final
+    `TimeStep` will equal `StepType.LAST`. All other `TimeStep`s in a sequence
+    will equal `StepType.MID.
+
+    Attributes:
+      step_type: a `Tensor` of `StepType` enum values.
+      reward: a `Tensor` of reward values from executing 'prev_action'.
+      discount: A discount value in the range `[0, 1]`.
+      observation: A (nested) 'Tensor' for observation
+      prev_action: A 'Tensor' for action from previous tiem step
```
emailweixu: tiem => time

.ci-cd/build.sh
```diff
@@ -44,6 +44,7 @@ function test() {
     python3 -m unittest -v \
         alf.environments.simple.noisy_array_test
     python3 -m unittest -v alf.tensor_specs_test
+    python3 -m unittest -v alf.data_structures_test
```
emailweixu:
```bash
python3 -m unittest -v \
    alf.environments.simple.noisy_array_test \
    alf.tensor_specs_test \
    alf.tensor_specs_test
```

alf/data_structures.py
```diff
@@ -42,55 +40,83 @@ def namedtuple(typename, field_names, default_value=None, default_values=()):
...
-    """
-    return ActionTimeStep(
-        step_type=time_step.step_type,
-        reward=time_step.reward,
-        discount=time_step.discount,
-        observation=time_step.observation,
-        prev_action=prev_action,
-        env_id=first_env_id + tf.range(time_step.step_type.shape[0]))
 
+PolicyStep = namedtuple('PolicyStep', ('action', 'state', 'info'))
```
emailweixu: default_value=(). Otherwise the default_value is None, which cannot be handled by nest.

alf/data_structures.py
runjerry: We do not have the old TimeStep, the new TimeStep has 'action' field, so maybe this function can be substituted by time_step._replace(action=action) ?

PR #369, opened 2020-02-19
projection net, encoding net, layers, and initializers

hnyu:
It has a dependency on encoding networks #366. For all networks, I always assume tensor outer_dims=0 to keep the APIs and code maintenance simpler. If there are use cases of multi outer_dims, the user can write a wrapper class based on specific problems. Also assume no nested input/output tensor specs for the same reason.

### DIFF
--- a/alf/layers.py
+++ b/alf/layers.py
@@ -37,7 +37,7 @@ class FC(nn.Module):
                  use_bias=True,
                  kernel_init_gain=1.0,
                  bias_init_value=0.0):
-        """A fully connected layer that's also responsible for activaiton and
+        """A fully connected layer that's also responsible for activation and
         customized weights initialization. An auto gain calculation might depend
         on the activation following the linear layer. Suggest using this wrapper
         module instead of nn.Linear if you really care about weight std after
@@ -173,7 +173,8 @@ class ConvTranspose2D(nn.Module):
         variance_scaling_init(
             self._conv_trans2d.weight.data,
             gain=kernel_init_gain,
-            nonlinearity=self._activation.__name__)
+            nonlinearity=self._activation.__name__,
+            transposed=True)
         if use_bias:
             nn.init.constant_(self._conv_trans2d.bias.data, bias_init_value)
 
--- a/alf/layers.py
+++ b/alf/layers.py
@@ -11,104 +11,155 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""Various layers."""
+"""Some basic layers."""
 
 import gin
-import six
 
-import tensorflow as tf
+import torch
+import torch.nn as nn
+
+from alf.networks.initializers import variance_scaling_init
 
 
-class ListWrapper(list):
-    """A wrapper of list.
-
-    It makes it possible to add some attributes to the object which is
-    necessary for keras Layer.
+def identity(x):
+    """PyTorch doesn't have an identity activation. This can be used as a
+    placeholder.
     """
+    return x
 
-    pass
 
-
-class Split(tf.keras.layers.Layer):
-    """Splits a tensor into sub tensors."""
-
-    def __init__(self, num_or_size_splits, axis):
-        """Create a Split layer.
-
-        If `num_or_size_splits` is an integer, then `input` is split along
-        dimension `axis` into `num_split` smaller tensors. This requires that
-        `num_split` evenly divides `value.shape[axis]`. If `num_or_size_splits`
-        is a 1-D Tensor (or list), we call it `size_splits` and `input` is split
-        into `len(size_splits)` elements. The shape of the `i`-th element has
-        the same size as the `value` except along dimension `axis` where the
-        size is `size_splits[i]`.
+@gin.configurable
+class FC(nn.Module):
+    def __init__(self,
+                 input_size,
+                 output_size,
+                 activation=identity,
+                 use_bias=True,
+                 kernel_init_gain=1.0,
+                 bias_init_value=0.0):
+        """A fully connected layer that's also responsible for activaiton and
+        customized weights initialization. An auto gain calculation might depend
+        on the activation following the linear layer. Suggest using this wrapper
+        module instead of nn.Linear if you really care about weight std after
+        init.
 
         Args:
-            num_or_size_splits (int|list[int]): Either an integer indicating the
-                number of splits along split_dim or a 1-D integer `Tensor` or Python
-                list containing the sizes of each output tensor along split_dim. If
-                a scalar then it must evenly divide `input.shape[axis]`; otherwise
-                the sum of sizes along the split dimension must match that of the
-                `input`.
-            axis: An integer or scalar `int32` `Tensor`. The dimension along which
-                to split. Must be in the range `[-rank(input), rank(input))`.
+            input_size (int): input size
+            output_size (int): output size
+            activation (torch.nn.functional):
+            use_bias (bool): whether use bias
+            kernel_init_gain (float): a scaling factor (gain) applied to
+                the std of kernel init distribution
+            bias_init_value (float): a constant
         """
-        self._num_or_size_splits = num_or_size_splits
-        self._axis = axis
-        super(Split, self).__init__()
+        super(FC, self).__init__()
+        self._activation = activation
+        self._linear = nn.Linear(input_size, output_size, bias=use_bias)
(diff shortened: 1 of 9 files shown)


alf/layers.py
```diff
@@ -11,104 +11,179 @@
...
+@gin.configurable
+class FC(nn.Module):
+    def __init__(self,
+                 input_size,
+                 output_size,
+                 activation=identity,
+                 use_bias=True,
+                 kernel_init_gain=1.0,
+                 bias_init_value=0.0):
+        """A fully connected layer that's also responsible for activaiton and
```
emailweixu: activaiton => activation

alf/layers.py
```diff
@@ -11,104 +11,179 @@
...
+        self._activation = activation
+        self._conv_trans2d = nn.ConvTranspose2d(
+            in_channels,
+            out_channels,
+            kernel_size,
+            stride=strides,
+            padding=padding,
+            bias=use_bias)
+        variance_scaling_init(
+            self._conv_trans2d.weight.data,
```
emailweixu: is weight.data transposed or not? If it is transposed, then the fanin fanout need to be switched.

alf/networks/encoding_networks.py
```diff
@@ -0,0 +1,304 @@
+# Copyright (c) 2020 Horizon Robotics. All Rights Reserved.
```
emailweixu: This is in the other PR #366. You either make 1 PR or have separate changes in 2 PRs.

alf/networks/projection_networks.py
```diff
@@ -0,0 +1,150 @@
...
+
+import alf.layers as layers
+from alf.tensor_specs import TensorSpec, BoundedTensorSpec
+
+
+def DiagMultivariateNormal(loc, scale_diag):
+    return td.Independent(td.Normal(loc, scale_diag), 1)
+
+
+class CategoricalProjectionNetwork(nn.Module):
```
emailweixu: tf_agent (Categorical|Normal)ProjectionNetwork are subclass of DistributionNetwork, which has a property output_spec. I'm not sure we will need it or not. Just keep it in mind that we may have to implement it later.

alf/networks/encoding_networks.py
hnyu: Yeah this PR depends on #366. I can remove the files, but it will make the review more difficult (you have to look at code in another PR).

alf/layers.py
hnyu: It's already transposed

hnyu: Yeah it's transposed.

alf/networks/encoding_networks.py
```diff
@@ -0,0 +1,304 @@
...
+                layers.Conv2D(
+                    input_channels,
+                    filters,
+                    kernel_size,
+                    activation=activation,
+                    strides=strides,
+                    padding=padding))
+            input_channels = filters
+
+    def output_shape(self, input_size):
```
emailweixu: Need to correctly handle when kernel_size, strides, padding are tuples

alf/networks/encoding_networks.py
```diff
@@ -0,0 +1,304 @@
...
+            H = (H1-1) * strides + HF - 2P
+
+        where H = output size, H1 = input size, HF = size of kernel, P = padding
+
+        Returns:
+            a tuple representing the output shape (C,H,W)
+        """
+        _, height, width = self._start_decoding_shape
+        for paras in self._transconv_layer_params:
+            filters, kernel_size, strides = paras[:3]
```
emailweixu: handle tuple

alf/networks/encoding_networks.py
```diff
@@ -0,0 +1,304 @@
...
+                 fc_layer_params=None,
+                 activation=layers.identity,
+                 last_layer_size=None,
+                 last_activation=None):
+        """Create an EncodingNetwork
+
+        This EncodingNetwork allows the last layer to have different settings
+        from the other layers.
+
+        Args:
```
emailweixu: TODO: Handle nested input_tensor_spec

alf/networks/initializers.py
```diff
@@ -0,0 +1,103 @@
...
+            so we need to compensate for that. Only used when mode=="fan_in".
+        nonlinearity (str): one of ['linear', 'conv1d', 'conv2d', 'conv3d',
+            'conv_transpose1d', 'conv_transpose2d', 'conv_transpose3d',
+            'sigmoid', 'tanh', 'relu', 'leaky_relu', 'identity']
+        nonlinearity_param (float): additional parameter of the nonlinearity;
+            currently only used by 'leaky_relu' as the negative slope (pytorch
+            default 0.01)
+        transposed (bool): a flag indicating if the weight tensor has been
+            tranposed (e.g., nn.ConvTranspose2d). In that case, `fan_in` and
+            `fan_out` should be swapped.
```
emailweixu: there are return at line 96 and 103. Need Returns comment section

alf/networks/projection_networks.py
```diff
@@ -0,0 +1,150 @@
...
+            self._std_transform = std_transform
+
+        self._means_projection_layer = layers.FC(
+            input_size,
+            action_spec.shape[0],
+            activation=activation,
+            kernel_init_gain=projection_output_init_gain)
+
+        # We always assume that std depends on states
+        self._std_projection_layer = layers.FC(
```
emailweixu: For many models, we don't use state dependent std

alf/networks/projection_networks_test.py
```diff
@@ -0,0 +1,132 @@
...
+        embedding = input_spec.ones(outer_dims=(1, ))
+
+        net = CategoricalProjectionNetwork(
+            input_size=input_spec.shape[0],
+            num_actions=5,
+            logits_init_output_factor=0)
+        dist = net(embedding)
+        self.assertTrue(torch.all(dist.probs == 0.2))
+
+    def test_zero_mean_projection_net(self):
```
emailweixu: what is the meaning of this function?

alf/networks/encoding_networks.py
hnyu: I don't think it's necessary to complicate the APIs in general. The users can write subclasses for specific problems. For now, unnested tensor specs is the common case.

alf/networks/encoding_networks.py
```diff
@@ -0,0 +1,314 @@
...
+                layers.Conv2D(
+                    input_channels,
+                    filters,
+                    kernel_size,
+                    activation=activation,
+                    strides=strides,
+                    padding=padding))
+            input_channels = filters
+
+    def output_shape(self, input_size):
```
le-horizon: I'm wondering, can we pass a 0 tensor of input_size to forward() and get the output_size from the output tensor?

That'll probably reduce code redundancy and won't add much overhead?

If we do this, then we can have a very simple output_shape function in the top level class, and all our custom nn.Module's can just inherit without explicitly implementing it?

hnyu: Calculating output_shape is usually required for conv networks because for other layers you will know its value during construction. As for why I didn't use forward to get the output_shape is mainly for testing purpose in encoding_networks_test.py. The code amount won't be much less by doing forward.

le-horizon: I see.  There are already two Modules that need output shape in this PR.  I worry less about code length, and more about code maintenance in the future.  If using forward can solve the problem once and for all, then why not?

hnyu: Because for encoding_networks_test.py, I compare the forwarding-computed shape with the maunally calculated one. It's basically a sanity check.

le-horizon: Then that means we can remove that test if we use forward to compute output shape in the output_shape() function?

hnyu: Then how do you check if the networks are implemented correctly?

le-horizon: Maybe move the manual computation logic into the test if you want.  Otherwise, just get a few example gold standard output shapes into the test?

le-horizon: The only downside I can think of is whenever output_shape() is called, especially when input size is large, gpu memory could rocket.  But it's hopefully just a one time cost.

hnyu: Also I personally don't feel comfortable that we are doing forward under the hood without telling the users. The log (error) message, if any, will be quite confusing, just like what we have now for prepare_specs.

le-horizon: Ok, fair enough.  We can always reconsider if needed.

PR #374, opened 2020-02-21
Algorithm

emailweixu:
Replace PolicyStep with AlgStep

### DIFF
--- a/alf/algorithms/algorithm.py
+++ b/alf/algorithms/algorithm.py
@@ -19,56 +19,19 @@ import copy
 import json
 import os
 
-import tensorflow as tf
-from tensorflow.python.util.serialization import get_json_type
+import torch
+import torch.nn as nn
 
-from tf_agents.utils import eager_utils
-
-from alf.data_structures import namedtuple, LossInfo
-import alf.utils
-
-AlgorithmStep = namedtuple("AlgorithmStep", ["outputs", "state", "info"])
+import alf
+from alf.data_structures import AlgStep, namedtuple, LossInfo
+from alf.utils import common
 
 
-def _is_alg(obj):
-    """Only return True if the obj in an instance of Algorithm."""
-    return isinstance(obj, Algorithm)
+def _get_optimizer_params(optimizer: torch.optim.Optimizer):
+    return set(sum([g['params'] for g in optimizer.param_groups], []))
 
 
-def _is_trainable_module(obj):
-    """Only return True if the module or var is trainable, to avoid
-    possible confusions shown in the optimizer info"""
-    return (isinstance(obj, tf.Module) and not isinstance(obj, Algorithm)
-            and obj.trainable_variables)
-
-
-def _is_trainable_var(obj):
-    """Only return True if the module or var is trainable, to avoid
-    possible confusions shown in the optimizer info"""
-    return isinstance(obj, tf.Variable) and obj.trainable
-
-
-def _get_trainable_vars(obj):
-    """Get trainable variable of obj, obj should be a trainable module
-    or variable
-
-    Args:
-        obj (tf.Module|tf.Variable)
-    Returns:
-        list[tf.Variable]
-    """
-    if obj is None:
-        return []
-    if _is_trainable_module(obj):
-        return obj.trainable_variables
-    elif _is_trainable_var(obj):
-        return [obj]
-    else:
-        raise ValueError(
-            "%s is not a trainable module or variable!" % obj.name)
-
-
-class Algorithm(tf.Module):
+class Algorithm(nn.Module):
     """Algorithm base class.
 
     Algorithm is a generic interface for supervised training algorithms.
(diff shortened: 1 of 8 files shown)


alf/algorithms/algorithm.py
```diff
@@ -206,124 +158,105 @@ def _trainable_attributes_to_ignore(self):
...
-                assert id(module) != id(self)
-                if isinstance(module, tf.Variable):
-                    vars.append(module)
-                elif isinstance(module, tf.Module):
-                    vars += list(module.trainable_variables)
-                else:
-                    raise ValueError("Unsupported module type %s" % module)
-            opt_and_var_sets.append((opt, vars))
-        return opt_and_var_sets
+    def optimizers(self, recurse=True):
```
Haichao-Zhang: ```recurse``` is not used

alf/utils/common.py
```diff
@@ -295,11 +292,11 @@ def cast_transformer(observation, dtype=tf.float32):
     """
 
     def _cast(obs):
-        if isinstance(obs, tf.Tensor):
-            return tf.cast(obs, dtype)
+        if isinstance(obs, torch.Tensor):
+            return torch.cast(obs, dtype)
```
Haichao-Zhang: torch does not have ```cast```.  maybe can use ```return obs.to(dtype)```

alf/algorithms/algorithm.py
```diff
@@ -206,124 +158,105 @@ def _trainable_attributes_to_ignore(self):
...
-                    optimizers.append(opt)
-                    algorithm_names.append(alg_name)
-                    module_sets.append(module_set)
-                else:
-                    module_sets[0].extend(module_set)
+                children.append(module)
+
+        for name, param in self.named_parameters(recurse=False):
+            if name in self._trainable_attributes_to_ignore():
+                pass
```
hnyu: pass or continue?

alf/algorithms/algorithm.py
```diff
@@ -150,43 +96,49 @@ def __init__(self,
...
+        retrieved in self.children() or self.parameters())
+
+        Args:
+            optimizer (Optimizer): optimizer
+            modules_and_params (list of Module or Parameter): The modules and
+                parameters to be optimized by `optimizer`
         """
-        return sum([var_set for _, var_set in self._get_opt_and_var_sets()],
-                   [])
+        params = []
```
hnyu: `params` not used anywhere in this function?

alf/algorithms/algorithm.py
emailweixu: fixed

alf/utils/common.py
emailweixu: fixed

alf/algorithms/algorithm.py
emailweixu: fixed

alf/utils/common.py
```diff
@@ -295,11 +292,11 @@ def cast_transformer(observation, dtype=tf.float32):
     """
 
     def _cast(obs):
-        if isinstance(obs, tf.Tensor):
-            return tf.cast(obs, dtype)
+        if isinstance(obs, torch.Tensor):
+            return torch.type(obs, dtype)
```
Haichao-Zhang: ```torch.type``` does not exist?

emailweixu: torch.type does exist

hnyu: I think you need `obs.type(dtype)`. torch.type doesn't exist.

emailweixu: Fixed

alf/algorithms/algorithm.py
```diff
@@ -150,43 +109,47 @@ def __init__(self,
...
-
-        self._cached_opt_and_var_sets = None
+        self._is_rnn = len(alf.nest.flatten(train_state_spec)) > 0
+
         self._gradient_clipping = gradient_clipping
         self._clip_by_global_norm = clip_by_global_norm
         self._debug_summaries = debug_summaries
+        self._default_optimizer = optimizer
+        self._optimizers = []
+        self._default_optimizer = optimizer
```
hnyu: duplicate with line 124

alf/algorithms/algorithm.py
```diff
@@ -206,124 +169,95 @@ def _trainable_attributes_to_ignore(self):
...
+        handled = set()
+        for child in self._get_children():
+            if child in handled:
                 continue
-            for leaf in tf.nest.flatten(attr):
-                if predicate(leaf) and not id(leaf) in ids:
-                    children.append(leaf)
-                    ids.add(id(leaf))
-        return children
+            handled.add(child)
```
hnyu: Is it possible to detect a loop and report errors if a grandchild contains the current algorithm as a child?

alf/algorithms/algorithm.py
```diff
@@ -188,21 +187,54 @@ def default_optimizer(self):
...
+        to_be_visited = [self]
+        while to_be_visited:
+            node = to_be_visited.pop(0)
+            visited.add(node)
+            for child in node._get_children():
+                assert child not in visited, (
+                    "There is a cycle in the "
+                    "algorithm tree caused by '%s'" % child.name)
+                if isinstance(child, Algorithm):
+                    to_be_visited.append(child)
```
hnyu: I don't think this BFS will work, because even when a child at two levels lower is a duplicate of a node at the current level, by the time we visit that child, all the current-level nodes will have been removed from the queue.

Issue #344, opened 2020-02-14
TTP: ValueNetwork


emailweixu:
Depends on

- [ ] #345 EncodingNetwork

Issue #340, opened 2020-02-14
TTP: ActorDistributionNetwork


emailweixu:
Depends on:

- [ ] #338 CategoricalProjectionNetwork
- [ ] #339 NormalProjectionNetwork
- [ ] #345 EncodingNetwork

hnyu:
While we will try to minimize the use of specs, in places where we have to use, I think we can directly rely on `gym.spaces.discrete.Discrete` for our network modules.

PR #377, opened 2020-02-21
value and actor distribution networks

hnyu:
Issue #344 and #340

### DIFF
--- a/alf/networks/actor_distribution_networks.py
+++ b/alf/networks/actor_distribution_networks.py
@@ -20,7 +20,7 @@ import torch.nn as nn
 
 from alf.networks import EncodingNetwork
 from alf.networks import NormalProjectionNetwork, CategoricalProjectionNetwork
-from alf.networks.value_networks import _create_lstm_cell_state_spec
+from alf.layers import _create_lstm_cell_state_spec
 from alf.tensor_specs import BoundedTensorSpec, TensorSpec
 
 
--- a/alf/networks/actor_distribution_networks.py
+++ b/alf/networks/actor_distribution_networks.py
@@ -1,0 +1,165 @@
+# Copyright (c) 2020 Horizon Robotics. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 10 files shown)


alf/networks/value_networks.py
```diff
@@ -0,0 +1,155 @@
...
+
+import torch
+import torch.nn as nn
+
+from alf.networks import EncodingNetwork
+import alf.layers as layers
+from alf.tensor_specs import TensorSpec
+
+
+def _create_lstm_cell_state_spec(hidden_size, dtype):
```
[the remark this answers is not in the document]
hnyu: I moved it to layers.py

alf/networks/value_networks.py
```diff
@@ -23,25 +23,6 @@
 from alf.tensor_specs import TensorSpec
```
le-horizon: Does this file also need to include
`from alf.layers import _create_lstm_cell_state_spec`
?

alf/networks/actor_distribution_networks.py
```diff
@@ -0,0 +1,165 @@
...
+        Args:
+            observation (torch.Tensor): consistent with `input_tensor_spec`
+            state: empty for API consistent with ActorRNNDistributionNetwork
+
+        Returns:
+            act_dist (torch.distributions): action distribution
+            state: empty
+        """
+        act_dist = self._projection_net(self._encoding_net(observation))
+        return act_dist, state
```
emailweixu: need to implement state_spec. Perhaps we should have a common base class Network which provides all the necessary interfaces.

alf/networks/actor_distribution_networks.py
```diff
@@ -0,0 +1,165 @@
...
+
+        self._lstm_cell = torch.nn.LSTMCell(
+            input_size=self._encoding_net.output_size,
+            hidden_size=lstm_hidden_size)
+        self._state_spec = _create_lstm_cell_state_spec(
+            lstm_hidden_size, input_tensor_spec.dtype)
+
+        input_tensor_spec = TensorSpec((lstm_hidden_size, ),
+                                       input_tensor_spec.dtype)
+        self._after_lstm_encoding_net = EncodingNetwork(
```
emailweixu: Perhaps we can have a LSTMEncodingNetwork to do all these work.

alf/networks/value_networks.py
hnyu: layers._create_lstm_cell_state_spec is what I used in this file

le-horizon: Ah, you are right.

hnyu:
Resolve conflict in build.sh

PR #382, opened 2020-02-25
pytorch metrics

le-horizon:
Hey All, this is my first pytorch PR.  Feedback welcome.

Btw, unit test passes.

Thanks,
Le


alf/data_structures.py
```diff
@@ -110,6 +111,39 @@ def is_last(self):
         raise ValueError('step_type is not a Torch Tensor')
 
 
+def _create_timestep(observation, prev_action, reward, discount, env_id,
+                     step_type):
+    discount = torch.tensor(discount)
+    # Warning: To copy construct from a tensor, it is recommended to use sourceTensor.clone().detach() or sourceTensor.clone().detach().requires_grad_(True), rather than torch.tensor(sourceTensor).
```
hnyu: 85c per line max.

alf/data_structures.py
```diff
@@ -110,6 +111,39 @@ def is_last(self):
...
+    # Input struct might contain tensor.
+    # TODO: remove warning
+    create_tensor = lambda t: t.clone().detach() if torch.is_tensor(
```
hnyu: Since there is no gradient required for any of the fields, I guess we don't have to do clone. Also, all these inputs should already be tensors (they are returned by the env @runjerry ).

alf/metrics/metrics.py
```diff
@@ -0,0 +1,232 @@
+# Copyright (c) 2020 Horizon Robotics. All Rights Reserved.
```
hnyu: Why not just combining `metrics.py` and `metric.py`?

alf/metrics/metric.py
```diff
@@ -0,0 +1,86 @@
...
+    def gen_summaries(self, train_step=None, step_metrics=()):
+        """Generates summaries against train_step and all step_metrics.
+    Args:
+      train_step: (Optional) Step counter for training iterations. If None, no
+        metric is generated against the global step.
+      step_metrics: (Optional) Iterable of step metrics to generate summaries
+        against.
+    Returns:
+      A list of summaries.
+    """
```
hnyu: Docstring format needs to be fixed

alf/metrics/metric.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def __call__(self, *args, **kwargs):
+        return self.call(*args, **kwargs)
+
+    def call(self, *args, **kwargs):
+        """Accumulates statistics for the metric.
+    Args:
+      *args:
+      **kwargs: A mini-batch of inputs to the Metric.
+    """
```
hnyu: Docstring format needs to be fixed. It seems that locally you didn't run pre-commit hooks before pushing.

alf/metrics/metric.py
```diff
@@ -0,0 +1,86 @@
...
+Code adapted from https://github.com/tensorflow/agents/blob/master/tf_agents/metrics/tf_metric.py
+"""
+
+import alf
+import os
+
+from torch import nn
+
+
+class StepMetric(nn.Module):
```
hnyu: What's the reason for deriving from `nn.Module`? Is it because for saving checkpoints?

alf/metrics/metrics.py
```diff
@@ -0,0 +1,232 @@
...
+
+
+class TCDeque(torch.nn.Module):
+    """Deque backed by torch.tensor storage."""
+
+    def __init__(self, max_len, dtype):
+        super().__init__()
+        shape = (max_len, )
+        self._dtype = dtype
+        self.max_len = torch.tensor(max_len, dtype=torch.int32)
```
hnyu: Probably `int64` will be safer

alf/metrics/metrics.py
```diff
@@ -0,0 +1,232 @@
...
+"""A set of metrics.
+Converted to PyTorch from the TF version.
+https://github.com/tensorflow/agents/blob/master/tf_agents/metrics/tf_metrics.py
+"""
+
+from alf.metrics import metric
+import torch
+
+
+class TCDeque(torch.nn.Module):
```
hnyu: The official abbr for `Torch` is `th`. I guess it's better to follow that way.

alf/data_structures.py
le-horizon: Removed the line.  No longer the case.

alf/data_structures.py
le-horizon: just using detach().
tf_agents has this, so just copying over in case needed.

alf/metrics/metric.py
le-horizon: Yes, I think only nn.Module can be checkpointed automatically?

alf/metrics/metric.py
le-horizon: Fixed.  I upgraded tf but forgot to install hooks.  Done it now.

alf/metrics/metric.py
le-horizon: Done.

alf/metrics/metrics.py
le-horizon: just doing it the same way as tf_agents.  Plus file becomes too long if we merge.

alf/metrics/metrics.py
le-horizon: Done.

alf/metrics/metrics.py
le-horizon: Done.

alf/metrics/metric.py
Haichao-Zhang: Can add tests for all these sub-classes that the state_dict() give you the correct things you want to save in checkpoint.

alf/metrics/metrics.py
```diff
@@ -0,0 +1,231 @@
...
+class THDeque(torch.nn.Module):
+    """Deque backed by torch.tensor storage."""
+
+    def __init__(self, max_len, dtype):
+        super().__init__()
+        shape = (max_len, )
+        self._dtype = dtype
+        self.max_len = torch.tensor(max_len, dtype=torch.int64)
+        self.buffer = torch.zeros(shape, dtype=dtype)
+        self.head = torch.zeros((), dtype=torch.int64)
```
emailweixu: I think you need to call register_buffer to make these tensors appear in state_dict. Should have a test to make sure we understand the pytorch mechanism correctly.

emailweixu: prefix the names of internal members with `_`

alf/metrics/metrics.py
```diff
@@ -0,0 +1,231 @@
...
+"""
+
+from alf.metrics import metric
+import torch
+
+
+class THDeque(torch.nn.Module):
+    """Deque backed by torch.tensor storage."""
+
+    def __init__(self, max_len, dtype):
```
emailweixu: comment arguments

alf/metrics/metrics.py
```diff
@@ -0,0 +1,231 @@
...
+        self._dtype = dtype
+        self.max_len = torch.tensor(max_len, dtype=torch.int64)
+        self.buffer = torch.zeros(shape, dtype=dtype)
+        self.head = torch.zeros((), dtype=torch.int64)
+
+    @property
+    def data(self):
+        return self.buffer[:self.length]
+
+    def append(self, value):
```
emailweixu: comment

alf/metrics/metrics.py
```diff
@@ -0,0 +1,231 @@
...
+        self.add(value)
+
+    def add(self, value):
```
emailweixu: comment. This might be confused with `append`

alf/metrics/metrics.py
```diff
@@ -0,0 +1,231 @@
...
+"""A set of metrics.
+Converted to PyTorch from the TF version.
+https://github.com/tensorflow/agents/blob/master/tf_agents/metrics/tf_metrics.py
+"""
+
+from alf.metrics import metric
+import torch
+
+
+class THDeque(torch.nn.Module):
```
emailweixu: Why is it called THDeque?

alf/metrics/metrics.py
```diff
@@ -0,0 +1,231 @@
...
+        return torch.min(self.head, self.max_len)
+
+    def clear(self):
+        self.head.fill_(0)
+        self.buffer.fill_(0)
+
+    def mean(self):
+        if self.head == 0:
+            return torch.zeros((), dtype=self._dtype)
+        return torch.mean(self.buffer[:self.length])
```
emailweixu: I cannot find where self.length is defined/updated. And if it is indeed a bug, unittest is not functional.

emailweixu: I cannot find where self.length is defined/updated. And if it is indeed a bug, unittest is not functional.

alf/metrics/metrics.py
le-horizon: Added tests.
Somehow earlier when I tried register_buffer with tensor, if member variable name starts with "_", it needs to be registered, otherwise, registration is automatic.  But with nn.Parameter, registration is automatic either ways.

le-horizon: Done

alf/metrics/metrics.py
le-horizon: `length` is a function decorated with `@property`.  Tests did fail when code wasn't correct.  So this should be Ok.

le-horizon: Should be good.

alf/metrics/metrics.py
le-horizon: TH is shorthand for torch, according to Haonan.

alf/metrics/metric.py
le-horizon: Good point, done.

alf/metrics/metrics.py
le-horizon: removed append, and renamed this to append

alf/metrics/metrics.py
le-horizon: Done

alf/metrics/metrics.py
le-horizon: Done

alf/metrics/metrics.py
```diff
@@ -0,0 +1,251 @@
...
+        """Constructor.
+
+        Args:
+            max_len (int): maximum length of the deque
+            dtype (torch.dtype): dtype of the content of the deque
+        """
+        super().__init__()
+        shape = (max_len, )
+        self.dtype = dtype
+        self._max_len = torch.nn.Parameter(
```
emailweixu: It's better to use register_buffer(), otherwise this may be picked up by optimizers. See https://pytorch.org/docs/stable/nn.html#torch.nn.Module.register_buffer

alf/metrics/metrics.py
```diff
@@ -0,0 +1,251 @@
...
+            torch.where(time_step.is_first(),
+                        torch.zeros_like(self._return_accumulator),
+                        time_step.reward))
+
+        # Add final returns to buffer.
+        last_episode_indices = torch.squeeze(
+            *torch.where(time_step.is_last())).type(torch.int64)
+        for indx in last_episode_indices:
+            self._buffer.append(
+                self._return_accumulator[indx].clone().detach())
```
emailweixu: no need to clone and detach.

alf/metrics/metrics.py
```diff
@@ -0,0 +1,251 @@
...
+            torch.ones_like(self._length_accumulator),
+            torch.zeros_like(self._length_accumulator))
+        self._length_accumulator.add_(non_boundary)
+
+        # Add lengths to buffer when we hit end of episode
+        is_last = time_step.is_last()
+        last_indices = torch.squeeze(*torch.where(is_last)).type(torch.int64)
+        for indx in last_indices:
+            self._buffer.append(
+                self._length_accumulator[indx].clone().detach())
```
emailweixu: remove clone detach

alf/metrics/metrics.py
```diff
@@ -0,0 +1,251 @@
...
+
+        # Clear length accumulator at the end of episodes.
+        self._length_accumulator[:] = torch.where(
```
emailweixu: It's better to use self._length_accumulator.scatter_() to do the update. Similar to the tf_agent implementation

alf/metrics/metrics.py
le-horizon: Done

alf/metrics/metrics.py
le-horizon: Done

alf/metrics/metrics.py
le-horizon: Done

alf/metrics/metrics.py
```diff
@@ -56,7 +56,7 @@ def append(self, value):
             Nothing
         """
         position = torch.remainder(self._head, self._max_len)
-        self._buffer[position] = value
+        self._buffer.scatter_(0, position, value)
```
emailweixu: "self._buffer[position] = value" is more readable

alf/metrics/metrics.py
```diff
@@ -223,23 +226,20 @@ def call(self, time_step):
...
         for indx in last_indices:
-            self._buffer.append(
-                self._length_accumulator[indx].clone().detach())
+            self._buffer.append(self._length_accumulator[indx])
 
         # Clear length accumulator at the end of episodes.
-        self._length_accumulator[:] = torch.where(
-            is_last, torch.zeros_like(self._length_accumulator),
-            self._length_accumulator)
+        self._length_accumulator.scatter_(
```
emailweixu: it seems that we can use the following:
```python
self._length_accumulator[last_indices] = torch.zeros_like(last_indices)
```

alf/metrics/metrics.py
```diff
@@ -34,12 +34,12 @@ def __init__(self, max_len, dtype):
...
         shape = (max_len, )
         self.dtype = dtype
-        self._max_len = torch.nn.Parameter(
-            torch.tensor(max_len, dtype=torch.int64), requires_grad=False)
-        self._buffer = torch.nn.Parameter(
-            torch.zeros(shape, dtype=dtype), requires_grad=False)
-        self._head = torch.nn.Parameter(
-            torch.zeros((), dtype=torch.int64), requires_grad=False)
+        self._max_len = torch.tensor(max_len, dtype=torch.int64)
+        self.register_buffer(tensor=self._max_len, name='max_len')
```
emailweixu:
```python
self.register_buffer('_max_len', torch.tensor(max_len, dtype=torch.int64))
```
otherwise self.max_len is still valid because register_buffer will add 'max_len' as attribute of self

alf/metrics/metrics.py
le-horizon: Done

alf/metrics/metrics.py
le-horizon: Done.  Interesting.

alf/metrics/metrics.py
le-horizon: Ah, only briefly read the API..

alf/metrics/metrics.py
le-horizon: Done

alf/metrics/metrics.py
le-horizon: Interesting.  Done.

le-horizon:
Any more comments @emailweixu and all?

alf/data_structures.py
```diff
@@ -110,6 +111,37 @@ def is_last(self):
         raise ValueError('step_type is not a Torch Tensor')
 
 
+def _create_timestep(observation, prev_action, reward, discount, env_id,
+                     step_type):
+    discount = torch.tensor(discount)
+    # as_tensor reuses the underlying data store of numpy array if possible.
+    create_tensor = lambda t: t.detach() if torch.is_tensor(
+        t) else torch.as_tensor(t).detach()
```
hnyu: detach() is not necessary for torch.as_tensor(t) because numpy doesn't support gradients

alf/data_structures.py
```diff
@@ -110,6 +111,37 @@ def is_last(self):
...
+    make_tensors = lambda struct: map_structure(create_tensor, struct)
+    return TimeStep(
+        step_type=step_type.view(discount.shape),
```
hnyu: Why do you want to use `view` here? Just `torch.as_tensor(step_type)` is OK

alf/data_structures.py
```diff
@@ -110,6 +111,37 @@ def is_last(self):
         raise ValueError('step_type is not a Torch Tensor')
 
 
+def _create_timestep(observation, prev_action, reward, discount, env_id,
+                     step_type):
+    discount = torch.tensor(discount)
```
hnyu: `tensor` -> `as_tensor`

alf/data_structures.py
```diff
@@ -110,6 +111,37 @@ def is_last(self):
...
+    create_tensor = lambda t: t.detach() if torch.is_tensor(
+        t) else torch.as_tensor(t).detach()
+    make_tensors = lambda struct: map_structure(create_tensor, struct)
+    return TimeStep(
+        step_type=step_type.view(discount.shape),
+        reward=make_tensors(reward),
+        discount=discount,
+        observation=make_tensors(observation),
+        prev_action=make_tensors(prev_action),
+        env_id=torch.tensor(env_id, dtype=torch.int64))
```
hnyu: `tensor` -> `as_tensor`

alf/metrics/metrics.py
```diff
@@ -0,0 +1,245 @@
...
+
+    def __init__(self,
+                 name='AverageEpisodeLength',
+                 prefix='Metrics',
+                 dtype=torch.float32,
+                 batch_size=1,
+                 buffer_size=10):
+        super(AverageEpisodeLengthMetric, self).__init__(
+            name=name, prefix=prefix)
+        # TODO: use tensor deque
```
hnyu: Remove TODO

alf/data_structures.py
```diff
@@ -110,6 +111,37 @@ def is_last(self):
         raise ValueError('step_type is not a Torch Tensor')
 
 
+def _create_timestep(observation, prev_action, reward, discount, env_id,
```
hnyu: If this function is only used for testing, then I guess it shouldn't belong to data_structures.py. Otherwise it should be moved to @runjerry 's implementation.

alf/metrics/metrics.py
```diff
@@ -0,0 +1,245 @@
...
+Converted to PyTorch from the TF version.
+https://github.com/tensorflow/agents/blob/master/tf_agents/metrics/tf_metrics.py
+"""
+
+from alf.metrics import metric
+import torch
+
+
+class THDeque(torch.nn.Module):
+    """Torch (TH) deque backed by torch.nn.Parameter storage."""
```
Haichao-Zhang: This comment is not accurate as you are not using ```torch.nn.Parameter storage```

emailweixu:
Please also address other comments.

alf/metrics/metrics.py
le-horizon: Good catch!

alf/metrics/metrics.py
le-horizon: Done.

alf/data_structures.py
le-horizon: Jerry is just using constructor for now.  Probably can be merged.  I'll leave it up to @runjerry?

alf/data_structures.py
le-horizon: Done.  Btw, discount is a scalar

alf/data_structures.py
le-horizon: Good point, done!

alf/data_structures.py
le-horizon: This scalar tensor needs to be reshaped, otherwise it cannot be concat with other similar shaped tensors.

alf/data_structures.py
le-horizon: Done

alf/data_structures.py
runjerry: I am not using it right now, but will leave it after the re-implementation.

PR #385, opened 2020-02-26
PR_value_ops

Haichao-Zhang:
Refactor some value-related operators to make it easier to understand

### DIFF
--- a/alf/utils/value_ops.py
+++ b/alf/utils/value_ops.py
@@ -144,33 +144,26 @@ def discounted_return(rewards, values, step_types, discounts, time_major=True):
         values = values.transpose(0, 1)
         step_types = step_types.transpose(0, 1)
 
-    assert values.shape[0] >= 2, "The sequence length needs to be \
-                                at least 2. Got {s}".format(s=values.shape[0])
+    assert values.shape[0] >= 2, ("The sequence length needs to be "
+                                  "at least 2. Got {s}".format(
+                                      s=values.shape[0]))
 
-    discounts = discounts[1:]
-    rewards = rewards[1:]
-    final_value = values[-1]
-    values = values[:-1]
-
-    step_types = step_types[:-1]
     is_lasts = (step_types == StepType.LAST).to(dtype=torch.float32)
 
-    def discounted_return_fn(acc_discounted_reward, reward, value, is_last,
-                             discount):
-        acc_discounted_value = acc_discounted_reward * discount + reward
-        return is_last * value + (1 - is_last) * acc_discounted_value
+    rets = torch.zeros(rewards.shape, dtype=rewards.dtype)
+    rets[-1] = values[-1]
 
-    returns = torch.zeros(rewards.shape, dtype=rewards.dtype)
-    ret = final_value
-    for t in reversed(range(rewards.shape[0])):
-        ret = discounted_return_fn(ret, rewards[t], values[t], is_lasts[t],
-                                   discounts[t])
-        returns[t] = ret
+    with torch.no_grad():
+        for t in reversed(range(rewards.shape[0] - 1)):
+            acc_value = rets[t + 1] * discounts[t + 1] + rewards[t + 1]
+            rets[t] = is_lasts[t] * values[t] + (1 - is_lasts[t]) * acc_value
+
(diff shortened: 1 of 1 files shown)


alf/utils/value_ops.py
```diff
@@ -187,18 +180,14 @@ def one_step_discounted_return(rewards, values, step_types, discounts):
...
-    rewards = rewards[1:]
-    values = values[1:]
-    step_types = step_types[:-1]
+    assert values.shape[0] >= 2, ("The sequence length needs to be "
+                                  "at least 2. Got {s}".format(
+                                      s=values.shape[0]))
 
     is_lasts = (step_types == StepType.LAST).to(dtype=torch.float32)
-    returns = rewards + (1 - is_lasts) * discounts * values
+    rets = rewards[1:] + (1 - is_lasts[:-1]) * discounts[1:] * values[1:]
```
emailweixu: It seems that `rets = (rewards + discounts * values)[1:]` should be fine with the understanding the the loss at StepType.LAST will be ignored. And the rets from LINE 188 at StepType.LAST is not correct either.

alf/utils/value_ops.py
```diff
@@ -242,32 +231,24 @@ def generalized_advantage_estimation(rewards,
...
 
-    advantages = torch.zeros(rewards.shape, dtype=rewards.dtype)
-    adv = 0
-    for t in reversed(range(rewards.shape[0])):
-        adv = weighted_cumulative_td_fn(adv, weighted_discounts[t], delta[t],
-                                        is_lasts[t])
-        advantages[t] = adv
+    with torch.no_grad():
+        for t in reversed(range(rewards.shape[0] - 1)):
+            delta = rewards[t + 1] + \
```
emailweixu: Do the following before the loop to utilize vector computation as more as possible.
```
delta = reward[1:] + discounts[1:] * values[1:] - values[:-1] 
```

Haichao-Zhang: Changed to vectorized computation for delta.

alf/utils/value_ops.py
Haichao-Zhang: Updated for better understanding

alf/utils/value_ops.py
```diff
@@ -144,39 +144,34 @@ def discounted_return(rewards, values, step_types, discounts, time_major=True):
...
     is_lasts = (step_types == StepType.LAST).to(dtype=torch.float32)
 
-    def discounted_return_fn(acc_discounted_reward, reward, value, is_last,
-                             discount):
-        acc_discounted_value = acc_discounted_reward * discount + reward
-        return is_last * value + (1 - is_last) * acc_discounted_value
+    rets = torch.zeros(rewards.shape, dtype=rewards.dtype)
+    rets[-1] = values[-1]
+
+    with torch.no_grad():
```
hnyu: Can remove `with torch.no_grad()` because you will detach anyway

alf/utils/value_ops.py
```diff
@@ -187,18 +182,17 @@ def one_step_discounted_return(rewards, values, step_types, discounts):
...
+                                  "at least 2. Got {s}".format(
+                                      s=values.shape[0]))
 
     is_lasts = (step_types == StepType.LAST).to(dtype=torch.float32)
-    returns = rewards + (1 - is_lasts) * discounts * values
 
-    return returns
+    # the following computation can be further simplified but is kept this way
+    # for better understanding
+    rets = (1 - is_lasts[:-1]) * rewards[1:] + discounts[1:] * values[1:] + \
```
hnyu: I think we can directly do `rets=rewards[1:] + discounts[1:] * values[1:]` and remove `is_lasts`

alf/utils/value_ops.py
Haichao-Zhang: ```with torch.no_grad()``` will save memory

alf/utils/value_ops.py
Haichao-Zhang: Added documentation on this for better understanding: 
``` return = next_reward + next_discount * next_value if is not the last step
    otherwise will set return = next_discount * next_value.```

PR #407, opened 2020-03-04
PPOAlgorithm

emailweixu:
Issue #407

Also fixed several bugs:
1. actor_critic_loss: view() function sometime fails in _normalize_advantages()
2. replay_buffer: need to explicit convert device
3. categorical_projection_network: make sure the event_shape is correct

### DIFF
--- a/alf/algorithms/ppo_algorithm_test.py
+++ b/alf/algorithms/ppo_algorithm_test.py
@@ -99,14 +99,12 @@ class PpoTest(alf.test.TestCase):
             logging.info("%d reward=%f", i,
                          float(eval_time_step.reward.mean()))
 
-        eval_env.reset()
-        eval_time_step = unroll(eval_env, algorithm, steps_per_episode - 1)
-        logging.info("%d reward=%f", i, float(eval_time_step.reward.mean()))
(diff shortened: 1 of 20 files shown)


alf/algorithms/ppo_algorithm_test.py
```diff
@@ -39,84 +36,90 @@ def create_algorithm(env, use_rnn=False, learning_rate=1e-1):
...
             logging.info("%d reward=%f", i,
-                         float(tf.reduce_mean(eval_time_step.reward)))
+                         float(eval_time_step.reward.mean()))
 
         eval_env.reset()
-        eval_time_step, _ = eval_driver.run(
-            max_num_steps=(steps_per_episode - 1) * batch_size)
-        logging.info("reward=%f", float(tf.reduce_mean(eval_time_step.reward)))
+        eval_time_step = unroll(eval_env, algorithm, steps_per_episode - 1)
+        logging.info("%d reward=%f", i, float(eval_time_step.reward.mean()))
```
Haichao-Zhang: Do we need to keep Line 102-104, as they are essentially the same as Line 97-101 for the last iteration?

Issue #412, opened 2020-03-04
TTP:  Support for  multi observations input


witwolf:
EncoidingNetwork , Actor*Network Critic*Network shoud support for multi observations input

similar to `observation_preprocessing_layers` `observation_preprocessing_combiner` in .xxNetwork in `tfagent `

witwolf:
solved by #417

Issue #410, opened 2020-03-04
TTP: normalizer and averager


hnyu:
Need by many algorithms like RND, ICM, DIAYN that involve intrinsic rewards.

PR #417, opened 2020-03-05
DataBuffer, Averager, Normalizer, DIAYN, RND, ICM, NestCombiner, and InputPreprocessor

hnyu:
#416, #410, #400, #401, #402, 

Didn't know that @witwolf  was also working on supporting nested inputs #412 ... Maybe can merge the two PRs.

Also make all networks derive from the base class `Network`. 

For now the algorithms don't come with trained results. I will add after Agent is implemented.


alf/algorithms/diayn_algorithm.py
```diff
@@ -108,37 +119,40 @@ def train_step(self,
         observation, skill = observations_aug
         prev_skill = state
 
+        # normalize observation for easier prediction
+        if self._observation_normalizer is not None:
+            observation = self._observation_normalizer.normalize(observation)
+
         if self._encoding_net is not None:
-            feature, _ = self._encoding_net(observation)
+            feature = self._encoding_net(observation)
```
emailweixu: Let's make the interface of all kinds of Network same: always have state_spec property, and forward() always have state argument, and return (output, state) tuple. The reason is to make switching between RNN and non-RNN network easier.

alf/algorithms/diayn_algorithm.py
```diff
@@ -108,37 +119,40 @@ def train_step(self,
...
+        return AlgStep(
+            output=(),
             state=skill,
             info=DIAYNInfo(
                 reward=intrinsic_reward,
                 loss=LossInfo(
-                    loss=skill_discriminate_loss,
-                    extra=dict(
-                        skill_discriminate_loss=skill_discriminate_loss))))
+                    loss=loss, extra=dict(skill_discriminate_loss=loss))))
```
emailweixu: can be optimized a little bit:
```python
return AlgStep(
    output=(),
    state=skill,
    info=DIAYNInfo(
        reward=intrinsic_reward,
        loss=loss))
```
And multiplying with `valid_masks` and make the LossInfo as `calc_loss()`.
This can reduce the number of operations because of batching. And it also reduces memory of the `extra` field.

alf/networks/utils.py
```diff
@@ -0,0 +1,56 @@
...
+            raise ValueError('Batch dims must be non-negative.')
+        self._batch_dims = batch_dims
+        self._original_tensor_shape = None
+
+    def flatten(self, tensor):
+        """Flattens and caches the tensor's batch_dims."""
+        if self._batch_dims == 1:
+            return tensor
+        self._original_tensor_shape = tensor.shape
+        return tensor.view(-1, *tensor.shape[self._batch_dims:])
```
emailweixu: It may not always possible to use `view()`.  `reshape()` is safer.  Whenever possible reshape will use view, otherwise it will copy. see https://pytorch.org/docs/stable/tensors.html#torch.Tensor.reshape

alf/utils/averager.py
```diff
@@ -14,33 +14,36 @@
...
+        spec (TensorSpec):
 
     Returns:
         the average tensor across outer dims
     """
     outer_dims = get_outer_rank(tensor, spec)
     batch_squash = BatchSquash(outer_dims)
     tensor = batch_squash.flatten(tensor)
-    return tf.reduce_mean(tensor, axis=0)
+    return torch.mean(tensor, dim=0)
```
emailweixu: tensor.mean(dim=list(range(outer_dims)))

alf/utils/data_buffer.py
```diff
@@ -81,21 +72,18 @@ def add_batch(self, batch):
...
+        batch_size = alf.nest.get_nest_batch_size(batch)
+        with alf.device(self._device):
+            n = torch.min(self._capacity, torch.as_tensor(batch_size))
+            indices = torch.arange(self._current_pos,
+                                   self._current_pos + n) % self._capacity
+            alf.nest.map_structure(
+                lambda buf, bat: buf.__setitem__(indices, bat[-n:].detach()),
+                self._buffer, batch)
+
+            self._current_pos = (self._current_pos + n) % self._capacity
```
emailweixu: use `self._current_pos.copy_()` so that the object `self._current_pos` refers to does not change,  which can reduce the overhead of book keeping.

alf/layers.py
```diff
@@ -29,6 +31,40 @@ def identity(x):
     return x
 
 
+def square(x):
```
emailweixu: seems to belong to math_ops.py

alf/algorithms/diayn_algorithm.py
```diff
@@ -11,18 +11,19 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-from collections import namedtuple
 
-import gin.tf
-import tensorflow as tf
+import gin
+import functools
```
emailweixu: not used

alf/algorithms/diayn_algorithm.py
```diff
@@ -11,18 +11,19 @@
...
 
-from tf_agents.networks.network import Network
-import tf_agents.specs.tensor_spec as tensor_spec
+import torch
 
-from alf.algorithms.algorithm import Algorithm, AlgorithmStep, LossInfo
+import alf
+from alf.tensor_specs import TensorSpec, BoundedTensorSpec
```
emailweixu: sort imports alphabetically

alf/algorithms/diayn_algorithm.py
hnyu: I cannot get StepType in calc_loss() for computing valid_masks?

alf/layers.py
```diff
@@ -29,6 +31,40 @@ def identity(x):
...
+    """torch doesn't have square."""
+    return torch.pow(x, 2)
+
+
+def batch_normalization(x, mean, variance, variance_epsilon):
```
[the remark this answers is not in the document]
hnyu: It's indeed literally "batch normalization"

alf/algorithms/diayn_algorithm.py
```diff
@@ -37,66 +38,76 @@ class DIAYNAlgorithm(Algorithm):
...
-                 feature_spec,
-                 hidden_size=256,
+                 skill_spec,
+                 encoding_net: EncodingNetwork,
                  reward_adapt_speed=8.0,
-                 encoding_net: Network = None,
-                 discriminator_net: Network = None,
+                 observation_spec=None,
+                 hidden_size=(),
+                 hidden_activation=torch.relu,
```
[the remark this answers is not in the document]
hnyu: The encoding network can handle images. To keep the API concise, I think there is no need for discriminator_net

alf/layers.py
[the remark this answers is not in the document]
Haichao-Zhang: I understand this function implements a specific way to normalize a batch of data, but the name ```batch_normalization``` is typically refers to the specific [operation](https://en.wikipedia.org/wiki/Batch_normalization), especially under the context of DNN. Since there are several difference with the standard version, it's better to use a different name, maybe ```batch_ standardization```.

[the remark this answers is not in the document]
hnyu: > I understand this function implements a specific way to normalize a batch of data, but the name `batch_normalization` typically refers to the specific [operation](https://en.wikipedia.org/wiki/Batch_normalization), especially under the context of DNN. Since there are several difference with the standard version, it's better to use a different name, maybe `batch_ standardization`.

That's what tf2 calls. I can change it to `normalize_along_batch_dims`.

alf/networks/network.py
```diff
@@ -133,12 +226,49 @@ def copy(self, **kwargs):
...
         return super(Network, self).__call__(inputs, *args, **kwargs)
 
+    @property
+    def nested_input_tensor_spec(self):
+        """Return the input tensor spec BEFORE preprocessings have been applied.
+        """
+        return self._nested_input_tensor_spec
+
+    @property
+    def input_tensor_spec(self):
```
emailweixu: I suggest this to be the original input tensor spec. i.e. (self._nested_input_tensor_spec) to reduce confusion. And it seems there is no need to expose the tensor spec after preprocessing.

alf/networks/network.py
```diff
@@ -133,12 +226,49 @@ def copy(self, **kwargs):
...
+                 skip_input_preprocessing=False,
+                 name="DistributionNetwork"):
+        super(DistributionNetwork, self).__init__(
+            input_tensor_spec, input_preprocessors, preprocessing_combiner,
+            skip_input_preprocessing, name)
+
+    @property
+    def output_spec(self):
+        if self._output_spec is None:
+            self._output_spec = DistributionSpec.from_distribution(
```
emailweixu: use `dist_utils.extract_spec()` instead in case that the output is nested.

alf/networks/preprocessors.py
```diff
@@ -0,0 +1,130 @@
...
+        return ret
+
+
+@gin.configurable
+class EmbeddingPreprocessor(InputPreprocessor):
+    """A preprocessor that converts the input to an embedding vector. This can
+    be used when the input is a discrete scalar, or a continuous vector to be
+    projected to a different dimension (to have the same length with other
+    vectors). Different from an `EncodingNetwork`, the input can be in the
+    original format from the environment.
```
emailweixu: Need to comment that this is not suitable for word embedding or discrete input with large number of discrete values. Or use [nn.Embedding](https://pytorch.org/docs/stable/nn.html#torch.nn.Embedding) to embed discrete input

alf/networks/network.py
```diff
@@ -84,35 +91,121 @@ def _capture_init(self, *args, **kwargs):
...
+                for different inputs by configuring a gin file without changing
+                the code. For example, embedding a discrete input before concatenating
+                it to another continuous vector.
+            preprocessing_combiner (NestCombiner): preprocessing called on
+                complex inputs. Note that this combiner must also accept
+                `input_tensor_spec` as the input to compute the processed
+                tensor spec. For example, see `alf.nest.utils.NestConcat`. This
+                arg is helpful if you want to combine inputs by configuring a
+                gin file without changing the code.
+            skip_input_preprocessing (bool): If True, skip preprocessing and
```
emailweixu: Can't this be inferred from `input_preprocessors`? if input_preprocessors is None, then skip.

alf/utils/tensor_utils.py
```diff
@@ -57,7 +57,7 @@ def explained_variance(ypred, y):
     Returns:
         1 - Var[y-ypred] / Var[y]
     """
-    ypred = ypred.view(-1)
-    y = y.view(-1)
+    ypred = torch.reshape(ypred, (-1, ))
+    y = torch.reshape(y, (-1, ))
```
emailweixu: line 60-61 can be removed if the following two lines are changed:
```python
    vary = torch.var(y, unbiased=False)
    return 1 - torch.var(y - ypred, unbiased=False) / (vary + 1e-30)
```

alf/data_structures.py
```diff
@@ -118,7 +118,7 @@ def _create_timestep(observation, prev_action, reward, discount, env_id,
     create_tensor = lambda t: torch.as_tensor(t).detach()
     make_tensors = lambda struct: map_structure(create_tensor, struct)
     return TimeStep(
-        step_type=step_type.view(discount.shape),
+        step_type=torch.reshape(step_type, discount.shape),
```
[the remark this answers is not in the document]
hnyu: I believe @le-horizon had some reason for the original `view`. This function is only used in `metrics_test`

alf/nest/utils.py
```diff
@@ -13,54 +13,110 @@
...
+        Returns:
+            tensor (Tensor): the combined result
+        """
+        pass
+
+    def __call__(self, nested):
+        """Combine all elements according to the method defined in `combine_flat`.
+
+        Args:
+            nested (nest): a nested structure; each element can be either a
```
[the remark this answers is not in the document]
hnyu: In the base class `Network __init__`, the function will automatically calculate the tensor spec after combining the nested input. So later the downstream networks can use `self._input_tensor_spec` it computes.

alf/data_structures.py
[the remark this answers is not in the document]
le-horizon: I think it was because step_type is a scalar, and cannot be concatenated with other tensors in the TimeStep nest.  View is necessary to avoid complaints.

alf/networks/encoding_networks.py
```diff
@@ -79,8 +85,9 @@ def __init__(self,
                     padding=padding))
             input_channels = filters
 
-    def output_shape(self):
-        """Return the output shape given the input image size.
+    @property
+    def output_spec(self):
```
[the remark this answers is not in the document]
hnyu: I think for some networks like `CriticNetwork` and `ValueNetwork`, their output spec is quite obvious. So there is no need to involve the `_test_forward`

alf/networks/network.py
hnyu: But then `self.input_tensor_spec` and `self._input_tensor_spec` return two different things. And the latter is always needed for constructing downstream networks as their input specs.

emailweixu: Perhaps change to a different name so not to be confused? For example, processed_tensor_spec

alf/networks/network.py
hnyu: If `input_preprocessors` is None, the user can still use the combiner (e.g., concat)

alf/algorithms/diayn_algorithm.py
```diff
@@ -106,39 +117,40 @@ def train_step(self,
...
+        skill_pred, _ = self._discriminator_net(feature)
 
-        skill_discriminate_loss = tf.nn.softmax_cross_entropy_with_logits(
-            labels=prev_skill, logits=skill_pred)
+        if self._skill_spec.is_discrete:
+            loss = torch.nn.CrossEntropyLoss(reduction='none')(
+                input=skill_pred, target=prev_skill.to(torch.int64))
+        else:
+            # nn.MSELoss doesn't support reducing along a dim
+            loss = torch.sum(math_ops.square(skill_pred - prev_skill), dim=-1)
```
Haichao-Zhang: Might be better to use ```mean``` for reducing. ```sum``` is more easily impacted by the dims of the skill vectors.

alf/networks/actor_distribution_networks.py
```diff
@@ -149,11 +199,11 @@ def forward(self, observation, state):
             act_dist (torch.distributions): action distribution
             new_state (nest[tuple]): the updated states
         """
-        encoding = self._encoding_net(observation)
-        encoding, state = self._lstm_encoding_net(encoding, state)
-        act_dist = self._projection_net(encoding)
+        observation, state = Network.forward(self, observation, state)
```
Haichao-Zhang: Is it possible to wrap this call (```Network.forward```) into a function like ```input_processing(...)``` in base ```Network``` class?  Then this line would be like ```observation, state = self.input_processing(self, observation, state)```, which is more natural and clear in terms what it does. 
The user can also override this ```input_processing``` function in their own class if customization is desirable.

alf/algorithms/diayn_algorithm.py
hnyu: It's nothing wrong with sum because it's essentially the log likelihood of a normal distribution with a fixed std.

alf/networks/actor_distribution_networks.py
hnyu: I put it in `Network.forward` because that can remind the user of calling it to some extent. Basically it's always beneficial to first call this before you do other things.

PR #419, opened 2020-03-05
Reset global counter

Haichao-Zhang:
1. Reset the global counter when a new training algorithm is created to reduce side effects
2. If recovering from checkpoint, set the global counter properly


emailweixu:
need to merge

PR #421, opened 2020-03-05
Network base class

emailweixu:
Need this to support copy() function.

### DIFF
--- a/alf/networks/network.py
+++ b/alf/networks/network.py
@@ -1,0 +1,150 @@
+# Copyright (c) 2020 Horizon Robotics. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Base extension to torch.nn.Module.
+   Adapted from tf_agents/tf_agents/networks/network.py
+"""
+
+import abc
+import functools
+import inspect
+import six
+import torch.nn as nn
+
+import alf
+
+
+class _NetworkMeta(abc.ABCMeta):
+    """Meta class for Network object.
+
+    We mainly use this class to capture all args to `__init__` of all `Network`
+    instances, and store them in `instance._saved_kwargs`.  This in turn is
+    used by the `instance.copy` method.
+
+    """
+
+    def __new__(mcs, classname, baseclasses, attrs):
+        """Control the creation of subclasses of the Network class.
+
+        Args:
+            classname: The name of the subclass being created.
+            baseclasses: A tuple of parent classes.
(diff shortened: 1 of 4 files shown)


alf/networks/network_test.py
```diff
@@ -0,0 +1,79 @@
+# coding=utf-8
+# Copyright 2018 The TF-Agents Authors.
```
Haichao-Zhang: This header needs to be updated?

alf/networks/network_test.py
```diff
@@ -0,0 +1,79 @@
...
+        self.assertRaises(TypeError, MockNetwork, 0, 1, 2, 3, 4, 5, 6)
+
+    def test_assert_input_spec(self):
+        spec = alf.TensorSpec([], torch.int32)
+        net = MockNetwork(spec, 1)
+        self.assertRaises(AssertionError, net, (1, 2), 2)
+
+
+if __name__ == '__main__':
+    NetworkTest().test_copy_works()
```
Haichao-Zhang: This line can be removed

alf/networks/network.py
```diff
@@ -0,0 +1,150 @@
...
+        Args:
+        **kwargs: Args to override when recreating this network.  Commonly
+            overridden args include 'name'.
+
+        Returns:
+        A shallow copy of this network.
+        """
+        return type(self)(**dict(self._saved_kwargs, **kwargs))
+
+    def __call__(self, inputs, *args, **kwargs):
```
hnyu: should be `forward`

alf/networks/network.py
```diff
@@ -0,0 +1,150 @@
...
+        alf.nest.assert_same_structure(inputs, self.input_tensor_spec)
```
hnyu: `inputs` has additional batch dims. We can remove this line for now. My local version has this check.

alf/networks/network.py
```diff
@@ -0,0 +1,150 @@
...
+        return super(Network, self).__call__(inputs, *args, **kwargs)
+
+
+class DistributionNetwork(Network):
+    """Base class for networks which generate Distributions as their output."""
+
+    def __init__(self, input_tensor_spec, state_spec, output_spec, name):
```
hnyu: Why do we need output_spec?

alf/networks/network.py
```diff
@@ -0,0 +1,150 @@
...
+
+        attrs["__init__"] = functools.update_wrapper(_capture_init, init)
+        return abc.ABCMeta.__new__(mcs, classname, baseclasses, attrs)
+
+
+@six.add_metaclass(_NetworkMeta)
+class Network(nn.Module):
+    """Base extension to nn.Module to simplify copy operations."""
+
+    def __init__(self, input_tensor_spec, state_spec, name):
```
hnyu: `state_spec` doesn't have to be always provided. Often we can infer from the network.

emailweixu: We need it to have state_spec property.

alf/networks/network.py
emailweixu: `__call__` is the right way. It will call base calss `__call__`, which will call `forward()` in turn.

alf/networks/network.py
emailweixu: You are right. Perhaps we don't need it.

alf/networks/network_test.py
Haichao-Zhang: This line is not removed yet:
```NetworkTest().test_copy_works()```

PR #422, opened 2020-03-06
support mutil observation for networks

witwolf:
add `preprocessing_layers`  and `preprocessing_combiner` to `EncodingNetwork` to handle muti observation input

And there is one thing i am not sure ,  should we make all `networks` inherit from  `Network
`   and copy fresh new instances for these layers   #421  . The benefit is that we can prevent some possible errors such as parameter sharing ...

TODO
make it flexible configure by `gin`, and i have an idea now , i'll try to add some functions to extend gin utils

### DIFF
--- a/alf/networks/encoding_networks.py
+++ b/alf/networks/encoding_networks.py
@@ -332,7 +332,6 @@ class MLPNetwork(nn.Module):
 
 @gin.configurable
 class EncodingNetwork(nn.Module):
-
     def __init__(self,
                  input_tensor_spec,
                  preprocessing_layers=None,
@@ -374,17 +373,16 @@ class EncodingNetwork(nn.Module):
                 lambda x: x or layers.identity, preprocessing_layers)
             output_tensor_spec = nest.map_structure(
                 lambda layer, input_spec: TensorSpec.from_tensor(
-                    tensor=layer(input_spec.zeros(outer_dims=(1,))),
-                    from_dim=1),
-                preprocessing_layers,
-                input_tensor_spec)
+                    tensor=layer(input_spec.zeros(outer_dims=(1, ))),
+                    from_dim=1), preprocessing_layers, input_tensor_spec)
 
             # It's a trick here, add all nn.Module instances in preprocessing_layers
             #   to variable  `self._preprocessing_module` so that `self.parameters()`
             #   can capture corresponding parameters
             self._preprocessing_module = nn.ModuleList(
-                list(filter(lambda x: isinstance(x, nn.Module),
-                       nest.flatten(preprocessing_layers))))
+                list(
+                    filter(lambda x: isinstance(x, nn.Module),
+                           nest.flatten(preprocessing_layers))))
             flat_input_tensor_spec = nest.flatten(output_tensor_spec)
             self._preprocessing_layers = preprocessing_layers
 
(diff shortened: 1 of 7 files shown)


hnyu:
Turns out that I am also working on this..

witwolf:
> Turns out that I am also working on this.. Should have created an issue in the project

唔。。#412 ,  I just  trying   .  Ignore this pr  if you have finish it or have a better implementation

hnyu:
> > Turns out that I am also working on this.. Should have created an issue in the project
> 
> 唔。。#412 , I just trying . Ignore this pr if you have finish it or have a better implementation

Sorry didn't see that ... Anyway I will make a push very soon and see whether the change is OK

hnyu:
perhaps you should use this to keep track of the progress:

https://github.com/HorizonRobotics/alf/projects/1

alf/networks/encoding_networks.py
```diff
@@ -248,7 +249,7 @@ def forward(self, inputs):
 
 
 @gin.configurable
-class EncodingNetwork(nn.Module):
+class MLPNetwork(nn.Module):
```
hnyu: Do you think it's necessary to have an MLP that just handles unnested input?

hnyu:
> 
> TODO
> make it flexible configure by `gin`, and i have an idea now , i'll try to add some functions to extend gin utils

Curious about what this idea is about?

alf/networks/encoding_networks.py
witwolf: `EncodingNetwork` can handle this casebut when we need preprocessing_layers to pre process multi inputs , we pass in (nested `EncodingNetwork`) without `preprocessing_layers` and `preprocessing_combiner` as `preprocessing_layers`

witwolf:
> Curious about what this idea is about?

Just add an interface
```
@gin.configurable
def gin_eval(source):
    """Evaluate the given source in the context of globals and locals.

    Args:
        source (str): string that will be evaluated
    """
    frame = _gin_call_context_frame()  # TODO
    return eval(source, frame.f_locals)
```

that cat meet the following tests
```python
    def test_refer_to_unregistered_func(self):
        """Test refer to unregister func by eval it with registered helper function
        `gin_eval`.

        builtin and 3rd-party functions are not registered in gin by default,
        we can not refer to it directly and it's also not unrealistic to register
        it explict by `gin.external_configurable(func)` for all functions we might
        refer to
        """

        @gin.configurable
        def _test(func):
            func()

        with self.assertRaises(ValueError):
            gin.parse_config([
                "_test/func=@list",
            ])

        gin.parse_config([
            "list/gin_eval.source='list'",
            "_test.func=@list/gin_eval()"
        ])
        _test()

    def _test_refer_to_local_values(self):
        """Test refer to local values
        Evaluate it in correct context frame
        """

        @gin.configurable
        def _add(a, b):
            return a + b

        def _test_add(a, b):
            return _add()

        gin.parse_config([
            "a/gin_eval.source='a'",
            "_add.a=@a/gin_eval()",
            "b/gin_eval.source='b'",
            "_add.b=@b/gin_eval()"
        ])

        self.assertEqual(1 + 2, _test_add(1, 2))
        self.assertEqual(2 + 3, _test_add(2, 3))

        @gin.configurable
        def _test_valid_prod(a, b, result):
            self.assertEqual(result, a * b)

        gin.parse_config([
            "_test_valid_prod/gin_eval.source='np.prod([a, b])'",
            "_test_valid_prod.result=@_test_valid_prod/gin_eval()"
        ])
        a, b = 1, 2
        _test_valid_prod(a, b)
        a, b = 3, 4
        _test_valid_prod(a, b)

```
not finished , have problem with getting correct context frame now

PR #433, opened 2020-03-07
Sac Algorithm

Haichao-Zhang:
+ Unit tested
+ Also tested with the sac_pendulum.gin with batch_env_temp, now all curves matches:
  1) return curve
  2) actor loss
  3) critic loss
  4) alpha loss
+ Further tested discrete action case with sac_cart_pole.gin, performs similar to TF version

### DIFF
--- a/alf/algorithms/sac_algorithm.py
+++ b/alf/algorithms/sac_algorithm.py
@@ -291,12 +291,12 @@ class SacAlgorithm(OffPolicyAlgorithm):
             target_critic_input, state=state.target_critic2)
 
         if not self._is_continuous:
-            exp_action = exp.action.to(torch.int32)
-            critic1 = critic1.gather(-1, exp_action.long())
-            critic2 = critic2.gather(-1, exp_action.long())
-            sampled_action = action.to(torch.int32)
-            target_critic1 = target_critic1.gather(-1, sampled_action.long())
-            target_critic2 = target_critic2.gather(-1, sampled_action.long())
+            exp_action = exp.action.view(critic1.shape[0], -1).long()
+            critic1 = critic1.gather(-1, exp_action)
+            critic2 = critic2.gather(-1, exp_action)
+            sampled_action = action.view(critic1.shape[0], -1).long()
+            target_critic1 = target_critic1.gather(-1, sampled_action)
+            target_critic2 = target_critic2.gather(-1, sampled_action)
 
         target_critic = torch.min(target_critic1, target_critic2).view(log_pi.shape) - \
                          (torch.exp(self._log_alpha) * log_pi).detach()
--- a/alf/algorithms/sac_algorithm.py
+++ b/alf/algorithms/sac_algorithm.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2019 Horizon Robotics. All Rights Reserved.
+# Copyright (c) 2020 Horizon Robotics. All Rights Reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -14,25 +14,22 @@
 """Soft Actor Critic Algorithm."""
 
 import numpy as np
-import tensorflow as tf
-import tensorflow_probability as tfp
-import gin.tf
+import gin
 
-from tf_agents.specs import tensor_spec
-from tf_agents.agents.ddpg.critic_network import CriticNetwork
-from tf_agents.agents.ddpg.critic_rnn_network import CriticRnnNetwork
-from tf_agents.networks.q_network import QNetwork
-from tf_agents.networks.q_rnn_network import QRnnNetwork
-from tf_agents.networks.actor_distribution_network import ActorDistributionNetwork
-from tf_agents.networks.actor_distribution_rnn_network import ActorDistributionRnnNetwork
-from tf_agents.networks.network import Network, DistributionNetwork
-from tf_agents.utils import common as tfa_common
+import torch
+import torch.nn as nn
+import torch.distributions as td
+from typing import Callable
 
+from alf.algorithms.config import TrainerConfig
 from alf.algorithms.off_policy_algorithm import OffPolicyAlgorithm
 from alf.algorithms.one_step_loss import OneStepTDLoss
 from alf.algorithms.rl_algorithm import RLAlgorithm
-from alf.data_structures import ActionTimeStep, Experience, LossInfo, namedtuple
-from alf.data_structures import PolicyStep, TrainingInfo
+from alf.data_structures import TimeStep, Experience, LossInfo, namedtuple
+from alf.data_structures import AlgStep, TrainingInfo
+from alf.nest import nest
+from alf.networks import ActorDistributionNetwork, CriticNetwork
+from alf.tensor_specs import TensorSpec, BoundedTensorSpec
 from alf.utils import losses, common, dist_utils
 
 SacShareState = namedtuple("SacShareState", ["actor"])
@@ -88,9 +85,11 @@ class SacAlgorithm(OffPolicyAlgorithm):
 
     def __init__(self,
                  observation_spec,
-                 action_spec,
-                 actor_network: DistributionNetwork,
-                 critic_network: Network,
+                 action_spec: BoundedTensorSpec,
+                 actor_network: ActorDistributionNetwork,
+                 critic_network: CriticNetwork,
+                 env=None,
+                 config: TrainerConfig = None,
                  critic_loss=None,
                  target_entropy=None,
                  initial_log_alpha=0.0,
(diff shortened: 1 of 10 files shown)


alf/algorithms/sac_algorithm_test.py
```diff
@@ -0,0 +1,161 @@
...
+        eval_time_step = unroll(eval_env, alg2, steps_per_episode - 1)
+        print(eval_time_step.reward.mean())
+
+        self.assertAlmostEqual(
+            1.0, float(eval_time_step.reward.mean()), delta=5e-1)
+
+
+if __name__ == '__main__':
+    SACAlgorithmTestDiscrete().test_sac_algorithm_discrete()
+    #unittest.main()
```
emailweixu: alf.test.main()

alf/algorithms/sac_algorithm.py
```diff
@@ -119,23 +126,18 @@ def __init__(self,
                 networks.
             target_update_period (int): Period for soft update of the target
                 networks.
-            dqda_clipping (float): when computing the actor loss, clips the
-                gradient dqda element-wise between [-dqda_clipping, dqda_clipping].
-                Does not perform clipping if dqda_clipping == 0.
-            actor_optimizer (tf.optimizers.Optimizer): The optimizer for actor.
-            critic_optimizer (tf.optimizers.Optimizer): The optimizer for critic.
-            alpha_optimizer (tf.optimizers.Optimizer): The optimizer for alpha.
+            actor_optimizer (torch.optim.optimizer): The optimizer for actor.
```
emailweixu: add back the comment for dqdq_clipping

alf/algorithms/sac_algorithm.py
```diff
@@ -119,23 +126,18 @@ def __init__(self,
...
-        critic_network2 = critic_network.copy(name='CriticNetwork2')
-        log_alpha = tfa_common.create_variable(
-            name='log_alpha',
-            initial_value=initial_log_alpha,
-            dtype=tf.float32,
-            trainable=True)
+        critic_network1 = critic_network.copy()
+        critic_network2 = critic_network.copy()
+
+        log_alpha = nn.Parameter(torch.Tensor([initial_log_alpha]))
```
emailweixu: `float(initial_log_alpha)` in case that an int value is provided

alf/utils/common.py
```diff
@@ -129,6 +131,38 @@ def as_list(x):
     return [x]
 
 
+class Periodically(nn.Module):
+    def __init__(self, body, period, name='periodically'):
+        """Periodically performs the ops defined in body.
+
+        Args:
+        body: callable to be performed every time
```
emailweixu: comment style. Indent and arg (type)

alf/utils/common.py
```diff
@@ -129,6 +131,38 @@ def as_list(x):
...
+            an internal counter is divisible by the period.
+        period (int): inverse frequency with which to perform the operation.
+        name: name of the object.
+
+        Raises:
+        TypeError: if body is not a callable.
+
+        Returns:
+        An op that periodically performs the specified body operation.
```
emailweixu: no Returns

alf/utils/common.py
```diff
@@ -854,18 +881,13 @@ def decorate(f):
 def set_random_seed(seed):
     """Set a seed for deterministic behaviors.
 
-    Note: If someone runs an experiment with a pre-selected manual seed, he can
-    definitely reproduce the results with the same seed; however, if he runs the
-    experiment with seed=None and re-run the experiments using the seed previously
-    returned from this function (e.g. the returned seed might be logged to
-    Tensorboard), and if cudnn is used in the code, then there is no guarantee
-    that the results will be reproduced with the recovered seed.
-
```
emailweixu: should not remove

alf/utils/common.py
```diff
@@ -854,18 +881,13 @@ def decorate(f):
...
     Args:
         seed (int|None): seed to be used. If None, a default seed based on
             pid and time will be used.
     Returns:
-        The seed being used if `seed` is None.
+        The seed being used if `seed` is None. Note that you may not reproduce
```
emailweixu: should not change

alf/utils/common.py
```diff
@@ -886,3 +908,53 @@ def log_metrics(metrics, prefix=''):
     """
     log = ['{0} = {1}'.format(m.name, m.result()) for m in metrics]
     logging.info('%s \n\t\t %s', prefix, '\n\t\t '.join(log))
+
+
+def assert_members_are_not_overridden(base_cls,
```
emailweixu: not used. remove

alf/algorithms/sac_algorithm_test.py
Haichao-Zhang: updated

alf/algorithms/sac_algorithm.py
Haichao-Zhang: Added

alf/algorithms/sac_algorithm.py
Haichao-Zhang: Added

alf/utils/common.py
Haichao-Zhang: Updated

alf/utils/common.py
Haichao-Zhang: Changed

alf/utils/common.py
Haichao-Zhang: Seems to be caused by a merge. Changed back.

alf/utils/common.py
Haichao-Zhang: Changed back.

alf/utils/common.py
Haichao-Zhang: There are a number of unintended changes in ```common.py``` due to a merge with PR_batch_env_temp for testing. They are changed back.

alf/utils/common.py
Haichao-Zhang: This is also removed.

alf/algorithms/sac_algorithm.py
```diff
@@ -361,9 +361,9 @@ def calc_loss(self, training_info: TrainingInfo):
         return LossInfo(
             loss=actor_loss.loss + critic_loss.loss + alpha_loss.loss,
             extra=SacLossInfo(
-                actor=actor_loss.extra.view(-1).mean(),
-                critic=critic_loss.extra.view(-1).mean(),
-                alpha=alpha_loss.extra.view(-1).mean()))
+                actor=actor_loss.extra,
+                critic=critic_loss.extra,
```
Haichao-Zhang: The previous mis-match on critic loss is because of this ```mean()```

Haichao-Zhang:
Now all curves matches, even though the initialization is different, suggesting it is not crucial for the performance.

alf/algorithms/sac_algorithm.py
```diff
@@ -224,63 +219,54 @@ def _actor_train_step(self, exp: Experience, state: SacActorState,
...
-                network_state=state.critic2)
+                exp.observation, state=state.critic2)
 
-            assert isinstance(action_distribution,
-                              tfp.distributions.Categorical), (
-                                  "Only `tfp.distributions.Categorical` " +
-                                  "was supported, received:" + str(
-                                      type(action_distribution)))
+            assert isinstance(
+                action_distribution, td.categorical.Categorical) or \
```
emailweixu: use this instead: https://github.com/HorizonRobotics/alf/blob/4723afdba6329459cc16f3204ae901117c86aa05/alf/utils/summary_utils.py#L283

alf/algorithms/sac_algorithm.py
```diff
@@ -224,63 +219,54 @@ def _actor_train_step(self, exp: Experience, state: SacActorState,
...
-            target_q_value = tf.stop_gradient(tf.minimum(critic1, critic2))
-            alpha = tf.stop_gradient(tf.exp(self._log_alpha))
-            actor_loss = tf.reduce_mean(
-                action_probs * (alpha * log_action_probs - target_q_value),
-                axis=-1)
+            target_q_value = torch.min(critic1, critic2).detach()
+            alpha = torch.exp(self._log_alpha)
+            actor_loss = torch.exp(log_action_probs) * (
+                alpha.detach() * log_action_probs - target_q_value)
+            actor_loss = actor_loss.view(actor_loss.size(0), -1).mean(1)
```
emailweixu: `view()` may not always work. Either avoid it or use reshape:
```python
actor_loss.mean(list(range(1, actor_loss.ndim)))
```

alf/algorithms/sac_algorithm.py
```diff
@@ -224,63 +219,54 @@ def _actor_train_step(self, exp: Experience, state: SacActorState,
...
+            critic1, critic1_state = self._critic_network1(
+                critic_input, state=state.critic1)
+            critic2, critic2_state = self._critic_network2(
+                critic_input, state=state.critic2)
 
-                target_q_value = tf.minimum(critic1, critic2)
+            target_q_value = torch.min(critic1, critic2)
 
-            dqda = tape.gradient(target_q_value, action)
+            dqda = torch.autograd.grad(target_q_value, action,
```
emailweixu: action might be nested.

alf/algorithms/sac_algorithm.py
Haichao-Zhang: Moved it to ```dist_utils.py``` as it is generally useful.

alf/algorithms/sac_algorithm.py
Haichao-Zhang: Changed according to suggestion.

alf/algorithms/sac_algorithm.py
Haichao-Zhang: Good point. Changed to properly handle nested action.

alf/algorithms/sac_algorithm.py
```diff
@@ -226,10 +226,10 @@ def _actor_train_step(self, exp: Experience, state: SacActorState,
 
             target_q_value = torch.min(critic1, critic2)
 
-            dqda = torch.autograd.grad(target_q_value, action,
-                                       torch.ones(target_q_value.size()))[0]
-
-            def actor_loss_fn(dqda, action):
+            def actor_loss_fn(action):
```
emailweixu: It's better to do grad only once if possible. The following might work:
```python
        dqda = alf.nest.pack_sequence_as(action,
            torch.autograd.grad(target_q_value, alf.nest.flatten(action)))
```

Haichao-Zhang: Now changed to one pass computation as suggested.

PR #435, opened 2020-03-07
Avoid side effects of encoding networks

Haichao-Zhang:
Encoding network modifies the input ```fc_layer_params``` (when it is a list) by appending the ```last_layer_size``` to it when specified.
This will have side effects whenever ```fc_layer_params``` is reused. For example, when network.copy() is called.

### DIFF
--- a/alf/networks/actor_distribution_networks.py
+++ b/alf/networks/actor_distribution_networks.py
@@ -40,7 +40,7 @@ class ActorDistributionNetwork(nn.Module):
         Args:
             input_tensor_spec (TensorSpec): the tensor spec of the input
             action_spec (TensorSpec): the action spec
-            conv_layer_params (list[tuple]): a list of tuples where each
+            conv_layer_params (tuple[tuple]): a tuple of tuples where each
                 tuple takes a format `(filters, kernel_size, strides, padding)`,
                 where `padding` is optional.
             fc_layer_params (tuple[int]): a tuple of integers representing hidden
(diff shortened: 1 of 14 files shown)


hnyu:
What about that we enforce all `xxx_params` to be tuples? Originally we supported list just for flexibility. But it seems that tuple is necessary to avoid side effect.

Haichao-Zhang:
Yes, we need to be careful when dealing with mutable inputs. This is ok currently and the last layer is the only place I found have this effect currently.
Perhaps we can do a refactor in the future after all parts are stabilized.

alf/networks/encoding_networks.py
```diff
@@ -298,17 +302,17 @@ def __init__(self,
             input_size = input_tensor_spec.shape[0]
 
         self._fc_layers = nn.ModuleList()
-        if fc_layer_params is None:
-            fc_layer_params = []
+        if _fc_layer_params is None:
+            _fc_layer_params = []
         else:
-            assert isinstance(fc_layer_params, (tuple, list))
```
hnyu: I mean just change this to `isinstance(fc_layer_params, tuple)` and problem solved.

Haichao-Zhang: Now enforcing all ```xxx_params``` to be tuples.

alf/networks/actor_distribution_networks.py
```diff
@@ -43,7 +43,7 @@ def __init__(self,
             conv_layer_params (list[tuple]): a list of tuples where each
```
hnyu: Should search for all `fc_layer_params` and `conv_layer_params` to change them to tuple

Haichao-Zhang: Done

alf/networks/critic_networks.py
```diff
@@ -133,20 +133,20 @@ def __init__(self,
...
                 hidden FC layer sizes for observations.
-            action_fc_layer_params (list[int]): a list of integers representing
+            action_fc_layer_params (tuple[int]): a tuple of integers representing
                 hidden FC layer sizes for actions.
-            joint_fc_layer_params (list[int]): a list of integers representing
+            joint_fc_layer_params (tuple[int]): a tuple of integers representing
                 hidden FC layer sizes FC layers after merging observations and
                 actions.
-            lstm_hidden_size (int or list[int] or tuple[int]): the hidden size(s)
+            lstm_hidden_size (int or tuple[int] or tuple[int]): the hidden size(s)
```
hnyu: remove one `tuple[int]`

alf/networks/encoding_networks.py
```diff
@@ -406,12 +406,12 @@ def forward(self, inputs, state):
         """
         Args:
             inputs (torch.Tensor):
-            state (list[tuple]): a list of tuples, where each tuple is a pair
+            state (tuple[tuple]): a tuple of tuples, where each tuple is a pair
```
hnyu: I believe that the state should still be a list

alf/networks/critic_networks.py
Haichao-Zhang: Removed

alf/networks/encoding_networks.py
Haichao-Zhang: Now keep it as list.

PR #439, opened 2020-03-10
Agent

hnyu:
RND on MRevenge has the performance just like the old version. 
Updated DIAYN because scale_distribution is turned off for now.
Will update the curves of ICM on super mario later


alf/algorithms/agent_helpers.py
```diff
@@ -0,0 +1,124 @@
...
+import torch
+
+import alf
+from alf.algorithms.rl_algorithm import RLAlgorithm
+from alf.data_structures import LossInfo
+from alf.utils.math_ops import add_ignore_empty
+
+
+class AgentStateSpecs(object):
+    def __init__(self):
```
runjerry: Is it possible to make it a more general helper class for future agents beyond the current one? Say users might want to create their own agent with different AgentState.

hnyu: I guess you mean with different algorithms. You need to specify how they interact with each other anyway, for example, one upstream algorithm will change the inputs of a downstream algorithm. So `__init__`, `preddict_step`, `rollout_step`, and `train_step` have to be reimplemented for a new agent.

emailweixu: I think @runjerry's suggestion is to pass AgentState as an argument to `__init__` so that it can handle different AgentState. and we can default it to be `alf.algorithms.agent.AgentState`

alf/algorithms/agent.py
```diff
@@ -53,10 +50,10 @@ class Agent(OnPolicyAlgorithm):
     def __init__(self,
                  observation_spec,
                  action_spec,
+                 env=None,
+                 goal_generator=None,
                  rl_algorithm_cls=ActorCriticAlgorithm,
-                 encoding_network: Network = None,
                  intrinsic_curiosity_module=None,
```
emailweixu: suggest to change a name for `intrinsic_curiosity_module`, as it suggests a specific algorithm. `intrinsic_reward_module` might be better.

alf/algorithms/agent.py
```diff
@@ -101,32 +104,31 @@ def __init__(self,
...
+                                                      "goal_generator")
             rl_observation_spec = [
                 observation_spec, goal_generator.action_spec
             ]
 
+        ## 2. rl algorithm
         rl_algorithm = rl_algorithm_cls(
             observation_spec=rl_observation_spec,
             action_spec=action_spec,
+            env=env,
```
emailweixu: no need to provide `env`? It's not root algorithm.

alf/algorithms/agent.py
```diff
@@ -53,10 +50,10 @@ class Agent(OnPolicyAlgorithm):
     def __init__(self,
                  observation_spec,
                  action_spec,
+                 env=None,
```
emailweixu, line 54: config needs to be provided.

alf/algorithms/agent.py
```diff
@@ -137,80 +139,71 @@ def __init__(self,
...
-            train_state_spec=train_state_spec,
             optimizer=[optimizer],
-            trainable_module_sets=[[encoding_network]],
             gradient_clipping=gradient_clipping,
             clip_by_global_norm=clip_by_global_norm,
             reward_shaping_fn=reward_shaping_fn,
             observation_transformer=observation_transformer,
             debug_summaries=debug_summaries,
-            name=name)
+            name=name,
```
emailweixu, line 157: `env` and `config`

alf/algorithms/agent.py
```diff
@@ -137,80 +139,71 @@ def __init__(self,
...
+        observation = time_step.observation
         if self._goal_generator is not None:
-            goal_step = self._goal_generator.predict(
-                time_step._replace(observation=observation),
-                state.goal_generator, epsilon_greedy)
+            goal_step = self._goal_generator.predict_step(
+                time_step, state.goal_generator, epsilon_greedy)
             new_state = new_state._replace(goal_generator=goal_step.state)
-            observation = [observation, goal_step.action]
+            observation = [time_step.observation, goal_step.output]
```
emailweixu:
```python
observation = [observation, goal_step.output]
```

alf/algorithms/agent.py
```diff
@@ -137,80 +139,71 @@ def __init__(self,
...
-        info = info._replace(rl=rl_step.info)
-
-        if self._entropy_target_algorithm:
-            # TODO: For off-policy training, skip entropy_target_algorithm
-            # during rollout()
+        if self._entropy_target_algorithm and self.is_on_policy():
+            # For off-policy training, skip the entropy target algorithm
+            # during `unroll()`.
             assert 'action_distribution' in rl_step.info._fields, (
                 "PolicyStep from rl_algorithm.rollout() does not contain "
```
emailweixu: PolicyStep => AlgStep

alf/algorithms/agent.py
```diff
@@ -272,74 +265,45 @@ def calc_training_reward(self, external_reward, info: AgentInfo):
...
+            external_reward, info.icm.reward if self._icm else None,
+            info.goal_generator.reward if
+            (self._goal_generator and
+             ('reward' in info.goal_generator._fields)) else None
+        ]
+        weights = [
+            self._extrinsic_reward_coef, self._intrinsic_reward_coef, 1.0
+        ]
+        names = ["extrinsic", "icm", "goal_generator"]
+
```
emailweixu, line 283: It's better to make code related to one algorithm adjacent to each other to minimize the effort of adding new algorithm.
```
rewards = [("extrinsic", external_reward, self._extrinsic_reward_conef)]
if self._icm:
   rewards.append(("icm", info.icm.reward, self._intrinsic_reward_coef))
if self._goal_generator and 'reward' in info.goal_generator._fields:
    rewards.append(("goal_generator", info.goal_generator.reward, 1.))
```

alf/algorithms/agent_helpers.py
```diff
@@ -0,0 +1,124 @@
...
+            loss=add_ignore_empty(loss_info.loss, new_loss_info.loss),
+            scalar_loss=add_ignore_empty(loss_info.scalar_loss,
+                                         new_loss_info.scalar_loss),
+            extra=loss_info.extra)
+
+    assert len(algorithms) > 0
+    assert len(algorithms) == len(names)
+    assert isinstance(algorithms[0], RLAlgorithm)
+
+    rl_loss_info = algorithms[0].calc_loss(
```
emailweixu: Perhaps we can use `if isintance(alg, RLAlgorithm)` for this. As `goal_generator` may also be `RLAlgorithm`

alf/algorithms/agent.py
```diff
@@ -272,74 +265,45 @@ def calc_training_reward(self, external_reward, info: AgentInfo):
...
+    def after_update(self, training_info):
+        self._rl_algorithm.after_update(
             training_info._replace(info=training_info.info.rl))
         if self._goal_generator:
-            self._goal_generator.after_train(
+            self._goal_generator.after_update(
                 training_info._replace(info=training_info.info.goal_generator))
+        if self._icm:
+            self._icm.after_update(
```
emailweixu: Only `RLAlgorithm` handles `TrainingInfo` per current interface.

alf/algorithms/agent.py
```diff
@@ -272,74 +265,45 @@ def calc_training_reward(self, external_reward, info: AgentInfo):
...
-    def after_train(self, training_info):
-        self._rl_algorithm.after_train(
+        # (algorithm, name)
+        algorithms = [(self._rl_algorithm, 'rl'), (self._icm, 'icm'),
+                      (self._goal_generator, 'goal_generator'),
+                      (self._entropy_target_algorithm, 'entropy_target')]
+        return accumulate_loss_info(training_info, *zip(*algorithms))
+
+    def after_update(self, training_info):
+        self._rl_algorithm.after_update(
```
emailweixu: `rollout_info` also needs to be replaced for completeness

alf/algorithms/agent.py
hnyu: yeah that's a bug fixed locally

alf/algorithms/agent.py
hnyu: fixed locally

alf/algorithms/agent.py
```diff
@@ -299,20 +294,19 @@ def calc_loss(self, training_info):
                 reward=self.calc_training_reward(training_info.reward,
                                                  training_info.info))
         # (algorithm, name)
-        algorithms = [(self._rl_algorithm, 'rl'), (self._icm, 'icm'),
-                      (self._goal_generator, 'goal_generator'),
-                      (self._entropy_target_algorithm, 'entropy_target')]
-        return accumulate_loss_info(training_info, *zip(*algorithms))
+        algorithms = [(self._rl_algorithm, 'rl'), (self._irm, 'irm')]
```
emailweixu: We don't always have self._irm

alf/algorithms/agent_helpers.py
```diff
@@ -64,61 +65,81 @@ def accumulate_algortihm_rewards(rewards, weights, names, summary_prefix,
...
+    assert loss_info is not None, "No loss info is calculated!"
     return loss_info
+
+
+def after_update(algorithms, names, training_info):
+    """For each provided algorithm, call its `after_update()` to do things after
+    the agent completes one gradient update (i.e. `update_with_gradient()`).
+
+    Args:
+        algorithms (list[Algorithm]): the list of algorithms whose `after_update`
```
emailweixu, line 137: list[RLAlgorithm]. Or update the code to differentiate RLAlgorithm with Algorithm

alf/algorithms/agent_helpers.py
```diff
@@ -64,61 +65,81 @@ def accumulate_algortihm_rewards(rewards, weights, names, summary_prefix,
...
+            is to be accumulated.
```
emailweixu: is to be accumulated => is to be called.

alf/algorithms/agent_helpers.py
hnyu: I think in general this API can be open to any Algorithm. For example, after each update I want to increase a counter for a non-RL algorithm. I've updated the code to handle differently between the two.

PR #442, opened 2020-03-10
Vtrace allsteps replay

le-horizon:
Hey Wei and Haonan, in this PR, I added CyclicOneTimeExperienceReplayer and set max_length of replay buffer to be unroll_length + 1, so that all timesteps can be used in training.

  To enable, when running async training, e.g. vtrace, simply set `AsyncOffPolicyDriver.exp_replayer="cycle_one_time"`

Training curve so far looks reasonable: 
  Sample efficiency is higher than vtrace(unroll 8) and vtrace(unroll 21), most likely due to adding back 1 timestep every 8 steps of rollout.
  Final performance is higher than unroll 8, but lower than unroll 21.
image

### DIFF
--- a/alf/drivers/async_off_policy_driver.py
+++ b/alf/drivers/async_off_policy_driver.py
@@ -83,7 +83,9 @@ class AsyncOffPolicyDriver(OffPolicyDriver):
             algorithm=algorithm,
             exp_replayer=exp_replayer,
             observers=observers,
-            metrics=metrics)
+            metrics=metrics,
+            unroll_length=unroll_length,
+            learn_queue_cap=learn_queue_cap)
 
         # create threads
         self._coord = tf.train.Coordinator()
@@ -131,11 +133,11 @@ class AsyncOffPolicyDriver(OffPolicyDriver):
         algorithm.set_metrics(self.get_metrics())
 
     def get_step_metrics(self):
-        """See PolicyDriver.get_step_metrics()"""
+        """See PolicyDriver.get_step_metrics()."""
         return self._log_thread.metrics[:2]
 
     def get_metrics(self):
-        """See PolicyDriver.get_metrics()"""
+        """See PolicyDriver.get_metrics()."""
         return self._log_thread.metrics
 
     def start(self):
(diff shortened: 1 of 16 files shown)


emailweixu:
Can you rebase first?

alf/experience_replayers/replay_buffer.py
```diff
@@ -175,28 +175,66 @@ def clear(self):
...
+                lambda buf: tf.concat([
+                    buf[:, pos:, ...], buf[:, :pos + size - self._max_length,
+                                           ...]
+                ],
+                                      axis=1), buffer)
         else:
+            # pull directly, not going over the end of the buffer
             return tf.nest.map_structure(lambda buf: buf[:, :size, ...],
-                                         self._buffer)
+                                         buffer)
```
emailweixu:
```python
buf[:, pos:pos+size, ...]
```
Test cases do not catch it.

alf/drivers/async_off_policy_driver.py
```diff
@@ -83,7 +83,9 @@ def __init__(self,
             algorithm=algorithm,
             exp_replayer=exp_replayer,
             observers=observers,
-            metrics=metrics)
+            metrics=metrics,
+            unroll_length=unroll_length,
+            learn_queue_cap=learn_queue_cap)
```
emailweixu: also need to update sync_off_policy_driver and sync_off_policy_trainer.

le-horizon: Done.  For SyncOffPolicyDriver, I'm setting learn_q_cap to be always 1, as we don't have multiple actors to gather experience from.

alf/experience_replayers/replay_buffer.py
le-horizon: `pos` is the ending position of the current buffer.  This line is correct, and is protected by the test line:170.

Since we only clear_all, when buffer is half full, we always read from 0th element in buf.  In this case, pos == size.

I haven't implemented anything that would lead to buffer beginning from an element that's not the 0th.  In order to cause that, we'll need to implement buffer.clear_n().  If you think this can be useful, I can add that, otherwise, I suggest we don't add unnecessary code?

PR #444, opened 2020-03-10
vae

hnyu:
Before VAE only handles flattened input, now it can handle nested inputs.

### DIFF
--- a/alf/algorithms/vae.py
+++ b/alf/algorithms/vae.py
@@ -15,98 +15,149 @@
 
 import gin
 
-import tensorflow as tf
-from tf_agents.networks import network
+import torch
+import torch.nn as nn
+
+from alf.algorithms.algorithm import Algorithm
+from alf.data_structures import AlgStep, LossInfo, namedtuple, TimeStep
+from alf.layers import FC
+from alf.networks import EncodingNetwork
+from alf.tensor_specs import TensorSpec
+from alf.utils import math_ops
+
+VAEInfo = namedtuple("VAEInfo", ["cost"], default_value=())
 
 
 @gin.configurable
-class VariationalAutoEncoder(tf.keras.Model):
-    """
-    VariationalAutoEncoder encodes data into diagonal multivariate gaussian, performs sampling with
-    reparametrization trick, and returns kl divergence between posterior and prior.
+class VariationalAutoEncoder(Algorithm):
+    """VariationalAutoEncoder encodes data into diagonal multivariate gaussian,
+    performs sampling with reparametrization trick, and returns kl divergence
+    between posterior and prior.
 
     Mathematically:
 
     log p(x) >= E_z log P(x|z) - beta KL(q(z|x) || prior(z))
 
-    sampling_forward method returns sampled z and KL, it is up to user of this class to use returned z to
-    decode and compute reconstructive loss to combine with kl loss returned here to optimize the whole network.
+    `sampling_forward()` method returns sampled z and KL, it is up to user of
+    this class to use returned z to decode and compute reconstructive loss to
+    combine with kl loss returned here to optimize the whole network.
 
-    See vae_test.py for example usages to train vanilla vae, conditional vae and vae with prior network
-    on mnist dataset.
-
-    comments: this class is not subclass of tf_agents.network, thus not copible.
+    See vae_test.py for example usages to train vanilla vae, conditional vae and
+    vae with prior network on mnist dataset.
     """
 
     def __init__(self,
-                 hidden_dim,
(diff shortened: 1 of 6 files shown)


alf/algorithms/vae.py
```diff
@@ -15,98 +15,149 @@
...
+            AlgStep:
+                output: the latent vector `z`
+                state: empty tuple ()
+                info (VAEInfo): kld_loss
+        """
+        z, kld_loss = self._sampling_forward(time_step.observation)
+        return AlgStep(output=z, state=state, info=VAEInfo(loss=kld_loss))
+
+    def calc_loss(self, info: VAEInfo):
+        loss = torch.mean(info.loss)
```
emailweixu: It's better to use batched loss so that mask can be applied at RLAlgorithm

alf/algorithms/vae.py
```diff
@@ -15,98 +15,149 @@
...
+        kl_div_loss = 0.5 * torch.sum(kl_div_loss, dim=-1)
         # reparameterization sampling: z = u + var ** 0.5 * eps
-        eps = tf.random.normal(
-            tf.shape(z_mean), dtype=tf.float32, mean=0., stddev=1.0)
-        z = z_mean + tf.exp(z_log_var * 0.5) * eps
+        eps = torch.randn(z_mean.shape)
+        z = z_mean + torch.exp(z_log_var * 0.5) * eps
         return z, self._beta * kl_div_loss
+
+    def train_step(self, time_step: TimeStep, state):
```
emailweixu: Doesn't need TimeStep as input.

alf/algorithms/vae_test.py
```diff
@@ -12,313 +12,141 @@
...
+            pr_train = pr_train[idx]
+            for i in range(0, x_train.shape[0], self._batch_size):
+                optimizer.zero_grad()
+                batch = x_train[i:i + self._batch_size]
+                y_batch = y_train[i:i + self._batch_size]
+                pr_batch = torch.nn.functional.one_hot(
+                    pr_train[i:i + self._batch_size],
+                    int(z_mean_prior_network.input_tensor_spec.shape[0])).to(
+                        torch.float32)
+                z, kl_loss = encoder._sampling_forward([pr_batch, batch])
```
emailweixu: should test `train_step` instead

alf/algorithms/vae.py
```diff
@@ -15,98 +15,149 @@
...
+        """
+        Args:
+            time_step (TimeStep): input time step data, where its `observation`
+                is augmented with a prior input if there is a prior network.
```
emailweixu: should make explicit how it is augmented: which one is first, which one is second

alf/algorithms/vae_test.py
```diff
@@ -12,313 +12,141 @@
...
-        hist = model.fit([self.y_train, self.x_train],
-                         epochs=self.epochs,
-                         batch_size=self.batch_size,
-                         validation_data=([self.y_test, self.x_test], None))
-
-        last_val_loss = hist.history['val_loss'][-1]
-        # cvae seems have the lowest errors with the same settings
-        print("loss: ", last_val_loss)
-        self.assertTrue(30.0 < last_val_loss < 31.0)
+        self._input_spec = TensorSpec((1, ))
```
emailweixu: call `super().setUp()`

alf/algorithms/vae_test.py
```diff
@@ -12,313 +12,141 @@
...
 import os
-import tensorflow as tf
-import alf.algorithms.vae as vae
 import numpy as np
+import tempfile
 
-from tensorflow.keras.datasets import mnist
-import matplotlib.pyplot as plt
+import torch
+import torchvision.datasets as datasets
```
emailweixu: not used

alf/algorithms/vae_test.py
```diff
@@ -12,313 +12,141 @@
...
-                         validation_data=([self.y_test, self.x_test], None))
-
-        last_val_loss = hist.history['val_loss'][-1]
-        # cvae seems have the lowest errors with the same settings
-        print("loss: ", last_val_loss)
-        self.assertTrue(30.0 < last_val_loss < 31.0)
+        self._input_spec = TensorSpec((1, ))
+        self._epochs = 10
+        self._batch_size = 100
+        self._latent_dim = 1
```
emailweixu: make it 2?

alf/algorithms/vae.py
hnyu: Why does this algorithm need mask? For any type of steps it can be trained.

Issue #434, opened 2020-03-07
extend gin


witwolf:

witwolf:
we use use gin heavily in alf, and we should make it more flexible to use 

i just add an interface
```
@gin.configurable
def gin_eval(source):
    """Evaluate the given source in the context of globals and locals.

    Args:
        source (str): string that will be evaluated
    """
    frame = _gin_call_context_frame()  # TODO
    return eval(source, frame.f_locals)
```

and needs  meeting the following tests
```python
    def test_refer_to_unregistered_func(self):
        """Test refer to unregister func by eval it with registered helper function
        `gin_eval`.

        builtin and 3rd-party functions are not registered in gin by default,
        we can not refer to it directly and it's also not unrealistic to register
        it explict by `gin.external_configurable(func)` for all functions we might
        refer to
        """

        @gin.configurable
        def _test(func):
            func()

        with self.assertRaises(ValueError):
            gin.parse_config([
                "_test/func=@list",
            ])

        gin.parse_config([
            "list/gin_eval.source='list'",
            "_test.func=@list/gin_eval()"
        ])
        _test()

    def _test_refer_to_local_values(self):
        """Test refer to local values
        Evaluate it in correct context frame
        """

        @gin.configurable
        def _add(a, b):
            return a + b

        def _test_add(a, b):
            return _add()

        gin.parse_config([
            "a/gin_eval.source='a'",
            "_add.a=@a/gin_eval()",
            "b/gin_eval.source='b'",
            "_add.b=@b/gin_eval()"
        ])

        self.assertEqual(1 + 2, _test_add(1, 2))
        self.assertEqual(2 + 3, _test_add(2, 3))

        @gin.configurable
        def _test_valid_prod(a, b, result):
            self.assertEqual(result, a * b)

        gin.parse_config([
            "_test_valid_prod/gin_eval.source='np.prod([a, b])'",
            "_test_valid_prod.result=@_test_valid_prod/gin_eval()"
        ])
        a, b = 1, 2
        _test_valid_prod(a, b)
        a, b = 3, 4
        _test_valid_prod(a, b)

```
not finished , have problem with getting correct context frame now

PR #445, opened 2020-03-11
add gin_eval to make gin config more flexiable

witwolf:
#434 

Just add a helper function to make refer unregistered function or class or non literal expression  as    parameter value a little easy .

Perhaps, we can try to extend gin-config to support new syntax

```
scope/fun.arg=#local_var_name  # or nested , eg [#ocal_var_name1 , #ocal_var_name2] ...
scope/fun.arg=#unregistered_func
```

### DIFF
--- a/alf/utils/gin_util_test.py
+++ b/alf/utils/gin_util_test.py
@@ -42,7 +42,7 @@ class GinUtilsTest(unittest.TestCase):
             ])
 
         gin.parse_config(
-            ["list/gin_eval.source='list'", "_test.func=@list/gin_eval()"])
+            ["list/gin_eval.str='list'", "_test.func=@list/gin_eval()"])
         _test()
 
     def test_refer_to_local_values(self):
@@ -59,8 +59,8 @@ class GinUtilsTest(unittest.TestCase):
             return _add()
 
         gin.parse_config([
-            "a/gin_eval.source='a'", "_add.a=@a/gin_eval()",
-            "b/gin_eval.source='b'", "_add.b=@b/gin_eval()"
+            "a/gin_eval.str='a'", "_add.a=@a/gin_eval()", "b/gin_eval.str='b'",
+            "_add.b=@b/gin_eval()"
         ])
 
         self.assertEqual(1 + 2, _test_add(1, 2))
@@ -71,7 +71,7 @@ class GinUtilsTest(unittest.TestCase):
             return value
 
         gin.parse_config([
-            "_value/gin_eval.source='np.prod([a,b])'",
+            "_value/gin_eval.str='np.prod([a,b])'",
             "_value.value=@_value/gin_eval()"
         ])
         a, b = 1, 2
--- a/alf/utils/gin_util_test.py
+++ b/alf/utils/gin_util_test.py
@@ -38,7 +38,7 @@ class GinUtilsTest(unittest.TestCase):
 
         with self.assertRaises(ValueError):
             gin.parse_config([
-                "_test/func=@list",
+                "_test.func=@list",
             ])
 
         gin.parse_config(
--- a/alf/utils/gin_util_test.py
+++ b/alf/utils/gin_util_test.py
@@ -1,0 +1,87 @@
+# Copyright (c) 2019 Horizon Robotics. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import gin
+import unittest
+import numpy as np
+import alf.utils.external_configurables
+
+
+class GinUtilsTest(unittest.TestCase):
+    """Tests for alf.utils.gin_utils
+    """
+
+    def test_refer_to_unregistered_func(self):
+        """Test refer to unregister func by eval it with registered helper function
+        `gin_eval`.
+
+        builtin and 3rd-party functions are not registered in gin by default,
+        we can not refer to it directly and it's also not unrealistic to register
+        it explict by `gin.external_configurable(func)` for all functions we might
+        refer to
+        """
+
+        @gin.configurable
+        def _test(func):
+            func()
+
+        with self.assertRaises(ValueError):
+            gin.parse_config([
+                "_test/func=@list",
+            ])
+
+        gin.parse_config(
+            ["list/gin_eval.source='list'", "_test.func=@list/gin_eval()"])
+        _test()
+
+    def test_refer_to_local_values(self):
+        """Test refer to local values
+
+        Passing expression as parameter value
+        """
+
+        @gin.configurable
+        def _add(a, b):
+            return a + b
+
+        def _test_add(a, b):
+            return _add()
+
+        gin.parse_config([
+            "a/gin_eval.source='a'", "_add.a=@a/gin_eval()",
+            "b/gin_eval.source='b'", "_add.b=@b/gin_eval()"
+        ])
+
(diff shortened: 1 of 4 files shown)


witwolf:
with this pr , Instance-related (not common utils functions) functions can be move to  .gin  files
eg:
https://github.com/HorizonRobotics/alf/blob/4d869b34dac7fec1cf02b5357c676a15956038a1/alf/utils/common.py#L619-L636

setup.py
```diff
@@ -18,19 +18,10 @@
...
-        'numpy',
-        'opencv-python >= 3.4.1.15',
-        'pathos == 0.2.4',
-        'pillow',
-        'psutil',
-        'pybullet == 2.5.0',
-        'tensorboard == 2.1.0',
-        'torch == 1.4.0',
-        'torchvision == 0.5.0',
+        'atari_py == 0.1.7', 'fasteners', 'gym == 0.12.5', 'matplotlib',
```
emailweixu: Can you keep the previous format so that changes are easy to spot?

alf/utils/gin_utils.py
```diff
@@ -53,3 +56,86 @@ def inoperative_config_str(max_line_length=80, continuation_indent=4):
     gin.config._OPERATIVE_CONFIG = operative_config
     gin.config._IMPORTED_MODULES = imported_module
     return inoperative_str
+
+
+def _config_gin_eval(func):
```
emailweixu: Can you describe the purpose of this function?

alf/utils/gin_utils.py
```diff
@@ -53,3 +56,86 @@ def inoperative_config_str(max_line_length=80, continuation_indent=4):
...
+    >>> @gin.configurable
+        def calc_arc_len(radius, radian):
+            return radius * radian
+    >>> r = 2
+    >>> calc_arc_len()
+
+    # Inside "config.gin"
+    radius/gin_eval.source="r"
+    radian/gin_eval.source="0.3*np.pi"
+    test/calc_arc.radius=@radius/gin_eval()
```
emailweixu: calc_arc => calc_arc_len

alf/utils/gin_util_test.py
```diff
@@ -0,0 +1,87 @@
...
+        refer to
+        """
+
+        @gin.configurable
+        def _test(func):
+            func()
+
+        with self.assertRaises(ValueError):
+            gin.parse_config([
+                "_test/func=@list",
```
hnyu: _test/func -> _test.func

setup.py
witwolf: It's auto re-formate by pre-commit hook yapf (0.28.0)

alf/utils/gin_utils.py
witwolf: fixed

alf/utils/gin_util_test.py
witwolf: fixed

setup.py
witwolf: passed ci code style checking job

alf/utils/gin_utils.py
```diff
@@ -53,3 +56,91 @@ def inoperative_config_str(max_line_length=80, continuation_indent=4):
...
+        name=name,
+        module=module,
+        whitelist=None,
+        blacklist=None,
+        selector=selector)
+    return gin_wrapper
+
+
+@_config_gin_eval
+def gin_eval(source):
```
hnyu: Now I actually think `source` would be better called `str`. It's shorter and can indicate the type of the argument. What do you think?

witwolf: renamed

witwolf: actually i just follow the signature of builtin function `eval(__source, __globals, __locals)`

alf/utils/gin_utils.py
```diff
@@ -53,3 +56,93 @@ def inoperative_config_str(max_line_length=80, continuation_indent=4):
     gin.config._OPERATIVE_CONFIG = operative_config
     gin.config._IMPORTED_MODULES = imported_module
     return inoperative_str
+
+
+def _config_gin_eval(func):
+    """Decorator to make function `gin_eval` configurable.
+    And it get and pass expected frame context variables `globals` and `locals`
```
emailweixu: What is the expected frame context supposed to be?

alf/utils/gin_utils.py
```diff
@@ -53,3 +56,93 @@ def inoperative_config_str(max_line_length=80, continuation_indent=4):
...
+                gin.config._CONFIG.get((partial_scope_str, selector), {}))
+        scope_str = partial_scope_str
+        operative_parameter_values = {}
+        operative_parameter_values.update(new_kwargs)
+        gin.config._OPERATIVE_CONFIG.setdefault(
+            (scope_str, selector), {}).update(operative_parameter_values)
+        new_kwargs = copy.deepcopy(new_kwargs)
+        # hack here ,injection was actually called at frame_trace[-7]
+        # (gin-config>=0.1.3,<=0.3.3)
+        frame = sys._getframe(7)
```
emailweixu: 7 seems very dangerous here. Is there any way to prevent error?

alf/utils/gin_utils.py
```diff
@@ -53,3 +56,93 @@ def inoperative_config_str(max_line_length=80, continuation_indent=4):
...
+        module=module,
+        whitelist=None,
+        blacklist=None,
+        selector=selector)
+    return gin_wrapper
+
+
+@_config_gin_eval
+def gin_eval(str):
+    """Evaluate the given source in the context of globals and locals.
```
emailweixu: Need to explain where are the globals and locals from.

setup.py
emailweixu: It's weird. I also use the save version according to CONTRIBUTING.md. But it doesn't format this way.

alf/utils/gin_utils.py
witwolf: Yes, only for gin-config>=0.1.3,<=0.3.3.  I'll  try to find a better way to extend gin config

emailweixu:
It would be much better if we can have an operator like ``, which evaluate the quoted string by python. So we can write something like:
```
r = 0.5
arc_len = `calc_arc_len(%r, 0.3*np.pi)`
```
Handling `%` is a little bit tricky though. Since it is a legit python operator and used as macro prefix for gin.

witwolf:
> It would be much better if we can have an operator like ``, which evaluate the quoted string by python. So we can write something like:
> 
> ```
> r = 0.5
> arc_len = `calc_arc_len(%r, 0.3*np.pi)`
> ```


Not complicated  to support this by modify the source code of gin-config,  just need to find a good way to extend gin in alf, i will implement the feature this week.

witwolf:
Seems  maintain a forked branch of `gin-config` instead of extending it inside code repo of `alf` 
is much simpler , it takes only a few dozen lines .  what do you thing ?  @emailweixu  

And   I have made some changes to extend extend gin to support

* refer to local or global variable (func)
* call registered function
* call unregistered function

  [Here is the code](https://github.com/witwolf/gin-config/pull/1) .

emailweixu:
> Seems maintain a forked branch of `gin-config` instead of extending it inside code repo of `alf`
> is much simpler , it takes only a few dozen lines . what do you thing ? @emailweixu
> 
> And I have made some changes to extend extend gin to support
> 
> * refer to local or global variable (func)
> * call registered function
> * call unregistered function
>   [Here is the code](https://github.com/witwolf/gin-config/pull/1) .

We want the configuration to be stable with respect to code changes. I feel allowing access to unregistered functions or variables may cause the config even more fragile. 

And for function call, it will be even better if keyword arguments can be supported .

We can have a forked copy. Then we have to handle the installation of gin differently in setup.py and CI.
Meanwhile, it will be better if you can push the change to the official version.

witwolf:
> And for function call, it will be even better if keyword arguments can be supported .
> Meanwhile, it will be better if you can push the change to the official version.

keyword arguments is now  supported, and i have created issues and pr to the official version
[Evaluate reference with kwargs ](https://github.com/google/gin-config/issues/58)
[Ordinary reference as value]( https://github.com/google/gin-config/issues/61)

> 
> We can have a forked copy. Then we have to handle the installation of gin differently in setup.py and CI.

ok, i'll update it

witwolf:
included in #545

PR #448, opened 2020-03-11
clipping gradient norms

hnyu:


alf/utils/tensor_utils.py
```diff
@@ -80,3 +81,105 @@ def to_tensor(data, dtype=None):
...
+
+def clip_by_norms(tensors, clip_norm):
+    """Clipping a nest of tensors *in place* by calling `clip_by_norm` on each
+    entry.
+
+    Args:
+        tensors (nested Tensor): a nest of tensors
+        clip_norm (float or Tensor): a positive scalar
+    """
+    assert alf.nest.is_nested(tensors), "tensors must be a nest!"
```
witwolf: it's seem this line can be removed , and `clip_by_norm` can be removed too ?

hnyu: Just in case these functions will be called separately, as in my test cases.

hnyu: simplified

alf/utils/tensor_utils.py
```diff
@@ -80,3 +81,105 @@ def to_tensor(data, dtype=None):
...
+    Returns:
+        global_norm (Tensor): a scalar tensor representing the global norm. If
+            `use_norm` is provided, it will be returned instead.
+    """
+    assert alf.nest.is_nested(tensors), "tensors must be a nest!"
+    if use_norm is None:
+        use_norm = global_norm(tensors)
+
+    clip_norm = torch.as_tensor(clip_norm)
+    assert len(clip_norm.shape) == 0, "clip_norm must be a scalar!"
```
emailweixu: assert clip_norm.ndim == 0

alf/utils/tensor_utils.py
```diff
@@ -80,3 +81,105 @@ def to_tensor(data, dtype=None):
...
+    assert clip_norm > 0, "clip_norm must be positive!"
+
+    scale = clip_norm / torch.max(clip_norm, use_norm)
+
+    def _clip(tensor):
+        if tensor is not None:
+            tensor.mul_(scale)
+
+    alf.nest.map_structure(_clip, tensors)
```
emailweixu: if scale < 1:

alf/utils/tensor_utils.py
```diff
@@ -80,3 +81,105 @@ def to_tensor(data, dtype=None):
...
+    Args:
+        tensors (nested Tensor): a nest of tensors to be clipped
+        clip_norm (float or Tensor): a positive floating scalar
+        use_norm (float or Tensor): the global norm to use. If None,
+            `global_norm()` will be used to compute the norm.
+    Returns:
+        global_norm (Tensor): a scalar tensor representing the global norm. If
+            `use_norm` is provided, it will be returned instead.
+    """
+    assert alf.nest.is_nested(tensors), "tensors must be a nest!"
```
emailweixu: why does it have to be nested?

alf/utils/tensor_utils.py
emailweixu: +1. Seems clip_by_norm is redundant.

alf/utils/tensor_utils.py
hnyu: Because if it's just a single tensor, then the name "global" seems confusing. If TF's version, it has to be a list.

alf/utils/tensor_utils.py
emailweixu: Why not do this?

hnyu: Strange. I remember that I did this change.

alf/utils/tensor_utils.py
hnyu: And also this one was changed, somehow the version was not correct

alf/utils/tensor_utils.py
hnyu: OK, I was about to push yesterday night, but VPN was down...

hnyu:
Another thing is that because for optimization we have to clip grad tensors in place, but for normal tensors that have grads, these in-place operations will fail. Maybe I can add a flag to support both.

PR #449, opened 2020-03-12
Pr sarsa algorithm

emailweixu:
This is based on my local tf version of sarsa. So I included in the first commit for reference.


alf/algorithms/sarsa_algorithm.py
```diff
@@ -90,179 +210,343 @@ def __init__(self,
...
-                                                  epsilon_greedy)
+            if epsilon_greedy == 1.0:
+                action = dist_utils.rsample_action_distribution(
+                    action_distribution)
+            else:
+                action = dist_utils.epsilon_greedy_sample(
+                    action_distribution, epsilon_greedy)
         else:
 
             def _sample(a, ou):
```
Haichao-Zhang: This is a questions shared with ddpg. The ```_sample``` function in tf version performs epsilon-greedy operation at the batch-level. In your pytorch implementation, this is operated as the sample-level. Do you think we should all switch to sample-level as you done here? Not sure if there is a clear difference empirically.

alf/utils/math_ops.py
```diff
@@ -87,3 +91,51 @@ def max_n(inputs):
 def square(x):
     """torch doesn't have square."""
     return torch.pow(x, 2)
+
+
+def min_n(inputs):
+    """Calculate the maximum of n Tensors
```
Haichao-Zhang: ```maximum``` -> ```minimum```

alf/utils/math_ops.py
```diff
@@ -87,3 +91,51 @@ def max_n(inputs):
...
+
+    Args:
+        inputs (list[Tensor]): list of Tensors, should have the same shape
+    Returns:
+        the elementwise maximum of all the tensors in `inputs`
```
Haichao-Zhang: ```maximum``` -> ```minimum```

alf/algorithms/sarsa_algorithm.py
emailweixu: tf version is also batch level

Haichao-Zhang: I might didn't make the point clear. For the ```_sample``` function, it seems:

TF (Line 161-164 of tf code) performs epsilon-greedy operation at batch-level (epsilon-greedy is on/off for the whole batch)
Pytorch (Line 341-346 of pytorch code): epsilon-greedy operation at sample-level (epsilon-greedy is on/off differently for samples within the batch).

Did I interpret the code incorrectly?

alf/algorithms/sarsa_algorithm.py
```diff
@@ -13,41 +13,136 @@
...
+class FastCriticBias(nn.Module):
+    """
+    The motivation of this is to prevent the critic value becomes greater and
+    greater as training goes on because of the max operation.
+
+    FastCriticBias estimates expected return `b` by minimizing the following
+    expectation:
+        min_b E(b - r - discount * b)^2
+    where r is reward. `b` can be analytically solved as:
+        b = E(r) / (1 - E(discount))
```
hnyu, line 62: discount is a constant? So E(discount) = discount here. Also in this objective the assumption seems to be that b_next is very close to b_current.

alf/algorithms/sarsa_algorithm.py
```diff
@@ -13,41 +13,136 @@
...
+                corresponding element in `mask` is True.
+        """
+        if torch.any(mask):
+            if isinstance(critics, torch.Tensor):
+                critics = [critics]
+            critics, reward, discount = math_ops.weighted_reduce_mean(
+                [critics, reward, discount], mask)
+            self._averager.average([critics, reward, discount])
+
+            critics, reward, discount = self._averager.get()
```
hnyu: These two lines can be combined into one because `average` by default will return.

alf/algorithms/sarsa_algorithm.py
```diff
@@ -13,41 +13,136 @@
...
+    default_value=())
+SarsaLossInfo = namedtuple('SarsaLossInfo', ['actor', 'critic', 'alpha'])
+
+nest_map = alf.nest.map_structure
+
+
+class FastCriticBias(nn.Module):
+    """
+    The motivation of this is to prevent the critic value becomes greater and
+    greater as training goes on because of the max operation.
```
hnyu, line 56: SARSA doesn't have any max operation? because the target critic uses the actual action

alf/algorithms/sarsa_algorithm.py
```diff
@@ -90,179 +210,343 @@ def __init__(self,
...
-            network_state=state.critic)
-        critic_state = tf.nest.map_structure(
-            lambda new_s, s: tf.where(not_first_step, new_s, s), critic_state,
-            state.critic)
+            states=state.critics)
+
+        if self._critic_bias is not None:
+            self._critic_bias.update(
+                critics=[
+                    c - b
```
hnyu, line 459: Why -b here instead of just using c which is the critic?

alf/algorithms/sarsa_algorithm.py
emailweixu: action is trained to maximize the q value. It is an implicit max operation. And for off-policy training, the action used to calculate the target_critic is not from experience, it is calculated every time.

alf/algorithms/sarsa_algorithm.py
emailweixu: It is because b is already added to the original critic by `self._calc_critics()`

alf/algorithms/sarsa_algorithm.py
emailweixu: discount may not be a constant. It is time_step.discount * discount.
Yes, there is such assumption. b=b_next is true for if MDP is ergodic. If not, then we hope they are close.

alf/algorithms/sarsa_algorithm.py
emailweixu: Fixed.

alf/algorithms/sarsa_algorithm.py
```diff
@@ -217,10 +218,9 @@ def __init__(self,
                 Does not perform clipping if dqda_clipping == 0.
             actor_optimizer (tf.optimizers.Optimizer): The optimizer for actor.
             critic_optimizer (tf.optimizers.Optimizer): The optimizer for critic
```
Haichao-Zhang: should update the document on all the three optimizers (```tf.optimizers.Optimizer```) to pytorch

alf/algorithms/sarsa_algorithm_test.py
```diff
@@ -0,0 +1,164 @@
...
+                logging.INFO,
+                "%d reward=%f" % (i, float(eval_time_step.reward.mean())),
+                n_seconds=1)
+
+        self.assertAlmostEqual(
+            1.0, float(eval_time_step.reward.mean()), delta=0.3)
+
+
+if __name__ == '__main__':
+    #    SarsaTest().test_sarsa()
```
Haichao-Zhang: can remove

PR #451, opened 2020-03-12
torch entropy target algorithm with tests

le-horizon:
@emailweixu and @hnyu,

I've only added some simple tests to make sure the python code can actually run.  No tests exist in the tf version.

I've removed SeedStream because torch seems to use a single random number generator per device, and doesn't have the problem that tf has, which could have two distributions generating random numbers using duplicate generators with the same seed.

I also changed to use td.TransformedDistribution check to handle scaled normal distributions.

@hnyu , could you test Agent with this and see if it works?

Thanks,
Le


le-horizon:
I've removed

alf/algorithms/entropy_target_algorithm.py
```diff
@@ -214,61 +203,62 @@ def adjust_alpha(self, entropy):
...
+        if self._stage == -1:
+            _free()
+        if self._stage >= 0:
+            _adjust()
+        alpha = torch.exp(self._log_alpha)
 
         if self._debug_summaries:
-            run_if(should_record_summaries(), _summarize)
+            if should_record_summaries():
+                with self.name_scope:
```
emailweixu: There is no self.name_scope. Use:
```python
```

alf/algorithms/entropy_target_algorithm.py
```diff
@@ -214,61 +203,62 @@ def adjust_alpha(self, entropy):
...
             crossing = above != previous_above
             update_rate = self._update_rate
-            update_rate = tf.where(crossing, 0.9 * update_rate, update_rate)
-            update_rate = tf.maximum(update_rate, self._slow_update_rate)
-            update_rate = tf.where(entropy < self._fast_stage_thresh,
-                                   np.float32(self._fast_update_rate),
-                                   update_rate)
-            self._update_rate.assign(update_rate)
-            above = tf.cast(above, tf.float32)
+            update_rate = torch.where(crossing, 0.9 * update_rate, update_rate)
```
emailweixu: the arguments for where should be Tensor

alf/algorithms/entropy_target_algorithm.py
```diff
@@ -214,61 +203,62 @@ def adjust_alpha(self, entropy):
...
+            update_rate = torch.max(update_rate, self._slow_update_rate)
```
emailweixu: arguments for max should be Tensors

alf/algorithms/entropy_target_algorithm_test.py
```diff
@@ -0,0 +1,65 @@
...
+        embedding = 10 * torch.rand(
+            (100, ) + self._input_tensor_spec.shape, dtype=torch.float32)
+
+        dist, _ = net(embedding)
+
+        alg_step = alg.train_step(dist, self._time_step.step_type)
+
+        info = EntropyTargetInfo(
+            step_type=self._time_step.step_type, loss=alg_step.info.loss)
+        alg.calc_loss(info)
```
emailweixu: need to make sure every stage is tested.

alf/utils/dist_utils.py
```diff
@@ -361,23 +359,84 @@ def estimated_entropy(dist,
...
+            # original distribution it transforms.
+            # TransformedDistribution is used by NormalProjectionNetwork with
+            # scale_distribution=True
+            dist = dist.base_dist
+        return (len(dist.batch_shape) + len(dist.event_shape) - len(
+            action_spec.shape))
+
+    def _compute_entropy(dist: td.Distribution, action_spec):
+        if isinstance(dist, td.TransformedDistribution):
+            entropy, entropy_for_gradient = estimated_entropy(dist)
```
hnyu: It might be possible to get a closed-form of entropy given a Normal as the base dist? It's better (lower variance) than this estimated one. Also, it seems that after transformations, the entropy actually changes, and we are computing the entropy of the base dist instead of the input dist.

hnyu: Something like what TFP does: https://github.com/tensorflow/probability/blob/356cfddef026b3339b8f2a81e600acd2ff8e22b4/tensorflow_probability/python/distributions/transformed_distribution.py#L636
(Probably it's complicated, but we need to spend time figuring out if the current estimation is the best way to do this).

le-horizon:
Fixed for Wei's comments.  Good points Haonan, should I leave them as todo here, given that we want to merge and use the algorithm first, before improving it?

Thanks,
Le

hnyu:
> Fixed for Wei's comments. Good points Haonan, should I leave them as todo here, given that we want to merge and use the algorithm first, before improving it?
> 
> Thanks,
> Le

Yeah probably for the future. But for now, can we wrap https://github.com/HorizonRobotics/alf/blob/9e3b5aa568495ec7f01ea578a06ed779f8f6ed6d/alf/networks/projection_networks.py#L185 as a new class (Maybe also called SquashSpecNormal or something similar) so that we only check and support for this class? The current way of checking TransformedDistribution seems too loose. If there are other transformed distributions, then report an error.

alf/algorithms/entropy_target_algorithm.py
```diff
@@ -176,99 +163,105 @@ def train_step(self, distribution, step_type):
...
+        num = torch.sum(mask)
         not_empty = num > 0
-        num = tf.maximum(num, 1)
-        entropy2 = tf.reduce_sum(tf.square(entropy)) / num
-        entropy = tf.reduce_sum(entropy) / num
-        entropy_std = tf.sqrt(tf.maximum(0.0, entropy2 - entropy * entropy))
-
-        run_if(not_empty, lambda: self.adjust_alpha(entropy))
+        num = max(num, 1)
+        entropy2 = torch.sum(entropy**2) / num
```
hnyu: suggest using math_ops.square because **2 is much slower than pow():https://github.com/pytorch/pytorch/issues/2618

alf/algorithms/entropy_target_algorithm.py
```diff
@@ -176,99 +163,105 @@ def train_step(self, distribution, step_type):
...
+        entropy = torch.sum(entropy) / num
+        entropy_std = math.sqrt(max(0.0, entropy2 - entropy * entropy))
```
hnyu: max -> torch.max, math.sqrt -> torch.sqrt

alf/algorithms/entropy_target_algorithm.py
```diff
@@ -176,99 +163,105 @@ def train_step(self, distribution, step_type):
...
-            with self.name_scope:
-                tf.summary.scalar("entropy_std", entropy_std)
+        if not_empty:
+            self.adjust_alpha(entropy)
 
         if self._debug_summaries:
-            run_if(
-                tf.logical_and(not_empty, should_record_summaries()),
-                _summarize)
+            if not_empty & should_record_summaries():
```
hnyu: & -> and

alf/algorithms/entropy_target_algorithm.py
```diff
@@ -176,99 +163,105 @@ def train_step(self, distribution, step_type):
...
             # -1 * increasing + 0.5 * (1 - increasing)
             update_rate = (
                 0.5 - 1.5 * increasing) * self._very_slow_update_rate
-            self._stage.assign_add(tf.cast(below, tf.int32))
-            self._log_alpha.assign(
-                tf.maximum(self._log_alpha + update_rate,
-                           np.float32(self._min_log_alpha)))
+            self._stage.add_(below.type(torch.int32))
+            self._log_alpha.fill_(
+                max(self._log_alpha + update_rate,
```
hnyu: max->torch.max

alf/algorithms/entropy_target_algorithm.py
```diff
@@ -176,99 +163,105 @@ def train_step(self, distribution, step_type):
...
             adjusted entropy regularization
         """
         prev_avg_entropy = self._avg_entropy.get()
         avg_entropy = self._avg_entropy.average(entropy)
 
         def _init_entropy():
-            self._max_entropy.assign(
-                tf.minimum(0.8 * avg_entropy, avg_entropy / 0.8))
-            self._stage.assign_add(1)
+            self._max_entropy.fill_(min(0.8 * avg_entropy, avg_entropy / 0.8))
```
hnyu: min->torch.min

alf/algorithms/entropy_target_algorithm.py
```diff
@@ -176,99 +163,105 @@ def train_step(self, distribution, step_type):
...
-            update_rate = tf.where(crossing, 0.9 * update_rate, update_rate)
-            update_rate = tf.maximum(update_rate, self._slow_update_rate)
-            update_rate = tf.where(entropy < self._fast_stage_thresh,
-                                   np.float32(self._fast_update_rate),
-                                   update_rate)
-            self._update_rate.assign(update_rate)
-            above = tf.cast(above, tf.float32)
+            update_rate = torch.where(crossing, 0.9 * update_rate, update_rate)
+            update_rate = torch.max(update_rate,
+                                    torch.tensor(self._slow_update_rate))
```
hnyu: I suggest converting arguments `self._slow_update_rate`, `self._fast_update_rate`, and `self._min_log_alpha` to tensors in `__init__` all the once, and avoiding converting them again and again during training.

alf/utils/dist_utils.py
```diff
@@ -361,23 +359,84 @@ def estimated_entropy(dist,
...
     else:
-        entropy = -dist.log_prob(tf.stop_gradient(single_action))
+        entropy = -dist.log_prob(single_action.detach())
         if check_numerics:
-            entropy = tf.debugging.check_numerics(entropy, 'entropy')
-        entropy_for_gradient = -0.5 * tf.math.square(entropy)
-        entropy = tf.reduce_mean(entropy, axis=0)
-        entropy_for_gradient = tf.reduce_mean(entropy_for_gradient, axis=0)
+            assert torch.all(torch.isfinite(entropy))
+        entropy_for_gradient = -0.5 * entropy**2
```
hnyu: **2 -> math_ops.square

alf/utils/dist_utils.py
```diff
@@ -361,23 +359,84 @@ def estimated_entropy(dist,
...
+        reduce_dims = list(range(outer_rank, rank))
+        entropy = torch.sum(entropy, reduce_dims)
+        entropy_for_gradient = torch.sum(entropy_for_gradient, reduce_dims)
+        return entropy, entropy_for_gradient
+
+    entropies = list(
+        map(_compute_entropy, nest.flatten(distributions),
+            nest.flatten(action_spec)))
+    entropies_for_gradient = [eg for e, eg in entropies]
+    entropies = [e for e, eg in entropies]
```
hnyu: No need to loop twice. `entropies, entropies_for_gradient = zip(*entropies)`

alf/utils/dist_utils.py
```diff
@@ -409,7 +468,7 @@ def calc_default_max_entropy(spec, fraction=0.8):
     assert fraction <= 1.0 and fraction > 0
     zeros = np.zeros(spec.shape)
     min_max = np.broadcast(spec.minimum, spec.maximum, zeros)
-    cont = tensor_spec.is_continuous(spec)
+    cont = spec.dtype.is_floating_point
```
hnyu: Use `spec.is_continuous`

alf/utils/dist_utils.py
```diff
@@ -386,7 +445,7 @@ def calc_default_target_entropy(spec):
     """
     zeros = np.zeros(spec.shape)
     min_max = np.broadcast(spec.minimum, spec.maximum, zeros)
-    cont = spec.is_continuous
+    cont = spec.dtype.is_floating_point
```
hnyu: Use `spec.is_continuous`

alf/algorithms/entropy_target_algorithm.py
hnyu: Use math_ops.square() instead because **2 is much slower pytorch/pytorch#2618

alf/utils/dist_utils.py
hnyu: Use math_ops.square() instead because **2 is much slower pytorch/pytorch#2618

alf/algorithms/entropy_target_algorithm.py
emailweixu: ** has special handing for 2. https://github.com/pytorch/pytorch/pull/2635/files/b03a18f13aebb3c34b80aebfad3e4f13836d0d8e#diff-3c0bce18bb4926c168da5d7acac34183

alf/utils/dist_utils.py
le-horizon: Noted.

alf/utils/dist_utils.py
le-horizon: Nice!

alf/utils/dist_utils.py
le-horizon: Done.

alf/utils/dist_utils.py
le-horizon: Done.

alf/algorithms/entropy_target_algorithm.py
le-horizon: Done.

alf/algorithms/entropy_target_algorithm.py
le-horizon: Done.

alf/algorithms/entropy_target_algorithm.py
le-horizon: Done.

alf/algorithms/entropy_target_algorithm.py
le-horizon: Done

alf/algorithms/entropy_target_algorithm.py
le-horizon: Done.

alf/algorithms/entropy_target_algorithm.py
le-horizon: Good point.

alf/utils/dist_utils.py
le-horizon: no longer the case?

alf/algorithms/entropy_target_algorithm.py
hnyu: OK maybe later they added this optimization. The link I showed was two years ago

alf/networks/projection_networks.py
```diff
@@ -182,7 +183,7 @@ def _normal_dist(self, means, stds):
             # kl_divergence(TransformedDistribution(Independent, transforms),
             #               TransformedDistribution(Independent, transforms))
             # ````
-            squashed_dist = td.TransformedDistribution(
+            squashed_dist = dist_utils.SquashSpecNormal(
```
hnyu: SquashSpecNormal should do the transformations in its `__init__` (the lines below should be moved to its construction. The API might be like `SquashSpecNormal(base_normal_dist, spec)` and the `action_means` and `action_magnitudes` will be computed inside it.

alf/utils/dist_utils.py
```diff
@@ -345,9 +345,7 @@ def estimated_entropy(dist,
     detail.
 
     Args:
-        dist (tfp.distributions.Distribution): concerned distribution
-        seed (Any): Any Python object convertible to string, supplying the
-            initial entropy.
+        dist (torch.distributions.Distribution): concerned distribution
         assume_reparametrization (bool): assume the sample from continuous
```
emailweixu: We don't need this. td.Distribution have `rsample()` and `sample()`. When rsample is available, we should use `rsample`, which is reparameterized sampling.

alf/utils/dist_utils.py
```diff
@@ -361,23 +359,104 @@ def estimated_entropy(dist,
...
+            action spec.
+    Returns:
+        tuple of (entropy, entropy_for_gradient). You should use entropy in
+        situations where its value is needed, and entropy_for_gradient where
+        you need to calculate the gradient of entropy.
+    """
+
+    def _calc_outer_rank(dist: td.Distribution, action_spec):
+        if isinstance(dist, SquashSpecNormal):
+            # TransformedDistribution does not implement the two necessary
```
emailweixu: This statement is not true for torch Distribution.

alf/utils/dist_utils.py
emailweixu: the entropy for transformed distribution is only available for affine transformation as noted in the linked code. Because of this, we can make the condition as the following:
```python
if (isinstance(dist, TransformedDistribution) and
   not (dist.bijector.is_constant_jacobian
           and  dist.bijector._is_injective)):
```
And we may remove `SqushSpecNormal`

alf/utils/dist_utils.py
```diff
@@ -361,23 +359,104 @@ def estimated_entropy(dist,
...
+        else:
+            entropy = dist.entropy()
+            entropy_for_gradient = entropy
+
+        outer_rank = _calc_outer_rank(dist, action_spec)
+        rank = len(entropy.shape)
+        reduce_dims = list(range(outer_rank, rank))
+        entropy = torch.sum(entropy, reduce_dims)
+        entropy_for_gradient = torch.sum(entropy_for_gradient, reduce_dims)
+        return entropy, entropy_for_gradient
```
emailweixu: line 445-450 can be removed. The entropy should already have correct shape because the use of IndependentDistribution.

alf/algorithms/entropy_target_algorithm.py
```diff
@@ -256,7 +256,7 @@ def _adjust():
         alpha = torch.exp(self._log_alpha)
 
         if self._debug_summaries and should_record_summaries():
-            with alf.summary.scope(self.__class__.__name__):
+            with alf.summary.scope(self.name):
```
emailweixu: self.name is not provided at `__init__`

alf/utils/dist_utils.py
emailweixu: You can take a look at the code of TransformedDistribution. It can only calculate entropy if (dist.bijector.is_constant_jacobian and  dist.bijector._is_injective)

alf/utils/dist_utils.py
emailweixu: _calc_outer_rank can be removed.

le-horizon: Done.

alf/utils/dist_utils.py
le-horizon: Replying thru chat.

alf/algorithms/entropy_target_algorithm.py
le-horizon: Oops, good point.

PR #457, opened 2020-03-13
add same padding mode as in TF

hnyu:
To reproduce some gin files, we might need this "same" padding mode. Because of PyTorch's implementation, the output size might be a little different from TF's counterpart. See comments in the code.

### DIFF
--- a/alf/networks/encoding_networks.py
+++ b/alf/networks/encoding_networks.py
@@ -42,6 +42,7 @@ class ImageEncodingNetwork(Network):
                  input_channels,
                  input_size,
                  conv_layer_params,
+                 same_padding=False,
                  activation=torch.relu,
                  flatten_output=False,
                  name="ImageEncodingNetwork"):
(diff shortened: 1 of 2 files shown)


alf/networks/encoding_networks_test.py
```diff
@@ -40,28 +40,33 @@ def test_empty_layers(self):
...
+            same_padding=same_padding,
             activation=torch.tanh,
             flatten_output=flatten_output)
 
         self.assertLen(list(network.parameters()), 4)  # two conv2d layers
 
         output, _ = network(img)
-        output_shape = (15, 34, 16)
+        if same_padding:
+            output_shape = (15, 34, 17)
```
emailweixu: This doesn't like same_padding. If same spatial dimension should be (32, 16)

hnyu: "same" padding is a wrong word. It's essentially taking a padding=kernel//2. If kernel is an odd number, then one additional padding is applied to the right. when strides=1, the output size will be the same; if not, it will not be the same output size. See my comments in ImageEncodingNetwork

hnyu: Made a minor change so that now if kernel size is an odd number and strides=1, then the output size will be the same. Note that because PyTorch doesn't support different paddings for left and right, it's impossible to still get the same output size if kernel size is an even number, by just configuring PyTorch conv2d or transposedconv2d layers. https://pytorch.org/docs/stable/nn.html#torch.nn.Conv2d

alf/networks/encoding_networks.py
```diff
@@ -57,12 +58,25 @@ def __init__(self,
 
         where H = output size, H1 = input size, HF = size of kernel, P = padding
 
+        Regarding padding: in the previous TF version, we have two padding modes:
+        "valid" and "same". For the former, we always have no padding (P=0); for
+        the latter, it's also called "half padding" (P=(HF-1)//2 when strides=1
```
Haichao-Zhang: ```(P=(HF-1)//2```  an extra `(` before ```P```

hnyu: It's not. The right ) is at the end.

alf/networks/encoding_networks.py
```diff
@@ -126,9 +145,18 @@ def __init__(self,
...
-            H = (H1-1) * strides + HF - 2P
+            H = (H1-1) * strides + HF - 2P + OP
 
-        where H = output size, H1 = input size, HF = size of kernel, P = padding
+        where H = output size, H1 = input size, HF = size of kernel, P = padding,
+        OP = output_padding (currently hardcoded to be 0 for this class).
+
+        Regarding padding: in the previous TF version, we have two padding modes:
+        "valid" and "same". For the former, we always have no padding (P=0); for
+        the latter, it's also called "half padding" (P=(HF-1)//2 when strides=1
```
Haichao-Zhang: (P=(HF-1)//2 an extra ( before P

PR #461, opened 2020-03-16
Fix normal projection network

Haichao-Zhang:
The purpose is for matching pytorch with tf version.

tanh_transforms = [td.AffineTransform(loc=0, scale=2),
                        td.SigmoidTransform(),
                        td.AffineTransform(loc=-1, scale=2]


hnyu:
Is this just for the purpose of comparable loss with TF's version? Or some benefit of tanh over sigmoid

Haichao-Zhang:
This PR is for matching the performance of Pytorch with TF.

PR #463, opened 2020-03-17
Stable tanh bijector

### LINKED ISSUE #455: TTP: stable tanh

Haichao-Zhang:


alf/utils/dist_utils.py
```diff
@@ -24,6 +24,48 @@
...
+
+    def _inverse(self, y):
+        # Based on https://github.com/tensorflow/agents/commit/dfb8c85a01d65832b05315928c010336df13f7b9#diff-a572e559b953f965c5c2cd1b9ded2c7b
+
+        # 0.99999997 is the maximum value such that atanh(x) is valid for both float32 and float64
+        def atanh(x):
+            return 0.5 * torch.log((1 + x) / (1 - x))
+
+        y = torch.where(
+            torch.abs(y) <= 1.0, torch.clamp(y, -0.99999997, 0.99999997), y)
```
hnyu, line 61: Shouldn't abs(y) <= 1.0 always be true?

alf/utils/dist_utils.py
```diff
@@ -24,6 +24,48 @@
...
+        return atanh(y)
+
+    def log_abs_det_jacobian(self, x, y):
+        return 2.0 * (torch.log(
+            torch.tensor(2.0, dtype=x.dtype, requires_grad=False).detach()) - x
```
hnyu: detach is unnecessary if requires_grad=False?

alf/utils/dist_utils.py
Haichao-Zhang: As an inverse function ```abs(y) <= 1``` should be true. As a general function ```y``` could be a general input with ```abs(y) > 1```.  In this case, for the invalid ```y``` values (```abs(y)>1```), we keep them unchanged and won't clamp them into valid range.

alf/utils/dist_utils.py
Haichao-Zhang: detach removed

hnyu (approved):

PR #466, opened 2020-03-17
Adam TF version

### LINKED ISSUE #465: TTP: TF_Adam

Haichao-Zhang:
Tensorflow and Pytorch implement Adam optimizer differently. The difference can not be eliminated by matching and setting hyper-parameters (e.g. eps). This PR is to provide a pytorch version of the Adam following TF's way of implementation. This might be useful for matching performance between the two.


hnyu:
Probably comment on which part is copied (don't need to be checked) so that to reduce reviewing cost?

Haichao-Zhang:
Now including original pytorch Adam as the first commit. So the difference from it can be easily viewed by selecting the second commit (AdamTF).

hnyu (approved):

PR #468, opened 2020-03-18
exp is on CPU, needs to be transferred to cuda before torch.where

le-horizon:
@hnyu this should fix the problem.  I tested locally.

Thanks,
Le

### DIFF
--- a/alf/metrics/metrics.py
+++ b/alf/metrics/metrics.py
@@ -163,20 +163,21 @@ class AverageReturnMetric(metric.StepMetric):
             The arguments, for easy chaining.
         """
         # Zero out batch indices where a new episode is starting.
+        is_first = exp.is_first().to('cuda')
         self._return_accumulator[:] = torch.where(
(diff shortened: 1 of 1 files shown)


hnyu:
Why exp is on CPU? The metrics during training seems OK, but just fails for eval? By default the tensors should use the default device which is CUDA.

le-horizon:
No idea why.  @emailweixu do you know why experience for eval env contains CPU tensors instead of GPU?

emailweixu:
> No idea why. @emailweixu do you know why experience for eval env contains CPU tensors instead of GPU?

training uses ProcessEnvironment, which explicitly convert to cuda
eval uses ThreadTorchEnvironment.

emailweixu:
We should fix ThreadTorchEnvironment intead.

le-horizon:
> We should fix ThreadTorchEnvironment intead.

@runjerry could you help explicitly convert experience tensors to cuda? as is done in ProcessEnvironment?

Thanks!

hnyu:
Refer to #474

PR #470, opened 2020-03-18
add action clipping wrapper

hnyu:
Tested with sac. Write gin file name to logging because sometimes if there are multiple running jobs, it will be easier to decide which is which from the screen.


alf/environments/torch_wrappers.py
```diff
@@ -89,6 +89,31 @@ def wrapped_env(self):
...
+        """Create an ActionClip torch wrapper.
+
+        Args:
+            env (TorchEnvironment): An TorchEnvironment instance to wrap.
+        """
+        super(ContinuousActionClip, self).__init__(env)
+
+    def _step(self, action):
+        action_spec = self.action_spec()
+        if action_spec.is_continuous:
```
emailweixu: action_spec might be nested

emailweixu (changes requested):

alf/environments/torch_wrappers.py
```diff
@@ -89,6 +89,36 @@ def wrapped_env(self):
...
+        action_spec = self.action_spec()
+
+        def _clip_action(spec, action):
+            # Check if the action is corrupted or not.
+            if torch.any(torch.isnan(action)):
+                raise ValueError(
+                    "NAN action detected! action: {}".format(action))
+            if spec.is_continuous:
+                if isinstance(spec, ts.BoundedTensorSpec):
+                    action = torch.max(
```
Haichao-Zhang: Might be able to use the function from spec_util.py:
https://github.com/HorizonRobotics/alf/blob/b1988ee0fa0a242263ed63ec014249664d8c48ba/alf/utils/spec_utils.py#L60-L70

emailweixu (dismissed):

hnyu:
resolved conflicts

Haichao-Zhang (approved):

PR #472, opened 2020-03-18
summarize_action()

### LINKED ISSUE #467: TTP: summarize_action
need to convert histogram_continuous() and histogram_discrete()

emailweixu:


alf/utils/summary_utils.py
```diff
@@ -108,47 +78,20 @@ def histogram_continuous(name,
...
+    if bucket_min is None:
+        bucket_min = data.min()
+    if bucket_max is None:
+        bucket_max = data.max()
+    bins = (
+        bucket_min +
+        (torch.arange(bucket_count + 1, dtype=torch.float64) / bucket_count) *
+        (bucket_max - bucket_min))
+    data = data.clamp(bucket_min, bucket_max)
+    alf.summary.histogram(name, data, step=step, bins=bins.cpu())
```
le-horizon, line 98: not sure why .cpu() is needed.

le-horizon (dismissed):
Not sure about .cpu(), otherwise looks good.

alf/utils/summary_utils.py
emailweixu: pytorch complains not cpu somewhere.

le-horizon: Could it be some other place is returning cpu instead of gpu tensors, like the environment case?
Ok to merge this change, just want to make sure there are no underlying problems.

le-horizon (approved):

Haichao-Zhang (approved):

alf/utils/summary_utils.py
emailweixu: bins indeed can be cuda.  The failure is because pytorch's tensorboard API want to call numpy() (fail if not cpu) on the tensor, which can be a cuda tensor.

PR #473, opened 2020-03-19
numerical gain for weight initialization

hnyu:
There is one issue with pytorch's sigmoid gain which is 1.0. I've corrected it.

### DIFF
--- a/alf/examples/icm_super_mario.gin
+++ b/alf/examples/icm_super_mario.gin
@@ -37,7 +37,8 @@ value/ValueNetwork.fc_layer_params=(256,256)
 value/ValueNetwork.conv_layer_params=%CONV_LAYER_PARAMS
 value/ValueNetwork.activation=@torch.nn.functional.elu
 
-ac/AdamTF.lr=3e-5
+ac/Adam.lr=3e-5
+ac/Adam.eps=1e-7
 
 icm/encoding_net_fc_layer_params=(256,)
 icm/EncodingNetwork.input_tensor_spec=%observation_spec
@@ -52,7 +53,7 @@ ICMAlgorithm.activation=@torch.nn.functional.elu
 
 ActorCriticAlgorithm.actor_network=@actor/ActorDistributionNetwork()
 ActorCriticAlgorithm.value_network=@value/ValueNetwork()
-Agent.optimizer=@ac/AdamTF()
+Agent.optimizer=@ac/Adam()
 Agent.intrinsic_reward_module=@ICMAlgorithm()
 Agent.observation_transformer=@image_scale_transformer
 ActorCriticLoss.entropy_regularization=0.01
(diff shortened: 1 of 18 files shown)


alf/networks/initializers.py
```diff
@@ -58,7 +89,7 @@ def variance_scaling_init(tensor,
         from alf.networks.initializers import variance_scaling_init
         layer = nn.Linear(2, 2)
         variance_scaling_init(layer.weight.data,
-                              nonlinearity="leaky_relu",
+                              nonlinearity=nn.functional.leaky_relu,
                               nonlinearity_param=0.01)
```
Haichao-Zhang: ```nonlinearity_param```  can be removed now?

alf/examples/icm_super_mario.gin
```diff
@@ -37,8 +37,7 @@ value/ValueNetwork.fc_layer_params=(256,256)
 value/ValueNetwork.conv_layer_params=%CONV_LAYER_PARAMS
 value/ValueNetwork.activation=@torch.nn.functional.elu
 
-ac/Adam.lr=3e-5
-ac/Adam.eps=1e-7
+ac/AdamTF.lr=3e-5
```
Haichao-Zhang: Can possibly move the changes in ```icm_super_mario.gin``` to another PR as they are less relevant to the numerical gain calculation.

Haichao-Zhang (changes requested):

alf/networks/initializers.py
```diff
@@ -19,16 +19,49 @@
...
+    Var(y) = Var(w) * E(x^2)
+
+    Args:
+        nonlinearity (Callable): any callable activation function
+        num_samples (int): number of samples for approximating the gain
+
+    Returns:
+        gain (float): a gain factor that will be applied to the init weights
+    """
+    tensor = torch.randn(num_samples)
```
emailweixu: we don't need to resort to sampling for doing this as its error is proportional to 1/sqrt(num_samples). We can do numerical integration. It's more accurate and don't have random behaviour.
```python
x=torch.arange(-r,r,delta)
variance = pdf(x)*(nonlinearity(x)**2)).sum()*delta
```

hnyu: Changed, the result seems slightly better.

alf/networks/initializers.py
```diff
@@ -23,25 +23,27 @@
...
+        dz (float): `dz` in the integration
+        r (float): `z` range will be `[-r, r]`
 
     Returns:
         gain (float): a gain factor that will be applied to the init weights
     """
-    tensor = torch.randn(num_samples)
-    tensor = nonlinearity(tensor)
-    return torch.sqrt(1.0 / torch.mean(tensor**2)).cpu().numpy()
+    dist = torch.distributions.normal.Normal(0, 1)
```
Haichao-Zhang, line 42: Which version is better, both in terms of speed and accuracy?

hnyu: Speed should be similar as they both involve 1000 samples. The integration has a better accuracy.

Haichao-Zhang (approved):

PR #474, opened 2020-03-19
fix the bug in TimeLimit that will lead to cpu backend error

runjerry:
The previous resetting StepType.LAST in TimeLimit torch wrapper leads to a cpu step_type when process env is involved.

### DIFF
--- a/alf/environments/torch_wrappers.py
+++ b/alf/environments/torch_wrappers.py
@@ -363,11 +363,14 @@ class NonEpisodicAgent(TorchEnvironmentBaseWrapper):
 
     def _step(self, action):
         time_step = self._env.step(action)
-        if time_step.step_type == StepType.LAST:
+        LAST = StepType.LAST
+        if alf.get_default_device() == "cuda":
+            LAST = LAST.cuda()
+        if time_step.step_type == LAST:
             # We set a non-zero discount so that the target value would not be
             # zero (non-episodic).
             time_step = time_step._replace(
-                discount=torch.tensor(self._discount, torch.float32))
+                discount=torch.as_tensor(self._discount, torch.float32))
         return time_step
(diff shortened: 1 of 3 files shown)


alf/environments/torch_wrappers.py
```diff
@@ -118,7 +118,10 @@ def _step(self, action):
 
         self._num_steps += 1
         if self._num_steps >= self._duration:
-            time_step = time_step._replace(step_type=StepType.LAST)
+            LAST = StepType.LAST
+            if time_step.step_type.is_cuda:
+                LAST = LAST.cuda()
+            time_step = time_step._replace(step_type=LAST)
```
hnyu: I think the more principled way (and less error-prone in the future) is to make sure that when you init step types in `data_structures.py` the tensors are on the right device? Otherwise if someone uses step types in other places, there will still be errors.

hnyu: Just check if cuda is available, if yes, then manually specify there devices

le-horizon: Probably due to StepType.LAST being a class constant and executed during import of the class, while set_default_device only happens when we start executing alf.set_default_device("cuda") in train.py?  So torch defaults the tensor to be CPU?

emailweixu: We can change them to numpy int: np.int32(0), np.int32(1), np.int32(2). Otherwise, we are going to be troubled by this again because we may mix them with other cuda tensor in one operation.

hnyu (approved):

alf/environments/torch_wrappers.py
runjerry: Updated, but need some luck to pass the ddpg test.

PR #478, opened 2020-03-20
convert tf thread based coordinator to process based coordinator

le-horizon:
Preparing for async.


alf/algorithms/ddpg_algorithm_test.py
```diff
@@ -85,7 +85,7 @@ def test_ddpg_algorithm(self):
         print(eval_time_step.reward.mean())
 
         self.assertAlmostEqual(
-            1.0, float(eval_time_step.reward.mean()), delta=3e-1)
+            1.0, float(eval_time_step.reward.mean()), delta=6e-1)
```
hnyu: should revert this change as Haichao now has a more reliable test.

alf/drivers/process_coordinator.py
```diff
@@ -0,0 +1,284 @@
+# Copyright (c) 2020 Horizon Robotics. All Rights Reserved.
```
hnyu: I believe now we no longer has the concept of `driver`? So probably move this file to some other place.

hnyu (changes requested):
To my understanding, most code is taken from TF. For an easier review, can you split the current commit into two? The first one is directly copying TF's thread coordinator code, and the second one is the modification you made? Then the diff will be easier to read.

alf/drivers/process_coordinator_test.py
```diff
@@ -0,0 +1,421 @@
...
+        m = MyAlgorithm()
+        m.share_memory()
+        coord = coordinator.Coordinator()
+        p = MyProcess(coord, kwargs={"m": m})
+        p.start()
+        coord.join([p])
+        # Registered Buffers are shared acrosses processes:
+        self.assertEqual(-3, m._m)
+        # Simple tensors are not shared:
+        self.assertEqual(0, m.x)
```
le-horizon: @emailweixu and @hnyu This is the proof of concept to share models/algorithms across processes.  It's the same thing as checkpointing: registered tensors/buffers can be shared and manipulated across processes.

Although for us, we probably only need other processes to read, instead of modify the shared model.

alf/drivers/process_coordinator.py
```diff
@@ -0,0 +1,331 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Coordinate asynchronous training process termination on request."""
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
```
emailweixu: The above 3 lines can be removed. We do not support python 2

alf/drivers/process_coordinator_test.py
```diff
@@ -0,0 +1,421 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+"""Tests for Process Coordinator."""
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
```
emailweixu: can be removed

alf/drivers/process_coordinator.py
```diff
@@ -0,0 +1,331 @@
...
+    def start_loop(self):
+        """Called when the process starts."""
+        pass
+
+    def stop_loop(self):
+        """Called when the process stops."""
+        pass
+
+    def run_loop(self):
+        """Called at 'timer_interval_secs' boundaries."""
```
emailweixu: what is `timer_interval_secs`?

alf/drivers/process_coordinator.py
```diff
@@ -0,0 +1,331 @@
...
+        with self._lock:
+            if self._exc_info_to_raise:
+                six.reraise(*self._exc_info_to_raise)
+
+
+class Process(mp.Process):
+    """A coordinated process class to execute acting loops.
+    """
+
+    def __init__(self, coord, target=None, args=(), kwargs={}):
```
emailweixu: comment

alf/drivers/process_coordinator_test.py
```diff
@@ -0,0 +1,421 @@
...
+        processes = [
+            Process(target=SleepABit, args=(0.01, )),
+            Process(target=SleepABit, args=(0.02, )),
+            Process(target=SleepABit, args=(0.01, ))
+        ]
+        for t in processes:
+            t.start()
+        for p in processes:
+            coord.register_process(p)
+        WaitForProcessesToRegister(coord, len(processes))
```
emailweixu: why do we need to wait here?

alf/drivers/process_coordinator_test.py
```diff
@@ -0,0 +1,421 @@
...
+import sys
+import time
+
+import torch
+
+from alf import test
+from alf.drivers import process_coordinator as coordinator
+
+
+def StopOnEvent(coord, wait_for_stop, set_when_stopped):
```
emailweixu: our naming convention is: stop_on_event

alf/algorithms/ddpg_algorithm_test.py
le-horizon: let me revert this after rebasing Haichao's change.

alf/drivers/process_coordinator_test.py
le-horizon: This is actually part of the test.  It only exits if registration is successful.

le-horizon (commented):
Thanks for the quick reviewing Wei and Haonan.

Good point Haonan.  This commit shows the change on top of TF's threading coordinator and test.
https://github.com/HorizonRobotics/alf/pull/478/commits/58560b95b6dd35acca6370a2d4758060cf53425d

Most of the changes are just naming changes.  Main change is for coordinator's member variables to be multiprocessing compatible - i.e. share memory via multiprocessing.Value().

I'll address all other comments later.  Replied some.

alf/drivers/process_coordinator_test.py
le-horizon: Btw, in order to avoid the case where a model somehow changes its member python variable instead of using a registered_buffer, we should probably declare all non-tensor member variables of all the nn.Module's to be immutable?  It's an easy mistake to make, and checkpointing would fail because of that.

le-horizon:
Dear reviewers,

I've addressed all comments.  Please take another look.

Thanks,
Le

alf/utils/process_coordinator_test.py
```diff
@@ -0,0 +1,432 @@
...
+        processes = [
+            Process(target=sleep_a_bit, args=(0.01, )),
+            Process(target=sleep_a_bit, args=(0.02, )),
+            Process(target=sleep_a_bit, args=(0.01, ))
+        ]
+        for t in processes:
+            t.start()
+        for p in processes:
+            coord.register_process(p)
+        wait_for_processes_to_register(coord, len(processes))
```
hnyu: Looks like this wait_for_processes_to_register won't be called automatically after a process is registered, which is dangerous. Should it be included in `register_process`?

alf/utils/process_coordinator_test.py
```diff
@@ -0,0 +1,432 @@
...
+        coord.join()
+        for t in processes:
+            self.assertFalse(t.is_alive())
+
+    def testJoinSomeRegistered(self):
+        coord = coordinator.Coordinator()
+        processes = [
+            Process(target=sleep_a_bit, args=(0.01, )),
```
hnyu: The sleeping time can be longer for join() to wait

alf/utils/process_coordinator_test.py
```diff
@@ -0,0 +1,432 @@
...
+        n.value -= 1
+
+
+class ProcessTest(test.TestCase):
+    def testTargetArgs(self):
+        n = Value(ctypes.c_int, 3)
+        coord = coordinator.Coordinator()
+        p = coordinator.Process.process(
+            coord, target=_stop_at_0, args=(coord, n))
+        coord.join([p])
```
hnyu: coord already registers this process when initializing the Process object, why do you still need to provide it as the argument for join()? It would be convenient if we can just join without any argument.

alf/utils/process_coordinator_test.py
```diff
@@ -0,0 +1,432 @@
...
+                self._m -= 1
+                self.x -= 1
+
+        m = MyAlgorithm()
+        m.share_memory()
+        coord = coordinator.Coordinator()
+        p = MyProcess(coord, kwargs={"m": m})
+        # A change in parent process is reflected in child process
+        # via share_memory
+        m._m.fill_(-1)
```
hnyu, line 408: no need to have this line; just check m._m equal to -3 after join()

alf/utils/process_coordinator_test.py
le-horizon: Good observation.  It used to be the case (in tf.coordinator) that sleep_a_bit will register itself with the coordinator.  But since we now register processes in the main process, it's sequential, and doesn't need the wait.

I can either 1) remove the wait here, or 2) let subprocesses register with coordinator (then we'll need to wait here).

I don't think it's possible to have coordinator wait for registration in register_process, because any process can cut in before coordinator knows how many processes are registering, hence it's unknown how many processes are going to register at the time of register_process.  It's only possible to wait after all processes are registered, and we know the total number of processes there are, maybe in join().

Probably removing the wait is more consistent with our usage in the async code where a global process generates all subprocesses and registers them in coord sequentially. And this is a simpler coding style that's less error prone as well.  What do you think?

alf/utils/process_coordinator_test.py
le-horizon: I'm not sure what you mean here.  Currently p0 and p2 registers very quickly and executes, p1 is joined later and we wait for p1 to end before join() returns.  I believe that's the intention of the code -- to test that we do wait for p1.

alf/utils/process_coordinator_test.py
le-horizon: Good point.  Removed.

alf/utils/process_coordinator_test.py
le-horizon: This is to test after subprocess p is already created, when we change model _m on parent process, the change is actually reflected in the subprocess.

None of the existing tests do that test.

le-horizon (commented):
Thanks for the quick turnaround Haonan.  Done some, answered some, and asked for clarification for a couple.

PTAL.

Thanks,
Le

alf/utils/process_coordinator_test.py
hnyu: If you want to do that, then this line should be moved after the process starts. Right now, the child process will inherit this -1.

alf/utils/process_coordinator_test.py
hnyu: I mean 0.01 might be too small so that the join() won't wait because all processes just end because it's called.

alf/utils/process_coordinator_test.py
hnyu: Yeah I think it's better to let the main process to do all the registrations, basically being done when you create a subprocess by passing the coord into the init.

alf/utils/process_coordinator.py
```diff
@@ -0,0 +1,337 @@
...
+        super().__init__()
+        self._coord = coord
+        # allow pass in target or overriding body
+        self._target = target or self.body
+        self._args = args
+        self._kwargs = kwargs
+        self._coord.register_process(self)
+
+    @staticmethod
+    def process(coord, target, args=(), kwargs={}):
```
hnyu: Also, I don't think that this function is necessary. It's better that p.start() won't be automatically called. In this way, the main process can manage a group of subprocesses and decide when to start.

le-horizon: removed.

alf/utils/process_coordinator_test.py
le-horizon: just to clarify, letting main process do registration means

```python
p = Process()
coord = Coord()
coord.register_process(p)
```

Passing coord into subprocess init would let subprocess register itself to the coordinator during subprocess.start().

alf/utils/process_coordinator_test.py
le-horizon: I see, increased to 0.02, as starting the 3 processes takes about 0.01 second, 0.02 should be enough.

alf/utils/process_coordinator_test.py
le-horizon: good point, done.

le-horizon (commented):
Hey Haonan, thanks for the careful review and comments.
One question and all others are fixed.

Thanks,
Le

alf/utils/process_coordinator_test.py
hnyu: I mean registering in init(). Passing coord into a subprocess's init will actually register it  in the main process, right? The init function is still inside the main process until you call p.start() to fork.

le-horizon: Oh, yes, you are right.  Do you also want me to use MyProcess.__init__ to register these simple test processes, or are you Ok with the current code (manually registering)?

I have other test code using MyProcess lower down in the file, which does registration in __init__.

hnyu: I think for testing code, either way is fine.

alf/utils/process_coordinator_test.py
```diff
@@ -406,10 +400,11 @@ def decrement(self):
         m.share_memory()
         coord = coordinator.Coordinator()
         p = MyProcess(coord, kwargs={"m": m})
+        p.start()
+        time.sleep(0.01)  # sleep just enough for init to execute
```
hnyu: init is executed in the main process and it will always end before p.start(). I guess this sleep is not necessary and the comment is confusing.

le-horizon: Good point.  I just want to test changing share_memory after subprocess started, but before subprocess start decrementing the value.  Changed comment.  PTAL.

hnyu (approved):

PR #479, opened 2020-03-20
improve entropy_target and unify algorithm step APIs

hnyu:
For EntropyTargetAlgorithm, in some extreme cases such as Gaussian policy with a bad initialization, the current entropy might be equal to the prev entropy. Then the old check will always return `increasing=False`, and `alpha` will be increased to a very large value.

Also make algorithms to have both `rollout_step` and `train_step` to unify the interfaces called by Agent.

### DIFF
--- a/alf/algorithms/entropy_target_algorithm.py
+++ b/alf/algorithms/entropy_target_algorithm.py
@@ -142,6 +142,23 @@ class EntropyTargetAlgorithm(Algorithm):
         self._slow_update_rate = torch.tensor(slow_update_rate)
         self._fast_update_rate = torch.tensor(fast_update_rate)
 
+    def rollout_step(self, distribution, step_type, on_policy_training):
+        """Rollout step.
+
+        Args:
+            distribution (nested Distribution): action distribution from the
+                policy.
+            step_type (StepType): the step type for the distributions.
+            on_policy_training (bool): If False, this step does nothing.
+
+        Returns:
+            AlgStep. `info` field is LossInfo, other fields are empty.
+        """
+        if on_policy_training:
+            return self.train_step(distribution, step_type)
+        else:
+            return AlgStep()
+
     def train_step(self, distribution, step_type):
         """Train step.
 
@@ -169,7 +186,7 @@ class EntropyTargetAlgorithm(Algorithm):
         Args:
             training_info (EntropyTargetInfo): for computing loss.
             valid_mask (tensor): valid mask to be applied on time steps.
-        
+
         Returns:
             LossInfo.
         """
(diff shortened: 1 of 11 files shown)


alf/algorithms/entropy_target_algorithm.py
```diff
@@ -142,6 +142,23 @@ def __init__(self,
...
+        """Rollout step.
+
+        Args:
+            distribution (nested Distribution): action distribution from the
+                policy.
+            step_type (StepType): the step type for the distributions.
+            on_policy_training (bool): If False, this step does nothing.
+
+        Returns:
+            AlgStep. `info` field is LossInfo, other fields are empty.
```
Haichao-Zhang: Can add ```All fields are empty If on_policy_training  is False```

alf/examples/sac.gin
```diff
@@ -0,0 +1,8 @@
+import alf.algorithms.agent
```
Haichao-Zhang, line 1: Not sure if it is good to use this shared sac.gin file. It does not save too many lines of shared code while making the original gin file less self-contained.

hnyu: > Not sure if it is good to use this shared sac.gin file. It does not save too many lines of shared code while making the original gin file less self-contained.

Just similar to 'ppo.gin'.

alf/examples/sac_bipedal_walker.gin
```diff
@@ -36,18 +30,13 @@ SacAlgorithm.alpha_optimizer=@alpha/Adam()
...
-TrainerConfig.trainer=@sync_off_policy_trainer
 TrainerConfig.initial_collect_steps=32000
 TrainerConfig.mini_batch_length=4
 TrainerConfig.unroll_length=1
 TrainerConfig.mini_batch_size=4096
 TrainerConfig.clear_replay_buffer=False
 TrainerConfig.num_updates_per_train_step=1
-TrainerConfig.algorithm_ctor=@Agent
 TrainerConfig.num_iterations=200000
 TrainerConfig.checkpoint_interval=100000
```
emailweixu: this is removed. now is num_checkpoints

hnyu: I know. This PR is just to include 'sac.gin' for all related gin files. I guess sac_bipedal_walker.gin hasn't been reproduced by anyone, so the configuration is untouched.

Haichao-Zhang: In this case, it might be better to keep these gin files as is until they are reproduced. Then a shared ```sac.gin``` might be introduced. 
This leads to a cleaner versioning of these files.

hnyu: Reverted some files. Was trying to save some hassles of making the change for the future.

Haichao-Zhang (approved):

PR #480, opened 2020-03-21
fix one_time exp replayer

hnyu:
Previously it's broken; now rl_algorithm will decide itself whether to use "uniform" or "one_time" given `config.whole_replay_buffer_training` and `config.clear_replay_buffer`.

### DIFF
--- a/alf/experience_replayers/experience_replay.py
+++ b/alf/experience_replayers/experience_replay.py
@@ -94,17 +94,21 @@ class OnetimeExperienceReplayer(ExperienceReplayer):
     """
 
     def __init__(self):
-        self._experience = None
-        self._batch_size = None
+        self.clear()
 
     def observe(self, exp):
-        # The shape is [learn_queue_cap, unroll_length, env_batch_size, ...]
-        exp = alf.nest.map_structure(lambda e: e.transpose(1, 2), exp)
-        # flatten the shape (num_envs, env_batch_size)
-        self._experience = alf.nest.map_structure(
-            lambda e: e.reshape(-1, *e.shape[2:]), exp)
+        # exp shape is [num_envs, ...]
+        if self._experience is None:
+            self._experience = alf.nest.map_structure(lambda e: [e.detach()],
+                                                      exp)
+            self._exp_sample = exp
+        else:
+            alf.nest.map_structure_up_to(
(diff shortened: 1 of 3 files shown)


alf/experience_replayers/experience_replay.py
```diff
@@ -120,10 +124,19 @@ def replay(self, sample_batch_size, mini_batch_length):
         raise NotImplementedError()  # Only supports replaying all!
 
     def replay_all(self):
-        return self._experience
+        assert self._experience, "No experience is observed yet!"
+        # batch major (B, T, ...)
+        exps = alf.nest.map_structure_up_to(
+            self._exp_sample, lambda e: torch.transpose(
+                torch.stack(e, dim=0), 0, 1), self._experience)
+        self.clear(
```
Haichao-Zhang: ```whole_replay_buffer_training = True``` does not necessarily need to clear all the buffer.
We have another flag ```clear_replay_buffer``` for explicitly specifying this. See:

https://github.com/HorizonRobotics/alf/blob/38c714928ab3e2f6a6c016dd9f7d71eba0ebf43a/alf/algorithms/off_policy_algorithm.py#L130-L133

hnyu: The name of this replayer is "one-time". So I believe that we should enforce this clear after replay.

Haichao-Zhang: Yes, but isn't it true that  ```one-time-replayer``` is equivalent to replayer with ```whole_replay_buffer_training = True``` and ```clear_replay_buffer = True```?

hnyu: I was thinking that "one-time" basically overwrites these two values just in case people set them in a wrong way. But by default these two flags are indeed right for "one-time".  Like there is no chance that `self._experience` will keep growing iteration by iteration.

emailweixu: I suggest we only keep whole_replay_buffer_training and clear_replay_buffer. Make sure their combination is valid. And choose appropriate replayer based on its value to make the config easy.

hnyu: updated

emailweixu (approved):
Need to keep in mind that `save_last_exp` option does not work for async training.

PR #483, opened 2020-03-23
compute rollout_info_spec for every RLAlgorithm

hnyu:
If there are several replay buffers, then it would be better that each rl algorithm will have rollout_info_spec computed.

Also the current RLAlgorithm seems not to be a strict abstract class. If a function decorated with `abstractmethod` is not implemented, no error is reported. Do we want to strictly check it?

### DIFF
--- a/alf/algorithms/actor_critic_algorithm.py
+++ b/alf/algorithms/actor_critic_algorithm.py
@@ -106,7 +106,7 @@ class ActorCriticAlgorithm(OnPolicyAlgorithm):
             state=ActorCriticState(actor=actor_state),
             info=ActorCriticInfo(action_distribution=action_dist))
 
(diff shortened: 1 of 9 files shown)


emailweixu:
As we discussed, if we want to make the interface more consistent, we can change RLAlgorithm in the following way

```python
def __init__(...):
    self._original_rollout_step = self.rollout_step
    self.rollout_step = self._rollout_step

def _rollout_step(self, ...):
       policy_step = self._original_rollout_step(time_step, state)
        if self._rollout_info_spec is None:
            self._rollout_info_spec = dist_utils.extract_spec(policy_step.info)
        return policy_step
```

emailweixu (approved):

PR #491, opened 2020-03-26
fix summarize_action

hnyu:
Introduced by #472 . The bins for histogram_discrete needs to be on CPU. Also spec.maximum is np.array and after it's added with 1, its type is np.int64 and different from spec.minimum. Torch.arange doesn't support this mixture of types.

Remove alf.TensorSpec and alf.BoundedTensorSpec from external_configurables because they can already be accessed as "TensorSpec" and "BoundedTensorSpec" in gin files.

### DIFF
--- a/alf/utils/external_configurables.py
+++ b/alf/utils/external_configurables.py
@@ -41,10 +41,6 @@ gin.external_configurable(torch.tanh, 'torch.tanh')
 gin.external_configurable(torch.relu, 'torch.relu')
 gin.external_configurable(torch.nn.functional.elu, 'torch.nn.functional.elu')
 
-gin.external_configurable(alf.TensorSpec, 'alf.TensorSpec')
-
-gin.external_configurable(alf.BoundedTensorSpec, 'alf.BoundedTensorSpec')
-
 gin.external_configurable(torch.nn.functional.softsign,
                           'torch.nn.funtional.softsign')
 
--- a/alf/examples/diayn_pendulum.gin
+++ b/alf/examples/diayn_pendulum.gin
@@ -18,10 +18,10 @@ create_environment.env_name='Pendulum-v0'
 # algorithm config
 observation_spec=@get_observation_spec()
 action_spec=@get_action_spec()
-goal/alf.TensorSpec.shape=(%skill_feature_size,)
+goal/TensorSpec.shape=(%skill_feature_size,)
 
 hidden_size=64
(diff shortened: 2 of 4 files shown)


alf/utils/external_configurables.py
```diff
@@ -41,10 +41,6 @@
 gin.external_configurable(torch.relu, 'torch.relu')
 gin.external_configurable(torch.nn.functional.elu, 'torch.nn.functional.elu')
 
-gin.external_configurable(alf.TensorSpec, 'alf.TensorSpec')
-
-gin.external_configurable(alf.BoundedTensorSpec, 'alf.BoundedTensorSpec')
-
```
Haichao-Zhang, line 47: Why these lines are removed?
Test cases are broken when these are removed. For example, in ```diayn_pendulum.gin```, ```goal/alf.TensorSpec.shape=(%skill_feature_size,)``` is used.
```ppo_rnd_mrevenge.gin``` and other cases may have a similar issue.

hnyu: Just use TensorSpec without prefixing 'alf.' If keeping these lines, without the prefix there is an ambiguity error.

Haichao-Zhang: It seems the ambiguity error are due to the following ```gin.configurable ``` introduced in #439:
https://github.com/HorizonRobotics/alf/blob/f4ae6a4ea9db05739ff63f0ef7923409cd4d5565/alf/tensor_specs.py#L30-L31
https://github.com/HorizonRobotics/alf/blob/f4ae6a4ea9db05739ff63f0ef7923409cd4d5565/alf/tensor_specs.py#L168-L169
Why are they necessary? Can we simply remove them to resolve the ambiguity error you mentioned? 
In addition, I feel the original form (```alf.TensorSpec```) might be more clear in indicating it is from alf and also configuring them in a centralized place (```external_configurables.py```) seems to be a good idea.

hnyu: I think external_configurables.py suggests that all those configurations inside that file have to be done in that way because they are "external" (third-party code). It's a backup plan if you can't directly decorate the code with 'gin.configurable'. We have our own implementation of TensorSpec, and it makes little sense to treat them as external. With TF being removed, TensorSpec always comes from alf, and you can pretend that there is no other implementation of 'TensorSpec'.

emailweixu (approved):

PR #496, opened 2020-03-27
make train_state_spec and rollout_state_spec different

hnyu:
Currently if we want to enable use_rollout_state, train_state_spec has to be the same with rollout_state_spec, which is both confusing and inefficient (regarding replay buffer storage) if there are lots of additional rollout states in rollout_state_spec. So an aggressive way is to prune rollout state before putting it into the replay buffer.

### DIFF
--- a/alf/nest/nest.py
+++ b/alf/nest/nest.py
@@ -291,11 +291,16 @@ def prune_nest_like(nest, slim_nest, value_to_match=None):
     `value_to_match`, then the corresponding field of `nest` will also be updated
     to this value.
 
-    Note: If a node is a list or unnamedtuple, then we will truncate that
-    sequence by the first several elements. In this case, make sure you know what
-    you're doing.
+    Note: If a node is a list or unnamedtuple, then we require their lengths are
+    equal.
 
-    For examples, check out `nest_test.py`.
+    For example:
+        x = dict(a=1, b=2)
+        y = dict(a=[])
+        z = prune_nest_like(x, y) # z is dict(a=1)
+
+        y2 = dict(a=[], b=())
+        z2 = prune_nest_like(x, y2, value_to_match=()) # z2 is dict(a=1, b=())
 
     Args:
         nest (nest): a nested structure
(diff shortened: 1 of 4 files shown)


alf/nest/nest.py
```diff
@@ -282,3 +282,67 @@ def find_field(nest, name, ignore_empty=True):
...
+    Returns:
+        pruned_nest (nest): the pruned nest that has the same set of fields with
+            `slim_nest`.
+    """
+
+    def _prune(nest, slim_nest):
+        if is_nested(nest) or is_nested(slim_nest):
+            assert_same_type(nest, slim_nest)
+            if isinstance(nest, list) or is_unnamedtuple(nest):
+                assert len(nest) >= len(slim_nest), \
```
emailweixu: I feel it's dangerous to allow different length.

alf/nest/nest.py
```diff
@@ -282,3 +282,67 @@ def find_field(nest, name, ignore_empty=True):
...
+    every corrsponding node, we only keep the fields that're contained in
+    `slim_nest`. In addition, if a field of `slim_nest` contains a value of
+    `value_to_match`, then the corresponding field of `nest` will also be updated
+    to this value.
+
+    Note: If a node is a list or unnamedtuple, then we will truncate that
+    sequence by the first several elements. In this case, make sure you know what
+    you're doing.
+
+    For examples, check out `nest_test.py`.
```
emailweixu: It's better to give an explicit example here.

alf/nest/nest.py
```diff
@@ -282,3 +282,67 @@ def find_field(nest, name, ignore_empty=True):
...
+            else:
+                ret = {}
+
+                nest_fields_values = dict(extract_fields_from_nest(nest))
+                slim_nest_fields_and_values = dict(
+                    extract_fields_from_nest(slim_nest))
+
+                assert (set(nest_fields_values.keys())
+                        >= set(slim_nest_fields_and_values.keys())), \
+                            "The fields of nest must be a subset of slim_nest!"
```
emailweixu: This check can be done at line 336 to give more informative message. 
```python
if field not in nest_fields_value:
   error message
else:
   nest_value = nest_fields_values[field]
```

alf/nest/nest.py
```diff
@@ -282,3 +282,67 @@ def find_field(nest, name, ignore_empty=True):
...
+
+                for field in slim_nest_fields_and_values.keys():
```
emailweixu:
```python
for field, slim_nest_value in slim_nest_fieds_and_values.items():
```

alf/nest/nest.py
```diff
@@ -282,3 +282,66 @@ def find_field(nest, name, ignore_empty=True):
...
+    `slim_nest`. In addition, if a field of `slim_nest` contains a value of
+    `value_to_match`, then the corresponding field of `nest` will also be updated
+    to this value.
+
+    Note: If a node is a list or unnamedtuple, then we require their lengths are
+    equal.
+
+    For example:
+        x = dict(a=1, b=2)
+        y = dict(a=[])
```
emailweixu: This is not a good example because x.a is not nested, but y.a is nested. Line 230 will fail.

hnyu: Fixed

emailweixu (approved):

PR #498, opened 2020-03-27
move action clip wrapper to gym wrappers

hnyu:
The reason is to prevent `time_step.prev_action` being clipped. Also added missing gradient_clipping comments

### DIFF
--- a/alf/algorithms/sac_algorithm.py
+++ b/alf/algorithms/sac_algorithm.py
@@ -101,6 +101,7 @@ class SacAlgorithm(OffPolicyAlgorithm):
                  critic_optimizer=None,
                  alpha_optimizer=None,
                  gradient_clipping=None,
+                 clip_by_global_norm=False,
                  debug_summaries=False,
                  name="SacAlgorithm"):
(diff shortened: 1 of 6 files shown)


alf/algorithms/sac_algorithm.py
```diff
@@ -134,6 +135,9 @@ def __init__(self,
             critic_optimizer (torch.optim.optimizer): The optimizer for critic.
             alpha_optimizer (torch.optim.optimizer): The optimizer for alpha.
             gradient_clipping (float): Norm length to clip gradients.
+            clip_by_global_norm (bool): If True, use `tensor_utils.clip_by_global_norm`
+                to clip gradient. If False, use `tensor_utils.clip_by_norms` for
+                each grad.
```
emailweixu, line 140: I was thinking of having an optimizer wrapper class (say, alf.optimizer.Optimizer) to handle things like clipping. So we only need to config the optimizer, and can reduce the arguments to Algorithms.

emailweixu (approved):

PR #501, opened 2020-03-29
sphinx documentation on readthedocs

hnyu:
Setting up documentation hosted on [readthedocs.org](https://readthedocs.org/). One example is 
https://alf-fork.readthedocs.io/en/latest/ which is generated for my forked version of ALF.

General workflow:
1. First we use `sphinx-apidoc` locally to convert docstrings to the restructuredtext (rst) format. This format is like Markdown but more powerful than it. Generally an auto-generated rst file will have some placeholders reserved for packages, modules, and functions, and they will be replaced by the update-to-date docstrings during a build phase later. So in principle if no new files or code structure is generated, the content of auto-generated rst files will stay the same. These files are always in `alf/docs/api`. To avoid generating documentations for testing cases, now I've moved all testing files to `alf/tests` and exclude it from the conversion.

2. I've added a precommit hook in `alf/tools/documentation/sphinx.hook` to run the `sphinx-apidoc` command every time before a commit is made.  In the case of code structure changes, it will update the auto-generated rst files and we need to re-commit the updated ones (like code format checking). Otherwise this command does nothing.

3. Besides the auto-generated rst files that have docstring placeholders, users can also manually write rst files just like Markdown files, and link them with each other. The entry point of the documentation is `alf/docs/index.rst`.

4. Unfortunately it seems that we need to convert most of existing Markdown files to rst format, if we want to put them on the documentation site. I've converted two of them: `async_training.rst` and `knowledge_base.rst`. You can help with other conversions.

5. Sphinx then builds the rst files into html files. The build will actually try to import all the python files and get the docstring info. For some files that still use TF or generate errors, the built html content will be empty. If you want to test the build locally, in `alf/docs` do `make html`. Then you will be able to view the generated html files locally.

6. Fortunately, our repo will be linked to readthedocs under the url https://alf.readthedocs.io/en/latest/ after this PR. Every time if there is some new commit to the pytorch branch (which branch can be set), readthedocs will be notified and re-build the rst files into html files and display the website at the url.

7. Finally, we can customize the documentation page appearance using the Sphinx configuration file `alf/docs/conf.py`. And we can customize how readthedocs builds and hosts the website using its configuration file `.readthedocs.yml`.

8. There are still lots of format errors in our docstrings. And we need take time to correct them.

9. For a normal ALF user or contributor, basically he needs to worry about nothing about this documentation and can submit PRs like before.

This PR just establishes the pipeline of documentation, lots of remaining work such as converting MD files to rst files and update our docstrings are to be done by all of us.

Let me know if there are questions or suggestions.


alf/algorithms/actor_critic_algorithm.py
```diff
@@ -45,8 +45,7 @@ def __init__(self,
                  optimizer=None,
                  debug_summaries=False,
                  name="ActorCriticAlgorithm"):
-        """Create an ActorCriticAlgorithm.
-
+        """
```
hnyu, line 48: Because sphinx will combine the docstrings of the Class declaration and init function, we need to remove duplicate sentences accordingly, for all our algorithms.

le-horizon:
Did you accidentally close this PR @hnyu ?

hnyu:
> Did you accidentally close this PR @hnyu ?

This one is closed because it's on my forked branch and CI won't run. I opened a new one #502 .

PR #502, opened 2020-03-29
sphinx documentation on readthedocs

hnyu:
Setting up documentation hosted on [readthedocs](https://readthedocs.org/) (RTD). 

General workflow:

0. Use `sphinx-quickstart` in `alf/docs` to initialize some files (e.g., Makefile, index.rst, etc) and do some basic setup. For a tutorial on this step, refer to https://medium.com/@richdayandnight/a-simple-tutorial-on-how-to-document-your-python-project-using-sphinx-and-rinohtype-177c22a15b5b 

L1. For a local build, we use the command 
```
sphinx-apidoc -f -o api ../alf `find .. -name '*_test.py'` ../alf/examples ../alf/bin --templatedir _templates 
```
 to convert docstrings to the restructuredtext (rst) format. This format is like Markdown but more powerful than it. The api files are always in `alf/docs/api`. I've excluded test files, examples files, and bin files during this process.

L2. Sphinx then builds the rst files into html files by `make html`. The build will actually try to import all the python files and get the docstring and code structure info. For some files that still use TF or throw errors, the built html content will be empty. 

3. Fortunately, our repo will be linked to RTD under the url https://alf.readthedocs.io/en/latest/ after this PR. Every time if there is some new commit to the pytorch branch (we can set which branch to use on RTD under "Advanced settings"), RTD will be notified and re-build the html files and display the website at the url. This will replace the two local steps L1-L2 above.

5. To configure RTD,  we use the Sphinx configuration file `alf/docs/conf.py`. And we can customize how RTD builds and hosts the website using its configuration file `.readthedocs.yml`. For the build, `alf/docs/requirements.txt` needs to be supplied and large packages should be avoided installation (make sure to check the building log on RTD to see if there is any import error!).

4. Users can also manually write rst files just like Markdown files, and link them with each other. The entry point of the documentation is `alf/docs/index.rst` (set by `conf.py master_doc`). Unfortunately it seems that we need to convert most of existing Markdown files to rst format, if we want to put them on the documentation site. I've converted two of them: `async_training.rst` and `knowledge_base.rst`. You can help with other conversions. Github can also render most rst syntax correctly.

6. There are still lots of format errors in our docstrings. And we need take time to correct them. We can use either RST-style docstrings or google-style docstrings to document parameters. But note that Github markdown syntax in docstrings will not be rendered correctly. [RST cheatsheet](https://thomas-cokelaer.info/tutorials/sphinx/rest_syntax.html).

7. **For a normal ALF user or contributor, basically he needs to worry about nothing about this documentation and can submit PRs like before.**

This PR just establishes the pipeline of documentation, lots of remaining work such as converting MD files to rst files and update our docstrings are to be done by all of us.

Let me know if there are questions or suggestions.


setup.py
```diff
@@ -28,7 +28,9 @@
         'pathos == 0.2.4',
         'pillow',
         'psutil',
-        'pybullet == 2.5.0',
+        #'pybullet == 2.5.0',
```
hnyu: For some reason, if readthedocs tries to install pybullet, there will be an error. And because pybullet is huge and rarely used by most training, here I change it to be installed manually by users.

le-horizon (commented):
Nice doc site generated, Haonan!

Some high level comments:

1) Looks like you don't need to move tests?
https://github.com/sphinx-doc/sphinx/issues/944

2) Since rst files are auto generated, shouldn't we avoid adding them into the alf repo unless we make real changes to them?  Otherwise, repo will look bloated?  We can add them to .gitignore file to ignore them locally so git status won't show them?  Maybe don't need to generate docs in precommit, because hopefully, most of the time, developers should not need to change any of the generated rst files?

3) maybe add some of the comments in your PR message into contributing.md?  Adding a section called "Build documentation"?

Thanks,
Le

hnyu:
> Nice doc site generated, Haonan!
> 
> Some high level comments:
> 
> 1. Looks like you don't need to move tests?
>    [sphinx-doc/sphinx#944](https://github.com/sphinx-doc/sphinx/issues/944)
> 2. Since rst files are auto generated, shouldn't we avoid adding them into the alf repo unless we make real changes to them?  Otherwise, repo will look bloated?  We can add them to .gitignore file to ignore them locally so git status won't show them?  Maybe don't need to generate docs in precommit, because hopefully, most of the time, developers should not need to change any of the generated rst files?
> 3. maybe add some of the comments in your PR message into contributing.md?  Adding a section called "Build documentation"?
> 
> Thanks,
> Le

1. Yes you can exclude certain files using regex. Our test files are scattered in multi-level directories and I think it's better to gather them in one place in general. I excluded `alf/tests` when running `sphinx-apidoc`.

2. The auto-generated rst files should actually be checked in. They won't be changed unless code structured is changed because there are placeholders reserved for docstrings. PyTorch also has them checked in. Right now, a normal developer doesn't indeed need to worry about this. Note that readthedocs won't generate rst files; it will only take rst files and build them into htmls. rst files have to be hosted on github. You can read again my comments above about the workflow.

3. Yeah I will do that, but in a different title like "How to set up Sphinx for documentation" for future projects. Again even though my comment looks like extremely long and complicated, I was just documenting how the pipeline was established. That's not the workflow a normal developer has to go through. There is basically nothing needed to be done for a normal user.

le-horizon:
> > Nice doc site generated, Haonan!
> > Some high level comments:
> > 
> > 1. Looks like you don't need to move tests?
> >    [sphinx-doc/sphinx#944](https://github.com/sphinx-doc/sphinx/issues/944)
> > 2. Since rst files are auto generated, shouldn't we avoid adding them into the alf repo unless we make real changes to them?  Otherwise, repo will look bloated?  We can add them to .gitignore file to ignore them locally so git status won't show them?  Maybe don't need to generate docs in precommit, because hopefully, most of the time, developers should not need to change any of the generated rst files?
> > 3. maybe add some of the comments in your PR message into contributing.md?  Adding a section called "Build documentation"?
> > 
> > Thanks,
> > Le
> 
> 1. Yes you can exclude certain files using regex. Our test files are scattered in multi-level directories and I think it's better to gather them in one place in general. I excluded `alf/tests` when running `sphinx-apidoc`.

Sounds good, just making sure sphinx is not the main reason for moving all tests.

> 2. The auto-generated rst files should actually be checked in. They won't be changed unless code structured is changed because there are placeholders reserved for docstrings. PyTorch also has them checked in. Right now, a normal developer doesn't indeed need to worry about this. Note that readthedocs won't generate rst files; it will only take rst files and build them into htmls. rst files have to be hosted on github. You can read again my comments above about the workflow.

In general, it's best not to check in generated files, unless they are testing templates.  Given that it's standard practice to check rst files in, and given that they are all in docs directory, it's probably Ok to have them checked-in.

Thanks for explaining.

> 3. Yeah I will do that, but in a different title like "How to set up Sphinx for documentation" for future projects. Again even though my comment looks like extremely long and complicated, I was just documenting how the pipeline was established. That's not the workflow a normal developer has to go through. There is basically nothing needed to be done for a normal user.

emailweixu:
> 1. Yes you can exclude certain files using regex. Our test files are scattered in multi-level directories and I think it's better to gather them in one place in general. I excluded `alf/tests` when running `sphinx-apidoc`.

Is it possible to exclude by pattern "*_test.py"? It's more convenient to have test located in the same directory of the file to be tested. You don't need to go to a separate directory to add or find the test.

docs/requirements.txt
```diff
@@ -0,0 +1,6 @@
+atari_py==0.1.7
```
Haichao-Zhang: Do we need to include these into docs? If not, we might be able to exclude it as well.

hnyu: I can remove atari_py. The other libs are definitely needed. This txt file will be read by readthedocs when building the project. When it tries to import our python files, they are needed. See `alf/.readthedocs.yml` when configuring the installation.

emailweixu (dismissed):

hnyu:
> 2. Since rst files are auto generated, shouldn't we avoid adding them into the alf repo unless we make real changes to them?  Otherwise, repo will look bloated?  We can add them to .gitignore file to ignore them locally so git status won't show them?  Maybe don't need to generate docs in precommit, because hopefully, most of the time, developers should not need to change any of the generated rst files?

After some search, I think it might be possible to let readthedocs generate the rst api files. 

https://github.com/readthedocs/readthedocs.org/issues/1139

I'm looking into this so that we don't have to check in the files and won't need the hook.

hnyu:
Finally get a better version to work. Now there is no generated rst file checked in. Also removed the precommit hook. I had to carefully edit `docs/requirements.txt` for building on readthedocs because a full torch-gpu version is too big to install and will fail.

docs/notes/sphinx_doc.rst
```diff
@@ -0,0 +1,44 @@
+How to preview this website locally
+===================================
+
+
+This website is automatically generated using `Sphinx
+<https://pythonhosted.org/an_example_pypi_project/sphinx.html>`_ on
+`<readthedocs.org>`_. It will get updated by RTD once a commit is pushed to ALF.
+So normally a ALF user/developer doesn't have to worry about keeping this
```
emailweixu: a ALF => an ALF

docs/notes/sphinx_doc.rst
```diff
@@ -0,0 +1,44 @@
...
+website update-to-update every time some code change is made. For a complete
```
emailweixu: update-to-update => up-to-date

docs/notes/sphinx_doc.rst
```diff
@@ -0,0 +1,44 @@
...
+    sphinx-apidoc -f -o api ../alf `find .. -name '*_test.py'` ../alf/examples --templatedir _templates
+    make html
+    cd _build/html
+
+Then open the file :code:`index.html` with a browser.
+
+2. VSCode users:
+
+Install the extension `reStructuredText` and open your vscode settings file
+:code:`settings.json` and add the following configuration:
```
emailweixu, line 30: Do we need to run any sphinx commands?

hnyu: No

docs/notes/sphinx_doc.rst
```diff
@@ -5,9 +5,9 @@ How to preview this website locally
 This website is automatically generated using `Sphinx
 <https://pythonhosted.org/an_example_pypi_project/sphinx.html>`_ on
 `<readthedocs.org>`_. It will get updated by RTD once a commit is pushed to ALF.
-So normally a ALF user/developer doesn't have to worry about keeping this
-website update-to-update every time some code change is made. For a complete
-setup process of using RTD, please refer to this `PR
+So normally an ALF user/developer doesn't have to worry about keeping this
+website up-to-update every time some code change is made. For a complete setup
```
emailweixu: up-to-update => up-to-date

CONTRIBUTING.md
```diff
@@ -1,64 +0,0 @@
-
```
le-horizon, line 1: instead of deleting and adding as new file, could you make it a git mv, so that the incremental/actual changes you made would show up correctly?

docs/async_training.md
```diff
@@ -1,116 +0,0 @@
-# Asynchronous training guide
```
le-horizon, line 1: same here

docs/knowledge_base.md
```diff
@@ -1,127 +0,0 @@
-# Knowlege Base
```
le-horizon, line 1: same here

le-horizon (commented):
Nice improvements Haonan, almost good!  3 small nits on file moving, should be easy to fix.

CONTRIBUTING.md
hnyu: That's what I did. Somehow it won't show the changes side by side. And the changes are not actually incremental. The rst syntax is quite different from md's. So it will render many changed lines.

emailweixu (approved):

PR #504, opened 2020-03-30
update dockerfile for CI

hnyu:
In order for #502 to have CI running, we need to first change the dockerfile.


Haichao-Zhang (approved):

Haichao-Zhang:
An updated docker image might need to be uploaded docker hub?

hnyu:
> An updated docker image might need to be uploaded docker hub?

I see. I will update that soon.

PR #507, opened 2020-03-31
Pr environment test

runjerry:
- add environment tests to ci
- make a copy after np.transpose in ImageChannelFirst wrapper, for addressing the slow down in subprocess pickle

### DIFF
--- a/alf/environments/suite_dmlab_test.py
+++ b/alf/environments/suite_dmlab_test.py
@@ -14,6 +14,7 @@
 
 from absl.testing import parameterized
 import functools
+import gin
 
 import alf
 from alf.environments import gym_wrappers, suite_dmlab, torch_environment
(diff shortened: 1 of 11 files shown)


alf/environments/suite_dmlab_test.py
```diff
@@ -26,8 +25,6 @@ def setUp(self):
         super().setUp()
         if not suite_dmlab.is_available():
             self.skipTest('suite_dmlab is not available.')
-        else:
```
emailweixu: Right now this test is skipped because dmlab is not installed. I suspect that removing this will make this test fail. I think a better way is removing gin.bind_parameter('torch.optim.Adam.params', [{'params': []}]) in alf/utils/external_configurable.py. We can add a class alf.optimizers.Adam derived from torch.optim.Adam with default value for `params`

runjerry: Add them back after Haichao's PR [508](https://github.com/HorizonRobotics/alf/pull/508)

le-horizon (approved):
Nice!

PR #509, opened 2020-04-01
transform_nest for nested namedtuple update

hnyu:
Modify the previous `transform_observation` to a general interface `transform_nest` so that we can use it to update deeply nested namedtuple. Also simplify the interfaces of `transform_observation` in the wrappers.

### DIFF
--- a/alf/nest/utils.py
+++ b/alf/nest/utils.py
@@ -153,3 +153,53 @@ def get_outer_rank(tensors, specs):
                 for r in outer_ranks]), ("Tensors have different "
                                          "outer_ranks %s" % outer_ranks)
     return outer_rank
+
+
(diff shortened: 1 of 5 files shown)


alf/nest/utils.py
```diff
@@ -153,3 +153,54 @@ def _get_outer_rank(tensor, spec):
                 for r in outer_ranks]), ("Tensors have different "
                                          "outer_ranks %s" % outer_ranks)
     return outer_rank
+
+
+def transform_nest(nested, field, func):
```
emailweixu, line 158: Should have unittest for this.

hnyu: added

emailweixu (approved):

PR #510, opened 2020-04-02
extract RingBuffer from ReplayBuffer for future async extensions and …

le-horizon:
…use to implement async Queue

Hey Wei and Haonan, I plan to reuse the ring buffer from replayer to implement queue.
Let me know how you like this plan.

Thanks,
Le


hnyu:
Why it's necessary to invent a new concept as RingBuffer? Even if you want to use ReplayBuffer to implement queue, you can just derive Queue from it.

alf/experience_replayers/replay_buffer.py
```diff
@@ -57,13 +57,13 @@ def __init__(self,
                  num_environments,
                  max_length=1024,
                  device="cpu",
-                 name="ReplayBuffer"):
-        """Create a ReplayBuffer.
+                 name="RingBuffer"):
+        """Create a RingBuffer.
```
hnyu: Remove "create a ..." template in `__init__`. We should do this for all init functions sooner or later. Sphinx will combine the docstrings of class and init into one. So this template is literally uninformative.

le-horizon:
> Why it's necessary to separate the two? Even if you want to use replay buffer to implement queue, you can just derive Queue from it.

Good point.  I just find it very hacky to let Queue inherit from ReplayBuffer.  It's probably best to have Queue and ReplayBuffer inherit from RingBuffer.

alf/experience_replayers/replay_buffer.py
le-horizon: Done.  Will remember to do this for others.

le-horizon:
Btw, I plan to create a directory, alf/modules, to put ring_buffer.py.  There are probably other data structures (nn.Modules) that we want to put there.  Does that make sense?

hnyu:
> Btw, I plan to create a directory, alf/modules, to put ring_buffer.py. There are probably other data structures (nn.Modules) that we want to put there. Does that make sense?

Currently different modules already have their own directories, e.g., networks, algorithms, metrics, etc. i don't think it makes sense to create a general one "modules"

hnyu:
> > Why it's necessary to separate the two? Even if you want to use replay buffer to implement queue, you can just derive Queue from it.
> 
> Good point. I just find it very hacky to let Queue inherit from ReplayBuffer. It's probably best to have Queue and ReplayBuffer inherit from RingBuffer.

Right now ReplayBuffer is basically identical to RingBuffer. There is no difference except their names.

le-horizon:
I expect differences to emerge as I add more capability into RingBuffer, some of which won't be needed in ReplayBuffer, e.g. block on enqueue, block on dequeue, locking mechanism etc..

As for directory, maybe utils then?

hnyu:
> I expect differences to emerge as I add more capability into RingBuffer, some of which won't be needed in ReplayBuffer, e.g. block on enqueue, block on dequeue etc..
> 
> As for directory, maybe utils then?

Then RingBuffer shouldn't be the base class for ReplayBuffer if it has more capacities? It should be the other way around. And you can put all queue-related stuff into `alf/queues`.

le-horizon:
ReplayBuffer is just a simple wrapper of RingBuffer in the current form.  I can also put a RingBuffer object as a member of ReplayBuffer, if you prefer that.
I plan to make RingBuffer multiprocessing safe, so queue can directly use it.

hnyu:
> ReplayBuffer is just a simple wrapper of RingBuffer in the current form. I can also put a RingBuffer object as a member of ReplayBuffer, if you prefer that.
> I plan to make RingBuffer multiprocessing safe, so queue can directly use it.

Yeah I mean if RingBuffer really has more capabilities, then there is no need to force ReplayBuffer to derive from it. Even if you make it as a member of ReplayBuffer, other capabilities won't be needed. I still think `RingBuffer(ReplayBuffer)` should be the correct way to do, according to your requirements. Or they two can share a minimal base class with minimal capabilities.

emailweixu:
I agree with Haonan that it doesn't seem it's necessary to derive ReplayBuffer from RingBuffer from this PR. They are the same thing. If you need to have a different RingBuffer, perhaps it makes more sense to make a PR until then.

le-horizon:
I'm glad we are talking about this early on.

In my view, RingBuffer is an underlying data structure to be used by different applications.
ReplayBuffer is at a much higher level than RingBuffer.  If we have utils/RingBuffer depend on experience_replayer/ReplayBuffer, then it's easy to cause circular dependencies like this: utils/RingBuffer -> exp/ReplayBuffer -> utils/something.

I'll send a more complete PR.

My current plan:
ReplayBuffer (some special logic on sampling exp) -> RingBuffer
Queue -> RingBuffer (need to add logic on dequeue, locking etc.)

I didn't say RingBuffer will be more capable than ReplayBuffer, just that it will be more capable than it is now, due to requirements from Queue, but ReplayBuffer will also have its special logic.

ReplayBuffer will also benefit from this change, because it will be multiprocessing safe if we add locking mechanism to RingBuffer.

Does that sound good to you guys?

hnyu:
> I'm glad we are talking about this early on.
> 
> In my view, RingBuffer is an underlying data structure to be used by different applications.
> ReplayBuffer is at a much higher level than RingBuffer. If we have utils/RingBuffer depend on experience_replayer/ReplayBuffer, then it's easy to cause circular dependencies like this: utils/RingBuffer -> exp/ReplayBuffer -> utils/something.
> 
> I'll send a more complete PR.
> 
> My current plan:
> ReplayBuffer (some special logic on sampling exp) -> RingBuffer
> Queue -> RingBuffer (need to add logic on dequeue, locking etc.)
> 
> I didn't say RingBuffer will be more capable than ReplayBuffer, just that it will be more capable than it is now, due to requirements from Queue, but ReplayBuffer will also have its special logic.
> 
> ReplayBuffer will also benefit from this change, because it will be multiprocessing safe if we add locking mechanism to RingBuffer.
> 
> Does that sound good to you guys?

I guess the bottom line is, if we want to separate RingBuffer and ReplayBuffer, then there should be some meaningful difference between the two. We can wait for your complete PR.

le-horizon:
Hey @emailweixu and @hnyu I've added dequeue to ringbuffer, and separated out functions that are specific to replaybuffer.  Please take a look again.

I plan to add multiprocessing safety to ringbuffer in the next change, so queue would be straightforward to add.

alf/experience_replayers/replay_buffer.py
```diff
@@ -181,37 +249,35 @@ def get_batch(self, batch_size, batch_length):
...
         max_size = self._current_size.max()
         assert size == max_size, (
-            "Not all environment have the same size. min_size: %s "
+            "Not all environments have the same size. min_size: %s "
             "max_size: %s" % (size, max_size))
 
+        # NOTE: this is not the proper way to gather all from a ring
+        # buffer whose data can start from the middle, so this is limited
+        # to the case where clear() is the only way to remove data from
+        # the buffer.
```
emailweixu, line 285: Perhaps you can assert _current_pos is 0?

alf/experience_replayers/replay_buffer.py
```diff
@@ -134,6 +144,63 @@ def add_batch(self, batch, env_ids=None):
...
+            current_size = self._current_size[env_ids]
+            min_size = current_size.min()
+            assert min_size >= 1, (
+                "Not all environments have enough data. The smallest data "
+                "size is: %s Try storing more data before calling dequeue" %
+                min_size)
+            pos = self._current_pos[env_ids] - current_size
+            pos = pos.reshape(-1, 1)  # [B, 1]
+            pos = pos % self._max_length
+            env_ids = env_ids.reshape(-1, 1).repeat(1, 1)  # [B, T]
```
emailweixu: why repeat(1, 1)?

alf/experience_replayers/replay_buffer.py
```diff
@@ -134,6 +144,63 @@ def add_batch(self, batch, env_ids=None):
             self._current_size[env_ids] = torch.min(
                 current_size + 1, torch.tensor(self._max_length))
 
+    def dequeue(self, env_ids=None):
+        """Return earliest batch and remove it from buffer."""
```
emailweixu: comment argument and return value (shape)

alf/experience_replayers/replay_buffer.py
emailweixu: where it T from? Is T equal to 1? Since there is only one step, we should not have T.

emailweixu (changes requested):

alf/experience_replayers/replay_buffer.py
le-horizon: Done.

alf/experience_replayers/replay_buffer.py
le-horizon: Good point.  Removed.  T==1

le-horizon: Good point.

alf/experience_replayers/replay_buffer.py
le-horizon: current_pos is the ending position of the data in the buffer, not the starting position, hence, current_pos == current_size when size < max_length.

le-horizon (commented):
Done all, PTAL.

Thanks,
Le

le-horizon:
Are there more comments guys?

alf/experience_replayers/replay_buffer.py
emailweixu: Right. That means we should assert current_size == current_pos?

alf/experience_replayers/replay_buffer.py
```diff
@@ -134,6 +144,72 @@ def add_batch(self, batch, env_ids=None):
...
+            assert len(env_ids.shape) == 1, "env_ids should be an 1D tensor"
+
+            current_size = self._current_size[env_ids]
+            min_size = current_size.min()
+            assert min_size >= 1, (
+                "Not all environments have enough data. The smallest data "
+                "size is: %s Try storing more data before calling dequeue" %
+                min_size)
+            pos = self._current_pos[env_ids] - current_size
+            pos = pos.reshape(-1, 1)  # [B, 1]
```
emailweixu: No need to do reshape at line 169, line 171, and line 173

le-horizon:
Done both.  PTAL?

Thanks,
Le

alf/experience_replayers/replay_buffer.py
```diff
@@ -166,18 +166,15 @@ def dequeue(self, env_ids=None):
                 "size is: %s Try storing more data before calling dequeue" %
                 min_size)
             pos = self._current_pos[env_ids] - current_size
-            pos = pos.reshape(-1, 1)  # [B, 1]
             pos = pos % self._max_length
-            env_ids = env_ids.reshape(-1, 1)  # [B, 1]
-            indices = env_ids * self._max_length + pos  # [B, 1]
-            indices = indices.reshape(-1)  # [B * 1]
+            indices = env_ids * self._max_length + pos  # shape [B*1]
```
emailweixu, line 170: [B*1] seems weird

alf/experience_replayers/replay_buffer.py
```diff
@@ -166,18 +166,15 @@ def dequeue(self, env_ids=None):
...
 
             batch_size = env_ids.shape[0]
             result = alf.nest.map_structure(
                 lambda buffer: buffer[indices].reshape(
                     batch_size, 1, *buffer.shape[1:]), self._flattened_buffer)
 
-            self._current_size[env_ids] = (current_size - 1).reshape(-1, 1)
+            self._current_size[env_ids] = (current_size - 1)
```
emailweixu, line 177: (current_size - 1) => current_size

emailweixu (approved):

le-horizon:
Good points Wei.  I'll fix these in the next bigger PR on atomic operations.

PR #511, opened 2020-04-02
Fix recording

emailweixu:
The cause of the bug is that social robot needs to be accessed from the thread environment. Previous code side-step the thread environment by directly accessing env.gym.

Also fixed a bug of checkpoint loading when the checkpoint is saved from cuda device but loaded using cpu device.

Issue #512


le-horizon (approved):
LG.  Nice fixes.

PR #514, opened 2020-04-04
gin configurable target entropy

hnyu:
Our current 0.01 for `calc_default_target_entropy` is too small. A number like 0.1 is better. So I made this value configurable. Also improve `sac_bipedal_walker.gin` so the wall-clock time for convergence is much (5 times) shorter now. 

Additionally make the assertion error in `nest.py` more friendly.


alf/algorithms/sac_algorithm.py
```diff
@@ -59,28 +61,29 @@
 
 @gin.configurable
 class SacAlgorithm(OffPolicyAlgorithm):
-    """Soft Actor Critic
+    """Soft Actor Critic algortihm, described in:
```
Haichao-Zhang: algortihm -> algorithm

alf/algorithms/sarsa_algorithm.py
```diff
@@ -196,13 +203,19 @@ def __init__(self,
...
-                            map(dist_utils.calc_default_target_entropy,
-                                flat_action_spec)))
+                        list(map(target_entropy, flat_action_spec)))
+                    logging.info(
+                        "Target entropy is calculated for {}: {}.".format(
+                            self.name, target_entropy))
+                else:
+                    logging.info(
+                        "User-supplied target entropy for {}: {}".format(
+                            self.name, target_entropy))
```
Haichao-Zhang: This part seems to be shared with SAC and potentially other future algorithms with target entropy. It might be better to put them into a single shared function?

alf/examples/sac_bipedal_walker.gin
```diff
@@ -46,5 +44,5 @@ TrainerConfig.num_checkpoints=5
 TrainerConfig.evaluate=False
 TrainerConfig.debug_summaries=True
 TrainerConfig.summarize_grads_and_vars=True
-TrainerConfig.summary_interval=1
-TrainerConfig.replay_buffer_length=1000000
+TrainerConfig.summary_interval=100
+TrainerConfig.replay_buffer_length=100000
```
Haichao-Zhang, line 48: Since now the parameters are changed, it is better to update the png curves to maintain the correspondence

Haichao-Zhang (changes requested):

emailweixu (approved):

Haichao-Zhang (approved):

PR #518, opened 2020-04-06
Temporally independent train_step

emailweixu:
For some algorithms, e.g. Sac, Sarsa, Ddpg, if RNN is not used, train_step() along a sequence can be independently evaluated. Previously, train_step() is called sequentially along the trajectory. In this PR, we add an option to calculate them with a single train_step() call, which can speed up the computation. With this PR, the speed of training using MultiStepTDLoss with SarsaAlgorithm is same as that of OneStepTDLoss. Without this PR, it's 30% slower.

With this option, SacAlgorithm and DdpgAlgorithm can also be faster.

### DIFF
--- a/alf/algorithms/config.py
+++ b/alf/algorithms/config.py
@@ -61,14 +61,16 @@ class TrainerConfig(object):
                 for the definition of FrameSkip.
             unroll_length (int):  number of time steps each environment proceeds per
                 iteration. The total number of time steps from all environments per
-                iteration can be computed as: ``num_envs * env_batch_size``
-                * `unroll_length`.
+                iteration can be computed as: ``num_envs * env_batch_size * unroll_length``.
             use_rollout_state (bool): Include the RNN state for the experiences
                 used for off-policy training
             temporally_independent_train_step (bool): If True, the ``train_step``
                 is called with all the experiences in one batch instead of being
                 called sequentially with ``mini_batch_length`` batches. Only used
-                by ``OffPolicyAlgorithm``.
+                by ``OffPolicyAlgorithm``. In general, this option can only be
+                used if the algorithm has no state. For Algorithm with state (e.g.
+                ``SarsaAlgorithm`` not using RNN), if there is no need to
+                recompute state at train_step, this option can also be used.
             num_checkpoints (int): how many checkpoints to save for the training
             evaluate (bool): A bool to evaluate when training
             eval_interval (int): evaluate every so many iteration
--- a/alf/algorithms/config.py
+++ b/alf/algorithms/config.py
@@ -17,17 +17,7 @@ import gin
 
 @gin.configurable
 class TrainerConfig(object):
-    """TrainerConfig
-
-    Note: This is a mixture collection configuration for all training setups,
-    not all parameter is operative and its not necessary to config it.
-
-    1. `num_steps_per_iter` is only for on_policy_trainer.
-
-    2. `initial_collect_steps`, `num_updates_per_train_step`, `mini_batch_length`,
-    `mini_batch_size`, `clear_replay_buffer`, `num_envs` are used by sync_off_policy_trainer and
-    async_off_policy_trainer.
-    """
+    """Configuration for training."""
 
     def __init__(self,
                  root_dir,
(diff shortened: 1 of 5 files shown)


alf/algorithms/config.py
```diff
@@ -58,24 +48,27 @@ def __init__(self,
...
             num_env_steps (int): number of environment steps (ignored if 0). The
-                total number of FRAMES will be (`num_env_steps`*`frame_skip`) for
+                total number of FRAMES will be (``num_env_steps*frame_skip``) for
                 calculating sample efficiency. See alf/environments/wrappers.py
                 for the definition of FrameSkip.
             unroll_length (int):  number of time steps each environment proceeds per
                 iteration. The total number of time steps from all environments per
-                iteration can be computed as: `num_envs` * `env_batch_size`
+                iteration can be computed as: ``num_envs * env_batch_size``
                 * `unroll_length`.
```
hnyu: double quotes

alf/algorithms/config.py
```diff
@@ -86,32 +79,34 @@ def __init__(self,
             num_eval_episodes (int) : number of episodes for one evaluation
             summary_interval (int): write summary every so many training steps
             update_counter_every_mini_batch (bool): whether to update counter
-                for every mini batch. The `summary_interval` is based on this
+                for every mini batch. The ``summary_interval`` is based on this
                 counter. Typically, this should be False. Set to True if you
                 want to have summary for every mini batch for the purpose of
-                debugging.
+                debugging. Only used by ``OffPolicyAlgorithm``.
```
hnyu, line 87: Instead of adding this sentence to every argument that is only used by OffPolicyAlgorithm, should we add a short prefix/suffix to the argument name to indicate this?

alf/algorithms/off_policy_algorithm.py
```diff
@@ -229,6 +229,39 @@ def _update(self, experience, weight):
...
+        return info
+
+    def _collect_train_info_parallelly(self, experience):
+        batch_size = experience.step_type.shape[1]
+        length = experience.step_type.shape[0]
+
+        exp = alf.nest.map_structure(lambda x: x.reshape(-1, *x.shape[2:]),
+                                     experience)
+
+        if self._use_rollout_state:
```
hnyu, line 241: Based on the code here,  I think we should add comments clarifying that `temporally_independent_train_step` can also be set to True if we don't recompute the state but just take the historical states from the replay buffer.

hnyu (changes requested):

alf/algorithms/off_policy_algorithm.py
```diff
@@ -229,6 +229,39 @@ def _update(self, experience, weight):
...
+            self._train_info_spec = dist_utils.extract_spec(policy_step.info)
+        info = dist_utils.distributions_to_params(policy_step.info)
+        info = alf.nest.map_structure(
+            lambda x: x.reshape(length, batch_size, *x.shape[1:]), info)
+        info = dist_utils.params_to_distributions(info, self.train_info_spec)
+        return info
+
+    def _update(self, experience, weight):
+        length = experience.step_type.shape[0]
+        if self._config.temporally_independent_train_step or length == 1:
```
hnyu, line 260: Do you think it's possible to automatically decide which one to choose depending on other arguments?

alf/algorithms/config.py
emailweixu: I think it's okay. Not that many options are only used by OffPolicyAlgorithm

alf/algorithms/off_policy_algorithm.py
emailweixu: Noted in TrainerConfig

alf/algorithms/off_policy_algorithm.py
emailweixu: I thought about it. But feel there is no reliable of doing this. For example SarsaAlgorithm has non-empty train_state even if RNN is not used, but we can still use this option. But in general, this option should not be used for algorithms with states.

alf/algorithms/config.py
emailweixu: updated.

hnyu (approved):

PR #520, opened 2020-04-07
remove some old tf files

hnyu:
These files will no longer be needed or have been converted.


emailweixu (commented):
action_encoder.py is used by merlin_algorithm

hnyu:
> action_encoder.py is used by merlin_algorithm

Can we just use EmbeddingPreprocessor to achieve a similar purpose?

le-horizon (dismissed):
Looks good to me.  Please address other reviewers' comments before push.

Thanks,
Le

emailweixu:
> > action_encoder.py is used by merlin_algorithm
> 
> Can we just use EmbeddingPreprocessor to achieve a similar purpose?

SimpleActionEncoder is parameterless can can handle nested action. It is easier to use.

emailweixu (approved):

PR #521, opened 2020-04-07
Update on last layer size and activation for EncodingNetwork

Haichao-Zhang:
This PR makes the change to ensure ```last_layer_size``` and ```last_activation``` are specified jointly to avoid potential errors due to the mis-understanding on its default behavior.


emailweixu (dismissed):
LGTM.

alf/networks/encoding_networks.py
```diff
@@ -281,39 +282,47 @@ def __init__(self,
         from the other layers.
```
hnyu: Remove "create an ..." template for all class init functions

Haichao-Zhang: Updated according to the comment.

hnyu (approved):

PR #522, opened 2020-04-07
Option to limit episode length for play

emailweixu:


alf/trainers/policy_trainer.py
```diff
@@ -331,12 +334,13 @@ def play(root_dir,
...
         episode_reward += float(time_step.reward)
 
-        if time_step.is_last():
+        if time_step.is_last() or episode_length >= max_episode_length > 0:
             logging.info("episode_length=%s episode_reward=%s" %
                          (episode_length, episode_reward))
             episode_reward = 0.
             episode_length = 0.
             episodes += 1
+            time_step = env.reset()
```
hnyu, line 343: seems no need to return time_step

alf/trainers/policy_trainer.py
```diff
@@ -267,6 +267,7 @@ def play(root_dir,
          checkpoint_step="latest",
          epsilon_greedy=0.1,
          num_episodes=10,
+         max_episode_length=0,
          sleep_time_per_step=0.01,
          record_file=None):
     """Play using the latest checkpoint under `train_dir`.
```
hnyu, line 273: docstring format needs to be updated

alf/trainers/policy_trainer.py
emailweixu: in case it ends because of max_episode_length, the previous time_step is not correct.

hnyu: I mean reset() should be called but time_step will always be overwritten.

hnyu (approved):

alf/trainers/policy_trainer.py
emailweixu: I don't understand what you mean.

hnyu: Sorry I misunderstood the code

PR #523, opened 2020-04-08
Reorganize Network

emailweixu:
1. Separate Network into Network and PreprocessorNetwork because not all Network need preprocessing capability.
2. Remove DistributionNetwork as it is essentially same as Network


Haichao-Zhang (dismissed):
Looks good. I noticed that somehow the following comments (from code base) are mis-matched:
https://github.com/HorizonRobotics/alf/blob/671bca2d37c47d63a81d9dd6ab2f6f87c630b1fc/alf/networks/actor_distribution_networks.py#L28-L29

https://github.com/HorizonRobotics/alf/blob/671bca2d37c47d63a81d9dd6ab2f6f87c630b1fc/alf/networks/actor_distribution_networks.py#L127-L128

Might be able to switch them while resolving the conflicts.

Haichao-Zhang (approved):

PR #524, opened 2020-04-08
allow flexible mini_batch_length

hnyu:
For one_time replayer, sometimes it might be difficult to require that mini_batch_length is divisible by experience length. It will be more flexible if we can remove the check.

Also fix total_size() for the one-time exp replayer. Before it returns the batch size. 

More docstring format corrections.

### DIFF
--- a/alf/algorithms/off_policy_algorithm.py
+++ b/alf/algorithms/off_policy_algorithm.py
@@ -155,6 +155,8 @@ class OffPolicyAlgorithm(RLAlgorithm):
                 "length=%s not a multiple of mini_batch_length=%s" %
                 (length, mini_batch_length))
             length = length // mini_batch_length * mini_batch_length
+            experience = alf.nest.map_structure(lambda x: x[:, :length, ...],
+                                                experience)
             common.warning_once(
                 "Experience length has been cut to %s" % length)
 
--- a/alf/algorithms/off_policy_algorithm.py
+++ b/alf/algorithms/off_policy_algorithm.py
@@ -150,9 +150,13 @@ class OffPolicyAlgorithm(RLAlgorithm):
 
         length = experience.step_type.shape[1]
         mini_batch_length = (mini_batch_length or length)
-        assert length % mini_batch_length == 0, (
-            "length=%s not a multiple of mini_batch_length=%s" %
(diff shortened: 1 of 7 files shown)


alf/algorithms/sac_algorithm.py
```diff
@@ -425,7 +426,7 @@ def _calc_critic_loss(self, training_info):
             value=critic_info.critic2,
             target_value=target_critic)
 
-        critic_loss = critic_loss1.loss + critic_loss2.loss
+        critic_loss = (critic_loss1.loss + critic_loss2.loss) / 2.
```
Haichao-Zhang: ``` / 2``` should be removed as each critic network is trained using its original loss instead of a scaled loss.

Haichao-Zhang (changes requested):

alf/algorithms/sac_algorithm.py
hnyu: The absolute scale is not important. We can also argument MSELoss * 0.5 is the way to go. More importantly, the current displayed critic_loss in TB is always doubled, which is confusing.

hnyu: Changed to only scaling the extra field for tensorboard display

Haichao-Zhang (dismissed):

alf/algorithms/off_policy_algorithm.py
```diff
@@ -150,9 +150,13 @@ def _train_experience(self, experience, num_updates, mini_batch_size,
...
-        assert length % mini_batch_length == 0, (
-            "length=%s not a multiple of mini_batch_length=%s" %
-            (length, mini_batch_length))
+        if length % mini_batch_length:
+            common.warning_once(
+                "length=%s not a multiple of mini_batch_length=%s" %
+                (length, mini_batch_length))
+            length = length // mini_batch_length * mini_batch_length
+            common.warning_once(
+                "Experience length has been cut to %s" % length)
```
emailweixu, line 161: Where is the logic for cutting experience?

hnyu: Somehow missed in this PR. Will update

emailweixu (approved):

PR #525, opened 2020-04-08
gradient clipping in optimizers

hnyu:
For gin files that haven't been updated, I didn't change the gradient clipping options.

### DIFF
--- a/alf/optimizers/optimizers.py
+++ b/alf/optimizers/optimizers.py
@@ -21,8 +21,6 @@ from alf.utils import common
 from alf.utils import tensor_utils
 from alf.optimizers.adam_tf import AdamTFUnwrapped
 
-optimizer_counter = 0
-
 
 def wrap_optimizer(cls):
     """A helper function to construct torch optimizers with
@@ -33,6 +31,7 @@ def wrap_optimizer(cls):
     """
     NewClsName = cls.__name__ + "_"
     NewCls = type(NewClsName, (cls, ), {})
+    NewCls.counter = 0
 
     @common.add_method(NewCls)
     def __init__(self,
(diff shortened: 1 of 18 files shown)


emailweixu (dismissed):

Haichao-Zhang (dismissed):

alf/optimizers/optimizers.py
```diff
@@ -16,24 +16,75 @@
...
+                to clip gradient. If False, use `tensor_utils.clip_by_norms` for
+                each grad.
+            name (str): the name displayed when summarizing the gradient norm. If
+                None, then a global name in the format of "class_name_i" will be
+                created, where "i" is the global optimizer id.
+        """
         super(NewCls, self).__init__([{'params': []}], **kwargs)
+        self._gradient_clipping = gradient_clipping
+        self._clip_by_global_norm = clip_by_global_norm
+        self._name = name
```
Haichao-Zhang, line 55: Is this ```name`` necessary?  Can use ```NewCls.\__name\__```?

hnyu: If there are several Adam optimizers then their global norm summary will have conficts. The user then has the chance to specify their own name, if he doesn't want the default numbered name

Haichao-Zhang: You can still keep the counter.

hnyu: > You can still keep the counter.

Even the numbered names will be confusing like you don't know Adam_0 corresponds to your higher-level algorithm or your lower-level one. The order of creating optimizers can be uncertain or hard to find out.

hnyu:
Resolved conflicts

alf/optimizers/optimizers.py
Haichao-Zhang: You are right. It's better to keep it if there are a lot of optimizers of the same type.

Haichao-Zhang (approved):

PR #526, opened 2020-04-08
RingBuffer atomic and blocking queueing operations

le-horizon:
Hey Guys,

Here is the follow up PR to make RingBuffer and ReplayBuffer multiprocessing safe.

It is nicely broken down into stages by several commits.

Thanks,
Le


alf/experience_replayers/replay_buffer.py
```diff
@@ -99,18 +144,72 @@ def _create_buffer(tensor_spec):
...
+            bool
+        """
+        current_size = self._current_size[env_ids]
+        max_size = current_size.max()
+        return max_size < self._max_length
+
+    def enqueue(self, batch, env_ids=None, blocking=False):
         """Add a batch of items to the buffer.
 
+        Args:
```
emailweixu, line 180: need to document the be behavior of block=False: it  always success even if there is no free slot. If there is no free slot, ...

The typical for block=False is that if there is not free space, it will not do anything and return False.

alf/experience_replayers/replay_buffer.py
```diff
@@ -167,20 +314,36 @@ def dequeue(self, env_ids=None):
...
-        self._current_pos.fill_(0)
+        with alf.device(self._device):
+            env_ids = self.check_convert_env_ids(env_ids)
+            self._current_size.scatter_(0, env_ids, 0)
+            self._current_pos.scatter_(0, env_ids, 0)
+            if self._dequeued:
+                self._dequeued.set()
+                self._enqueued.clear()
+
+    def stop(self):
```
emailweixu, line 346: comment

alf/experience_replayers/replay_buffer.py
```diff
@@ -167,20 +314,36 @@ def dequeue(self, env_ids=None):
...
+        self._stop.set()
+
+    def revive(self):
```
emailweixu, line 357: comment

alf/experience_replayers/replay_buffer.py
```diff
@@ -196,17 +359,20 @@ def __init__(self,
...
             device=device,
+            allow_multiprocess=allow_multiprocess,
             name=name)
 
-    def add_batch(self, batch, env_ids=None):
-        self.enqueue(batch, env_ids)
+    def add_batch(self, batch, env_ids=None, blocking=False):
+        self.enqueue(batch, env_ids, blocking=blocking)
 
+    @atomic
```
emailweixu, line 391: Is there readwrite lock? If so, we can just use readlock for reading operations so that multiple reading operations do not block each other.

le-horizon: According to https://stackoverflow.com/questions/53119297/python-multiprocessing-read-write-lock
We need to write our own lock.
Do you think it's good to have for us?
Compared with env.step(), policy.step(), or train.step(), queueing time will likely be negligible?

Addressed the other comments.

emailweixu (approved):

PR #528, opened 2020-04-10
fix minor bugs

hnyu:
By default RlAlgorithm has a cast_transformer which can produce errors if some algorithm does take discrete observation inputs. 

Also fix the error when info.loss is () while info.scalar_loss is not.

Also handle the case where mini_batch_length is greater than experience length.

### DIFF
--- a/alf/algorithms/off_policy_algorithm.py
+++ b/alf/algorithms/off_policy_algorithm.py
@@ -154,6 +154,7 @@ class OffPolicyAlgorithm(RLAlgorithm):
             common.warning_once(
                 "length=%s not a multiple of mini_batch_length=%s" %
                 (length, mini_batch_length))
+            mini_batch_length = min(mini_batch_length, length)
             length = length // mini_batch_length * mini_batch_length
             experience = alf.nest.map_structure(lambda x: x[:, :length, ...],
                                                 experience)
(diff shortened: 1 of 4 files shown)


alf/algorithms/off_policy_algorithm.py
```diff
@@ -154,6 +154,7 @@ def _train_experience(self, experience, num_updates, mini_batch_size,
             common.warning_once(
                 "length=%s not a multiple of mini_batch_length=%s" %
                 (length, mini_batch_length))
+            mini_batch_length = min(mini_batch_length, length)
```
emailweixu: I am worried that this may causes problem. At least should give a warning.

hnyu:
Finally fix the TB display of SAC alpha. Currently in every `_alpha_train_step()` there is a summary of `alpha`, which results in a vertical line segment at each counter if `mini_batch_length > 1`. Now put `alpha` in the info.

Actually since `alpha` never changes during the several train steps before gradient update, there should only be one value (but multiple points) at each counter value. I realize that somehow if there is no smoothing of the curve, then it looks normal; but after smoothing vertical lines appear. This TB smoothing behavior can be confusing when debugging an algorithm.

emailweixu (approved):

PR #530, opened 2020-04-10
move RingBuffer to utils.data_buffer.py

le-horizon:
Hey Haonan and Wei,

This change moves the RingBuffer class to a better place utils/data_buffer.py, and also uses RingBuffer as the underlying store for DataBuffer.

Let me know if there are things you may want changed.

Thanks,
Le


emailweixu:
Did you make any changes beside moving?

le-horizon:
> Did you make any changes beside moving?

Yes, this change now uses RingBuffer (num_environments == 1) as the underlying store for DataBuffer, which means DataBuffer._buffer[0] should be used instead of DataBuffer._buffer.  Same for _current_pos and _current_size.  I added test coverage in data_buffer_test to cover DataBuffer.get_batch.

The first commit is the move.
The next two are the real change.

hnyu:
> > Did you make any changes beside moving?
> 
> Yes, this change now uses RingBuffer (num_environments == 1) as the underlying store for DataBuffer, which means DataBuffer._buffer[0] should be used instead of DataBuffer._buffer. Same for _current_pos and _current_size. I added test coverage in data_buffer_test to cover DataBuffer.get_batch.
> 
> The first commit is the move.
> The next two are the real change.

Have changed all the places that use DataBuffer?

I don't think [0] is elegant. Maybe even though DataBuffer derives from RingBuffer, we should not suggest directly using _buffer, _current_pos, and _current_size as they are arrays. For example, we can add three properties buffer, current_pos, and current_size that will call [0] inside. This will make the code cleaner.

le-horizon:
> > The first commit is the move.
> > The next two are the real change.
> 
> Have changed all the places that use DataBuffer?
> 

I haven't checked calling sites, but as long as callers only access the class by exposed properties, everything should work just fine.

Should I manually check?  Tests are all passing.

> I don't think [0] is elegant. Maybe even though DataBuffer derives from RingBuffer, we should not suggest directly using _buffer, _current_pos, and _current_size as they are arrays. For example, we can add three properties buffer, current_pos, and current_size that will call [0] inside. This will make the code cleaner.

Good point.  Done.  Buffer is trickier due to being a nest.  Every time map_structure needs to be called, which I think is perhaps unnecessary and can be inefficient?

alf/utils/data_buffer.py
```diff
@@ -388,6 +388,7 @@ def __init__(self,
             name=name)
         self._capacity = torch.as_tensor(
             self._max_length, dtype=torch.int64, device=device)
+        self.buffer = alf.nest.map_structure(lambda buf: buf[0], self._buffer)
```
hnyu: Still should prefix with "_" because it's a private member? Maybe with a different name

hnyu (changes requested):
Please make sure to correct docstrings in the files you have updated.

le-horizon:
Addressed all @hnyu , PTAL.

Thanks,
Le

le-horizon:
@hnyu , just FYI, I merged in your change 

https://github.com/HorizonRobotics/alf/pull/537/commits/722b0bee591a60b8828d8ef50e64c0c45901a436

and removed the TODO.

hnyu:
Docstrings in data_buffer.py are still not updated yet?

le-horizon:
@hnyu I did a rough look at first, and didn't find artifacts.  I guess my eyes are not yet trained on detecting those.  I looked more closely this time.  PTAL.  Thanks.

hnyu (approved):

PR #537, opened 2020-04-14
add after_train_iter

hnyu:
Added a hook `after_train_iter` to be called after each training iteration. The next PR would be to make non-RL algorithms also make use of most of `train_from_unroll` and `train_from_replay_buffer` (put common code into helper functions). 

A major change is that currently we unroll for off-policy algorithms with `torch.no_grad()`. Now that part also has grads for optionally training on-policy algorithms in `after_train_iter`. 

Example usage (for now assume that `self._predictor` derives from `OnPolicyAlgorithm` or `OffPolicyAlgorithm` in order to use the two functions):

```python

# in an off-policy agent.py:
def after_train_iter(self, training_info):
     self._predictor.train_from_unroll(getattr(training_info, "predictor"))
```
or 
```python
# in an off-policy agent.py:
def after_train_iter(self, training_info):
     self._predictor.train_from_replay_buffer() # has its own replay buffer
```

### DIFF
--- a/alf/algorithms/agent_helpers.py
+++ b/alf/algorithms/agent_helpers.py
@@ -23,61 +23,8 @@ from alf.data_structures import LossInfo
 from alf.utils.math_ops import add_ignore_empty
 
 
-class AgentStateSpecs(object):
-    def __init__(self, state_ctor):
-        """Create three state specs given the state creator."""
-        self.train_state_spec = state_ctor()
-        self.rollout_state_spec = state_ctor()
-        self.predict_state_spec = state_ctor()
-
-    def collect_algorithm_state_specs(self, alg, alg_field):
-        """Collect state specs from algorithms. For code conciseness, we collect
-        all three state specs even though some of them will not be used during
-        `unroll` or `train`.
-
-        Args:
-            alg (Algorithm):
-            alg_field (str): the corresponding algorithm field in an
-                `AgentState`.
-        """
-        self.train_state_spec = self.train_state_spec._replace(
-            **{alg_field: alg.train_state_spec})
-        self.rollout_state_spec = self.rollout_state_spec._replace(
-            **{alg_field: alg.rollout_state_spec})
-        self.predict_state_spec = self.predict_state_spec._replace(
-            **{alg_field: alg.predict_state_spec})
-
-
-def accumulate_algortihm_rewards(rewards, weights, names, summary_prefix,
-                                 summarize_fn):
-    """Sum a list of rewards by their weights. Also summarize the rewards
-    statistics given their names.
-
-    Args:
-        rewards (list[Tensor]): a list of rewards tensors
-        weights (list[float]): a list of floating numbers
-        names (list[str]): a list of reward names
-        summary_prefix (str): a string prefix for summary
-        summarize_fn (Callable): a summarize function that accepts a name and
-            a reward.
-
-    Returns:
-        A single reward after accumulation.
-    """
-    assert len(rewards) > 0
-
-    reward = torch.zeros_like(rewards[0])
-    for r, w, name in zip(rewards, weights, names):
-        summarize_fn(os.path.join(summary_prefix, name), r)
-        reward += w * r
-
-    if len(rewards) > 1:
-        summarize_fn(os.path.join(summary_prefix, "overall"), reward)
-    return reward
-
-
 def _make_rl_training_info(training_info, name):
-    """Given an agent's training info, extracts the `info` and `rollout_info`
+    """Given an agent's training info, extracts the ``info`` and ``rollout_info``
     fields for an RL algorithm."""
     if training_info.rollout_info == ():
         rollout_info = ()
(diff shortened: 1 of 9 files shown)


hnyu:
@le-horizon  My last commit adds a comment about removing the tmp solution of detaching in RLAlgorithm.observe() after you fix the detach issue

le-horizon:
> @le-horizon My last commit adds a comment about removing the tmp solution of detaching in RLAlgorithm.observe() after you fix the detach issue

Roger!

alf/algorithms/algorithm.py
```diff
@@ -737,17 +737,41 @@ def update_with_gradient(self, loss_info, valid_masks=None, weight=1.0):
...
 
         Args:
-            training_info (nested Tensor): information collected for training.
+            training_info (TrainingInfo): information collected for training.
                 It is batched from each ``info`` returned by ``rollout_step()`` or
                 ``train_step()``.
         """
         pass
 
+    def after_train_iter(self, training_info=None):
```
emailweixu, line 751: Can you also update the comments for RLAlgorithm about all the major hooks?

alf/algorithms/off_policy_algorithm.py
```diff
@@ -105,18 +101,41 @@ def _train_iter_off_policy(self):
...
+                training_info = self.unroll(config.unroll_length)
+                self.summarize_rollout(training_info)
+                self.summarize_metrics()
+                if self._exp_replayer.total_size >= config.initial_collect_steps:
+                    break
+
+        steps = self.train_from_replay_buffer(True)
+
+        with record_time("time/after_train_iter"):
+            # skip the first iteration as it can unroll quite long
```
emailweixu: I don't understand the reason to skip.

alf/algorithms/off_policy_algorithm.py
```diff
@@ -105,18 +101,41 @@ def _train_iter_off_policy(self):
         """User may override this for their own training procedure."""
         config: TrainerConfig = self._config
 
+        first_iter = (alf.summary.get_global_counter() == 0)
+
         if not config.update_counter_every_mini_batch:
             alf.summary.get_global_counter().add_(1)
 
-        with torch.no_grad():
```
emailweixu, line 111: As a note, it will cost more memory usage without "with no_grad()".

alf/algorithms/algorithm.py
```diff
@@ -737,17 +737,41 @@ def update_with_gradient(self, loss_info, valid_masks=None, weight=1.0):
...
+        is always only one gradient update in each training iteration. Otherwise
+        it's less frequently called than ``after_update``.
+
+        Args:
+            training_info (TrainingInfo): information collected during ``unroll()``.
+                Note that it won't contain the field ``rollout_info`` because this
+                is the info collected just from the unroll but not from a replay
+                buffer. So if ``training_info`` is used, make sure you are doing
+                on-policy training in this function; if it's None, then only
+                off-policy training is allowed.
```
emailweixu: comment when training_info is None.

alf/algorithms/off_policy_algorithm.py
hnyu: Because we have init_collect steps. It might be thousands of steps. I've chopped init_collect_steps into multiples of unroll_length.

alf/algorithms/algorithm.py
hnyu: updated

alf/algorithms/off_policy_algorithm.py
hnyu: Yeah I know that. Potentially we could optimize this part by setting a flag indicating whether we should unroll with grads or not depending on the user's need for training sub-algorithms. But if `unroll_length` is small then it's probably not a big issue.

alf/algorithms/sac_algorithm_test.py
```diff
@@ -87,7 +87,7 @@ def test_sac_algorithm(self):
             name="MySAC")
 
         eval_env.reset()
-        for i in range(200):
+        for i in range(700):
```
Haichao-Zhang, line 90: what's the reason for increasing this number?

alf/algorithms/agent_helpers.py
```diff
@@ -23,126 +23,164 @@
...
+
+    def state_specs(self):
+        """Return the state specs collected from child algorithms."""
+        return dict(
+            train_state_spec=self._train_state_spec,
+            rollout_state_spec=self._rollout_state_spec,
+            predict_state_spec=self._predict_state_spec)
+
+    @staticmethod
+    def accumulate_algortihm_rewards(rewards, weights, names, summary_prefix,
```
Haichao-Zhang: ```accumulate_algortihm_rewards``` -> ```accumulate_algorithm_rewards```; and all the places calling this method

Haichao-Zhang (changes requested):

alf/algorithms/sac_algorithm_test.py
hnyu: Now init_collect_steps is divided into multiple iterations

Haichao-Zhang: I see. Do you think counting the ```initial_collection_steps``` into ```num_of_iters``` is a bit confusing?
For example, in this case, the first 500 training iterations are essentially for data collection and not doing any training, which might not be very clear when configuring gins.

hnyu: I don't see another better way. I think we always choose a large enough num_iterations in a gin file. The point is, initial_collection_steps has to be divided if grads are recorded for a flexible training of sub-algorithms.  Otherwise the graph will be really huge.

emailweixu: Perhaps update the comment about `num_iterations` at TrainerConfig?

hnyu: updated

Haichao-Zhang (approved):

emailweixu (approved):

PR #539, opened 2020-04-15
add config flag unroll_with_grad

hnyu:
for saving memory when using off-policy algorithms that have a very large `unroll_length`, e.g., PPO.

### DIFF
--- a/alf/algorithms/config.py
+++ b/alf/algorithms/config.py
@@ -67,11 +67,12 @@ class TrainerConfig(object):
                 iteration. The total number of time steps from all environments per
                 iteration can be computed as: ``num_envs * env_batch_size * unroll_length``.
             unroll_with_grad (bool): a bool flag indicating whether we require
-                grad during ``unroll()``. ``OnPolicyAlgorithm`` always unrolls with
-                grads. Thus this flag is only used by ``OffPolicyAlgorithm``. In
-                most cases it's unnecessary and turned off for saving memory.
-                However, when there is an on-policy sub-algorithm, then we can
-                enable this flag for its training.
+                grad during ``unroll()``. This flag is only used by
+                ``OffPolicyAlgorithm`` where unrolling with grads is usually
+                unnecessary and turned off for saving memory. However, when there
+                is an on-policy sub-algorithm, we can enable this flag for its
+                training. ``OnPolicyAlgorithm`` always unrolls with grads and this
+                flag doesn't apply to it.
             use_rollout_state (bool): If True, when off-policy training, the RNN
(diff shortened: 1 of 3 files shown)


alf/algorithms/off_policy_algorithm.py
```diff
@@ -104,17 +104,27 @@ def _train_iter_off_policy(self):
...
-            training_info = self.unroll(config.unroll_length)
-            self.summarize_rollout(training_info)
-            self.summarize_metrics()
+        def _unroll():
+            with record_time("time/unroll"):
+                training_info = self.unroll(config.unroll_length)
+                self.summarize_rollout(training_info)
+                self.summarize_metrics()
+
+        if config.unroll_with_grad:
```
emailweixu: use `with torch.set_grad_enabled(is_train):` to simplify code

hnyu: changed

alf/algorithms/off_policy_algorithm.py
```diff
@@ -104,17 +104,21 @@ def _train_iter_off_policy(self):
...
         with record_time("time/after_train_iter"):
-            training_info = training_info._replace(
-                rollout_info=(), info=training_info.rollout_info)
-            self.after_train_iter(training_info)
+            if config.unroll_with_grad:
+                training_info = training_info._replace(
+                    rollout_info=(), info=training_info.rollout_info)
+                self.after_train_iter(training_info)
+            else:
+                self.after_train_iter()  # only off-policy training
```
emailweixu, line 121: need to update the comment for after_train_iter about when training_info is None

alf/algorithms/config.py
```diff
@@ -65,6 +66,12 @@ def __init__(self,
             unroll_length (int):  number of time steps each environment proceeds per
                 iteration. The total number of time steps from all environments per
                 iteration can be computed as: ``num_envs * env_batch_size * unroll_length``.
+            unroll_with_grad (bool): a bool flag indicating whether we require
+                grad during ``unroll()``. ``OnPolicyAlgorithm`` always unrolls with
+                grads. Thus this flag is only used by ``OffPolicyAlgorithm``. In
```
Haichao-Zhang: Suggest to move ```This flag is only used by OffPolicyAlgorithm ...enable this flag for its training``` immediately after ```we require
                grad during unroll()```. And move The explanation to ```OnPolicyAlgorithm``` case to the end.

hnyu: updated

Haichao-Zhang (approved):

PR #540, opened 2020-04-15
fix the shape of SAC alpha when collect info in parallel

hnyu:
Currently tensors in `info` computed by `train_step` are not checked for their shapes in `_collect_train_info_parallelly`. I added an assertion for that. Fix the shape of `alpha` returned by SACAlgorithm.

### DIFF
--- a/alf/algorithms/sac_algorithm.py
+++ b/alf/algorithms/sac_algorithm.py
@@ -359,8 +359,9 @@ class SacAlgorithm(OffPolicyAlgorithm):
         info = SacAlphaInfo(
             loss=LossInfo(
                 loss=alpha_loss,
-                extra=dict(alpha_loss=alpha_loss, alpha=self._log_alpha.
-                           exp())))
+                extra=dict(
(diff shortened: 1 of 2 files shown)


alf/algorithms/sac_algorithm.py
```diff
@@ -359,8 +359,9 @@ def _alpha_train_step(self, log_pi):
         info = SacAlphaInfo(
             loss=LossInfo(
                 loss=alpha_loss,
-                extra=dict(alpha_loss=alpha_loss, alpha=self._log_alpha.
-                           exp())))
+                extra=dict(
+                    alpha_loss=alpha_loss,
+                    alpha=self._log_alpha.exp().repeat(alpha_loss.shape[0]))))
```
emailweixu, line 364: use expand() instead. expand does not copy. And we should add a comment somewhere about the shapes of the tensor in info.

hnyu:
close this PR because alpha is summarized in another PR #544

PR #541, opened 2020-04-15
inplace relu & record gpu memory usage

### LINKED ISSUE #534: Use nn.ReLU(inplace=True) as default activation
This can save quite some memory, especially for ConvNets

emailweixu:
Also added
1. log observation_spec and action_spec in the beginning
2. no_grad for trac to save memory


alf/networks/initializers.py
```diff
@@ -41,8 +41,11 @@ def _numerical_calculate_gain(nonlinearity, dz=0.01, r=5.0):
     """
     dist = torch.distributions.normal.Normal(0, 1)
     z = torch.arange(-r, r, dz)
+    # `nonlinearity` might be an inplace op, need to use `z` be for applying
```
hnyu: be for -> before

hnyu (approved):

PR #542, opened 2020-04-16
refactor train_iter

hnyu:
1. remove `TrainingInfo` and use `Experience` and `AlgStep.info`
~~2. add a function `create_experience_data` so that an algorithm can create its own experience structure. Now `TimeStep` and `Experience` are also created from it. If you think this is unnecessary, I can revert this change.~~
3. moved `_train_from_unroll` and `_train_from_replay_buffer` to `Algorithm`. Now `on_policy_algorithm.py` and `off_policy_algorithm.py` are very simple.
4. `experience_spec` is not manually created but inferred when an exp is observed for the first time.
~~5. `calc_loss` of `RLAlgorithm` will take both `Experience` and `AlgStep.info` while `calc_loss` of `Algorithm` only takes `AlgStep.info`.~~
6. The rest is basically changing the interfaces at different places. Also correct docstrings as many as I can.

### DIFF
--- a/alf/algorithms/agent_helpers.py
+++ b/alf/algorithms/agent_helpers.py
@@ -122,10 +122,10 @@ class AgentHelper(object):
         def _update_loss(loss_info, algorithm, name):
             info = getattr(train_info, name)
             if algorithm.is_rl():
-                new_loss_info = algorithm.calc_loss(
-                    _make_rl_experience(experience, name), info)
+                exp = _make_rl_experience(experience, name)
             else:
-                new_loss_info = algorithm.calc_loss(info)
+                exp = experience._replace(rollout_info=())
+            new_loss_info = algorithm.calc_loss(exp, info)
             if loss_info is None:
                 return new_loss_info._replace(
                     extra={name: new_loss_info.extra})
@@ -160,9 +160,10 @@ class AgentHelper(object):
             field = self._get_algorithm_field(alg)
             info = getattr(train_info, field)
             if alg.is_rl():
-                alg.after_update(_make_rl_experience(experience, field), info)
+                exp = _make_rl_experience(experience, field)
             else:
-                alg.after_update(experience, info)
+                exp = experience._replace(rollout_info=())
+            alg.after_update(exp, info)
 
     def after_train_iter(self, algorithms, experience=None, train_info=None):
         """For each provided algorithm, call its ``after_train_iter()`` to do
@@ -182,9 +183,9 @@ class AgentHelper(object):
             if experience is not None:
                 info = getattr(train_info, field)
                 if alg.is_rl():
-                    alg.after_train_iter(
-                        _make_rl_experience(experience, field), info)
+                    exp = _make_rl_experience(experience, field)
                 else:
-                    alg.after_train_iter(experience, info)
+                    exp = experience._replace(rollout_info=())
+                alg.after_train_iter(exp, info)
             else:
                 alg.after_train_iter()
--- a/alf/algorithms/agent_helpers.py
+++ b/alf/algorithms/agent_helpers.py
@@ -22,8 +22,8 @@ from alf.data_structures import LossInfo
 from alf.utils.math_ops import add_ignore_empty
 
 
-def _make_rl_experience(experience, name):
-    """Given an experience, extracts the ``rollout_info`` field for an RL
+def _make_alg_experience(experience, name):
+    """Given an experience, extracts the ``rollout_info`` field for an
     algorithm.
     """
     if experience.rollout_info == ():
@@ -121,10 +121,7 @@ class AgentHelper(object):
 
         def _update_loss(loss_info, algorithm, name):
             info = getattr(train_info, name)
-            if algorithm.is_rl():
-                exp = _make_rl_experience(experience, name)
-            else:
-                exp = experience._replace(rollout_info=())
+            exp = _make_alg_experience(experience, name)
             new_loss_info = algorithm.calc_loss(exp, info)
             if loss_info is None:
                 return new_loss_info._replace(
@@ -159,13 +156,10 @@ class AgentHelper(object):
         for alg in algorithms:
             field = self._get_algorithm_field(alg)
             info = getattr(train_info, field)
-            if alg.is_rl():
-                exp = _make_rl_experience(experience, field)
-            else:
-                exp = experience._replace(rollout_info=())
+            exp = _make_alg_experience(experience, field)
             alg.after_update(exp, info)
 
-    def after_train_iter(self, algorithms, experience=None, train_info=None):
+    def after_train_iter(self, algorithms, experience, train_info=None):
         """For each provided algorithm, call its ``after_train_iter()`` to do
         things after the agent finishes one training iteration (i.e.,
         ``train_iter()``).
@@ -178,14 +172,10 @@ class AgentHelper(object):
                 algorithms. It is batched from each ``AlgStep.info`` returned by
                 ``rollout_step()``.
         """
+        assert experience.rollout_info == (), (
+            "'experience' should always be collected from 'unroll()' and its "
+            "'rollout_info' field should have been moved to 'train_info'!")
         for alg in algorithms:
             field = self._get_algorithm_field(alg)
-            if experience is not None:
-                info = getattr(train_info, field)
-                if alg.is_rl():
-                    exp = _make_rl_experience(experience, field)
-                else:
-                    exp = experience._replace(rollout_info=())
-                alg.after_train_iter(exp, info)
-            else:
-                alg.after_train_iter()
+            info = (None if train_info is None else getattr(train_info, field))
+            alg.after_train_iter(experience, info)
--- a/alf/algorithms/agent_helpers.py
+++ b/alf/algorithms/agent_helpers.py
@@ -18,20 +18,19 @@ import os
 import torch
 
 import alf
-from alf.algorithms.rl_algorithm import RLAlgorithm
 from alf.data_structures import LossInfo
 from alf.utils.math_ops import add_ignore_empty
 
 
-def _make_rl_training_info(training_info, name):
-    """Given an agent's training info, extracts the ``info`` and ``rollout_info``
-    fields for an RL algorithm."""
-    if training_info.rollout_info == ():
+def _make_rl_experience(experience, name):
+    """Given an experience, extracts the ``rollout_info`` field for an RL
+    algorithm.
+    """
+    if experience.rollout_info == ():
         rollout_info = ()
     else:
-        rollout_info = getattr(training_info.rollout_info, name)
-    info = getattr(training_info.info, name)
-    return training_info._replace(info=info, rollout_info=rollout_info)
+        rollout_info = getattr(experience.rollout_info, name)
+    return experience._replace(rollout_info=rollout_info)
 
 
 class AgentHelper(object):
@@ -103,7 +102,7 @@ class AgentHelper(object):
             summarize_fn(os.path.join(summary_prefix, "overall"), reward)
         return reward
 
-    def accumulate_loss_info(self, algorithms, training_info):
+    def accumulate_loss_info(self, algorithms, experience, train_info):
         """Given an overall Agent training info that contains various training infos
         for different algorithms, compute the accumulated loss info for updating
         parameters.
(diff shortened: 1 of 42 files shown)


emailweixu:
> 5. `calc_loss` of `RLAlgorithm` will take both `Experience` and `AlgStep.info` while `calc_loss` of `Algorithm` only takes `AlgStep.info`.

We can make the same interface same. Of course, for Algorithm, `Experience` is what retrieved from the ReplayBuffer, not same as the Experience for RLAlgorithm

emailweixu:
> 2. add a function `create_experience_data` so that an algorithm can create its own experience structure. Now `TimeStep` and `Experience` are also created from it. If you think this is unnecessary, I can revert this change.

Right, it seems unnecessary.

hnyu:
> > 1. add a function `create_experience_data` so that an algorithm can create its own experience structure. Now `TimeStep` and `Experience` are also created from it. If you think this is unnecessary, I can revert this change.
> 
> Right, it seems unnecessary.

reverted

hnyu:
> > 1. `calc_loss` of `RLAlgorithm` will take both `Experience` and `AlgStep.info` while `calc_loss` of `Algorithm` only takes `AlgStep.info`.
> 
> We can make the same interface same. Of course, for Algorithm, `Experience` is what retrieved from the ReplayBuffer, not same as the Experience for RLAlgorithm

updated. For `Algorithm` I will strip the `rollout_info` from `experience` before it being passed to the `calc_loss`.

alf/algorithms/agent_helpers.py
```diff
@@ -103,29 +102,30 @@ def accumulate_algorithm_rewards(rewards, weights, names, summary_prefix,
...
             LossInfo: the accumulated loss info.
         """
 
-        def _update_loss(loss_info, training_info, algorithm, name):
-            if isinstance(algorithm, RLAlgorithm):
-                new_loss_info = algorithm.calc_loss(
-                    _make_rl_training_info(training_info, name))
+        def _update_loss(loss_info, algorithm, name):
+            info = getattr(train_info, name)
+            if algorithm.is_rl():
```
emailweixu: why need to handle rl and non-rl algorithm differently here?

alf/algorithms/on_policy_algorithm.py
```diff
@@ -69,38 +69,19 @@ def _train_iter_on_policy(self):
...
+            train_info = experience.rollout_info
+            experience = experience._replace(rollout_info=())
+            steps = self.train_from_unroll(experience, train_info)
 
         with record_time("time/after_train_iter"):
-            # Here we don't pass ``training_info`` to disable another on-policy
-            # training because otherwise it will backprop on the same graph twice,
-            # which is unnecessary because we could have simply merged the two
-            # trainings into the parent's ``rollout_step``.
+            # Here we don't pass ``experience`` and ```train_info`` to disable
```
emailweixu: It's okay to pass experience because people may choose to use it for different purpose.

alf/algorithms/sarsa_algorithm.py
```diff
@@ -408,34 +405,34 @@ def calc_loss(self, training_info: TrainingInfo):
...
-            loss_info = self._critic_losses[i](shifted_training_info, critic,
+                info.target_critics[..., i])
+            loss_info = self._critic_losses[i](shifted_experience, critic,
                                                target_critic)
             critic_losses.append(nest_map(lambda l: l[:-1], loss_info.loss))
 
         critic_loss = math_ops.add_n(critic_losses)
 
-        not_first_step = training_info.step_type != StepType.FIRST
+        not_first_step = ~experience.is_first()
```
emailweixu: Although this is shorter, but it's slower as it involves two operations.

alf/algorithms/trac_algorithm.py
```diff
@@ -177,18 +181,19 @@ def after_update(self, training_info):
...
 
     @torch.no_grad()
     def _calc_change(self, exp_array):
         """Calculate the distance between old/new action distributions.
 
         The squared distance is:
-        ||logits_1 - logits_2||^2 for Categorical distribution
-        ||loc_1 - loc_2||^2 for Deterministic distribution
-        KL(d1||d2) + KL(d2||d1) for others
+
```
emailweixu, line 193: remove squared in the above sentence.

alf/networks/critic_networks.py
```diff
@@ -94,6 +94,12 @@ def __init__(self,
             raise ValueError(
                 'Only a single action is supported by this network')
 
+        if action_spec.is_discrete:
```
emailweixu: discrete can still be used, right? For high dimensional discrete action, QNetwork may not be the best choice.

alf/utils/averager.py
```diff
@@ -124,21 +124,27 @@ def __init__(self,
 class EMAverager(nn.Module):
     """Class for exponential moving average.
 
-    x_t = (1-update_rate)* x_{t-1} + update_Rate * x
-    The average is corrected by a mass as x_t / mass_t, and the mass is
+    .. code-block:: python
```
emailweixu: math equation seems better

alf/utils/dist_utils.py
```diff
@@ -238,62 +249,61 @@ def extract_distribution_parameters(dist: td.Distribution):
...
-    Extract TensorSpec or DistributionSpec for each element of a nested structure.
-    It assumes that the first dimension of each element is the batch size.
+    Extract ``TensorSpec`` or ``DistributionSpec`` for each element of a nested
+    structure. It assumes that the first dimension of each element is the batch
+    size.
 
     Args:
         nests (nested structure): each leaf node of the nested structure is a
-            Tensor or Distribution of the same batch size
+            Tensor or Distribution of the same batch size/
```
emailweixu: / => .

alf/algorithms/algorithm.py
```diff
@@ -763,39 +980,265 @@ def after_train_iter(self, training_info=None):
...
+
+        Args:
+            experience (Experience): collected during ``unroll()``.
+            train_info (nest): ``AlgStep.info`` returned by ``rollout_step()``.
+
+        Returns:
+            int: number of steps that have been trained
+        """
+        valid_masks = (experience.step_type != StepType.LAST).to(torch.float32)
+        if self.is_rl():
```
emailweixu, line 1075: If non RL algorithm is allowed, experience may not always has step_type

alf/algorithms/algorithm.py
```diff
@@ -763,39 +980,265 @@ def after_train_iter(self, training_info=None):
...
+            loss_info = self.calc_loss(experience, train_info)
+        else:
+            loss_info = self.calc_loss(train_info)
```
emailweixu: calc_loss(experience, train_info)

alf/algorithms/algorithm.py
```diff
@@ -763,39 +980,265 @@ def after_train_iter(self, training_info=None):
...
+        self.after_update(experience, train_info)
+        self.summarize_train(experience, train_info, loss_info, params)
+        return torch.tensor(experience.step_type.shape).prod()
+
+    def train_from_replay_buffer(self, update_global_counter=False):
+        """This function can be called by any algorithm that has its own
+        replay buffer configured. There are several parameters specified in
+        ``self._config`` that will affect how the training is performed:
+
+        - ``initial_collect_steps``: only start sampling and training after so
```
emailweixu: "only start sampling and training" => only start training

alf/algorithms/algorithm.py
```diff
@@ -763,39 +980,265 @@ def after_train_iter(self, training_info=None):
...
+        length = experience.step_type.shape[0]
+        if self._config.temporally_independent_train_step or length == 1:
+            train_info = self._collect_train_info_parallelly(experience)
+        else:
+            train_info = self._collect_train_info_sequentially(experience)
+
+        experience = dist_utils.params_to_distributions(
+            experience, self.processed_experience_spec)
+
+        if self.is_rl():
```
emailweixu, line 1289: no need to distinguish rl/non-rl

alf/algorithms/rl_algorithm.py
```diff
@@ -58,9 +55,9 @@ class RLAlgorithm(Algorithm):
     5. ``update_with_gradient()``: Do one gradient update based on the loss. It is
        used by the default ``train_iter()`` implementation. You can override to
        implement your own ``update_with_gradient()``.
-    6. ``calc_loss()``: calculate loss based the training_info collected from
-       ``rollout_step()`` or ``train_step()``. It is used by the default
-       implementation of ``train_iter()``. If you want to use the default
+    6. ``calc_loss()``: calculate loss based the ``experience`` and the ``train_info``
+       collected from ``rollout_step()`` or ``train_step()``. It is used by the
+       default implementation of ``train_iter()``. If you want to use the default
```
emailweixu, line 60: perhaps we should also have a similar paragraph of description about Algorithm?

alf/algorithms/agent_helpers.py
hnyu: already changed to the same interface for both algorithms.

alf/algorithms/on_policy_algorithm.py
hnyu: So basically after_train_iter always gets the two arguments wherever it's called?

alf/networks/critic_networks.py
hnyu: If action spec is discrete then it has to go through a preprocessor first. Which is not supported by CriticNetwork for now. Right now the EncodingNetwork `self._action_encoder` will complain the shape is not correct. I can add an argument `action_input_processors` for this class.

alf/algorithms/rl_algorithm.py
hnyu: added

hnyu:
One issue we have currently is that for a sub-algorithm to observe a time step, it has to do this in the parent's ``rollout_step``. However, the time step has already been transformed in ``unroll`` and the transformed one will be stored in the sub-algorithm's buffer. When calling ``train_from_replay_buffer`` the experiences will be transformed again, which would potentially cause errors. Is there any suggestion for this?

alf/algorithms/algorithm.py
```diff
@@ -1200,16 +1248,17 @@ def _collect_train_info_sequentially(self, experience):
...
 
         exp = alf.nest.map_structure(lambda x: x.reshape(-1, *x.shape[2:]),
                                      experience)
 
         if self._use_rollout_state:
             policy_state = exp.state
         else:
-            policy_state = self.get_initial_train_state(exp.step_type.shape[0])
+            size = alf.nest.get_nest_size(exp, dim=0)
+            policy_state = self.get_initial_train_state(size)
```
emailweixu, line 1264: Seems to me that the original code has a bug. Should use batch_size instead?

alf/networks/critic_networks.py
```diff
@@ -29,43 +29,62 @@
...
 @gin.configurable
 class CriticNetwork(Network):
-    """Create an instance of CriticNetwork."""
+    """Creates an instance of ``CriticNetwork`` for estimating action-value of
+    continuous or discrete actions. The action-value is defined as the expected
+    return starting from the given input observation and taking the given action.
+    This module takes observation as input and action as input and outputs an
+    action-value tensor with the shape of ``[batch_size]``.
+
+    Currently there seems no need for this class to handle nested inputs; If
```
emailweixu: It seems that it can handle nested observation.

alf/networks/critic_networks.py
```diff
@@ -204,43 +217,48 @@ def forward(self, inputs, state=()):
...
 @gin.configurable
 class CriticRNNNetwork(Network):
-    """Creates a critic network with RNN."""
+    """Creates an instance of ``CriticRNNNetwork`` for estimating action-value
+    of continuous or discrete actions. The action-value is defined as the
+    expected return starting from the given inputs (observation and state) and
+    taking the given action. It takes observation and state as input and outputs
+    an action-value tensor with the shape of [batch_size].
+
+    Currently there seems no need for this class to handle nested inputs; If
```
emailweixu: can handle nested observation.

alf/algorithms/algorithm.py
hnyu: Nothing is wrong here. This is after reshaping so now `size=length*batch_size` because of collecting info in parallel.

alf/algorithms/rl_algorithm.py
```diff
@@ -451,8 +451,13 @@ def unroll(self, unroll_length):
             policy_state = common.reset_state_if_necessary(
                 policy_state, initial_state, time_step.is_first())
             transformed_time_step = self.transform_timestep(time_step)
+            # save the untransformed time step in case that sub-algorithms need
+            # to store it in replay buffers
+            transformed_time_step._replace(untransformed_time_step=time_step)
```
emailweixu:
```
 transformed_time_step = transformed_time_step._replace(untransformed_time_step=time_step)
```

hnyu: fixed

emailweixu (approved):
Try to avoid forced push in the future.  Use "git pull" to update your local branch, not "git pull --rebase"

PR #543, opened 2020-04-16
Dequeue many

le-horizon:
This is needed for async to dequeue an unroll length of observation from queue.

### DIFF
--- a/alf/experience_replayers/replay_buffer.py
+++ b/alf/experience_replayers/replay_buffer.py
@@ -85,9 +85,8 @@ class ReplayBuffer(RingBuffer):
             pos = pos + torch.arange(batch_length).unsqueeze(0)  # [B, T]
             pos = pos % self._max_length
             env_ids = env_ids.reshape(-1, 1).repeat(1, batch_length)  # [B, T]
-            result = alf.nest.map_structure(
-                lambda buffer: buffer.__getitem__((env_ids, pos)).reshape(
-                    batch_size, batch_length, *buffer.shape[2:]), self._buffer)
+            result = self.get_data_by_indices(env_ids, pos, batch_size,
+                                              batch_length)
         return convert_device(result)
 
     @atomic
--- a/alf/experience_replayers/replay_buffer.py
+++ b/alf/experience_replayers/replay_buffer.py
@@ -85,8 +85,8 @@ class ReplayBuffer(RingBuffer):
             pos = pos + torch.arange(batch_length).unsqueeze(0)  # [B, T]
             pos = pos % self._max_length
             env_ids = env_ids.reshape(-1, 1).repeat(1, batch_length)  # [B, T]
-            result = self.get_data_by_indices(env_ids, pos, batch_size,
-                                              batch_length)
+            result = alf.nest.map_structure(lambda b: b[(env_ids, pos)],
+                                            self._buffer)
         return convert_device(result)
 
     @atomic
--- a/alf/experience_replayers/replay_buffer.py
+++ b/alf/experience_replayers/replay_buffer.py
@@ -85,12 +85,9 @@ class ReplayBuffer(RingBuffer):
             pos = pos + torch.arange(batch_length).unsqueeze(0)  # [B, T]
             pos = pos % self._max_length
             env_ids = env_ids.reshape(-1, 1).repeat(1, batch_length)  # [B, T]
-            indices = env_ids * self._max_length + pos  # [B, T]
-            indices = indices.reshape(-1)  # [B*T]
             result = alf.nest.map_structure(
-                lambda buffer: buffer[indices].reshape(
-                    batch_size, batch_length, *buffer.shape[1:]),
-                self._flattened_buffer)
+                lambda buffer: buffer.__getitem__((env_ids, pos)).reshape(
+                    batch_size, batch_length, *buffer.shape[2:]), self._buffer)
         return convert_device(result)
 
     @atomic
--- a/alf/utils/data_buffer.py
+++ b/alf/utils/data_buffer.py
@@ -132,8 +132,6 @@ class RingBuffer(nn.Module):
                     num_environments, dtype=torch.int64))
             self._buffer = alf.nest.map_structure(_create_buffer, data_spec)
             bs = BatchSquash(batch_dims=2)
-            self._flattened_buffer = alf.nest.map_structure(
-                lambda x: bs.flatten(x), self._buffer)
 
         if allow_multiprocess:
             self.share_memory()
@@ -207,12 +205,6 @@ class RingBuffer(nn.Module):
         Args:
             batch (Tensor): shape should be
                 ``[batch_size] + tensor_space.shape``.
-                If we ever want to enqueue ``n`` steps (or
-                ``[batch_size, n]`` shaped batch), we'll need to use something
-                like this:
-                ``buf.index_put_((env_ids.reshape(batch_size, 1).repeat(1,
-                batch_size), _current_pos[env_ids] - n to _current_pos[env_ids]
-                ), batch)``
             env_ids (Tensor): If ``None``, ``batch_size`` must be
                 ``num_environments``. If not ``None``, its shape should be
                 ``[batch_size]``. We assume there are no duplicate ids in
@@ -233,10 +225,10 @@ class RingBuffer(nn.Module):
                 "There are duplicated ids in env_ids %s" % env_ids)
 
             current_pos = self._current_pos[env_ids]
-            indices = env_ids * self._max_length + current_pos
             alf.nest.map_structure(
-                lambda buf, bat: buf.__setitem__(indices, bat.detach()),
-                self._flattened_buffer, batch)
+                lambda buf, bat: buf.__setitem__((env_ids, current_pos),
+                                                 bat.detach()), self._buffer,
+                batch)
 
             self._current_pos[env_ids] = (current_pos + 1) % self._max_length
             current_size = self._current_size[env_ids]
@@ -279,6 +271,7 @@ class RingBuffer(nn.Module):
             ``AssertionError`` when not enough data is present, in non-blocking
             mode.
         """
+        assert n <= self._max_length
         if blocking:
             assert self._allow_multiprocess, [
                 "Set allow_multiprocess", "to enable blocking mode."
@@ -322,8 +315,7 @@ class RingBuffer(nn.Module):
                 "size is: %s Try storing more data before calling dequeue" %
                 min_size)
             batch_size = env_ids.shape[0]
-            pos = self._current_pos[env_ids] - current_size
-            pos = pos % self._max_length
+            pos = self._current_pos[env_ids] - current_size  # mod done later
             b_indices = env_ids.reshape(batch_size, 1).repeat(1, n)
             t_range = torch.as_tensor([range(n)] * batch_size)
             t_indices = (pos.reshape(batch_size, 1).repeat(1, n) +
--- a/alf/utils/data_buffer.py
+++ b/alf/utils/data_buffer.py
@@ -207,7 +207,7 @@ class RingBuffer(nn.Module):
         Args:
             batch (Tensor): shape should be
                 ``[batch_size] + tensor_space.shape``.
-                If we ever want to enqueue ``n`` batches (or
+                If we ever want to enqueue ``n`` steps (or
                 ``[batch_size, n]`` shaped batch), we'll need to use something
                 like this:
                 ``buf.index_put_((env_ids.reshape(batch_size, 1).repeat(1,
@@ -248,12 +248,12 @@ class RingBuffer(nn.Module):
                 self._dequeued.clear()
 
     def has_data(self, env_ids, n):
-        """Check ``n`` batches of data available for ``env_ids``.
+        """Check ``n`` steps of data available for ``env_ids``.
 
         Args:
             env_ids (Tensor): Assumed not ``None``, properly checked by
                 ``check_convert_env_ids()``.
-            n (int): Number of batches to check.
+            n (int): Number of time steps to check.
         Returns:
             bool
         """
@@ -262,7 +262,7 @@ class RingBuffer(nn.Module):
         return min_size >= n
 
     def dequeue(self, env_ids=None, n=1, blocking=False):
-        """Return earliest ``n`` batches and mark them removed in the buffer.
+        """Return earliest ``n`` steps and mark them removed in the buffer.
 
         Args:
             env_ids (Tensor): If None, ``batch_size`` must be num_environments.
@@ -274,7 +274,7 @@ class RingBuffer(nn.Module):
                 dequeue.
         Returns:
             nested Tensors or None when blocking dequeue gets terminated by
-            stop event. The shapes of the Tensors are ``[batch_size, ...]``.
+            stop event. The shapes of the Tensors are ``[batch_size, n, ...]``.
         Raises:
             ``AssertionError`` when not enough data is present, in non-blocking
             mode.
@@ -300,7 +300,7 @@ class RingBuffer(nn.Module):
 
     @atomic
     def _dequeue(self, env_ids=None, n=1):
-        """Return earliest ``n`` batches and mark them removed in the buffer.
+        """Return earliest ``n`` steps and mark them removed in the buffer.
 
         Args:
             env_ids (Tensor): If None, ``batch_size`` must be num_environments.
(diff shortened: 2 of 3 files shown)


alf/utils/data_buffer.py
```diff
@@ -207,7 +206,13 @@ def _enqueue(self, batch, env_ids=None):
...
         Args:
             batch (Tensor): shape should be
-                ``[batch_size] + tensor_space.shape``
+                ``[batch_size] + tensor_space.shape``.
+                If we ever want to enqueue ``n`` batches (or
+                ``[batch_size, n]`` shaped batch), we'll need to use something
+                like this:
+                ``buf.index_put_((env_ids.reshape(batch_size, 1).repeat(1,
+                batch_size), _current_pos[env_ids] - n to _current_pos[env_ids]
+                ), batch)``
```
emailweixu: What does a user need to do to enqueue n batches? How does he access buf?

alf/utils/data_buffer.py
```diff
@@ -242,33 +247,37 @@ def _enqueue(self, batch, env_ids=None):
                 self._enqueued.set()
                 self._dequeued.clear()
 
-    def has_data(self, env_ids):
-        """Check one batch of data available for ``env_ids``.
+    def has_data(self, env_ids, n):
+        """Check ``n`` batches of data available for ``env_ids``.
```
emailweixu: "n steps" is less confusing.

alf/utils/data_buffer.py
```diff
@@ -307,26 +317,58 @@ def _dequeue(self, env_ids=None):
...
-            indices = env_ids * self._max_length + pos  # shape [B]
 
-            batch_size = env_ids.shape[0]
-            result = alf.nest.map_structure(
-                lambda buffer: buffer[indices].reshape(
-                    batch_size, 1, *buffer.shape[1:]), self._flattened_buffer)
+            def _create_empty(buf):
+                return []
+
+            result_arr = alf.nest.map_structure(_create_empty, self._buffer)
```
emailweixu: map_structure(lambda x: [], self._buffer)

alf/utils/data_buffer.py
```diff
@@ -287,17 +296,18 @@ def dequeue(self, env_ids=None, blocking=False):
...
+        """Return earliest ``n`` batches and mark them removed in the buffer.
 
         Args:
             env_ids (Tensor): If None, ``batch_size`` must be num_environments.
                 If not None, dequeue from these environments. We assume there
                 is no duplicate ids in ``env_id``. ``result[i]`` will be from
                 environment env_ids[i].
+            n (int): Number of batches to dequeue.
         Returns:
             nested Tensors of shapes ``[batch_size, ...]``.
```
emailweixu: need to update the comment for shape

alf/utils/data_buffer.py
```diff
@@ -307,26 +317,58 @@ def _dequeue(self, env_ids=None):
...
-            batch_size = env_ids.shape[0]
-            result = alf.nest.map_structure(
-                lambda buffer: buffer[indices].reshape(
-                    batch_size, 1, *buffer.shape[1:]), self._flattened_buffer)
+            def _create_empty(buf):
+                return []
+
+            result_arr = alf.nest.map_structure(_create_empty, self._buffer)
+
+            for _ in range(n):
```
emailweixu: better to not to use loop. It is slow. It should be possible to do all n steps at the same time.

emailweixu (changes requested):

alf/utils/data_buffer.py
le-horizon: If it so happens that each unroll gives out several steps of data, we can enqueue them all together.  I don't think we need to do it now, but did a bit research, so noting it down here.

DataBuffer's add_batch actually does this.  We could potentially move that logic into RingBuffer.

alf/utils/data_buffer.py
le-horizon: Done.

alf/utils/data_buffer.py
le-horizon: Good catch.

alf/utils/data_buffer.py
le-horizon: nice!

alf/utils/data_buffer.py
le-horizon: Good point.  I tried before without success, but finally found out __getitem__((x,y)) can do the trick.

Btw, I'm not sure why flatten_buffer is used to store and retrieve items from buffer, in this change I've changed to use _buffer instead.  Seems more straightforward.

Let me know if this is Ok.

le-horizon (commented):
Thanks for the very helpful comments, Wei.

Done all, PTAL.

Le

alf/utils/data_buffer.py
emailweixu: Nice! I didn't know that `__getitem__()` can take tuple. That's why flattened_buffer is used for ReplayBuffer.get_batch(). Can you also change that to use `__getitem__()`?

alf/utils/data_buffer.py
```diff
@@ -207,7 +206,13 @@ def _enqueue(self, batch, env_ids=None):
 
         Args:
             batch (Tensor): shape should be
-                ``[batch_size] + tensor_space.shape``
+                ``[batch_size] + tensor_space.shape``.
```
hnyu: tensor_space -> tensor_spec

alf/utils/data_buffer.py
hnyu: I don't quite understand the code here. Is it valid python code?

alf/utils/data_buffer.py
```diff
@@ -242,33 +247,37 @@ def _enqueue(self, batch, env_ids=None):
                 self._enqueued.set()
                 self._dequeued.clear()
 
-    def has_data(self, env_ids):
-        """Check one batch of data available for ``env_ids``.
+    def has_data(self, env_ids, n):
```
hnyu: can default n to 1

alf/utils/data_buffer.py
```diff
@@ -287,46 +296,62 @@ def dequeue(self, env_ids=None, blocking=False):
...
+                self._buffer)
+            self._current_size[env_ids] = current_size - n
             # set flags if they exist to unblock potential consumers
             if self._dequeued:
                 self._dequeued.set()
                 self._enqueued.clear()
         return convert_device(result)
 
+    @atomic
+    def pop(self, n, env_ids=None):
```
hnyu: rename to `pop_up_to` or `dequeue_up_to`? otherwise it will be confusing

alf/utils/data_buffer.py
```diff
@@ -287,46 +296,62 @@ def dequeue(self, env_ids=None, blocking=False):
...
+        """Mark as removed up to the last n batches.
```
hnyu: n -> \``n\``

hnyu (changes requested):

alf/experience_replayers/replay_buffer.py
```diff
@@ -85,12 +85,9 @@ def get_batch(self, batch_size, batch_length):
             pos = pos + torch.arange(batch_length).unsqueeze(0)  # [B, T]
             pos = pos % self._max_length
             env_ids = env_ids.reshape(-1, 1).repeat(1, batch_length)  # [B, T]
-            indices = env_ids * self._max_length + pos  # [B, T]
-            indices = indices.reshape(-1)  # [B*T]
             result = alf.nest.map_structure(
-                lambda buffer: buffer[indices].reshape(
-                    batch_size, batch_length, *buffer.shape[1:]),
-                self._flattened_buffer)
+                lambda buffer: buffer.__getitem__((env_ids, pos)).reshape(
```
emailweixu: Will buffer[(env_ids, pos)] work?

alf/utils/data_buffer.py
```diff
@@ -242,43 +239,48 @@ def _enqueue(self, batch, env_ids=None):
...
+        """Return earliest ``n`` steps and mark them removed in the buffer.
 
         Args:
             env_ids (Tensor): If None, ``batch_size`` must be num_environments.
                 If not None, dequeue from these environments. We assume there
                 is no duplicate ids in ``env_id``. ``result[i]`` will be from
                 environment ``env_ids[i]``.
-            blocking (bool): If ``True``, blocks if there is no free slot to
-                add data.
+            n (int): Number of batches to dequeue.
```
emailweixu: Number of steps ...

alf/utils/data_buffer.py
```diff
@@ -287,46 +289,61 @@ def dequeue(self, env_ids=None, blocking=False):
...
-        """Return earliest batch and remove it from buffer.
+    def _dequeue(self, env_ids=None, n=1):
+        """Return earliest ``n`` steps and mark them removed in the buffer.
 
         Args:
             env_ids (Tensor): If None, ``batch_size`` must be num_environments.
                 If not None, dequeue from these environments. We assume there
                 is no duplicate ids in ``env_id``. ``result[i]`` will be from
                 environment env_ids[i].
+            n (int): Number of batches to dequeue.
```
emailweixu: steps

alf/utils/data_buffer.py
```diff
@@ -287,46 +289,61 @@ def dequeue(self, env_ids=None, blocking=False):
...
                 "Not all environments have enough data. The smallest data "
                 "size is: %s Try storing more data before calling dequeue" %
                 min_size)
-            pos = self._current_pos[env_ids] - current_size
-            pos = pos % self._max_length
-            indices = env_ids * self._max_length + pos  # shape [B]
-
             batch_size = env_ids.shape[0]
+            pos = self._current_pos[env_ids] - current_size  # mod done later
+            b_indices = env_ids.reshape(batch_size, 1).repeat(1, n)
```
emailweixu: use expand instead of repeat to save memory and to be more efficient.

alf/utils/data_buffer.py
```diff
@@ -287,46 +289,61 @@ def dequeue(self, env_ids=None, blocking=False):
...
+            t_range = torch.as_tensor([range(n)] * batch_size)
```
emailweixu: Use torch function whenever possible: `torch.arange(n).reshape(1, -1)`
`as_tensor([range(n)] * batch_size)` is slow.

alf/utils/data_buffer.py
```diff
@@ -287,46 +289,61 @@ def dequeue(self, env_ids=None, blocking=False):
...
+            t_indices = (pos.reshape(batch_size, 1).repeat(1, n) +
```
emailweixu: `pos.reshape(batch_size, 1) + t_range` is more concise and efficient because of the broadcast mechanism (https://pytorch.org/docs/stable/notes/broadcasting.html)

alf/utils/data_buffer.py
```diff
@@ -132,8 +132,6 @@ def _create_buffer(tensor_spec):
                     num_environments, dtype=torch.int64))
             self._buffer = alf.nest.map_structure(_create_buffer, data_spec)
             bs = BatchSquash(batch_dims=2)
```
emailweixu: bs is not used. And BatchSquash imported at the beginning can be removed.

alf/utils/data_buffer.py
le-horizon: good point.  was confused by that name.

alf/utils/data_buffer.py
le-horizon: Removed.  __setitem__((x,y)) can achieve this.

alf/utils/data_buffer.py
le-horizon: done.

alf/utils/data_buffer.py
le-horizon: good point.

alf/utils/data_buffer.py
le-horizon: Done.

le-horizon (commented):
Done all.
I've removed flatted_buffer from both RingBuffer and ReplayBuffer.

torch.tensor.__getitem__ is not documented well..  I assumed it has the same indexing scheme as __setitem__ in _enqueue, and it works!  Hopefully this API won't change as pytorch evolves..

alf/experience_replayers/replay_buffer.py
le-horizon: Yes!  It's the same thing..  Changed.

alf/utils/data_buffer.py
le-horizon: Done.

alf/utils/data_buffer.py
le-horizon: Done

alf/utils/data_buffer.py
le-horizon: Good point.

alf/utils/data_buffer.py
le-horizon: interesting!

alf/utils/data_buffer.py
le-horizon: Done.

alf/utils/data_buffer.py
le-horizon: good point

le-horizon (commented):
Done all.  PTAL.

Thanks for all the comments!
Le

alf/utils/data_buffer.py
```diff
@@ -287,46 +287,77 @@ def dequeue(self, env_ids=None, blocking=False):
...
             if self._dequeued:
                 self._dequeued.set()
                 self._enqueued.clear()
         return convert_device(result)
 
+    def get_data_by_indices(self, b_indices, t_indices, batch_size, n):
+        """Return data by batch and time indices.
+
+        Args:
+            b_indices (Tensor): Batch indices of shape ``[B, 1]``.
```
emailweixu: incorrect comment about shape

alf/utils/data_buffer.py
```diff
@@ -287,46 +287,77 @@ def dequeue(self, env_ids=None, blocking=False):
...
+            t_indices (Tensor): Time indices of shape ``[n, 1]``.
```
emailweixu: incorrect comment about shape

alf/utils/data_buffer.py
```diff
@@ -287,46 +287,77 @@ def dequeue(self, env_ids=None, blocking=False):
...
+                Should equal ``b_indices.shape[0]``.
```
emailweixu: Is this comment about batch_size?

alf/utils/data_buffer.py
le-horizon: Oops

alf/utils/data_buffer.py
le-horizon: Done

alf/utils/data_buffer.py
le-horizon: Accidentally moved the line..

le-horizon (commented):
Done.

alf/utils/data_buffer.py
```diff
@@ -287,46 +287,77 @@ def dequeue(self, env_ids=None, blocking=False):
...
+    def get_data_by_indices(self, b_indices, t_indices, batch_size, n):
+        """Return data by batch and time indices.
+
+        Args:
+            b_indices (Tensor): Batch indices of shape ``[B, n]``.
+            t_indices (Tensor): Time indices of shape ``[B, n]``.
+            batch_size (int): Number of batches to retrieve.
+                Should equal ``*_indices.shape[0]``.
+            n (int): Number of time steps to retrieve.
+                Should equal ``*_indices.shape[1]``.
```
emailweixu: If so, why should we pass these two values? Just get them from indices.

alf/utils/data_buffer.py
```diff
@@ -287,46 +287,77 @@ def dequeue(self, env_ids=None, blocking=False):
...
+        Returns:
+            nested Tensors of shape ``[batch_size, n, ...]``.
+        """
+        return alf.nest.map_structure(
+            lambda buffer: buffer[(b_indices, t_indices)].reshape(
```
emailweixu: perhaps reshape() is unnecessary.

emailweixu: and if so, there's not much sense to have this function because it's shot enough to write:
```python
alf.nest.map_structure(lambda b: b[(b_indices, t_indices)], self._buffer)
```

le-horizon:
Nice!  Done all.

emailweixu (approved):

le-horizon:
@hnyu your approval is needed.  Thanks!

hnyu (approved):

PR #544, opened 2020-04-17
Efficient softsign & auto temporally_independent_train_step

emailweixu:


alf/utils/math_ops.py
```diff
@@ -198,3 +198,58 @@ def shuffle(values):
...
+
+
+class Softsign_(torch.autograd.Function):
+    r"""softsign(input) -> Tensor
+
+    Applies element-wise inplace, the function :math:`\text{SoftSign}(x) = \frac{x}{1 + |x|}`
+    
+    The `current pytorch implementation of softsign
+    <https://pytorch.org/docs/stable/_modules/torch/nn/functional.html#softsign>`_
+    is inefficient for backward and does not have an inplace version. Hence we
```
hnyu: The reason why the current softsign is inefficient is because there is no explicit gradient defined?

alf/utils/math_ops.py
```diff
@@ -198,3 +198,58 @@ def shuffle(values):
...
+
+softsign_ = Softsign_.apply
+
+
+class Softsign(torch.autograd.Function):
+    r"""softsign(input) -> Tensor
+
+    Applies element-wise, the function :math:`\text{SoftSign}(x) = \frac{x}{1 + |x|}`
+    
+    Compared to Softsign_, this uses more memory but is faster and has higher precision
```
hnyu: Softsign_ -> \``Softsign_``

alf/utils/math_ops.py
```diff
@@ -198,3 +198,58 @@ def shuffle(values):
     batch_size = alf.nest.get_nest_batch_size(values)
     indices = torch.randperm(batch_size)
     return nest_map(lambda value: value[indices], values)
+
+
+class Softsign_(torch.autograd.Function):
+    r"""softsign(input) -> Tensor
```
hnyu: Try to rephrase this in a sentence because it will look better in a doc.

alf/utils/math_ops.py
emailweixu: Right. It relies on AD to do it. And because it involve several steps of computation. All the intermediate result need to be kept.

hnyu (approved):

alf/utils/math_ops.py
```diff
@@ -198,3 +198,59 @@ def shuffle(values):
...
+    implementation.
+
+    Reference:
+    `PyTorch: Defining New Autograd Functions
+    <https://pytorch.org/tutorials/beginner/examples_autograd/two_layer_net_custom_function.html>`_
+    """
+
+    @staticmethod
+    def forward(ctx, input):
+        output = torch.div(input, input.abs() + 1, out=input)
```
Haichao-Zhang, line 221: Just curious, what is ```, out=input)``` used for?

emailweixu: the result of the operation will be written to input

PR #547, opened 2020-04-22
PR output destination folder name in logging info

Haichao-Zhang:
Adding the destination folder name into logging info to make it more informative. 
This is useful in cases when running multiple training experiments using the same gin file.

### DIFF
--- a/alf/trainers/policy_trainer.py
+++ b/alf/trainers/policy_trainer.py
@@ -164,13 +164,13 @@ class Trainer(object):
             t = time.time() - t0
             logging.log_every_n_seconds(
                 logging.INFO,
-                '%s: %s time=%.3f throughput=%0.2f' %
-                (common.get_gin_file(), iter_num, t, int(train_steps) / t),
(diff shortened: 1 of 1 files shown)


alf/trainers/policy_trainer.py
```diff
@@ -164,13 +164,13 @@ def _train(self):
             t = time.time() - t0
             logging.log_every_n_seconds(
                 logging.INFO,
-                '%s: %s time=%.3f throughput=%0.2f' %
-                (common.get_gin_file(), iter_num, t, int(train_steps) / t),
+                '%s -> %s: %s time=%.3f throughput=%0.2f' %
+                (common.get_gin_file(), [os.path.basename(self._root_dir)],
```
hnyu, line 168: Why do you need a list around the directory here?

Haichao-Zhang: Just to make the printed information looks good, more symmetric around ```->```.

hnyu (approved):

PR #549, opened 2020-04-23
parallel networks for SAC

hnyu:


emailweixu:
Currently, QNetwork does not implement make_parallel. Thus for discrete actions, use_parallel_network does not have any effect.

emailweixu:
> Currently, QNetwork does not implement make_parallel. Thus for discrete actions, use_parallel_network does not have any effect.

Can be dismissed. You actually implemented ParallelQNetwork

alf/layers.py
```diff
@@ -126,6 +126,11 @@ def __init__(self,
             bias_init_value (float): a constant
         """
         super(FC, self).__init__()
+        # get the argument list with vals
+        self._kwargs = locals()
```
emailweixu: suggest to make a copy of locals(). Otherwise, future addition of local variable will change self._kwargs

hnyu: updated

alf/networks/critic_networks.py
```diff
@@ -171,19 +174,18 @@ class ParallelCriticNetwork(Network):
...
-                 name='ParallelCriticNetwork'):
+                 name="ParallelCriticNetwork"):
         """
         It create a parallelized version of ``critic_network``.
 
         Args:
             critic_network (CriticNetwork): non-parallelized critic network
             n (int): make ``n`` replicas from ``critic_network`` with different
                 initialization.
-            name (str): name of this network
```
Haichao-Zhang, line 182: why remove this comment?

alf/networks/q_networks.py
```diff
@@ -20,14 +20,14 @@
...
-from alf.networks import EncodingNetwork, LSTMEncodingNetwork
-from alf.networks import PreprocessorNetwork
+from alf.networks import EncodingNetwork, LSTMEncodingNetwork, ParallelEncodingNetwork
+from alf.networks import Network
 from alf.tensor_specs import TensorSpec, BoundedTensorSpec
 import alf.utils.math_ops as math_ops
 
 
 @gin.configurable
-class QNetwork(PreprocessorNetwork):
```
Haichao-Zhang, line 30: We need to derive from ```PreprocessorNetwork``` instead of ```Network```?

emailweixu (approved):

alf/networks/critic_networks.py
hnyu: It's quite obvious

alf/networks/q_networks.py
hnyu: preprocessing is handled by `self._encoding_net`; so no need to derive from that

PR #550, opened 2020-04-23
Improve checkpoint loading related information

Haichao-Zhang:
Improve the error information so that it is more informative when checkpoint loading is failed.
Also updated the readme on the proper setting of ```root_dir```.

### DIFF
--- a/alf/trainers/policy_trainer.py
+++ b/alf/trainers/policy_trainer.py
@@ -213,7 +213,15 @@ class Trainer(object):
             algorithm=self._algorithm,
             metrics=nn.ModuleList(self._algorithm.get_metrics()))
 
-        recovered_global_step = checkpointer.load()
+        try:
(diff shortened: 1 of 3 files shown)


alf/trainers/policy_trainer.py
```diff
@@ -213,7 +213,15 @@ def _restore_checkpoint(self):
             algorithm=self._algorithm,
             metrics=nn.ModuleList(self._algorithm.get_metrics()))
 
-        recovered_global_step = checkpointer.load()
+        try:
```
emailweixu, line 216: Do we still have  the original error message about which parameter is the problem? It is still useful for debugging in some cases.

Haichao-Zhang: Good point, Now also appended the original error message in the new commit. 
So now we have both the high-level guidance and low-level details.

emailweixu (approved):

PR #551, opened 2020-04-24
summarize env info

hnyu:
Allow users return extra env info (e.g., "success", "number of rooms", etc) that's meaningful for episodic accumulation, and summarize it with `AverageEnvInfoMetric`. Did a refactor of metrics so that some common base code can be reused. 

Since now `TimeStep` has new fields, I've turned off the ``flatten`` flag when creating environments so that the time steps of different envs won't be flattened before being stacked together. The reason is that with ``flatten=True``, ``self._time_step_spec`` has to be consistent with `TimeStep`, which is hard to track in the code. The benefit of ``flatten=True`` is unclear yet at this point, so disabling it for now.


alf/environments/parallel_torch_environment.py
```diff
@@ -155,7 +153,14 @@ def _stack_time_steps(self, time_steps):
         """Given a list of TimeStep, combine to one with a batch dimension."""
         if self._flatten:
             return nest.fast_map_structure_flatten(
-                lambda *arrays: torch.stack(arrays), self._time_step_spec,
+                # TODO: right now ``self._time_step_spec`` is independenty defined
```
le-horizon, line 159: Let's throw an error before return, to make sure this is not called?

alf/environments/suite_robotics.py
```diff
@@ -0,0 +1,145 @@
...
+            env_id=env_id,
+            discount=discount,
+            max_episode_steps=max_episode_steps,
+            gym_env_wrappers=gym_env_wrappers,
+            torch_env_wrappers=torch_env_wrappers)
+
+    # concat robot's observation and the goal location
+    env = FlattenDictWrapper(env, ["observation", "desired_goal"])
+    env = SuccessWrapper(env)
+    #env = SparseReward(env)
```
le-horizon: Maybe add a config option for this?

alf/environments/torch_environment.py
```diff
@@ -141,50 +143,38 @@ def current_time_step(self):
...
 
         Returns:
-            A `TimeStep` namedtuple containing:
-                step_type: a `Tensor` of `StepType` enum values.
-                reward: a `Tensor` of reward values from executing 'prev_action'.
-                discount: A discount value in the range `[0, 1]`.
-                observation: A (nested) 'Tensor' for observation
-                prev_action: A (nested) 'Tensor' for action from previous time step
-                env_id: the ID of the environment from which this time_step is
+            TimeStep:
```
le-horizon, line 153: This looks weird when coding":"
Is there a way to improve over this?

alf/environments/torch_environment.py
```diff
@@ -141,50 +143,38 @@ def current_time_step(self):
...
 
         Returns:
-          A `TimeStep` namedtuple containing:
-              step_type: a `Tensor` of `StepType` enum values.
-              reward: a `Tensor` of reward values from executing 'prev_action'.
-              discount: A discount value in the range `[0, 1]`.
-              observation: A (nested) 'Tensor' for observation
-              prev_action: A (nested) 'Tensor' for action from previous time step
-              env_id: the ID of the environment from which this time_step is
+            TimeStep:
```
le-horizon, line 177: same here?

alf/environments/suite_robotics.py
```diff
@@ -0,0 +1,145 @@
...
+
+def is_available():
+    return mujoco_py is not None
+
+
+class SparseReward(gym.Wrapper):
+    """Convert the original :math:`-1/0` rewards to :math:`0/1`.
+    """
+
+    def __init__(self, env):
```
le-horizon, line 50: maybe use threshold = 0.5 as an argument, < thresh gives 0, above gives 1?
This makes the wrapper less task dependent, and also avoids potential problems with float comparison.

alf/examples/sac_fetchreach.gin
```diff
@@ -0,0 +1,57 @@
+
+include 'sac.gin'
+
+import alf.environments.suite_socialbot
```
le-horizon: Is this used?

alf/metrics/metrics.py
```diff
@@ -18,57 +18,44 @@
...
-        self._head.add_(1)
+        self._buf.add_batch(value)
 
-    @property
-    def length(self):
-        return torch.min(self._head, self._max_len)
+    def mean(self):
+        if self._buf.current_size == 0:  # avoid nan
+            return torch.as_tensor(0, dtype=self._dtype)
+        return self._buf.get_all()[:self._buf.current_size].mean()
```
le-horizon, line 55: `self._buf.get_batch(self._buf.current_size).mean()` is perhaps safer, in case some entries get deleted from buffer.

get_all simply returns the whole buffer, and probably shouldn't be exposed for external use.

le-horizon (commented):
Nice refactor and changes Haonan!  Mostly good, some nits.

alf/environments/torch_environment.py
hnyu: What do you mean?

alf/environments/suite_robotics.py
hnyu: This wrapper is only meant for openai Robotics environment. it shouldn't be used for other envs. https://openai.com/blog/ingredients-for-robotics-research/

alf/examples/sac_fetchreach.gin
hnyu: removed

alf/metrics/metrics.py
hnyu: get_batch will randomly sample with replacement. Here I need to average over all elements in the buffer. And DataBuffer can only overflow data but not delete data.

alf/environments/parallel_torch_environment.py
hnyu: added

alf/environments/suite_robotics.py
hnyu: added

alf/environments/torch_environment.py
le-horizon: The comment ends with ":".  Looks incomplete.  Is that for formatting?  Maybe I'll just have to get used to it..

alf/metrics/metrics.py
le-horizon: You are right, let's leave as is then.

le-horizon (commented):
Some replies.  Btw, I don't see your recent change?

alf/environments/torch_environment.py
hnyu: The correct format for "Returns" is `type: description`. When the type says everything we can just ignore the `description`, but ':' is still needed in order for sphinx to interpret the token before ':' is a type. Otherwise it's just plain text. You can just take a look at other similar cases and see how it shows in the rendered page.

le-horizon (dismissed):
LGTM.  Nice refactor and improvements!

alf/environments/torch_gym_wrapper.py
```diff
@@ -129,17 +129,28 @@ def __init__(self,
             self._gym_env.observation_space, simplify_box_bounds)
         self._action_spec = tensor_spec_from_gym_space(
             self._gym_env.action_space, simplify_box_bounds)
-        self._flat_obs_spec = nest.flatten(self._observation_spec)
         self._time_step_spec = ds.time_step_spec(self._observation_spec,
                                                  self._action_spec)
         self._info = None
         self._done = True
+        self._zero_info = self._obtain_zero_info()
```
emailweixu, line 136: need to reset() after this?

alf/examples/sac_fetchreach.gin
```diff
@@ -0,0 +1,56 @@
...
+
+# training config
+TrainerConfig.temporally_independent_train_step=True
+TrainerConfig.initial_collect_steps=3000
+TrainerConfig.mini_batch_length=2
+TrainerConfig.mini_batch_size=256
+TrainerConfig.unroll_length=1
+TrainerConfig.num_updates_per_train_step=5
+TrainerConfig.whole_replay_buffer_training=False
+TrainerConfig.clear_replay_buffer=False
```
emailweixu: Shouldn't these two be in sac.gin?

emailweixu (approved):

Issue #552, opened 2020-04-24
Naive/Parallel Networks load_state_dict bug


jesbu1:
In the latest pytorch branch, when using a parallel network and the `observation_input_preprocessors` includes an `EmbeddingPreprocessor`,  this error occurs:

`  File "/home/jessezhang/alf-private/alf/utils/common.py", line 150, in get_target_updater
    target_model.load_state_dict(model.state_dict())
  File "/home/jessezhang/miniconda3/envs/alf-private/lib/python3.6/site-packages/torch/nn/modules/module.py", line 830, in load_state_dict
    self.__class__.__name__, "\n\t".join(error_msgs)))
RuntimeError: Error(s) in loading state_dict for NaiveParallelNetwork:
        Missing key(s) in state_dict: "_networks.1._obs_encoder._input_preprocessor_modules.0._embedding_net.weight".`

It seems the input_preprocessor embedding_net is not copied over correctly when making it parallel.

jesbu1:
@Haichao-Zhang @hnyu please take a look at this, thanks!

Haichao-Zhang:
Looking into it @jesbu1

Haichao-Zhang:
For now you can simply pull https://github.com/HorizonRobotics/alf/pull/554 and you should be able to train. Feel free to reopen if you encounter any further problems.

Haichao-Zhang:
There might be  some other issues and working on it.

Haichao-Zhang:
#554 is one way to solve the reported problem and I have verified on my self-constructed problem.

PR #554, opened 2020-04-25
Parameter copy in the presence of duplicates

Haichao-Zhang:
Pytorch ```load_state_dict()``` handles parameters shared among multiple modules by duplicating them with names with different prefixes. ```state_dict()``` is implemented in a similar way.

ALF handle the shared parameters with only a single copy by having its own version of ```load_state_dict```/ ```state_dict``` and have only overridden the ```state_dict``` function of ```nn.Module``` previously.

This was not a problem when we call both functions from ```alf.algorithm``` with shared parameters included in its sub-modules/networks, as our own version of ```load_state_dict``` and ```state_dict``` will be used. But in other cases, there might be problems.
We have now overridden both functions of ```nn.Module``` to ensure they are paired.

The problem reported in #552 is caused by the issue explained above. More concretely, making parallel ```EncodingNetwork``` in the presence of  ```input_preprocessor``` will use the ```NaivePrallelNetwork```. In this case, the ```input_preprocessor``` is a module that is shared among all naive network replicas. And calling ```load_state_dict```/```state_dict``` in ```get_target_updater``` used an unpaired version of them.

### DIFF
--- a/alf/algorithms/algorithm.py
+++ b/alf/algorithms/algorithm.py
@@ -641,6 +641,7 @@ class Algorithm(nn.Module):
 
         return destination
 
+    @common.add_method(nn.Module)
     def load_state_dict(self, state_dict, strict=True):
         """Load state dictionary for the algorithm.
 
--- a/alf/utils/checkpoint_utils_test.py
+++ b/alf/utils/checkpoint_utils_test.py
@@ -12,21 +12,31 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from absl.testing import parameterized
+from collections import OrderedDict
+import numpy as np
+import functools
 import json
-import warnings
+import os
 import shutil
 import tempfile
-import os
-import numpy as np
-from collections import OrderedDict
+import warnings
 
-import alf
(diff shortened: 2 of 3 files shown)


hnyu:
Why are the parameters of preprocessors not included in state_dict()?

Haichao-Zhang:
It is included, just not the duplicated ones.
Will push a an alternative solution in an update commit.

hnyu (dismissed):

emailweixu:
Shouldn't NaivePralleNetwork make n copies of the input preprocessors?

Haichao-Zhang:
> Shouldn't NaivePralleNetwork make n copies of the input preprocessors?

That can be done in another PR if that is a desired feature.
Currently, it is not. As the ```input_preprocessor``` is one input argument of the ```EncodingNetwork/PreprocessorNetwork```, ```NaivePrallelNetwork```
 will use the captured input parameters for constructing each replica of the base network.
In this case, only the same captured```input_preprocessor``` is used for each replica.

emailweixu (approved):

PR #555, opened 2020-04-25
Fix the output size issue of parallel encoding network

Haichao-Zhang:
For the ```ParallelEncodingNetwork```, when only specifying its ```input_tensor_spec``` and ```n``` values (e.g. ```ParallelEncodingNetwork(input_tensor_spec, n)```), we will get a dummy network that has no actual layers. And we can later use it as ```output, _= ParallelEncodingNetwork(input)```.

In this case, the ```output``` should be properly adjusted from the ```input``` taking ```n``` into consideration. By doing this, 
* we will have the ```output``` shape that is consistent with the ```_output_spec``` property of ```ParallelEncodingNetwork```,
* the shape of the ```output``` will be the same as the one from a ```NaiveParallelNetwork```.

### DIFF
--- a/alf/networks/encoding_networks.py
+++ b/alf/networks/encoding_networks.py
@@ -554,8 +554,7 @@ class ParallelEncodingNetwork(PreprocessorNetwork):
         z, state = super().forward(inputs, state, max_outer_rank=2)
         if len(self._fc_layers) == 0:
             if inputs.ndim == 2:
-                z = z.unsqueeze(0).expand(self._n, *z.shape)
-                z = z.transpose(0, 1)
+                z = z.unsqueeze(1).expand(-1, self._n, *z.shape[1:])
         else:
             for fc in self._fc_layers:
                 z = fc(z)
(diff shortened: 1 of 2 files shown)


alf/networks/encoding_networks.py
```diff
@@ -551,8 +552,13 @@ def forward(self, inputs, state=()):
         """
         # call super to preprocess inputs
         z, state = super().forward(inputs, state, max_outer_rank=2)
-        for fc in self._fc_layers:
-            z = fc(z)
+        if len(self._fc_layers) == 0:
+            if inputs.ndim == 2:
+                z = z.unsqueeze(0).expand(self._n, *z.shape)
```
hnyu: `z.unsqueeze(1).expand(-1, self._n, *z.shape[1:])` to avoid an additional transpose

Haichao-Zhang: Changed

hnyu (approved):

PR #556, opened 2020-04-25
support mixed actions

hnyu:
Support mixed actions for ActorDistributionNetwork and CriticNetwork. 

SAC can now support mixed policies. Other algorithms are to be updated accordingly.


alf/algorithms/sac_algorithm.py
```diff
@@ -290,32 +287,71 @@ def actor_loss_fn(dqda, action):
...
+                    loss += alpha * log_pi
+                    return loss
+
+                return nest.map_structure(actor_loss_fn, dqda, action, log_pi)
+
+            def _train_discrete_actor(log_pi):
+                def actor_loss_fn(log_pi):
+                    # See ``notes/estimating_derivative_of_expectation.rst``
+                    # for how this loss formula is derived.
+                    loss = log_pi * (alpha * log_pi - target_q_value).detach()
```
emailweixu: does this make a difference compared to the previous version?

hnyu: On carpole, this version will converge slower than using a QNetwork, but eventually got 200. 

This version:
Screen Shot 2020-04-28 at 11 35 31 AM

Previous:
Screen Shot 2020-04-28 at 11 21 16 AM

alf/algorithms/sac_algorithm.py
```diff
@@ -248,40 +258,27 @@ def rollout_step(self, time_step: TimeStep, state):
...
-            actor_loss += alpha * log_pi
-        else:
-            critics, critics_state = self._critic_networks(
-                exp.observation, state=state.critics)
+        critics, critics_state = self._compute_critics(
+            self._critic_networks, exp.observation, action, state.critics)
+        alpha = torch.exp(self._log_alpha).detach()
 
+        if self._use_q_network:
+            # Make the entire Categorial dist shape close to the target Q value
```
Haichao-Zhang: Categorial -> Categorical

docs/notes/estimating_derivative_of_expectation.rst
```diff
@@ -0,0 +1,167 @@
...
+
+    \begin{array}{ll}
+        \frac{\partial L_{\theta}}{\partial \theta} &= \sum_k
+        \left[\frac{\partial p_{\theta}(k|s)}{\partial \theta}(\alpha\log
+        p_{\theta}(k|s) - Q(s,k)) + \alpha p_{\theta}(k|s)\frac{\partial \log
+        p_{\theta}(k|s)}{\partial \theta}\right]\\
+        &=\sum_k
+        \left[p_{\theta}(k|s)\frac{ \partial \log p_{\theta}(k|s)}{\partial
+        \theta}(\alpha\log p_{\theta}(k|s) - Q(s,k))\right] + \underbrace{\alpha\sum_k\left[
+        p_{\theta}(k|s)\frac{\partial \log p_{\theta}(k|s)}{\partial
```
Haichao-Zhang: Can you add some explanation in the doc on the second term why it is zero?

alf/algorithms/sac_algorithm.py
Haichao-Zhang: To handle a mixed action case, it seems possible to use the original critic network for continuous part of the actions, and the q-network for discrete part of the actions,  if that is beneficial for the final performance?

hnyu: If the critic network and q-network are separate, then learning either of them will be non-stationary because the reward is decided by both actions and the other network is always changing (multi-agent learning). It might work if these two networks share some early layers, but it seems to me that the sharing is not straightforward.

alf/algorithms/sac_algorithm.py
```diff
@@ -248,40 +258,27 @@ def rollout_step(self, time_step: TimeStep, state):
...
-            actor_loss += alpha * log_pi
-        else:
-            critics, critics_state = self._critic_networks(
-                exp.observation, state=state.critics)
+        critics, critics_state = self._compute_critics(
+            self._critic_networks, exp.observation, action, state.critics)
+        alpha = torch.exp(self._log_alpha).detach()
 
+        if self._use_q_network:
+            # Make the entire Categorical dist shape close to the target Q value
```
emailweixu: For this case, p(a|s) have a close form solution p(a|s) = exp(Q(a|s) / alpha), which should be much better. Because of this, for mixed case, it's better to model the q values as Q(a_d|a_c, s), where a_c is continuous action, and a_d is discrete action.

hnyu: Yeah adding a_c as a conditional input to Q will theoretically avoid the non-stationarity issue. But I guess we need to do some experiment to see how this compares with the common critic solution.

I think the more difficult-to-handle case is that we have multiple discrete actions? Unless we manually merge them into one big cross product, then it's not easy to form conditional Q values.

alf/algorithms/sac_algorithm.py
emailweixu: Similar to policy gradient, subtracting a base might be helpful for convergence because of lower variance

alf/algorithms/sac_algorithm.py
emailweixu: For the pure discrete case, it's almost certainly better because there is no need to train an actor network. Thus it's likely to better for the mixed case if the discrete action is a scalar.
Currently, out main use cases for discrete action is still the scalar. We can deal with multidimensional discrete action when we actually encounter it.

hnyu: The actor network as a function approximator still has to be trained across different states s if we are sampling actions from it, even though there is a closed form solution for every given Q(a_d|s). It's just that for every training loss, the entire distribution is considered because we could enumerate all the probability entries since it's discrete and finite. SGD is still needed to gradually adjust the actor given Q values. I think the inefficiency of having a critic network for discrete actions is mainly due to the empirical estimation of the actor loss instead of an accurate calculation of the actor loss. So the question is really how much error will be introduced by this empirical estimation vs. introduced by a conditional Q(a_d|a_c,s).

But I think for discrete actions in our current SAC implementation, we could get rid of the actor network (or make a fake one) so that we always sample from the Q values using exp(Q(a|s) / alpha)?

emailweixu: For discrete case, we don't need actor loss because the action distribution is exp(Q(a|s) / alpha). For mixed case, the loss is only for the continuous action a_c. Once a_c is sampled, a_d is based on exp(Q(a_d|a_c, s)/alpha).

hnyu: Still, the learning of discrete actions will make the learning of continuous actions non-stationary because the dependency is one-way in Q(a_d|a_c,s). That is, the same continuous action made under the same state would generate different rewards and critic values. Do you think this is a problem? 

BTW, this kind of implementation Q(a_d|a_c, s) is essentially equivalent to having two separate SAC algorithms with continuous and discrete actions respectively in an Agent. The second SAC algorithm takes  the first one's output as part of the inputs. Seems that there is no need to merge them into one SACAlgorithm.

hnyu:
Now removed actor training for discrete actions. On sac_cart_pole.gin, its performance is the same as before. Haven't tested on games that have a large number of discrete actions.

image

alf/algorithms/sac_algorithm.py
```diff
@@ -32,23 +33,21 @@
...
 from alf.networks import ActorDistributionNetwork, CriticNetwork
+from alf.networks import QNetwork, QRNNNetwork
 from alf.tensor_specs import TensorSpec, BoundedTensorSpec
 from alf.utils import losses, common, dist_utils, math_ops
 
-SacShareState = namedtuple("SacShareState", ["actor"])
+ActionType = Enum('ActionType', ('Discrete', 'Continuous', 'Mixed'))
 
-SacActorState = namedtuple("SacActorState", ["critics"])
+SacActionState = namedtuple(
```
hnyu, line 43: Simplified state and info structures. Before was too cluttered.

alf/algorithms/sac_algorithm.py
```diff
@@ -168,20 +199,26 @@ def __init__(self,
...
         log_alpha = nn.Parameter(torch.Tensor([float(initial_log_alpha)]))
 
         super().__init__(
             observation_spec,
             action_spec,
             train_state_spec=SacState(
-                share=SacShareState(actor=actor_network.state_spec),
-                actor=SacActorState(critics=critic_networks.state_spec),
+                action=SacActionState(
+                    actor_network=(() if self._act_type == ActionType.Discrete
```
emailweixu, line 220: It's unexpected that self._act_type is specified in _make_parallel_critic_networks()

alf/algorithms/sac_algorithm.py
```diff
@@ -228,168 +261,293 @@ def __init__(self,
...
+                discrete_action = dist_utils.epsilon_greedy_sample(
+                    discrete_action_dist, epsilon_greedy)
+            else:
+                discrete_action = dist_utils.sample_action_distribution(
+                    discrete_action_dist)
+
+        if self._act_type == ActionType.Mixed:
+            action_dist = type(self._action_spec)((discrete_action_dist,
+                                                   continuous_action_dist))
+            action = type(action_dist)((discrete_action, continuous_action))
```
emailweixu: why not `type(self._action_spec)`? Feel it's easier to understand without the unnecessary dependency on action_dist

alf/algorithms/sac_algorithm.py
```diff
@@ -228,168 +261,293 @@ def __init__(self,
...
+            discrete_action_dist = td.Categorical(logits=logits)
+            if eps_greedy_sampling:
+                discrete_action = dist_utils.epsilon_greedy_sample(
+                    discrete_action_dist, epsilon_greedy)
+            else:
+                discrete_action = dist_utils.sample_action_distribution(
+                    discrete_action_dist)
+
+        if self._act_type == ActionType.Mixed:
+            action_dist = type(self._action_spec)((discrete_action_dist,
```
emailweixu, line 398: This is not the actual distribution. The way this is constructed make discrete_action and continuous action independent with each other given the observation.

alf/algorithms/sac_algorithm.py
```diff
@@ -228,168 +261,293 @@ def __init__(self,
...
+        elif self._act_type == ActionType.Mixed:
+            observation = (observation, action[1])  # continuous action
+        # discrete/mixed: critics shape [B, replicas, num_actions]
+        # continuous: critics shape [B, replicas]
+        critics, critics_state = critic_net(observation, state=critics_state)
+        return critics, critics_state
+
+    def _actor_train_step(self, exp: Experience, state, action, log_pi_sum):
+        if self._act_type == ActionType.Discrete:
+            # Pure discrete case doesn't need learn an actor network
```
emailweixu: need => need to

alf/algorithms/sac_algorithm.py
```diff
@@ -228,168 +261,293 @@ def __init__(self,
...
+            return (), ()
+
+        critics, critics_state = self._compute_critics(
```
emailweixu: There seems to be duplicated computation for this because for mixed and discrete action, when the action is generated, this is already computed.

alf/algorithms/sac_algorithm.py
```diff
@@ -228,168 +261,293 @@ def __init__(self,
...
-            action = dist_utils.rsample_action_distribution(
-                action_distribution)
-        else:
-            action = dist_utils.sample_action_distribution(action_distribution)
+        (action_distribution, action, action_state) = self._predict_action(
+            exp.observation, state=state.action)
+
+        log_pi = nest.map_structure(dist_utils.compute_log_probability,
+                                    action_distribution, action)
+        log_pi_sum = math_ops.add_n(nest.flatten(log_pi))
```
emailweixu: I feel it's better to keep using the name `log_pi`. The meaning is still correct, it is `\log \pi(a|s)`

alf/algorithms/sac_algorithm.py
```diff
@@ -228,168 +261,293 @@ def __init__(self,
...
+                dqda = torch.clamp(dqda, -self._dqda_clipping,
+                                   self._dqda_clipping)
+            loss = 0.5 * losses.element_wise_squared_loss(
+                (dqda + action).detach(), action)
+            return loss.sum(list(range(1, loss.ndim)))
+
+        actor_loss = nest.map_structure(actor_loss_fn, dqda, action)
+        alpha = torch.exp(self._log_alpha).detach()
+        actor_loss = (
+            math_ops.add_n(nest.flatten(actor_loss)) + alpha * log_pi_sum)
```
emailweixu: log_pi from discrete action should be detached.

alf/algorithms/sac_algorithm.py
hnyu: I agree. The discrete action distribution is only valid for a particular continuous action. If another continuous action is sampled from the continuous distribution, then it's no longer valid. The distributions returned will only be used for computing logP and summary purpose. The action pair however is still valid. 

For computing logP, given that specific pair of actions, the probability should be correct.

I don't see any easier way of returning a valid joint distribution object here, unless we define a new distribution class that wraps this whole action sampling process using Q values inside.

alf/algorithms/sac_algorithm.py
hnyu: Good catch. Actually because I always detach the Q values (as logits) when constructing the discrete distribution (it's never trained), so this makes no difference.

alf/utils/common.py
```diff
@@ -378,6 +380,47 @@ def _process(line):
     return '\n'.join(output_lines)
 
 
+@gin.configurable
+class ObservationNormalizer(object):
```
emailweixu: Should this need to be checkpointable?

alf/algorithms/sac_algorithm.py
```diff
@@ -228,102 +261,252 @@ def __init__(self,
...
+            critics_state = ()
+
+        if self._act_type == ActionType.Mixed:
+            # pick values according to discrete actions
+            critics = self._select_q_value(action[0], critics).squeeze(-1)
+            action = action[1]
+
+        target_q_value = critics.min(dim=1)[0]
+
+        dqda = nest.pack_sequence_as(
```
emailweixu: This is worth to be a helper function:  ``grad(nest, objective)``. It's used at several places: ddpg, sarsa, sac.

alf/algorithms/sac_algorithm.py
```diff
@@ -228,102 +261,252 @@ def __init__(self,
...
+                dqda = torch.clamp(dqda, -self._dqda_clipping,
+                                   self._dqda_clipping)
+            loss = 0.5 * losses.element_wise_squared_loss(
+                (dqda + action).detach(), action)
+            return loss.sum(list(range(1, loss.ndim)))
+
+        actor_loss = nest.map_structure(actor_loss_fn, dqda, action)
+        alpha = torch.exp(self._log_alpha).detach()
+        actor_loss = (
+            math_ops.add_n(nest.flatten(actor_loss)) + alpha * log_pi)
```
emailweixu: as we discussed, having neg_entropy as a separate loss item is informative.

hnyu: I've put entropy in SacAlphaInfo because discrete actions won't have an actor loss

alf/algorithms/sac_algorithm.py
```diff
@@ -523,7 +520,7 @@ def _critic_train_step(self, exp: Experience, state: SacCriticState,
     def _alpha_train_step(self, log_pi):
         alpha_loss = self._log_alpha * (
             -log_pi - self._target_entropy).detach()
-        return alpha_loss
+        return SacAlphaInfo(loss=alpha_loss, entropy=-log_pi)
```
emailweixu: Other losses (e.g. ActorCriticLoss) use neg_entropy. Let's keep them consistent.

alf/algorithms/sac_algorithm.py
emailweixu: I think we should not include alpha*log_pi in actor_loss. It makes it hard to interpret the curve of actor_loss. Without log_pi in actor loss, small actor loss value means that dqda is small, which means the action is close to local maximum of the critic function.

emailweixu (approved):

PR #558, opened 2020-04-28
Fix the parallel net name error and some other changes

Haichao-Zhang:
Previously, if we construct ```EncodingNetwork``` with an explicit ```name``` argument as ```network = EncodingNetwork(... name='enc_net')```, there will be issues later when we call ```pnet = network.make_parallel(n)```.

Also did some change to improve the way to name a ```NaiveParallelNetwork```,  so when we call ```nnet = alf.networks.network.NaiveParallelNetwork(network)```, the name of ```nnet``` will be adapted to the name of the input ```network```.

### DIFF
--- a/alf/networks/network.py
+++ b/alf/networks/network.py
@@ -202,7 +202,7 @@ class Network(nn.Module):
 class NaiveParallelNetwork(Network):
     """Naive implementation of parallel network."""
 
-    def __init__(self, network, n, name="naive_parallel"):
+    def __init__(self, network, n, name=None):
         """
         A parallel network has ``n`` copies of network with the same structure but
         different indepently initialized parameters.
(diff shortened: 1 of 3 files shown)


alf/networks/network.py
```diff
@@ -215,18 +215,21 @@ def __init__(self, network, n, name="NaiveParallelNetwork"):
             network (Network): the parallel network will have ``n`` copies of
                 ``network``.
             n (int): ``n`` copies of ``network``
+            name(str): a string that will be added as a prefix to the name of
```
emailweixu: It's better to keep the semantics of name unchanged. The caller still have the control of the name. We make the name default to `None` to be `name + '_' network.name`

Haichao-Zhang: Right, the expected behavior is more natural in this way. Changed as suggested.

emailweixu (approved):

PR #560, opened 2020-05-01
Preprocessor copy

Haichao-Zhang:
This PR makes the following changes:
1) It derives ```Preprocessor``` from ```alf.Network``` instead of ```nn.Module```, to enable a consistent behavior as other types of networks.
2) We now convert back to using an instance of ```Preprocessor``` (```proc_inst```) as input to the constructor of ```PreprocessorNetwork``` and the ```proc_inst.copy()``` function is called during construction.  Therefore, when calling ```proc_network.copy()```, the corresponding ```proc_inst``` is also copied by default.
3) If the ```Preprocessor``` instance (or other types networks) is to be shared, we can now call ```proc_inst.singleton()``` before passing it to the constructor of ```PreprocessorNetwork```.

A possible example for using a shared preprocessor would be:
```
input_preprocessors={
    "camera": @ResnetEncodingNetwork(
        input_tensor_spec=alf.nest.get_field(%observation_spec, "camera")).singleton(),
    "collision": SequentialNetwork(torch.nn.Flatten()),
    "gnss": SequentialNetwork(torch.nn.Flatten()),
    "imu": SequentialNetwork(torch.nn.Flatten()),
    "goal": SequentialNetwork(torch.nn.Flatten()),
    "speed": SequentialNetwork(torch.nn.Flatten()),
}
```


emailweixu:
If preprocessor is a Network, it can support copy even if it is contructed.

Haichao-Zhang:
> If preprocessor is a Network, it can support copy even if it is contructed.

That's true when it is used alone. When we use a constructed preprocessor instance as an input to construct another network as ```proc_inst=InputPreprocessor(), net=Network(...,proc_inst)```, when we call ```net2=net.copy()``` later, the same ```proc_inst``` is shared between ```net``` and ```net2```.

But as you suggested, there could be alternative ways to handle this case by leveraging ```proc_inst.copy()```.

emailweixu:
I thought about this again when I was trying to write the gin config for Carla, which has a lot of different observations. I feel that use Network instance as input processor is more convenient so that I write something like:
```python
input_preprocessors={
    "camera": @ResnetEncodingNetwork(
        input_tensor_spec=alf.nest.get_field(%observation_spec, "camera")),
    "collision": SequentialNetwork(torch.nn.Flatten()),
    "gnss": SequentialNetwork(torch.nn.Flatten()),
    "imu": SequentialNetwork(torch.nn.Flatten()),
    "goal": SequentialNetwork(torch.nn.Flatten()),
    "speed": SequentialNetwork(torch.nn.Flatten()),
}
```
where SequentialNetwork is similar to torch.nn.Sequential

alf/networks/preprocessors.py
```diff
@@ -129,3 +129,110 @@ def _preprocess(self, tensor):
...
+                # preprocessing. If it does change, then you should consider
+                # defining an `InputPreprocessor` instead.
+                return spec
+            self._input_preprocessor_modules.append(preproc)
+            return preproc(spec)
+
+        self._input_preprocessors = None
+        if input_preprocessor_ctors is not None:
+            input_preprocessor_ctors = alf.nest.pack_sequence_as(
+                input_tensor_spec, alf.nest.flatten(input_preprocessor_ctors))
```
emailweixu: what is the purpose of this line? Shouldn't input_preprocessor_ctors has same structure as input_tensor_spec?

Haichao-Zhang:
> I thought about this again when I was trying to write the gin config for Carla, which has a lot of different observations. I feel that use Network instance as input processor is more convenient so that I write something like:
> 
> ```python
> input_preprocessors={
>     "camera": @ResnetEncodingNetwork(
>         input_tensor_spec=alf.nest.get_field(%observation_spec, "camera")),
>     "collision": SequentialNetwork(torch.nn.Flatten()),
>     "gnss": SequentialNetwork(torch.nn.Flatten()),
>     "imu": SequentialNetwork(torch.nn.Flatten()),
>     "goal": SequentialNetwork(torch.nn.Flatten()),
>     "speed": SequentialNetwork(torch.nn.Flatten()),
> }
> ```
> 
> where SequentialNetwork is similar to torch.nn.Sequential

I have now changed back to using instance as input to make it easier to config.

alf/networks/preprocessors.py
```diff
@@ -22,10 +22,11 @@
 import alf
 from alf.tensor_specs import TensorSpec, BoundedTensorSpec
 from alf.nest.utils import get_outer_rank
+from alf.networks.network import Network, SingletonInstanceNetwork
 import alf.utils.math_ops as math_ops
 
 
-class InputPreprocessor(nn.Module):
+class InputPreprocessor(Network):
```
emailweixu, line 29: Since it's Network now, forward() should follow the interface for Network, i.e., return output, state

alf/networks/preprocessors.py
```diff
@@ -129,3 +129,161 @@ def _preprocess(self, tensor):
...
+
+    def _preprocess(self, tensor):
+        assert get_outer_rank(tensor, self._input_tensor_spec) == 1, \
+            "Only supports one outer rank (batch dim)!"
+        ret = self._embedding_net(tensor)
+        # EncodingNetwork returns a pair
+        return (ret if self._input_tensor_spec.is_discrete else ret[0])
+
+
+class PreprocessorNetwork(Network):
```
hnyu: It's not proper to move PreprocessorNetwork here in preprocessors.py. PreprocessorNetwork is a network that has preprocessors, but itself is not a preprocessor.

alf/networks/preprocessors.py
```diff
@@ -129,3 +129,161 @@ def _preprocess(self, tensor):
         ret = self._embedding_net(tensor)
         # EncodingNetwork returns a pair
         return (ret if self._input_tensor_spec.is_discrete else ret[0])
+
+
+@gin.configurable
+class SharedEmbeddingPreprocessor(InputPreprocessor, SingletonInstanceNetwork):
```
hnyu: This inheritance is confusing. InputPreprocessor is not singleton while SingletonInstanceNetwork is. And mostly what this class does is EmbeddingPreprocessor's thing. Is there any better way of constructing this class?

It seems to me that there is no need for the existence of SingletonInstanceNetwork which makes the inheritance complex. We can directly assign the flag `self._singleton_instance` in different network constructors. 

Maybe it's better as below to reuse code:

```python
class SharedEmbeddingPreprocessor(EmbeddingPreprocessor):
     def __init__(self, ...):
           super().__init__(...)
           self._singleton_instance = True
```

hnyu (changes requested):

alf/networks/preprocessors.py
Haichao-Zhang: It is indeed unnecessary. The file you commented on is an intermediate/previous version.

alf/networks/preprocessors.py
Haichao-Zhang: I have simplified the code where ```SharedEmbeddingPreprocessor``` can be removed completely. Will push a commit soon.

alf/networks/preprocessors.py
Haichao-Zhang: Putting ```PreprocessorNetwork``` in ```network.py``` does not seem right as there will be circular dependency.

hnyu: Try not doing `from .. .import`. instead just do `alf.networks....`

Haichao-Zhang:
Push a commit to address the review comments.

alf/networks/preprocessors.py
Haichao-Zhang: Corrected.

alf/networks/preprocessors.py
Haichao-Zhang: Now move ```PreprocessorNetwork``` out of ```preprocessors.py``` to a separate file.

alf/networks/preprocessors.py
Haichao-Zhang: To make it easier for review on the preprocessor network part, we can use this commit 6327bcc77815165daeddf76a21079d24b1ab1f17 to highlight the diff on this part.

hnyu (approved):

alf/networks/preprocessor_networks.py
```diff
@@ -0,0 +1,137 @@
...
+            inputs (nested Tensor): inputs to the network
+            state (nested Tensor): RNN state of the network
+            min_outer_rank (int): the minimal outer rank allowed
+            max_outer_rank (int): the maximal outer rank allowed
+        Returns:
+            Tensor: tensor after preprocessing.
+        """
+        if self._input_preprocessors:
+            inputs = alf.nest.map_structure(
+                lambda preproc, tensor: preproc(tensor),
```
emailweixu, line 123: if preproc is Network, we should use preproc(tensor)[0]

hnyu: If it's always stateless, then combiner will ignore the empty state.

emailweixu: I see.

Haichao-Zhang: Will improve on this.

PR #564, opened 2020-05-07
faster trunc_norm init

hnyu:
Copied from torch.nn.init.py https://github.com/pytorch/pytorch/pull/32397/

Tested offline; consistent with the scipy way; much faster than before.

>>> import torch
>>> from scipy.stats import truncnorm
>>> x = torch.zeros([1000000])
>>> x = trunc_normal_(x)
>>> y = torch.as_tensor(truncnorm.rvs(-2, 2, size=[1000000]))
>>> x.std()
tensor(0.8801)
>>> y.std()
tensor(0.8796, dtype=torch.float64)
>>> truncnorm.std(-2, 2)
0.8796256610342398


Haichao-Zhang (approved):
LGTM

PR #565, opened 2020-05-10
summarize env_info during eval()

hnyu:
Previously only summarize env info for training. Now can do this during eval(), e.g., evaluation success rate.

### DIFF
--- a/alf/trainers/policy_trainer.py
+++ b/alf/trainers/policy_trainer.py
@@ -64,20 +64,6 @@ class Trainer(object):
         self._evaluate = config.evaluate
         self._eval_interval = config.eval_interval
         self._num_eval_episodes = config.num_eval_episodes
-        eval_metrics = None
-        eval_summary_writer = None
-        if self._evaluate:
-            eval_metrics = [
-                alf.metrics.AverageReturnMetric(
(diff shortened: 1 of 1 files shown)


alf/trainers/policy_trainer.py
```diff
@@ -266,11 +267,8 @@ def _step(algorithm, env, time_step, policy_state, epsilon_greedy, metrics):
     policy_step = algorithm.predict_step(transformed_time_step, policy_state,
                                          epsilon_greedy)
     next_time_step = env.step(policy_step.output)
-
-    exp = alf.data_structures.make_experience(time_step, policy_step,
-                                              policy_state)
     for metric in metrics:
-        metric(exp)
+        metric(time_step)
```
Haichao-Zhang, line 271: The ```metric()``` assumes ```Experience``` as input. Might be better to follow this assumption?

hnyu: It's already changed in one PR. Right now only exp replay buffer's observe will take exps.

Haichao-Zhang: I see. Then it's good.

Haichao-Zhang (approved):

PR #566, opened 2020-05-11
correct episode_length during play

hnyu:
before the length output by logging is always missing one, which is confusing

### DIFF
--- a/alf/data_structures.py
+++ b/alf/data_structures.py
@@ -92,8 +92,10 @@ class TimeStep(
     - prev_action: A (nested) ``Tensor`` for action from previous time step.
     - env_id: A scalar ``Tensor`` of the environment ID of the time step.
     - untransformed: a nest that represents the entire time step itself *before*
-      any transformation; used for experience replay.
-    - env_info: A dictionary containing some environment information.
+      any transformation (e.g., observation or reward transformation); used for
+      experience replay observing by subalgorithms.
+    - env_info: A dictionary containing information returned by Gym environments'
+      ``info``.
     """
 
     def is_first(self):
--- a/alf/trainers/policy_trainer.py
+++ b/alf/trainers/policy_trainer.py
@@ -338,6 +338,7 @@ def play(root_dir,
                 policy_state=policy_state,
                 epsilon_greedy=epsilon_greedy,
                 metrics=[])
+        episode_length += 1
         if recorder:
             recorder.capture_frame()
         else:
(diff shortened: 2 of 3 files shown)


emailweixu:
In metric, the last step is not counted for episode length

hnyu:
> In metric, the last step is not counted for episode length

Just checked, I think the metric does count the last step for episode length, which is also consistent with what I observe in my training TB (it doesn't count the first step, however). There the robotics runs always has an episode length of 50. When I played, it always says episode_length=49. 

(In our TimeLimit wrapper, `episode_length` means how many steps the agent takes. So for an episode of length `N`, there are `N+1` time steps involved. )

I believe making this quantity--either in TB or displayed by logging--consistent with max_episode_steps would be better. Some key statistics, like the final success signal might be what the user want to emphasize on. It'll be confusing if the length displayed is one missing.

emailweixu:
> > In metric, the last step is not counted for episode length
> 
> Just checked, I think the metric does count the last step for episode length, which is also consistent with what I observe in my training TB (it doesn't count the first step, however). There the robotics runs always has an episode length of 50. When I played, it always says episode_length=49.
> 
> (In our TimeLimit wrapper, `episode_length` means how many steps the agent takes. So for an episode of length `N`, there are `N+1` time steps involved. )
> 
> I believe making this quantity--either in TB or displayed by logging--consistent with max_episode_steps would be better. Some key statistics, like the final success signal might be what the user want to emphasize on. It'll be confusing if the length displayed is one missing.

Actually, It's the first step not counted for episode length.

hnyu:
> > > In metric, the last step is not counted for episode length
> > 
> > 
> > Just checked, I think the metric does count the last step for episode length, which is also consistent with what I observe in my training TB (it doesn't count the first step, however). There the robotics runs always has an episode length of 50. When I played, it always says episode_length=49.
> > (In our TimeLimit wrapper, `episode_length` means how many steps the agent takes. So for an episode of length `N`, there are `N+1` time steps involved. )
> > I believe making this quantity--either in TB or displayed by logging--consistent with max_episode_steps would be better. Some key statistics, like the final success signal might be what the user want to emphasize on. It'll be confusing if the length displayed is one missing.
> 
> Actually, It's the first step not counted for episode length.

Right, it's the first `time_step` that's ignored. But the total episode length in metric is still equal to `max_episode_steps` (number of action steps). (#time_steps = episode_length + 1). In our TimeLimit wrapper, `max_episode_steps` means number of action steps, not number of time steps.

emailweixu (approved):

PR #567, opened 2020-05-12
Preprocessor return first

Haichao-Zhang:
Added some checks and codes to make it explicit that current ```InputPreprocessor```s are stateless.

### DIFF
--- a/alf/networks/preprocessor_networks.py
+++ b/alf/networks/preprocessor_networks.py
@@ -72,6 +72,9 @@ class PreprocessorNetwork(Network):
                 # In this case we just assume the spec won't change after the
                 # preprocessing. If it does change, then you should consider
                 # defining an input preprocessor network instead.
+                assert not isinstance(
+                    preproc, nn.Module), ("To use trainable preprocessors, "
+                                          "please derive from alf.Network")
                 return spec
             return preproc.output_spec
 
--- a/alf/networks/preprocessor_networks.py
+++ b/alf/networks/preprocessor_networks.py
@@ -27,6 +27,7 @@ from alf.initializers import variance_scaling_init
 from alf.nest.utils import get_outer_rank
 from alf.networks.preprocessors import InputPreprocessor
 from alf.tensor_specs import TensorSpec
+from alf.utils import common
 
 from .network import Network
 
@@ -87,6 +88,7 @@ class PreprocessorNetwork(Network):
                 elif isinstance(preproc, InputPreprocessor):
                     preproc = preproc.copy()
                     self._input_preprocessor_modules.append(preproc)
+                    preproc = common.return_first(preproc)
                 return preproc
 
             self._input_preprocessors = alf.nest.map_structure(
--- a/alf/networks/preprocessor_networks.py
+++ b/alf/networks/preprocessor_networks.py
@@ -27,7 +27,6 @@ from alf.initializers import variance_scaling_init
 from alf.nest.utils import get_outer_rank
 from alf.networks.network import Network
 from alf.tensor_specs import TensorSpec
-from alf.utils import common
 
 from .network import Network
 
@@ -72,7 +71,7 @@ class PreprocessorNetwork(Network):
             if not isinstance(preproc, Network):
                 # In this case we just assume the spec won't change after the
                 # preprocessing. If it does change, then you should consider
-                # defining a input preprocessor network instead.
+                # defining an input preprocessor network instead.
                 return spec
             return preproc.output_spec
 
@@ -84,12 +83,13 @@ class PreprocessorNetwork(Network):
             def _return_or_copy_preprocessor(preproc):
                 if preproc is None:
                     # allow None as a placeholder in the nest
-                    preproc = math_ops.identity
+                    return math_ops.identity
                 elif isinstance(preproc, Network):
                     preproc = preproc.copy()
                     self._input_preprocessor_modules.append(preproc)
-                    preproc = common.return_first(preproc)
-                return preproc
+                    return lambda x: preproc(x)[0]
+                else:
+                    return preproc
 
             self._input_preprocessors = alf.nest.map_structure(
                 _return_or_copy_preprocessor, input_preprocessors)
@@ -97,7 +97,7 @@ class PreprocessorNetwork(Network):
         self._preprocessing_combiner = preprocessing_combiner
         if alf.nest.is_nested(input_tensor_spec):
             assert preprocessing_combiner is not None, \
-                ("When a nested input tensor spec is provided, a input " +
+                ("When a nested input tensor spec is provided, an input " +
                 "preprocessing combiner must also be provided!")
             input_tensor_spec = preprocessing_combiner(input_tensor_spec)
         else:
@@ -124,6 +124,7 @@ class PreprocessorNetwork(Network):
             inputs = alf.nest.map_structure(
                 lambda preproc, tensor: preproc(tensor),
                 self._input_preprocessors, inputs)
+
         proc_inputs = self._preprocessing_combiner(inputs)
         outer_rank = get_outer_rank(proc_inputs,
                                     self._processed_input_tensor_spec)
--- a/alf/networks/preprocessor_networks.py
+++ b/alf/networks/preprocessor_networks.py
@@ -25,7 +25,7 @@ import alf.utils.math_ops as math_ops
 import alf.nest as nest
 from alf.initializers import variance_scaling_init
 from alf.nest.utils import get_outer_rank
-from alf.networks.preprocessors import InputPreprocessor
+from alf.networks.network import Network
 from alf.tensor_specs import TensorSpec
 from alf.utils import common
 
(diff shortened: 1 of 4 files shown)


hnyu:
Is defining a new decorator an overkill for this purpose?

Haichao-Zhang:
> Is defining a new decorator an overkill for this purpose?

Defining a functional wrapper makes the code clean and the logic the same for all the preprocessors, without the need to do type-specific processings later when forward:
https://github.com/HorizonRobotics/alf/blob/aafcc5f79db8ebbafcf3a37565f896cc75deb691/alf/networks/preprocessor_networks.py#L122-L124

hnyu:
> > Is defining a new decorator an overkill for this purpose?
> 
> Defining a functional wrapper makes the code clean and the logic the same for all the preprocessors, without the need to do type-specific processings later when forward:
> https://github.com/HorizonRobotics/alf/blob/aafcc5f79db8ebbafcf3a37565f896cc75deb691/alf/networks/preprocessor_networks.py#L122-L124

I mean we could just do `preproc(tensor)[0]` here. It's not used anywhere else.

Haichao-Zhang:
> > > Is defining a new decorator an overkill for this purpose?
> > 
> > 
> > Defining a functional wrapper makes the code clean and the logic the same for all the preprocessors, without the need to do type-specific processings later when forward:
> > https://github.com/HorizonRobotics/alf/blob/aafcc5f79db8ebbafcf3a37565f896cc75deb691/alf/networks/preprocessor_networks.py#L122-L124
> 
> I mean we could just do `preproc(tensor)[0]` here. It's not used anywhere else.


You cannot do that directly. It will slice Tensor outputs.

As I mentioned in the last reply, you have to do type-specific processing here.
Basically check if it is an instance of ```InputPreprocessors``` and apply the op only in that case.

Defining a functional wrapper makes the code clean and the logic the same for all the preprocessors, without the need to do type-specific processings later when forward.

hnyu:
> > > > Is defining a new decorator an overkill for this purpose?
> > > 
> > > 
> > > Defining a functional wrapper makes the code clean and the logic the same for all the preprocessors, without the need to do type-specific processings later when forward:
> > > https://github.com/HorizonRobotics/alf/blob/aafcc5f79db8ebbafcf3a37565f896cc75deb691/alf/networks/preprocessor_networks.py#L122-L124
> > 
> > 
> > I mean we could just do `preproc(tensor)[0]` here. It's not used anywhere else.
> 
> You cannot do that directly. It will slice Tensor outputs.
> 
> As I mentioned in the last reply, you have to do type-specific processing here.
> Basically check if it is an instance of `InputPreprocessors` and apply the op only in that case.
> 
> Defining a functional wrapper makes the code clean and the logic the same for all the preprocessors, without the need to do type-specific processings later when forward.

OK I see. If we have other customized preprocessor, then that will be a problem.

But what about in forward() just return the output? Now I don't see it's necessary to keep forward() the same API with other networks.

With this decorator, it's basically that you are reverting the API, in an awkward way..

Haichao-Zhang:
> > > > > Is defining a new decorator an overkill for this purpose?
> > > > 
> > > > 
> > > > Defining a functional wrapper makes the code clean and the logic the same for all the preprocessors, without the need to do type-specific processings later when forward:
> > > > https://github.com/HorizonRobotics/alf/blob/aafcc5f79db8ebbafcf3a37565f896cc75deb691/alf/networks/preprocessor_networks.py#L122-L124
> > > 
> > > 
> > > I mean we could just do `preproc(tensor)[0]` here. It's not used anywhere else.
> > 
> > 
> > You cannot do that directly. It will slice Tensor outputs.
> > As I mentioned in the last reply, you have to do type-specific processing here.
> > Basically check if it is an instance of `InputPreprocessors` and apply the op only in that case.
> > Defining a functional wrapper makes the code clean and the logic the same for all the preprocessors, without the need to do type-specific processings later when forward.
> 
> OK I see. If we have other customized preprocessor, then that will be a problem.
> 
> But what about in forward() just return the output? Now I don't see it's necessary to keep forward() the same API with other networks.
> 
> With this decorator, it's basically that you are reverting the API, in an awkward way..

It is indeed a compromise between complying with the Network API and explicit communication.

alf/networks/preprocessor_networks.py
```diff
@@ -87,6 +88,7 @@ def _return_or_copy_preprocessor(preproc):
                 elif isinstance(preproc, InputPreprocessor):
                     preproc = preproc.copy()
                     self._input_preprocessor_modules.append(preproc)
+                    preproc = common.return_first(preproc)
```
emailweixu: Seems using a decorator is not necessary, we can simply do:
```
elif isinstance(preproc, Network）：
   ...
   preproc = lambda x: preproc(x)[0]
```
Also at line 72, we use:
```
if not instance(preproc, Network):
 ...
else:
   return proproc.output_spec
```
Network always output_spec. We don't need to require preprocessor to calculate output spec given input spec.
And since we can use Network as preprocessor. It doesn't seem necessary to have class InputPreprocessor.

emailweixu (changes requested):

alf/networks/preprocessor_networks.py
Haichao-Zhang: Right, InputPreprocessor seems to be unnecessary now. Will try to remove it.

alf/networks/preprocessor_networks.py
```diff
@@ -68,12 +68,12 @@ def __init__(self,
         self._input_preprocessor_modules = nn.ModuleList()
 
         def _get_preprocessed_spec(preproc, spec):
```
hnyu, line 70: spec is no longer needed

hnyu (changes requested):

alf/networks/preprocessor_networks.py
Haichao-Zhang: I didn't quite understand your comment. This is getting the preprocessed spec from the original spec. You mean this step is not needed?

alf/networks/preprocessor_networks.py
Haichao-Zhang: Both the InputPreprocessor and decorator are removed now.

alf/networks/preprocessor_networks.py
hnyu: Ignore the comment above. What if preproc is a layer as in Wei's carla gin file? The if condition here basically excludes the possibility of using non-network modules as preprocessors?

I think a better way is still that we generate a zero tensor and do a forward here.

alf/networks/preprocessor_networks.py
```diff
@@ -83,19 +83,21 @@ def _get_preprocessed_spec(preproc, spec):
...
                     # allow None as a placeholder in the nest
-                    preproc = math_ops.identity
-                elif isinstance(preproc, InputPreprocessor):
+                    return math_ops.identity
+                elif isinstance(preproc, Network):
                     preproc = preproc.copy()
                     self._input_preprocessor_modules.append(preproc)
-                return preproc
+                    return lambda x: preproc(x)[0]
+                else:
```
hnyu, line 94: Also here we should consider preproc as a nn.Module if now preproc can be any trainable thing

Haichao-Zhang: It is not a generic nn.Module, still Network.

hnyu: What if network1 has
```python
input_preprocessors={
    "camera": @ResnetEncodingNetwork(
        input_tensor_spec=alf.nest.get_field(%observation_spec, "camera"), output_size=%encoding_dim, output_activation=alf.math.identity),
    "radar": torch.nn.Sequential(alf.layers.Reshape([-1]), alf.layers.FC(@flattened_size(%radar_spec), %encoding_dim, use_bias=False)),
    "collision": torch.nn.Sequential(alf.layers.Reshape([-1]), alf.layers.FC(@flattened_size(%collision_spec), %encoding_dim, use_bias=False)),
    "gnss": torch.nn.Sequential(alf.layers.Reshape([-1]), alf.layers.FC(@flattened_size(%goal_spec), %encoding_dim, use_bias=False)),
    "imu": torch.nn.Sequential(alf.layers.Reshape([-1]), alf.layers.FC(@flattened_size(%imu_spec), %encoding_dim, use_bias=False)),
    "goal": torch.nn.Sequential(alf.layers.Reshape([-1]), alf.layers.FC(@flattened_size(%goal_spec), %encoding_dim, use_bias=False)),
    "speed": torch.nn.Sequential(alf.layers.Reshape([-1]), alf.layers.FC(@flattened_size(%speed_spec), %encoding_dim, use_bias=False)),
}
```

And then I copy network1 to network2? The layers of network2 are shared with network1.

alf/networks/preprocessor_networks.py
Haichao-Zhang: Generic layer or nn.Module is not supported currently.
I guess something like a SequentialNetwork which can wrap layers/modules into a Network could be used in the case you mentioned.

hnyu: I think then it's best to directly assert False in the case of non-Network instead of doing something that the user isn't aware of, or potentially crash the code when the return spec is incorrect.

alf/networks/preprocessor_networks.py
emailweixu: I also added a Network call SequentialNetwork which accept a list of torch layers. In PreprocessNetwork, if the preprocessor is not a Network, I construct a SequentialNetork from it.

alf/networks/preprocessor_networks.py
Haichao-Zhang: Good point. Now added.

alf/networks/preprocessor_networks.py
Haichao-Zhang: That sounds great. It will bridge the gap between nn.module and Network.

hnyu (approved):

emailweixu (approved):

PR #569, opened 2020-05-14
Make networks more configurable from gin

emailweixu:
Added get_field() to get a field from a nest so that we can use it to get the spec to construct Network
Added SequentialNetwork() to use an arbitrary sequence of torch layer as a Network

### DIFF
--- a/alf/algorithms/rl_algorithm.py
+++ b/alf/algorithms/rl_algorithm.py
@@ -232,9 +232,7 @@ class RLAlgorithm(Algorithm):
                                         'action_distribution')
             if len(field) == 1:
                 summary_utils.summarize_action_dist(
-                    action_distributions=field[0],
-                    action_specs=self._action_spec,
-                    name="rollout_action_dist")
+                    action_distributions=field[0], name="rollout_action_dist")
 
     def summarize_train(self, experience, train_info, loss_info, params):
         """Generate summaries for training & loss info after each gradient update.
@@ -267,8 +265,7 @@ class RLAlgorithm(Algorithm):
         if self._config.summarize_action_distributions:
             field = alf.nest.find_field(train_info, 'action_distribution')
             if len(field) == 1:
-                summary_utils.summarize_action_dist(field[0],
-                                                    self._action_spec)
+                summary_utils.summarize_action_dist(field[0])
 
     def summarize_metrics(self):
         """Generate summaries for metrics ``AverageEpisodeLength``,
@@ -467,6 +464,9 @@ class RLAlgorithm(Algorithm):
                 reward=transformed_time_step.reward,
                 discount=transformed_time_step.discount,
                 step_type=transformed_time_step.step_type,
+                state=policy_state,
+                prev_action=transformed_time_step.prev_action,
+                observation=transformed_time_step.observation,
                 rollout_info=dist_utils.distributions_to_params(
                     policy_step.info),
                 env_id=transformed_time_step.env_id)
--- a/alf/nest/nest.py
+++ b/alf/nest/nest.py
@@ -458,3 +458,33 @@ def prune_nest_like(nest, slim_nest, value_to_match=None):
         raise AssertionError(
             "prune_nest_like() fails between {} and {}".format(
                 nest, slim_nest))
+
+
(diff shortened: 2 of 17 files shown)


alf/networks/network.py
```diff
@@ -291,3 +292,40 @@ def forward(self, inputs, state=()):
...
+        super().__init__(input_tensor_spec, name)
+        if len(layers) == 1 and isinstance(layers[0], nn.Sequential):
+            self._layers = layers[0]
+        else:
+            self._layers = nn.Sequential(*layers)
+
+    def forward(self, input, state=()):
+        return self._layers(input), state
+
+    def copy(self, name=None):
```
Haichao-Zhang, line 320: Maybe can add a description to the ```copy``` function, which is shallow as well, similar to Network.copy.

alf/networks/preprocessor_networks.py
```diff
@@ -72,30 +72,34 @@ def _get_preprocessed_spec(preproc, spec):
...
-                    return math_ops.identity
+                    return lambda x: (x, None)
                 elif isinstance(preproc, Network):
                     preproc = preproc.copy()
                     self._input_preprocessor_modules.append(preproc)
-                    return lambda x: preproc(x)[0]
-                else:
                     return preproc
+                elif isinstance(preproc, nn.Module):
+                    preproc = SequentialNetwork(input_spec, [preproc])
```
Haichao-Zhang: In this case (```preproc``` is ```nn.Module```), if we call ```preprocessor_network.copy()```, will/should ```preproc``` be copied?

alf/nest/nest.py
```diff
@@ -458,3 +458,33 @@ def _prune(nest, slim_nest):
         raise AssertionError(
             "prune_nest_like() fails between {} and {}".format(
                 nest, slim_nest))
+
+
+def get_field(nested, field):
```
hnyu, line 463: There is already a function called "find_field" in this file. What's the difference between them?

alf/networks/preprocessor_networks.py
emailweixu: preproc will be copied, same as Network as input processor.

Haichao-Zhang: The current implementation will not copy preproc in the case as follows
```
        preproc = torch.nn.Linear(10, 10)

        net = EncodingNetwork(input_tensor_spec=TensorSpec((10, )),
                              input_preprocessors=preproc,
                              fc_layer_params=(10, 10))

        copied_net = net.copy()
```

```preproc``` will be shared between ```net``` and ```copied_net```.

File: alf/nest/nest.py (hunk 3 above)
Comment (emailweixu) [reply alf/nest/nest.py:463]: find_field() finds all the fields with the given name. Here field is a path uniquely identifying the field in the nest.

File: alf/networks/preprocessor_networks.py (hunk 2 above)
Comment (emailweixu) [reply alf/networks/preprocessor_networks.py]: You are right. What behavior do you think is reasonable? Or we can remove the ability of using nn.Module here. The user can always use SequentialNetwork himself in the gin.

Comment (Haichao-Zhang) [reply alf/networks/preprocessor_networks.py]: I think we should make the expected behavior consistent with others, i.e., should be copied in this case as you said.
Being able to use nn.Module would be good to make the gin file simpler.
Maybe we can do ```preproc = SequentialNetwork(input_spec, [preproc]).copy()``` for now?

**Review by Haichao-Zhang — DISMISSED** [review]

File: alf/algorithms/rl_algorithm.py [hunk 4]
```diff
@@ -467,6 +464,9 @@ def unroll(self, unroll_length):
                 reward=transformed_time_step.reward,
                 discount=transformed_time_step.discount,
                 step_type=transformed_time_step.step_type,
+                state=policy_state,
+                prev_action=transformed_time_step.prev_action,
+                observation=transformed_time_step.observation,
```
hnyu, line 469: Where are these fields needed? I think they are no longer needed

emailweixu: after_train_iter might use them.

hnyu (approved):

PR #572, opened 2020-05-17
add ddpg_fetchslide, sac_fetchreach, sac_fetchslide

hnyu:
Reproduced DDPG's performance on FetchSlide-v1. Also wrote a SAC's version. The key is:
* Large batch size (5000) + large learning rate (1e-3)
* Delayed target network update (period=40)

What have been verified unnecessary but were emphasized in the original paper:
* observation normalization. Performance almost didn't change with/without this. But I still added an option in SAC/DDPG for observation normalization.
* action l2 cost. Made the performance worse actually.
* target value clipping. Didn't try in the first place. 

Added an option for taking uniform random actions in DDPG (required for reproducing). 

For observation normalization, because usually we only want to update the normalizer's stats during unroll (the first time an exp is seen) but not during training, I added a decorator to mark the training/inference code so that the normalizer could know when it's called whether the agent is learning or predicting. The decorator might be useful for other modules that behave differently during training and inference.

One last bit of experience: for off-policy algorithms, if `unroll_length` is small, then to avoid frequently switch between unroll and train, we could increase `unroll_length` to `K * unroll_length` and also change `num_updates_per_train_step` to `K * num_updates_per_train_step`. When `unroll_length` is small and `K` is large, this will speed up the training. The training logic will slightly change, though (from "online" RL towards a flavor of ["batch" RL](https://nervanasystems.github.io/coach/features/batch_rl.html)).


alf/examples/ddpg_fetchslide.gin
```diff
@@ -0,0 +1,51 @@
...
+
+# training config
+TrainerConfig.temporally_independent_train_step=True
+TrainerConfig.initial_collect_steps=10000
+TrainerConfig.mini_batch_length=2
+TrainerConfig.mini_batch_size=5000
+TrainerConfig.unroll_length=50
+TrainerConfig.num_updates_per_train_step=40
+TrainerConfig.whole_replay_buffer_training=False
+TrainerConfig.clear_replay_buffer=False
```
emailweixu: line 41-42 should be in ddpg.gin, since they are common to all ddpg

alf/examples/ddpg_fetchslide.gin
```diff
@@ -0,0 +1,51 @@
...
+DdpgAlgorithm.actor_optimizer=@AdamTF()
+DdpgAlgorithm.critic_optimizer=@AdamTF()
+DdpgAlgorithm.rollout_random_action=0.3
+DdpgAlgorithm.use_parallel_network=True
+DdpgAlgorithm.target_update_tau=0.05
+DdpgAlgorithm.target_update_period=40
+OneStepTDLoss.gamma=0.98
+
+# training config
+TrainerConfig.temporally_independent_train_step=True
```
emailweixu: no need to set this. It can be automatically inferred.

alf/examples/ddpg_fetchslide.gin
```diff
@@ -0,0 +1,51 @@
...
+TrainerConfig.initial_collect_steps=10000
+TrainerConfig.mini_batch_length=2
+TrainerConfig.mini_batch_size=5000
+TrainerConfig.unroll_length=50
+TrainerConfig.num_updates_per_train_step=40
```
emailweixu, line 36: It seems to me that this should be similar to "unroll_length=1, num_updates_per_train_step=1"

alf/examples/sac_fetchreach.gin
```diff
@@ -32,23 +32,27 @@ SacAlgorithm.actor_optimizer=@AdamTF()
...
-sac/calc_default_target_entropy.min_prob=0.03
+sac/calc_default_target_entropy.min_prob=0.2
 SacAlgorithm.target_entropy=@sac/calc_default_target_entropy
 SacAlgorithm.target_update_tau=0.05
+SacAlgorithm.target_update_period=40
+OneStepTDLoss=0.98
 
 
 # training config
 TrainerConfig.temporally_independent_train_step=True
```
emailweixu: can be inferred.

alf/algorithms/algorithm.py
```diff
@@ -1070,6 +1075,7 @@ def calc_loss(self, experience, train_info):
             " calc_loss() to generate LossInfo from train_info")
         return train_info
 
+    @common.mark_training
```
emailweixu, line 1078: I feel this is fragile as there may be sub-algorithms which call train_from_replay_buffer().

alf/examples/ddpg_fetchslide.gin
```diff
@@ -0,0 +1,51 @@
...
+
+import alf.utils.dist_utils
+import alf.environments.suite_robotics
+
+# environment config
+create_environment.env_load_fn=@suite_robotics.load
+create_environment.num_parallel_environments=20
+create_environment.env_name='FetchSlide-v1'
+
+hidden_layers=(256, 256, 256)
```
le-horizon, line 12: HER paper uses 3 layers of 64.  Appendix A.
https://arxiv.org/pdf/1707.01495.pdf
But their code uses 256:
https://github.com/openai/baselines/blob/6c44fb28fecdb666182e2951b7b1cbe9cf198ff1/baselines/her/experiment/config.py#L53

Maybe add a note to explain we are using the value in the code?

alf/examples/ddpg_fetchslide.gin
```diff
@@ -0,0 +1,51 @@
...
+AdamTF.lr=1e-3
+
+critic/CriticNetwork.input_tensor_spec=(%observation_spec, %action_spec)
+critic/CriticNetwork.joint_fc_layer_params=%hidden_layers
+
+DdpgAlgorithm.actor_network=@actor/ActorNetwork()
+DdpgAlgorithm.critic_network=@critic/CriticNetwork()
+DdpgAlgorithm.actor_optimizer=@AdamTF()
+DdpgAlgorithm.critic_optimizer=@AdamTF()
+DdpgAlgorithm.rollout_random_action=0.3
```
le-horizon, line 25: Same here? 0.2 in the paper.

alf/examples/ddpg_fetchslide.gin
hnyu: From my experience, this will be faster. The reason of having "num_updates_per_train_step=40" is to exactly replicate what they have in the paper (40 batches per cycle/unroll).

alf/examples/ddpg_fetchslide.gin
```diff
@@ -0,0 +1,51 @@
+
+include 'ddpg.gin'
+
+import alf.utils.dist_utils
+import alf.environments.suite_robotics
+
+# environment config
+create_environment.env_load_fn=@suite_robotics.load
+create_environment.num_parallel_environments=20
```
le-horizon, line 9: from what I can see, their paper says using 8 threads, and code here says 1 env per thread: https://github.com/openai/baselines/blob/ea25b9e8b234e6ee1bca43083f8f3cf974143998/baselines/common/cmd_util.py#L167

alf/algorithms/algorithm.py
hnyu: As long as the main algorithm and sub-algorithm don't run in parallel, this should be fine because we also want the sub-algorithm to know whether it's training or not. In my use case, the main algorithm and sub-algorithm share an observation normalizer which only updates the stats in the main algorithm rollout_step.

alf/examples/ddpg_fetchslide.gin
```diff
@@ -0,0 +1,51 @@
...
+TrainerConfig.clear_replay_buffer=False
+TrainerConfig.num_iterations=5000
+TrainerConfig.num_checkpoints=5
+TrainerConfig.evaluate=True
+TrainerConfig.eval_interval=60
+TrainerConfig.debug_summaries=True
+TrainerConfig.summarize_grads_and_vars=0
+TrainerConfig.summary_interval=30
+TrainerConfig.use_rollout_state=True
+TrainerConfig.replay_buffer_length=100000
```
le-horizon, line 45: shouldn't this be 1e6?

alf/algorithms/algorithm.py
emailweixu: It's possible that a sub-algorithm calls train_from_replay_buffer() during a train_from_unroll() from the higher level algorithm. The training flag will be changed by the sub-algorithm.

emailweixu: It's better to pass a flag to transform_timestep() when it's called.

hnyu: > It's possible that a sub-algorithm calls train_from_replay_buffer() during a train_from_unroll() from the higher level algorithm. The training flag will be changed by the sub-algorithm.

All actual training iters happen sequentially, at least for the current code. So there is no nested train iters. The reason why I didn't use a flag is because the observation_transformer can be any function (e.g., image_scale_transformer, cast_transformer). So the API is hard to be unified. Right now for any transformer we directly call it on observation as the sole argument.  (Unless we also set some global training flag inside transform_timestep instead of passing the flag to transformer.)

Also we do need some convenient way to know whether we are training or not because some modules with eval mode can behave differently.

hnyu: Or maybe maintain a stack of training flags?

alf/utils/common.py
```diff
@@ -393,6 +393,14 @@ def __init__(self,
...
 
+        The normalizer by default doesn't automatically update the mean and std.
+        Instead, it will check when ``self.forward()`` is called, whether an
+        algorithm is unrolling or training. It only updates the mean and std
+        during unroll. This is the suggested way of using an observation
+        normalizer (i.e., update the stats when encountering new data for the
+        first time). This same strategy has been used by OpenAI's baselines for
+        training their Robotics environments.
+
         Args:
```
le-horizon, line 404: Where is ObservationNormalizer called?

emailweixu (dismissed):

alf/algorithms/algorithm.py
le-horizon: If obs_normalizer is the only place training flag is needed, then since it's not really useful, maybe we just remove from this PR?

le-horizon (commented):
Nice change.  Some nits and a couple suggestions.

alf/examples/ddpg_fetchslide.gin
hnyu: The original HER paper never experimented on Fetch envs. I reproduced the results in the paper https://arxiv.org/abs/1802.09464 as clarified in the updated README.

alf/algorithms/algorithm.py
hnyu: Added just for potential future use. Sooner or later we will face this issue.

alf/examples/ddpg_fetchslide.gin
hnyu: same above

alf/examples/ddpg_fetchslide.gin
hnyu: same above.

alf/examples/ddpg_fetchslide.gin
hnyu: We need to multiply this with num_envs. Currently this have 2M (100000 * 20) steps in the buffer. They actually have 19 * 1M = 19M in total (in the Fetch env paper). I didn't follow exactly here because I think that large number is unnecessary.

alf/utils/common.py
hnyu: Expected to be called in Algorithm's transform_time_step as self._observation_transformer. This implementation doesn't actually use it as no performance gain was observed. Keep it here for future use.

alf/examples/ddpg_fetchslide.gin
hnyu: https://github.com/openai/baselines/tree/ea25b9e8b234e6ee1bca43083f8f3cf974143998/baselines/her  See "Reproducing results" section

emailweixu (approved):

PR #574, opened 2020-05-18
Simplify gin config for ac, ppo, ddpg and sarsa

emailweixu:
so that we don't need to set specs for various networks in the gin config.
TODO for sac since it's quite complicated.

### DIFF
--- a/alf/algorithms/actor_critic_algorithm.py
+++ b/alf/algorithms/actor_critic_algorithm.py
@@ -36,8 +36,8 @@ class ActorCriticAlgorithm(OnPolicyAlgorithm):
     def __init__(self,
                  observation_spec,
                  action_spec,
-                 actor_network: ActorDistributionNetwork,
-                 value_network: ValueNetwork,
+                 actor_network_ctor=ActorDistributionNetwork,
+                 value_network_ctor=ValueNetwork,
                  env=None,
                  config: TrainerConfig = None,
                  loss=None,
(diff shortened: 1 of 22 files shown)


alf/algorithms/actor_critic_algorithm.py
```diff
@@ -70,6 +74,10 @@ def __init__(self,
             debug_summaries (bool): True if debug summaries should be created.
             name (str): Name of this algorithm.
             """
+        actor_network = actor_network_ctor(
+            input_tensor_spec=observation_spec, action_spec=action_spec)
+        value_network = value_network_ctor(input_tensor_spec=observation_spec)
+
         super(ActorCriticAlgorithm, self).__init__(
             observation_spec=observation_spec,
```
hnyu, line 82: I feel that it's unnecessary to pass observation_spec to the base RLAlgorithm class.

hnyu (changes requested):
Some PPO gin files have not been updated

emailweixu:
> Some PPO gin files have not been updated

Those ppo gin files are not converted from tf version yet.

alf/algorithms/actor_critic_algorithm.py
emailweixu: It doesn't hurt to leave it there. It was once useful.

Haichao-Zhang (approved):

hnyu (approved):

PR #579, opened 2020-05-21
Handling compass being nan from IMUSensor

emailweixu:
Also do not use docker to run carla if already in docker container.
And make sure vehicles not reset to a location colliding with other vehicles.

### DIFF
--- a/alf/environments/suite_carla.py
+++ b/alf/environments/suite_carla.py
@@ -56,6 +56,7 @@ except ImportError:
 
 import alf
 import alf.data_structures as ds
+from alf.utils import common
 from .torch_environment import TorchEnvironment
 from .process_environment import array_to_tensor, tensor_to_array
 from .suite_socialbot import _get_unused_port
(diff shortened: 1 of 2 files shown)


alf/environments/suite_carla.py
```diff
@@ -334,11 +335,15 @@ def _IMU_callback(weak_self, sensor_data):
         self = weak_self()
         if not self:
             return
+        if not math.isnan(sensor_data.compass):
+            self._compass = sensor_data.compass
```
le-horizon, line 339: maybe initialize self._compass to some default value in __init__()?

le-horizon (approved):
LG.  one nit.

alf/environments/suite_carla.py
emailweixu: "self._compass  = 0" is there already

le-horizon: Ah, missed that..  LG then.

alf/utils/common.py
```diff
@@ -856,3 +856,26 @@ def flattened_size(spec):
...
+
+def is_inside_docker_container():
+    """Return whether the current process is running inside a docker container.
+
+    See discussions at `<https://stackoverflow.com/questions/23513045/how-to-check-if-a-process-is-running-inside-docker-container>`_
+    """
+    return os.path.exists("/.dockerenv")
+
+
+def check_numerics(nested):
```
hnyu, line 869: There are several other places in the current code that have a simpler check_numerics. Maybe replace them with this function?

emailweixu: This is good for nested one. The other places are not checking nested.

hnyu (approved):

PR #580, opened 2020-05-21
trainer progress

hnyu:
Try to provide a way for any code to have access to the current training progress so that some scheduling can be done regarding the progress. To make it easily accessible anywhere, I have to create a class variable in `Trainer`.

Also before `iter_num` was not checkpointed and when a job continues from a checkpoint, another `num_iterations` will be run, which is inconsistent with `env_steps` (checkpointed).

### DIFF
--- a/alf/trainers/policy_trainer.py
+++ b/alf/trainers/policy_trainer.py
@@ -34,19 +34,27 @@ from alf.utils.checkpoint_utils import Checkpointer
 from alf.utils.summary_utils import record_time
 
 
-class TrainerProgress(nn.Module):
+class _TrainerProgress(nn.Module):
     def __init__(self):
-        super(TrainerProgress, self).__init__()
+        super(_TrainerProgress, self).__init__()
         self.register_buffer("_iter_num", torch.zeros((), dtype=torch.int64))
         self.register_buffer("_env_steps", torch.zeros((), dtype=torch.int64))
         self._num_iterations = None
         self._num_env_steps = None
+        self._progress = None
 
     def set_termination_criterion(self, num_iterations, num_env_steps):
         self._num_iterations = float(num_iterations)
         self._num_env_steps = float(num_env_steps)
+        # might be loaded from a checkpoint, so we update first
+        self.update()
 
-    def progress(self):
+    def update(self, iter_num=None, env_steps=None):
+        if iter_num is not None:
+            self._iter_num.fill_(iter_num)
+        if env_steps is not None:
+            self._env_steps.fill_(env_steps)
+
         assert not (self._num_iterations is None
                     and self._num_env_steps is None), (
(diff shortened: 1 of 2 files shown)


alf/trainers/policy_trainer.py
```diff
@@ -34,9 +34,38 @@
 from alf.utils.summary_utils import record_time
 
 
+class TrainerProgress(nn.Module):
+    def __init__(self):
```
Haichao-Zhang, line 38: It might be better to use ```global_counter``` for this purpose instead of introducing another counter.

hnyu: global_counter is not equal to iter_num because when update_every_minibatch is true it's a multiple of iter_num (its purpose is for summary and shouldn't be coupled with training iterations). Also currently iter_num is not checkpointed, which is inconsistent with env_steps.

alf/trainers/policy_trainer.py
```diff
@@ -34,9 +34,38 @@
 from alf.utils.summary_utils import record_time
 
 
+class TrainerProgress(nn.Module):
```
emailweixu: This class is not public interface. So prefix with "_" as _TrainerProgress.

alf/trainers/policy_trainer.py
```diff
@@ -200,6 +242,10 @@ def _markdownify(paragraph):
             env_steps_metric = self._algorithm.get_step_metrics()[1]
             total_time_steps = env_steps_metric.result()
             iter_num += 1
+
+            self._trainer_progress._iter_num.fill_(iter_num)
+            self._trainer_progress._env_steps.fill_(total_time_steps)
```
emailweixu: Update using a member function of TrainerProgress:
```
 self._trainer_progress.update(iter_num, total_time_steps)
```
update() can also do the calculation of progress() so it not calculated every time it's called.

alf/trainers/policy_trainer.py
```diff
@@ -200,6 +255,9 @@ def _markdownify(paragraph):
             env_steps_metric = self._algorithm.get_step_metrics()[1]
             total_time_steps = env_steps_metric.result()
             iter_num += 1
+
+            self._trainer_progress.update(iter_num, total_time_steps)
+
             if ((self._num_iterations and iter_num >= self._num_iterations)
                     or (self._num_env_steps
                         and total_time_steps >= self._num_env_steps)):
```
Haichao-Zhang, line 263: Can use ```progress()```  to simplify Line 261-263

hnyu: Might have floating error when the integer number is huge. If we really care about stopping at that number, then integer comparison might be better.

Haichao-Zhang (approved):

PR #584, opened 2020-05-23
global_norm  handles empty list and optimizer check

hnyu:
Another unsolved issue is that get_optimizer_info won't obtain optimizer info for a subalgorithm that's an ignored attribute (e.g., self-training), because setup_optimizers will not include ignored attributes. So TB optimizers tab won't show anything for it .

### DIFF
--- a/alf/algorithms/algorithm.py
+++ b/alf/algorithms/algorithm.py
@@ -527,7 +527,11 @@ class Algorithm(nn.Module):
         return opts
 
     def get_optimizer_info(self):
-        """Return the optimizer info for all the modules in a string."""
+        """Return the optimizer info for all the modules in a string.
+
+        TODO: for a subalgorithm that's an ignored attribute, its optimizer info
+        won't be obtained.
+        """
         unhandled = self._setup_optimizers()
 
         optimizer_info = []
@@ -992,6 +996,7 @@ class Algorithm(nn.Module):
         loss = weight * loss_info.loss
 
         unhandled = self._setup_optimizers()
+        unhandled = [(self._param_to_name[p], p) for p in unhandled]
         assert not unhandled, ("'%s' has some modules/parameters do not have "
                                "optimizer: %s" % (self.name, unhandled))
         optimizers = self.optimizers()
@@ -1002,8 +1007,14 @@ class Algorithm(nn.Module):
 
         all_params = []
         for optimizer in optimizers:
+            params = []
             for param_group in optimizer.param_groups:
-                all_params.extend(param_group['params'])
+                params.extend(param_group['params'])
+            assert params, (
+                "The recorded optimizer '" + optimizer.name +
+                "' haven't been used for learning any parameters! Please check."
+            )
+            all_params.extend(params)
             optimizer.step()
 
         all_params = [(self._param_to_name[p], p) for p in all_params]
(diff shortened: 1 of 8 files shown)


alf/utils/tensor_utils.py
```diff
@@ -142,6 +142,8 @@ def global_norm(tensors):
     """
     assert alf.nest.is_nested(tensors), "tensors must be a nest!"
     tensors = alf.nest.flatten(tensors)
+    if not tensors:
+        return torch.zeros((), dtype=torch.float32)
```
hnyu, line 146: If not check here, torch.sqrt will complain the arg is int for an empty list.

alf/trainers/policy_trainer.py
```diff
@@ -101,8 +101,9 @@ def __init__(self, config: TrainerConfig):
         self._random_seed = config.random_seed
         self._num_iterations = config.num_iterations
         self._num_env_steps = config.num_env_steps
-        assert self._num_iterations + self._num_env_steps > 0, \
-            "Must provide #iterations or #env_steps for training!"
+        assert (self._num_iterations + self._num_env_steps > 0
+                and self._num_iterations * self._num_env_steps == 0), \
```
hnyu, line 105: These two need to be exclusive because otherwise the checkpoint interval will function incorrectly.

alf/utils/tensor_utils.py
emailweixu: why not return 0 here?

alf/algorithms/sac_algorithm_test.py
```diff
@@ -223,7 +223,6 @@ def test_sac_algorithm_discrete(self, use_parallel_network):
             use_parallel_network=use_parallel_network,
             env=env,
             config=config,
-            actor_optimizer=alf.optimizers.Adam(lr=1e-2),
```
emailweixu, line 226: why is this removed?

alf/utils/tensor_utils.py
hnyu: This function can potentially be used for other purposes so it's better to return a tensor.

alf/algorithms/sac_algorithm_test.py
hnyu: In the discrete case, there is no actor network. So this optimizer will not be used. Now any unused optimizer will be checked and produce an error.

alf/algorithms/algorithm.py
```diff
@@ -1002,8 +1007,14 @@ def update_with_gradient(self, loss_info, valid_masks=None, weight=1.0):
 
         all_params = []
         for optimizer in optimizers:
+            params = []
             for param_group in optimizer.param_groups:
-                all_params.extend(param_group['params'])
+                params.extend(param_group['params'])
+            assert params, (
+                "The recorded optimizer '" + optimizer.name +
+                "' haven't been used for learning any parameters! Please check."
```
Haichao-Zhang, line 1015: haven't -> hasn't

alf/networks/critic_networks_test.py
```diff
@@ -118,13 +118,15 @@ def test_make_parallel(self):
 
         def _train(pnet, name):
             t0 = time.time()
-            optimizer = alf.optimizers.AdamTF(list(pnet.parameters()), lr=1e-4)
+            optimizer = alf.optimizers.AdamTF(lr=1e-4)
+            optimizer.add_param_group({'params': list(pnet.parameters())})
```
Haichao-Zhang, line 122: what's the reason for this change?

hnyu: Before the init is wrong because the API is no longer like that. You can still specify params but have to provide the key.

emailweixu (approved):

PR #585, opened 2020-05-26
Fixed Basis Decoding

Haichao-Zhang:
Adding support on using pre-specified basis set for decoding the network output.

### DIFF
--- a/alf/layers.py
+++ b/alf/layers.py
@@ -111,6 +111,112 @@ class OneHot(nn.Module):
 
 
 @gin.configurable
+class FixedDecodingLayer(nn.Module):
+    def __init__(
+            self,
+            input_size,
+            output_size,
+            basis_type="rbf",
+            min_v=-1.,
+            max_v=1.,
+            sigma=None,
+    ):
+        """A layer that uses a set of fixed basis for decoding the inputs.
+        Args:
+            input_size (int): the size of input to be decoded, representing the
+                number of representation coefficients
+            output_size (int): the size of the decoded output
+            basis_type (str): the type of basis to be used for decoding
+                - "poly": polynomials basis
+                - "rbf": radial basis functions
+                - "haar": haar wavelet basis
+            min_v (float): lower bound of the support for the input
+            max_v (float): upper bound of the support for the input
+            sigma (float): the bandwidth parameter used for RBF basis. If None,
+                a default value of ``sigma = (max_v - min_v) / output_size``
+                will be used.
+        """
+        # get the argument list with vals
+        self._kwargs = copy.deepcopy(locals())
+        self._kwargs.pop('self')
+        self._kwargs.pop('__class__')
+
+        super(FixedDecodingLayer, self).__init__()
+
+        assert input_size > 0, "input_size should be at least one"
+        assert basis_type in {"poly", "rbf", "haar"
+                              }, ("the specified method "
+                                  "{} is not supported".format(basis_type))
+
+        self._B = nn.Linear(input_size, output_size, bias=False)
+
+        def _poly_matrix(n, D):
+            # non-square matrix [n, D + 1]
+            x = torch.linspace(min_v, max_v, n)
+            B = torch.empty(n, D + 1)
+            for d in range(D + 1):
+                B[:, d] = x**d
+            return B
+
+        def _rbf_matrix(n, sigma=1.0):
+            # square matrix [n, n]
+            x = torch.linspace(min_v, max_v, n)
+            B = torch.empty(n, n)
+            for d in range(n):
+                B[:, d] = torch.exp(-(x - x[d])**2 / sigma)
+            return B
+
+        def _haar_matrix(n):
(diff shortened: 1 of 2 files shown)


alf/layers.py
```diff
@@ -110,6 +110,112 @@ def forward(self, input):
...
+        assert input_size > 0, "input_size should be at least one"
+        assert basis_type in {"poly", "rbf", "haar"
+                              }, ("the specified method "
+                                  "{} is not supported".format(basis_type))
+
+        self._B = nn.Linear(input_size, output_size, bias=False)
+
+        def _poly_matrix(n, D):
+            # non-square matrix [n, D + 1]
+            x = torch.linspace(min_v, max_v, n)
```
emailweixu: we don't need min_v, max_v. Can simply assume min_v =-1, max_v=1. And it may be better to scale B[:,d]  by 1/n^2 or 1/2^n which encode our prior knowledge of preferring lower order polynomial. And we can do similar scaling for haar basis.

alf/layers.py
```diff
@@ -110,6 +110,112 @@ def forward(self, input):
...
+            B = torch.empty(n, D + 1)
+            for d in range(D + 1):
+                B[:, d] = x**d
+            return B
```
emailweixu: numpy has good support for different polynomials:
```python
B = np.polynomial.polynomial.polyvander(x, D)
```
You can also try other polynomial such as Chebyshev polynomial:
```python
B = np.polynomial.chebyshev.chebvander(x, D)
```

alf/layers.py
Haichao-Zhang: Removed min_v and max_v. Added basis scaling according to frequencies.

alf/layers.py
Haichao-Zhang: Good point. We could leverage numpy for constructing more types of basis. Added Chebyshev polynomial.

alf/layers.py
```diff
@@ -110,6 +111,127 @@ def forward(self, input):
...
+        assert basis_type in {"poly", "cheb", "rbf", "haar"
+                              }, ("the specified method "
+                                  "{} is not supported".format(basis_type))
+
+        self._B = nn.Linear(input_size, output_size, bias=False)
+
+        def _polyvander_matrix(n, D, tau=tau):
+            # non-square matrix [n, D + 1]
+            x = torch.linspace(-1, 1, n)
+            B = torch.empty(n, D + 1)
```
emailweixu: B = np.polynomial.polynomial.polyvander(x, D)

alf/layers.py
```diff
@@ -110,6 +112,128 @@ def forward(self, input):
...
+                B[:, d] = x**d
+            # weight for encoding the preference to low-frequency basis
+            exp_factor = torch.arange(D + 1).float()
+            basis_weight = tau**exp_factor
+            return B * basis_weight
+
+        def _chebvander_matrix(n, D, tau=tau):
+            # non-square matrix [n, D + 1]
+            x = np.linspace(-1, 1, n)
+            B = torch.from_numpy(np.polynomial.chebyshev.chebvander(x, D)).to(
```
emailweixu: as_tensor() can convert to the default device by default.

alf/layers_test.py
```diff
@@ -151,6 +151,64 @@ def test_parallel_conv_transpose(self,
...
+        ("rbf", 7, 8, 0.1),
+        ("haar", 7, 7, None),
+        ("unimplemented", 3, 8, None),
+    )
+    def test_fixed_decoding_layer(self, basis_type, input_size, output_size,
+                                  sigma):
+        batch_size = 3
+
+        if (basis_type == "rbf" and input_size != output_size) or \
+           (basis_type == "haar" and (input_size & (input_size - 1)) != 0) or \
```
emailweixu, line 169: for haar, please compare to an explicit matrix to make sure it's correctly constructed.

alf/layers.py
Haichao-Zhang: Changed to use as_tensor().

alf/layers.py
Haichao-Zhang: Now changed to using np.
Previously I didn't change as I thought these two are equal and there might be some benefits if we use torch directly.

alf/layers_test.py
Haichao-Zhang: Now added explicit tests against several explicit Haar matrices.

alf/layers.py
```diff
@@ -205,7 +202,7 @@ def _kron(A, B):
                 h = torch.cat((h_n, h_i), dim=1)
                 return h
 
-            B = _get_haar_matrix(n)
+            B = _get_haar_matrix(n) / torch.sqrt(torch.Tensor([n]))
```
Haichao-Zhang, line 205: The term ``1 / torch.sqrt(torch.Tensor([n]))`` was not included before as it is a global scaling. Now added to make it fully compatible with explicit Haar matrices.

emailweixu (approved):

PR #586, opened 2020-05-28
segment tree

### LINKED ISSUE #562: Prioritized Replay

emailweixu:
This data structure is for importance sampling of prioritized replay buffer.

### DIFF
--- a/alf/experience_replayers/segment_tree.py
+++ b/alf/experience_replayers/segment_tree.py
@@ -46,11 +46,11 @@ class SegmentTree(nn.Module):
                                  torch.zeros((2 * capacity, ), dtype=dtype))
         self._op = op
         self._capacity = capacity
-        self._leftest_leaf = 1
+        self._leftmost_leaf = 1
         self._depth = 0
-        while self._leftest_leaf < capacity:
-            self._leftest_leaf *= 2
-            if self._leftest_leaf < capacity:
+        while self._leftmost_leaf < capacity:
+            self._leftmost_leaf *= 2
+            if self._leftmost_leaf < capacity:
                 self._depth += 1
 
     def __setitem__(self, indices, values):
@@ -63,6 +63,9 @@ class SegmentTree(nn.Module):
         """
 
         def _step(indices):
+            """
+            Calculate the parent value from its children.
+            """
             indices = indices / 2
             indices = torch.unique(indices)
             left = self._values[indices * 2]
@@ -86,7 +89,7 @@ class SegmentTree(nn.Module):
             indices = self._index_to_leaf(indices)
             self._values[indices] = values
 
-            num_large = (indices >= self._leftest_leaf).to(torch.int64).sum()
+            num_large = (indices >= self._leftmost_leaf).to(torch.int64).sum()
             if num_large > 0:
                 large_indices = indices[:num_large]
                 small_indices = indices[num_large:]
@@ -116,12 +119,12 @@ class SegmentTree(nn.Module):
         """
         Make sure idx=0 is the leftest leaf.
         """
-        idx += self._leftest_leaf
+        idx += self._leftmost_leaf
         idx = torch.where(idx >= 2 * self._capacity, idx - self._capacity, idx)
         return idx
 
     def _leaf_to_index(self, leaf):
-        idx = leaf - self._leftest_leaf
+        idx = leaf - self._leftmost_leaf
         idx = torch.where(idx < 0, idx + self._capacity, idx)
         return idx
 
@@ -169,6 +172,12 @@ class SumSegmentTree(SegmentTree):
         """
 
         def _step(indices, thresholds):
+            """Choose one of the children of each index based on threshold.
+
+            If threshold is greater than or equal to the
+            left child, choose the right child and update threhsold to threhsold - left_child.
+            Otherwise choose left child and keep threshold unchanged.
+            """
             indices *= 2
             left = self._values[indices]
             greater = thresholds >= left
--- a/alf/experience_replayers/segment_tree.py
+++ b/alf/experience_replayers/segment_tree.py
@@ -82,8 +82,8 @@ class SegmentTree(nn.Module):
             assert indices.ndim == 1
             assert values.ndim == 1
             assert indices.shape == values.shape, (
-                "indices and values should be 1-D tensor with the length. Got "
-                "%s and %s." % (indices.shape, values.shape))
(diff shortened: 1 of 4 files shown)


alf/experience_replayers/segment_tree.py
```diff
@@ -0,0 +1,216 @@
...
+                 name="SegmentTree"):
+        super().__init__()
+        self._name = name
+        self._device = device
+        with alf.device(self._device):
+            self.register_buffer("_values",
+                                 torch.zeros((2 * capacity, ), dtype=dtype))
+        self._op = op
+        self._capacity = capacity
+        self._leftest_leaf = 1
```
hnyu: Maybe "leftmost" is a more proper word?

alf/experience_replayers/segment_tree.py
```diff
@@ -0,0 +1,216 @@
...
+class SegmentTree(nn.Module):
+    """
+    Data structure to allow efficient calculate the summary statistics over a
+    segment of elements.
+    See https://en.wikipedia.org/wiki/Segment_tree for detail.
+
+    In this implementation, ``values[1]`` is the root. ``values[capacity: 2*capacity]``
+    are the leaves. The two children of an internal node ``values[i]`` are ``values[2*i]``
+    and ``values[2*i+1]``. And ``values[i]`` is set to ``op(values[2*i], values[2*i+1])``
+    """
```
hnyu, line 35: Should also comment that the leaves represent scalars instead of intervals in this case.

alf/experience_replayers/segment_tree.py
```diff
@@ -0,0 +1,216 @@
...
+    def __init__(self,
+                 capacity,
+                 dtype=torch.float32,
+                 device="cpu",
+                 name="SumSegmentTree"):
+        super().__init__(
+            capacity, torch.add, dtype=dtype, device=device, name=name)
+
+    def __setitem__(self, indices, values):
+        assert values.min() >= 0
```
hnyu, line 158: should be > 0 instead so that find_sum_bound has a unique solution?

alf/experience_replayers/segment_tree.py
```diff
@@ -0,0 +1,216 @@
...
+    def __setitem__(self, indices, values):
+        """Set the value of leaves and update the internal nodes.
+
+        Args:
+            indices (Tensor): 1-D int64 Tensor. Its values should be in range
+                [0, capacity).
+            values (Tensor): 1-D Tensor with the same shape as ``indices``
+        """
+
+        def _step(indices):
```
hnyu, line 67: Probably add a comment for this function, describing the process.

alf/experience_replayers/segment_tree.py
```diff
@@ -0,0 +1,216 @@
...
+        values[result[i]] will never be 0.
+
+        Args:
+            thresholds (Tensor): 1-D Tensor. All the elements in `thresholds`
+                should be smaller than self.summary()
+        Returns:
+            Tensor: 1-D int64 Tensor with the same shape as ``thesholds``
+        """
+
+        def _step(indices, thresholds):
```
hnyu, line 176: Probably add a comment for this function, describing the process.

hnyu (changes requested):

hnyu:
I think a more difficult question is, given that our replay buffer is circular, how can we maintain this segment tree when idx is also circular?

alf/experience_replayers/segment_tree.py
emailweixu: updated.

alf/experience_replayers/segment_tree.py
emailweixu: updated.

alf/experience_replayers/segment_tree.py
```diff
@@ -0,0 +1,227 @@
...
+            return indices
+
+        with alf.device(self._device):
+            indices = convert_device(indices)
+            values = convert_device(values)
+
+            assert indices.ndim == 1
+            assert values.ndim == 1
+            assert indices.shape == values.shape, (
+                "indices and values should be 1-D tensor with the length. Got "
```
le-horizon: the *same* length

alf/experience_replayers/segment_tree.py
```diff
@@ -0,0 +1,227 @@
...
+                "%s and %s." % (indices.shape, values.shape))
+            op = self._op
```
le-horizon, line 87: remove this line and use self._op in _step?

alf/experience_replayers/segment_tree.py
emailweixu: 0 is okay. find_sum_bound() will never return an index with zero value.

alf/experience_replayers/segment_tree.py
```diff
@@ -0,0 +1,227 @@
...
+        with alf.device(self._device):
+            idx = convert_device(idx)
+            assert 0 <= idx.min()
+            assert idx.max() < self._capacity
+            result = self._values[self._index_to_leaf(idx)]
+        return convert_device(result)
+
+    def _index_to_leaf(self, idx):
+        """
+        Make sure idx=0 is the leftest leaf.
```
le-horizon: leftest -> leftmost

emailweixu:
> I think a more difficult question is, given that our replay buffer is circular, how can we maintain this segment tree when idx is also circular?

This is not a problem. Simply overwriting the value of the previous index should work.

alf/experience_replayers/segment_tree.py
```diff
@@ -0,0 +1,227 @@
...
+        result[i] is the minimum idx such that
+            thresholds[i] < values[0] + ... + values[idx]
+
+        values[result[i]] will never be 0.
+
+        Args:
+            thresholds (Tensor): 1-D Tensor. All the elements in `thresholds`
+                should be smaller than self.summary()
+        Returns:
+            Tensor: 1-D int64 Tensor with the same shape as ``thesholds``
```
le-horizon: th*r*esholds

hnyu (dismissed):

le-horizon (commented):
LG.  Mostly small nits.

alf/experience_replayers/segment_tree.py
emailweixu: Python does not do compile optimization. So sometime I do optimization for python.

le-horizon: Haha interesting.

le-horizon (approved):
LG

PR #587, opened 2020-05-28
Her

le-horizon:
Hi Wei and Haonan,

This is the her experience replay code so far.

I also refactored gin files to reduce duplication.

ddpg_fetchpush with 20 envs usually gets this:
image

her with k=0.1 sometimes is better and sometimes never learns:
image

Thanks,
Le


alf/experience_replayers/replay_buffer.py
```diff
@@ -18,10 +18,78 @@
 import torch.nn as nn
 
 import alf
+from alf import data_structures as ds
 from alf.nest.utils import convert_device
 from alf.utils.data_buffer import atomic, RingBuffer
 
 
+def get_nest_attr(obj, path):
```
emailweixu: Seems we already have this: `alf.nest.get_field()`

alf/experience_replayers/replay_buffer.py
```diff
@@ -18,10 +18,78 @@
...
+    for e in levels:
+        if e[0] == ":":  # dict element
+            r = r[e[1:]]
+        else:  # namedtuple element
+            r = getattr(r, e)
+    return r
+
+
+def recursive_set(obj, levels, value):
+    """Set element value specified by path in nest obj.
```
emailweixu: Can use `alf.nest.utils.transform_nest`

alf/experience_replayers/replay_buffer.py
```diff
@@ -32,21 +100,216 @@ def __init__(self,
                  max_length=1024,
                  device="cpu",
                  allow_multiprocess=False,
+                 her_k=0,
```
emailweixu: It seems to me HER should be subclass or a wrapper of ReplayBuffer. ReplayBuffer is supposed to be more general.

emailweixu (changes requested):

alf/experience_replayers/replay_buffer.py
le-horizon: Good points.  Done.  I originally tried it and somehow an error appeared..  Maybe some other bug caused it.

alf/experience_replayers/replay_buffer.py
le-horizon: I can subclass as HERReplayBuffer and have all experience_replayers use that instead of ReplayBuffer, since HERReplayBuffer is efficient when k=0.

Will that be Ok?

le-horizon (commented):
Done and one question.

alf/experience_replayers/replay_buffer.py
emailweixu: We are going to have PrioritizedReplayBuffer. We need to a way to flexibly combine these things:
plain replay buffer
HER with plain replay buffer
prioritized replay buffer
HER with prioritized replay buffer

le-horizon: Then having two flags controlling HER and Prioritized replay is probably the best solution at the moment.  Prioritization can be done in this function: _get_sample_batch, which is independent of HER.  If things become even more complex, then maybe we could use a chain of inheritance ReplayBuffer=>HERReplayBuffer=>PrioritizedHERReplayBuffer, and have experience_replayers use the most capable replay_buffer.

alf/environments/suite_robotics.py
```diff
@@ -92,12 +92,19 @@ def __init__(self, env):
         super().__init__(env)
 
     def observation(self, observation):
-        return np.clip(observation, -200., 200.)
+        if isinstance(observation, dict):
+            for k, v in observation.items():
+                observation[k] = np.clip(v, -200., 200.)
```
emailweixu: (-200, 200) should be configurable.

alf/environments/suite_robotics.py
```diff
@@ -146,9 +153,11 @@ def env_ctor(env_id=None):
             torch_env_wrappers=torch_env_wrappers)
 
     # concat robot's observation and the goal location
-    env = FlattenDictWrapper(env, ["observation", "desired_goal"])
+    if concat_desired_goal:
+        env = FlattenDictWrapper(env, ["observation", "desired_goal"])
     env = SuccessWrapper(env, max_episode_steps)
-    env = ObservationClipWrapper(env)
+    if clip_observation:
```
emailweixu: No need to write code for this. This one can be configured using gym_env_wrappers.

alf/environments/torch_gym_wrapper.py
```diff
@@ -82,8 +82,10 @@ def torch_dtype(data):
     elif isinstance(space, gym.spaces.Tuple):
         return tuple([tensor_spec_from_gym_space(s) for s in space.spaces])
     elif isinstance(space, gym.spaces.Dict):
-        return collections.OrderedDict([(key, tensor_spec_from_gym_space(s))
-                                        for key, s in space.spaces.items()])
+        return {
```
emailweixu: Now OrderedDict is supported

alf/examples/ddpg.gin
```diff
@@ -8,6 +8,9 @@ create_environment.num_parallel_environments=%NUM_PARALLEL_ENVIRONMENTS
 observation_spec=@get_observation_spec()
 action_spec=@get_action_spec()
 
+DdpgAlgorithm.observation_spec=%observation_spec
+DdpgAlgorithm.action_spec=%action_spec
```
emailweixu: no need to have these two lines.

alf/examples/ddpg_fetchreach.gin
```diff
@@ -0,0 +1,45 @@
...
+
+# override algorithm and training config
+actor/ActorNetwork.fc_layer_params=%hidden_layers
+critic/CriticNetwork.joint_fc_layer_params=%hidden_layers
+
+DdpgAlgorithm.actor_network_ctor=@actor/ActorNetwork
+DdpgAlgorithm.critic_network_ctor=@critic/CriticNetwork
+DdpgAlgorithm.actor_optimizer=@AdamTF()
+DdpgAlgorithm.critic_optimizer=@AdamTF()
+DdpgAlgorithm.use_parallel_network=True
```
emailweixu: No need to use parallel network since num_critic_replicas=1

alf/environments/suite_robotics.py
hnyu: It's hard-coded for Fetch environments in the paper. Maybe this file should renamed to suite_fetch.py

alf/examples/sac_fetchpush.gin
```diff
@@ -0,0 +1,2 @@
+include 'sac_fetchreach.gin'
```
hnyu, line 1: I don't think it's necessary to check in lots of gin files for all the fetch environments? Typically an example gin file should come with training figures and gif files.

alf/experience_replayers/replay_buffer.py
```diff
@@ -32,21 +34,216 @@ def __init__(self,
...
+                 step_type_field="step_type",
+                 achieved_goal_field="observation.achieved_goal",
+                 desired_goal_field="observation.desired_goal",
```
hnyu: I also don't think it's a good idea to add many seemly irrelevant arguments like "achieved_goal_field" in a general class like ReplayBuffer. The current implementation is overly bloated.

alf/networks/encoding_networks.py
```diff
@@ -803,7 +803,7 @@ def forward(self, inputs, state=()):
         # call super to preprocess inputs
         z, state = super().forward(inputs, state, max_outer_rank=2)
         if self._img_encoding_net is None and len(self._fc_layers) == 0:
-            if inputs.ndim == 2:
+            if len(alf.nest.get_nest_shape(inputs)) == 2:
```
hnyu: This check is not general enough. Imagine that I have several inputs but get_nest_shape will return the shape of an arbitrary input.

alf/utils/data_buffer.py
```diff
@@ -133,6 +133,10 @@ def _create_buffer(tensor_spec):
         if allow_multiprocess:
             self.share_memory()
 
+    def circular(self, t):
```
hnyu: prefix an '_' as it's private use.

hnyu (changes requested):

hnyu:
FYI, we noticed that ddpg_fetchpush might not learn anything at all sometimes, as shown below with 20 envs (one run out of three):

image

This phenomenon is also observed in FetchReach where one DDPG run out of three learns nothing.

le-horizon:
> FYI, we noticed that ddpg_fetchpush might not learn anything at all sometimes, as shown below with 20 envs (one run out of three):
> 
> image
> 
> This phenomenon is also observed in FetchReach where one DDPG run out of three learns nothing.

Oh, then maybe it's the same problem HER is seeing.

I tried softsign, it's better, as gradients are no longer exactly zero, but still, they are very close to zero, and agent isn't learning.  I'm trying lowering AdamTF.lr next.
image

alf/examples/ddpg_fetchreach.gin
le-horizon: Done.

alf/networks/encoding_networks.py
le-horizon: I wonder what would be a good solution here.  Using any field that's not explicitly specified would be risky?

alf/utils/data_buffer.py
le-horizon: It's used by subclass as well.

le-horizon (commented):
Implemented the more efficient version of her sampling with the help of indexes to the beginning and end of episodes.

Time spent on HER is now negligible.

Thanks,
Le

alf/networks/encoding_networks.py
emailweixu: ImageEncodingNetwork takes only one input

alf/experience_replayers/replay_buffer.py
```diff
@@ -180,6 +266,8 @@ def get_batch(self, batch_size, batch_length):
                         all non-zero importance weights in the buffer.
         """
         with alf.device(self._device):
+            result, env_ids, pos = self._get_sample_batch(
+                batch_size, batch_length)
```
emailweixu: There is already uniform_sample() or prioritized_sample() in the lines below.

alf/networks/encoding_networks.py
le-horizon: This is the inputs of the ParallelEncodingNetwork, with self._img_encoding_net == None?  Not sure what you mean here.

alf/environments/suite_robotics.py
le-horizon: Good point.  Done.

le-horizon (commented):
Thanks for the comments Wei.  Addressed all.
Have one question.  PTAL.

Thanks,
Le

alf/networks/encoding_networks.py
le-horizon: Using z.ndim in the end as suggested by Wei.  @hnyu let me know if you have a better idea.

le-horizon (commented):
Added two small fixes.

Thanks,
Le

alf/environments/gym_wrappers.py
```diff
@@ -145,7 +145,8 @@ def transform_space(self, observation_space):
         return observation_space
 
     def transform_observation(self, observation):
-        transpose = self._need_channel_transpose(observation.shape)
+        transpose = self._need_channel_transpose(
+            alf.nest.get_nest_shape(observation))
```
emailweixu: I think observation is not nested.

alf/environments/gym_wrappers.py
```diff
@@ -547,7 +551,9 @@ def _clip_action(space, action):
                 raise ValueError(
                     "NAN action detected! action: {}".format(action))
             if isinstance(space, gym.spaces.Box):
-                action = np.maximum(np.minimum(action, space.high), space.low)
+                action = np.maximum(
+                    np.minimum(action, np.minimum(space.high, self.max_v)),
```
emailweixu: can be saved at line 545.

alf/environments/suite_robotics.py
emailweixu: I still think this should be parameter passed from `__init__`.

alf/utils/data_buffer.py
```diff
@@ -227,7 +231,7 @@ def _enqueue(self, batch, env_ids=None):
                                                  bat.detach()), self._buffer,
                 batch)
 
-            self._current_pos[env_ids] = (current_pos + 1) % self._max_length
+            self._current_pos[env_ids] = self.circular(current_pos + 1)
```
emailweixu: perhaps we don't even apply circular here. We only need to apply circular when we want to use position to access buffer elements. Perhaps code will be cleaner and easier to understand that way.

alf/experience_replayers/replay_buffer.py
```diff
@@ -253,6 +343,148 @@ def _prioritized_sample(self, batch_size, batch_length):
...
+                not FIRST steps.  We need to update last step pos for all
+                these env_ids[non_first_idx].
+            pos (tensor): position of the stored batch.
+            env_ids (tensor): env_ids of the stored batch.
+        """
+        _env_ids = env_ids[non_first]
+        _pos = pos[non_first]
+        prev_pos = self.circular(_pos - 1)
+        prev_first = self._index[(_env_ids, prev_pos)]
+        padded_prev_first = self.pad(prev_first, _env_ids)
```
emailweixu: If we never apply circular() to position, we don't need to pad.

alf/experience_replayers/replay_buffer.py
```diff
@@ -253,6 +343,148 @@ def _prioritized_sample(self, batch_size, batch_length):
         indices = self._sum_tree.find_sum_bound(r)
         return self._index_to_env_id_pos(indices)
 
+    def pad(self, x, env_ids):
```
emailweixu: prefix with "_" for private function

alf/environments/gym_wrappers.py
le-horizon: Somehow observation is a dict.  The transform function isn't recursing into dict, only namedtuple it seems.

alf/environments/gym_wrappers.py
le-horizon: Good point.

alf/environments/suite_robotics.py
le-horizon: Done.  I'm fine either ways, but as the code is getting complicated due to nest, it's better not to hard code it.

alf/utils/data_buffer.py
le-horizon: Good point.  Done.
It's a bit tricky because stored pos become outdated as new data are added into the buffer.  So comparing two pos stored in buffer would typically require padding to be correct.

alf/experience_replayers/replay_buffer.py
le-horizon: Done.

alf/experience_replayers/replay_buffer.py
le-horizon: True here, but not always true in general, because pos in _indexed_pos becomes outdated as we add new data.

le-horizon (commented):
Addressed all comments.
@hnyu could you help check the action_l2 cost in actor_networks.py to see if that's how I should implement it?

Thanks,
Le

alf/experience_replayers/replay_buffer.py
```diff
@@ -27,13 +27,19 @@
 from .segment_tree import SumSegmentTree, MaxSegmentTree, MinSegmentTree
 
 BatchInfo = namedtuple(
-    "BatchInfo", ["env_ids", "positions", "importance_weights"],
-    default_value=())
+    "BatchInfo", ["env_ids", "index", "importance_weights"], default_value=())
```
emailweixu: We can still use positions so that we can know whether a position is outdated (removed) during update_priority()

le-horizon (commented):
Addressed comments.  Please take another look.

Thanks,
Le

alf/experience_replayers/replay_buffer.py
```diff
@@ -203,16 +204,16 @@ def _change_mini_batch_length(self, mini_batch_length):
     # ``prioritized_sample()`` for the idx sampled to be still valid
```
emailweixu: not ``idx`` anymore. It's possible to check whether positions are still valid by comparing with the current_pos.

alf/experience_replayers/replay_buffer.py
```diff
@@ -254,7 +360,164 @@ def _prioritized_sample(self, batch_size, batch_length):
...
+    def _pad(self, x, env_ids):
+        """Make ``x`` (any index) absolute positions in the RingBuffer.
+
+        This is the reverse of ``circular()``, and is useful when trying to
+        compute the distance from one index to the other.  This is done by
+        adding multiples of _max_length back to the index.
+
+        NOTE, this operation depends on the _current_pos of the RingBuffer,
+        and can generate different results if new data are added to the buffer.
+        """
```
emailweixu, line 364: position can be calculated more efficiently:
position = idx + n L
current_pos - L <= idx + n L < current_pos
current_pos - idx - 1 - L < n L <= current_pos - idx - 1
n = (current_pos - idx - 1) / L
position =  ((current_pos - idx - 1) / L) * L + idx

alf/utils/data_buffer.py
hnyu: It's fine to have '_' even for subclass.

alf/networks/actor_networks.py
```diff
@@ -134,6 +134,8 @@ def forward(self, observation, state=()):
 
         observation, state = super().forward(observation, state)
         encoded_obs, _ = self._encoding_net(observation)
+        if isinstance(info, dict):
+            info["pre_activation"] = encoded_obs
```
hnyu: In principle shouldn't change the code of actor_networks.py. Should only modify ddpg's code.

alf/algorithms/ddpg_algorithm.py
```diff
@@ -256,25 +259,29 @@ def _critic_train_step(self, exp: Experience, state: DdpgCriticState):
...
 
-        def actor_loss_fn(dqda, action):
+        def actor_loss_fn(dqda, action, pre_activation):
             if self._dqda_clipping:
                 dqda = torch.clamp(dqda, -self._dqda_clipping,
                                    self._dqda_clipping)
             loss = 0.5 * losses.element_wise_squared_loss(
                 (dqda + action).detach(), action)
             loss = loss.sum(list(range(1, loss.ndim)))
+            if self.action_l2 > 0:
```
hnyu: Prefix `action_l2` with "_". And just need to compute a cost based on the current `action`, assuming the computation of `action` is differentiable (which is true if we use `ActorNetwork` or `rsample()`). But in practice I didn't observe much benefit of having this action cost.

hnyu (changes requested):

alf/experience_replayers/replay_buffer.py
```diff
@@ -231,9 +339,7 @@ def _uniform_sample(self, batch_size, batch_length):
         num_positions = num_positions[env_ids]
         pos = (r * num_positions).to(torch.int64)
         pos += (self._current_pos - self._current_size)[env_ids]
-        pos = pos % self._max_length
-
-        return env_ids, pos
+        return env_ids, self.circular(pos)
```
emailweixu: We can still return `pos` so that we don't need to have _pad() at line 297. To be consistent with this, _priorized_sample() need to change to return pos too. Or perhaps it's better to return BatchInfo from both functions.

alf/experience_replayers/replay_buffer.py
```diff
@@ -254,7 +360,164 @@ def _prioritized_sample(self, batch_size, batch_length):
...
+            idx (tensor): indices of the sampled data, of shape
+                [batch_size, batch_length]
+        Returns:
+            nested Tensors. The shapes are [batch_size, batch_length, ...]
+        """
+        if self._her_k == 0:
+            return result
+
+        # TODO: add support for batch_length > 2.
+        assert batch_length == 2
```
emailweixu: does batch_length of 1 also work?

alf/algorithms/ddpg_algorithm.py
```diff
@@ -132,6 +134,7 @@ def __init__(self,
         else:
             critic_networks = alf.networks.NaiveParallelNetwork(
                 critic_network, num_critic_replicas)
+        self.action_l2 = action_l2
```
emailweixu: self._action_l2

alf/algorithms/ddpg_algorithm.py
```diff
@@ -272,6 +275,9 @@ def actor_loss_fn(dqda, action):
             loss = 0.5 * losses.element_wise_squared_loss(
                 (dqda + action).detach(), action)
             loss = loss.sum(list(range(1, loss.ndim)))
+            if self.action_l2 > 0:
+                assert action.requires_grad
+                loss += self.action_l2 * torch.norm(action)
```
emailweixu: This is not preactivation as commented at line 122.  And it should be per action norm: `torch.norm(action, dim=1)`

alf/environments/gym_wrappers.py
emailweixu: I checked transform_nest, it does recurse into dict.  If not, should fix transform_nest instead of fixing here.

alf/utils/data_buffer.py
```diff
@@ -467,8 +478,7 @@ def get_batch_by_indices(self, indices):
         with alf.device(self._device):
             indices = convert_device(indices)
             indices.copy_(
-                (indices +
-                 (self.current_pos - self.current_size)) % self._capacity)
+                self.circular(indices + self.current_pos - self.current_size))
```
emailweixu: better not to use `copy_` as it changes value of indices, which is typically unexpected by the caller.

alf/experience_replayers/replay_buffer.py
emailweixu: her_k => her_proportion to be more meaningful.

alf/experience_replayers/replay_buffer.py
```diff
@@ -41,28 +51,39 @@ def __init__(self,
...
-            device (str): A torch device to place the Variables and ops.
-            allow_multiprocess (bool): if ``True``, allows multiple processes
-                to write and read the buffer asynchronously.
-            name (str): name of the replay buffer.
+            device (string): "cpu" or "cuda" where tensors are created.
+            allow_multiprocess (bool): whether multiprocessing is supported.
+            her_k (float): proportion of hindsight relabeled experience.
+            step_type_field (string): path to the step_type field in exp nest.
+                This and the following fields are for hindsight relabeling.
+            achieved_goal_field (string): path to the achieved_goal field in
```
emailweixu: string => str. str is the type name for string.

alf/experience_replayers/replay_buffer.py
```diff
@@ -41,28 +51,39 @@ def __init__(self,
...
+                exp nest.
+            desired_goal_field (string): path to the desired_goal field in the
+                exp nest.
+            reward_field (string): path to the reward field in the exp nest.
+                are for hindsight experience replay.
+            reward_fn (callable): function to recompute reward based on
```
emailweixu: callable => Callable. `typing.Callable` is the actual type.

alf/experience_replayers/replay_buffer.py
```diff
@@ -254,7 +362,187 @@ def _prioritized_sample(self, batch_size, batch_length):
...
+        NOTE, this operation depends on the _current_pos of the RingBuffer,
+        and can generate different results if new data are added to the buffer.
+        """
+        multiples = (
+            self._current_pos[env_ids] / self._max_length) * self._max_length
+        idx = self.circular(x)
+        return torch.where(idx < self.circular(self._current_pos[env_ids]),
+                           idx + multiples, idx + multiples - self._max_length)
+
+    def store_episode_end_pos(self, non_first, pos, env_ids):
```
emailweixu: should this be a private function?

alf/experience_replayers/replay_buffer.py
```diff
@@ -254,7 +362,187 @@ def _prioritized_sample(self, batch_size, batch_length):
...
+        # (The case where current step is a ``FIRST`` step can be safely
+        # ignored because the first_step_pos points to episode end which
+        # is guarranteed to be existing.)
+        headless, = torch.where(
+            self._current_pos[env_ids] > first_step_pos + self._max_length)
+        headless_env_ids = env_ids[headless]
+        result[headless] = self._headless_indexed_pos[headless_env_ids]
+        return result
+
+    def distance_to_episode_end(self, idx, env_ids):
```
emailweixu: distance_to_episode_end => steps_to_episode_end to be more accurate.

alf/experience_replayers/replay_buffer.py
```diff
@@ -254,7 +362,187 @@ def _prioritized_sample(self, batch_size, batch_length):
...
+        self._indexed_pos[(_env_ids[has_head],
+                           prev_first_idx[has_head])] = _pos[has_head]
+        # For a headless episode whose ``FIRST`` step was overwritten by new
+        # data, the current step has to belong to the same episode as all the
+        # other steps in the buffer, i.e. episode is longer than max_length of
+        # the buffer.  This means prev_first <= _pos - max_length.
+        headless, = torch.where(torch.logical_not(has_head_cond))
+        self._headless_indexed_pos[_env_ids[headless]] = _pos[headless]
+
+    def get_episode_end_pos(self, idx, env_ids):
```
emailweixu: _get_episode_end_pose to make it private function. Otherwise, need to document idx and env_ids. For a public function, it might be better to use position instead of idx as argument.

alf/experience_replayers/replay_buffer.py
```diff
@@ -254,7 +362,187 @@ def _prioritized_sample(self, batch_size, batch_length):
...
+        """Use _indexed_pos and _headless_indexed_pos to get last step pos.
+
+        NOTE, the positions returned are the stored positions, are
+        possibly outdated, and need to be padded with self._pad().
```
emailweixu: How can they be outdated?

alf/experience_replayers/replay_buffer.py
```diff
@@ -254,7 +362,187 @@ def _prioritized_sample(self, batch_size, batch_length):
...
+        """
+        # Look up ``FIRST`` step position, then look up the stored last pos
+        first_step_pos = self._indexed_pos[(env_ids, idx)]
+        first_step_idx = self.circular(first_step_pos)
+        result = self._indexed_pos[(env_ids, first_step_idx)].clone()
```
emailweixu: Why clone?

alf/experience_replayers/replay_buffer.py
```diff
@@ -254,7 +362,187 @@ def _prioritized_sample(self, batch_size, batch_length):
...
+                     result_desired_goal[non_her_indices],
+                     env_ids[non_her_indices], idx[non_her_indices])
+            logging.warning(msg)
+            # assert False, msg
+
+        if alf.summary.get_global_counter() == 64:
+            step_types = alf.nest.get_field(self._buffer,
+                                            self._step_type_field)
+            logging.info("step_types:\n", step_types)
+            logging.info("achieved goals:\n", achieved_goals)
```
emailweixu: should be removed?

emailweixu (changes requested):

le-horizon:
Hey @hnyu and @emailweixu, I addressed your concern about polluting the ReplayBuffer's interface.  Please take another look and see if you have more suggestions.

Thanks,
Le

alf/algorithms/ddpg_algorithm.py
le-horizon: done

alf/utils/data_buffer.py
le-horizon: good catch.

alf/experience_replayers/replay_buffer.py
```diff
@@ -70,19 +67,10 @@ def __init__(self,
...
-                exp nest.
-            desired_goal_field (string): path to the desired_goal field in the
-                exp nest.
-            reward_field (string): path to the reward field in the exp nest.
-                are for hindsight experience replay.
-            reward_fn (callable): function to recompute reward based on
-                achieve_goal and desired_goal.  Default gives reward 0 when
-                L2 distance less than 0.05 and -1 otherwise, same as is done in
-                suite_robotics environments.
+            postprocess_exp_fn (callable): function to postprocess experience.
```
emailweixu, line 73: Comment the expected arguments and return of this function.

alf/experience_replayers/replay_buffer.py
```diff
@@ -312,9 +283,9 @@ def get_batch(self, batch_size, batch_length):
                 info = info._replace(
                     importance_weights=self._sum_tree[indices] * avg_weight)
 
-            if self._her_k > 0:
-                result = self._hindsight_relabel(batch_size, batch_length,
-                                                 result, out_env_ids, idx)
+            if self._postprocess_exp_fn:
+                result, info = self._postprocess_exp_fn(
+                    self, result, info, device=self._device)
```
emailweixu: No need to pass device. It can be obtained from self

alf/experience_replayers/replay_buffer.py
```diff
@@ -296,7 +267,7 @@ def get_batch(self, batch_size, batch_length):
 
             start_idx = idx
             info = BatchInfo(
-                env_ids=env_ids, positions=self._pad(idx, env_ids))
+                env_ids=env_ids, positions=self._pad(start_idx, env_ids))
```
emailweixu: return `BatchInfo` directly from _prioritized_sample() and _uniform_sample() so that we can avoid redundant _pad() and circular()

alf/algorithms/ddpg_algorithm.py
```diff
@@ -194,8 +194,10 @@ def __init__(self,
         self._dqda_clipping = dqda_clipping
 
     def predict_step(self, time_step: TimeStep, state, epsilon_greedy=1.):
+        self._actor_network.eval = True
```
emailweixu: torch.nn.Module already has function eval() and train() https://pytorch.org/docs/stable/nn.html#torch.nn.Module.eval for this purpose.

alf/algorithms/ddpg_algorithm.py
```diff
@@ -277,7 +279,8 @@ def actor_loss_fn(dqda, action):
             loss = loss.sum(list(range(1, loss.ndim)))
             if self.action_l2 > 0:
                 assert action.requires_grad
-                loss += self.action_l2 * torch.norm(action)
+                loss += self.action_l2 * (torch.mean(
+                    torch.norm(action, dim=-1)**2))
```
emailweixu: In case that action is multi-dimensional: `torch.norm(list(range(1, action.ndim))`

alf/networks/actor_networks.py
```diff
@@ -133,13 +137,31 @@ def forward(self, observation, state=()):
...
+            alf.summary.scalar(name=name, data=_info["pre_activation"].norm())
+
             action = spec_utils.scale_to_spec(action, spec)
+
+            a_name = ('summarize_output/' + self.name + '.action_layer.' +
+                      str(i) + '.action.output_norm')
+            if self.eval:
+                a_name += ".eval"
+            alf.summary.scalar(name=a_name, data=action.norm())
+
```
emailweixu, line 155: By default, this should not be enabled  because it might slow down training a lot.

alf/environments/gym_wrappers.py
emailweixu: What do you think about this?
> I checked transform_nest, it does recurse into dict. If not, should fix transform_nest instead of fixing here.

alf/experience_replayers/replay_buffer.py
emailweixu: not idx anymore

alf/experience_replayers/replay_buffer.py
emailweixu: simplify as above

alf/experience_replayers/replay_buffer.py
emailweixu: make it private

alf/experience_replayers/replay_buffer.py
emailweixu: make it private

emailweixu (commented):
Please also address previous review comments.

alf/environments/gym_wrappers.py
le-horizon: Good point.  Found the problem.  it's due to field being None, in which case the code now is skipping any processing.

alf/algorithms/ddpg_algorithm.py
le-horizon: interesting.

alf/algorithms/ddpg_algorithm.py
le-horizon: done.

alf/networks/actor_networks.py
le-horizon: Done.

alf/experience_replayers/replay_buffer.py
le-horizon: Good point, done.

alf/experience_replayers/replay_buffer.py
le-horizon: done

alf/experience_replayers/replay_buffer.py
le-horizon: Good point.  I thought changes to result's content can alter _index_pos

alf/experience_replayers/replay_buffer.py
le-horizon: You are right.  The episode end pos shouldn't be out of date, unless async is going on.  The content in _indexed_pos can be outdated in the headless case.

alf/experience_replayers/replay_buffer.py
le-horizon: Done.

alf/experience_replayers/replay_buffer.py
le-horizon: Done

alf/experience_replayers/replay_buffer.py
le-horizon: Done

alf/experience_replayers/replay_buffer.py
le-horizon: length == 1 will also work, but will that be useful?

alf/experience_replayers/replay_buffer.py
le-horizon: Done.

alf/algorithms/ddpg_algorithm.py
le-horizon: Good catch.

alf/experience_replayers/replay_buffer.py
le-horizon: Done.

alf/experience_replayers/replay_buffer.py
le-horizon: Done.

alf/experience_replayers/replay_buffer.py
le-horizon: Done.

alf/experience_replayers/replay_buffer.py
le-horizon: updated code so only valid positions are updated.

alf/experience_replayers/replay_buffer.py
le-horizon: Nice!  git collapsed this file, so I missed a bunch of comments from previous reviews..

le-horizon (commented):
Missed a bunch of comments due to github collapsing the long file automatically.. Should've addressed all comments this time.

Thanks,
Le

alf/networks/actor_networks.py
```diff
@@ -137,28 +135,30 @@ def forward(self, observation, state=()):
         """
 
         observation, state = super().forward(observation, state)
-        self._encoding_net.eval = self.eval
+        self._encoding_net.train(self.training)
```
emailweixu: I believe nn.Module.train() will update submodules recursively. So no need to do it here again.

alf/algorithms/ddpg_algorithm.py
```diff
@@ -194,10 +194,10 @@ def __init__(self,
         self._dqda_clipping = dqda_clipping
 
     def predict_step(self, time_step: TimeStep, state, epsilon_greedy=1.):
-        self._actor_network.eval = True
+        self._actor_network.eval()
```
emailweixu: do this at policy_trainer.py _step()

alf/algorithms/ddpg_algorithm.py
```diff
@@ -277,10 +277,13 @@ def actor_loss_fn(dqda, action):
...
             loss = loss.sum(list(range(1, loss.ndim)))
-            if self.action_l2 > 0:
+            if self._action_l2 > 0:
                 assert action.requires_grad
-                loss += self.action_l2 * (torch.mean(
-                    torch.norm(action, dim=-1)**2))
+                assert action.ndim < 3, (
+                    "Action has ndim: {}.  Tensor.norm ".format(action.ndim) +
+                    "will incorrectly take the norm of the last dim.")
+                loss += self._action_l2 * (torch.mean(
```
emailweixu: The shape of loss is [B]. We should add the action_l2 loss separately to each sample:
```
loss += self._action_l2 * (action * action).sum(list(range(1, action.ndim)))
```

alf/networks/actor_networks.py
```diff
@@ -137,28 +135,30 @@ def forward(self, observation, state=()):
...
-            if self.eval:
-                name += ".eval"
-            alf.summary.scalar(name=name, data=_info["pre_activation"].norm())
+            if alf.summary.should_summarize_output():
+                name = ('summarize_output/' + self.name + '.action_layer.' +
+                        str(i) + '.pre_activation.output_norm')
+                if not self.training:
+                    name += ".eval"
+                alf.summary.scalar(
+                    name=name, data=_info["pre_activation"].norm())
```
emailweixu: it's better to be average across samples so it can be easily compared if batch size is changed.

alf/algorithms/ddpg_algorithm.py
le-horizon: good point.

alf/algorithms/ddpg_algorithm.py
le-horizon: Ah, it's so easy to make a mistake or several mistakes!

alf/networks/actor_networks.py
le-horizon: Good point.

alf/networks/actor_networks.py
le-horizon: Good point.

alf/experience_replayers/replay_buffer.py
le-horizon: Done

alf/experience_replayers/replay_buffer.py
le-horizon: Done

le-horizon (commented):
Fixed again.  Also running more experiments to verify performance.

Thanks,
Le

emailweixu (dismissed):

le-horizon:
@hnyu still need your approval.  Let me know if you have more comments.
Btw, github page of this PR is slow to load, so refresh if first time took too long and didn't load.

Looks like we broke github :P

alf/layers.py
```diff
@@ -286,8 +286,11 @@ def reset_parameters(self):
         if self._use_bias:
             nn.init.constant_(self._linear.bias.data, self._bias_init_value)
 
-    def forward(self, inputs):
-        return self._activation(self._linear(inputs))
+    def forward(self, inputs, info=None):
```
hnyu: This seems quite hacky. The other layers don't have this signature.

alf/networks/actor_networks.py
```diff
@@ -136,10 +138,34 @@ def forward(self, observation, state=()):
...
         actions = []
+        i = 0
         for layer, spec in zip(self._action_layers, self._flat_action_spec):
-            action = layer(encoded_obs)
+            _info = {} if alf.summary.should_summarize_output() else None
+            action = layer(encoded_obs, info=_info)
+            if alf.summary.should_summarize_output():
+                name = ('summarize_output/' + self.name + '.action_layer.' +
+                        str(i) + '.pre_activation.output_norm')
+                if not self.training:
```
hnyu: Where is this self.training defined?

le-horizon: This is from nn.Module, set by nn.Module.train(mode).

alf/layers.py
le-horizon: Yes, agree.  Let me remove this for now.

le-horizon:
@hnyu Done.

hnyu (approved):

PR #588, opened 2020-05-30
prioritized replay

### LINKED ISSUE #562: Prioritized Replay

emailweixu:
This PR is only for replay buffer. Not integrated with Algorithm yet.

### DIFF
--- a/alf/experience_replayers/replay_buffer.py
+++ b/alf/experience_replayers/replay_buffer.py
@@ -31,18 +31,24 @@ BatchInfo = namedtuple(
 
 
 class ReplayBuffer(RingBuffer):
-    """Replay buffer with RingBuffer as implementation.
-    """
+    """Replay buffer with RingBuffer as implementation."""
 
     def __init__(self,
                  data_spec,
                  num_environments,
                  max_length=1024,
                  prioritized_sampling=False,
+                 initial_priority=1.0,
                  device="cpu",
                  allow_multiprocess=False,
                  name="ReplayBuffer"):
         """
+        If ``prioritized_sampling`` is set to True, instead of sampling experiences
+        uniformly from the replay buffer, ``get_batch()`` samples experiences
+        with probability proportional to the priority of each experience. The
+        initial priority of a new expeirence added to the buffer is set to
+        ``initial_priority``. The priorities can be updated using ``update_priority()``.
+
         Args:
             data_spec (nested TensorSpec): spec describing a single item that
                 can be stored in this buffer.
@@ -50,6 +56,9 @@ class ReplayBuffer(RingBuffer):
             max_length (int): The maximum number of items that can be stored
                 for a single environment.
             prioritized_sampling (bool): Use prioritized sampling if this is True.
+            initial_priority (float): initial priority used for new experiences.
+                The actual initial priority used for new experience is the maximum
+                of this value and the current maximum priority of all experiences.
             device (str): A torch device to place the Variables and ops.
             allow_multiprocess (bool): if ``True``, allows multiple processes
                 to write and read the buffer asynchronously.
@@ -67,8 +76,10 @@ class ReplayBuffer(RingBuffer):
         if prioritized_sampling:
             self._mini_batch_length = 1
             tree_size = self._max_length * num_environments
-            self._sum_tree = SumSegmentTree(tree_size)
-            self._max_tree = MaxSegmentTree(tree_size)
+            self._sum_tree = SumSegmentTree(tree_size, device=device)
+            self._max_tree = MaxSegmentTree(tree_size, device=device)
+            self._initial_priority = torch.tensor(
+                initial_priority, dtype=torch.float32, device=device)
 
     def add_batch(self, batch, env_ids=None, blocking=False):
         with alf.device(self._device):
@@ -85,7 +96,8 @@ class ReplayBuffer(RingBuffer):
         be used for training sooner. We make it at least 1.0 so that it can never
         be very small.
         """
-        return torch.max(self._max_tree.summary(), torch.tensor(1.0))
+        return convert_device(
+            torch.max(self._max_tree.summary(), self._initial_priority))
 
     def _initialize_priority(self, env_ids):
         last_pos = (self._current_pos[env_ids] -
--- a/alf/experience_replayers/replay_buffer.py
+++ b/alf/experience_replayers/replay_buffer.py
@@ -228,8 +228,8 @@ class ReplayBuffer(RingBuffer):
             "batch_length=%s" % batch_length)
 
         r = torch.rand((batch_size, ))
-        r = r + torch.arange(batch_size, dtype=torch.float32)
-        r = r * (self._sum_tree.summary() / batch_size)
+        r = (r + torch.arange(batch_size, dtype=torch.float32)) / batch_size
+        r = r * self._sum_tree.summary()
         indices = self._sum_tree.find_sum_bound(r)
         return self._index_to_env_id_pos(indices)
 
--- a/alf/experience_replayers/replay_buffer.py
+++ b/alf/experience_replayers/replay_buffer.py
@@ -18,8 +18,16 @@ import torch
 import torch.nn as nn
 
 import alf
+from alf.data_structures import namedtuple
 from alf.nest.utils import convert_device
+from alf.utils.common import warning_once
 from alf.utils.data_buffer import atomic, RingBuffer
+
+from .segment_tree import SumSegmentTree, MaxSegmentTree, MinSegmentTree
+
+BatchInfo = namedtuple(
+    "BatchInfo", ["env_ids", "positions", "importance_weights"],
+    default_value=())
 
 
 class ReplayBuffer(RingBuffer):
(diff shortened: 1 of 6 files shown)


alf/experience_replayers/segment_tree.py
```diff
@@ -186,12 +186,20 @@ def _step(indices, thresholds):
             """
             indices *= 2
             left = self._values[indices]
-            greater = thresholds >= left
+            right = self._values[indices + 1]
+            # The condition (thresholds >= left) * (right > 0) is only possible
```
hnyu: (thresholds >= left) * (right == 0) is only possible ...

alf/experience_replayers/segment_tree_test.py
```diff
@@ -65,7 +65,15 @@ def test_sum_tree(self):
 
             thresh = tree.summary().reshape(1)
             self.assertEqual(tree.find_sum_bound(thresh), size - 1)
-            self.assertEqual(tree.find_sum_bound(thresh + 1), size - 1)
+            self.assertRaises(ValueError, tree.find_sum_bound, thresh + 1)
+
+    def test_boundary(self):
+        tree = SumSegmentTree(10)
+        v = torch.tensor([0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.0])
```
hnyu: [0.1] * 9 + [0.0]

alf/experience_replayers/replay_buffer.py
```diff
@@ -41,8 +63,87 @@ def __init__(self,
...
+        return torch.max(self._max_tree.summary(), torch.tensor(1.0))
+
+    def _initialize_priority(self, env_ids):
+        last_pos = (self._current_pos[env_ids] -
+                    self._mini_batch_length) % self._max_length
+        indices = self._env_id_pos_to_index(env_ids, last_pos)
+        values = torch.full(indices.shape, self.initial_priority)
+
+        if self._mini_batch_length > 1:
+            last_pos = (self._current_pos[env_ids] - 1) % self._max_length
```
hnyu, line 109: Why this additional value updating? Seems that it sets the end of a mini_batch_length to 0 but doesn't set the middle values.

alf/experience_replayers/replay_buffer.py
```diff
@@ -18,9 +18,17 @@
...
 
+from .segment_tree import SumSegmentTree, MaxSegmentTree, MinSegmentTree
+
+BatchInfo = namedtuple(
+    "BatchInfo", ["env_ids", "positions", "importance_weights"],
+    default_value=())
+
 
 class ReplayBuffer(RingBuffer):
     """Replay buffer with RingBuffer as implementation.
```
hnyu: Add comment on what behavior to expect for this replay buffer if `prioritized_sampling=True` for mini_batch_len=1 and >1

alf/experience_replayers/replay_buffer.py
```diff
@@ -55,39 +156,82 @@ def get_batch(self, batch_size, batch_length):
...
+
+    def _prioritized_sample(self, batch_size, batch_length):
+        if batch_length != self._mini_batch_length:
+            if self._mini_batch_length > 1:
+                warning_once(
+                    "It is not advisable to use different batch_length"
+                    "for different calls to get_batch(). Previous batch_length=%d "
+                    "new batch_length=%d" % (self._mini_batch_length,
+                                             batch_length))
+            self._change_mini_batch_length(batch_length)
```
hnyu, line 243: If I've added many batches with mini_batch_len=1 but sample with mini_batch_len>1, will there be any problem?

hnyu (changes requested):

alf/experience_replayers/replay_buffer_test.py
```diff
@@ -137,6 +137,71 @@ def test_replay_buffer(self, allow_multiprocess):
...
+        batch1 = get_batch([1], self.dim, x=0.25, t=0)
+        replay_buffer.add_batch(batch1, batch1.env_id)
+
+        batch, batch_info = replay_buffer.get_batch(1, 1)
+        self.assertEqual(batch_info.env_ids,
+                         torch.tensor([1], dtype=torch.int64))
+        self.assertEqual(batch_info.importance_weights, torch.tensor([1.]))
+        self.assertRaises(AssertionError, replay_buffer.get_batch, 1, 2)
+
+        batch2 = get_batch([1], self.dim, x=0.5, t=1)
```
le-horizon, line 158: batch2 is not used?

alf/experience_replayers/replay_buffer.py
le-horizon: We currently only support adding batch_length==1.  It is sampling that this comment is talking about.

alf/experience_replayers/replay_buffer.py
```diff
@@ -41,8 +63,87 @@ def __init__(self,
             allow_multiprocess=allow_multiprocess,
             name=name)
 
+        self._prioritized_sampling = prioritized_sampling
+        if prioritized_sampling:
+            self._mini_batch_length = 1
```
le-horizon, line 77: The smallest length we can possibly train is 2?  Why do we set 1 here?

alf/experience_replayers/replay_buffer.py
```diff
@@ -41,8 +63,87 @@ def __init__(self,
...
+            tree_size = self._max_length * num_environments
+            self._sum_tree = SumSegmentTree(tree_size)
+            self._max_tree = MaxSegmentTree(tree_size)
+            self._min_tree = MinSegmentTree(tree_size)
```
le-horizon: Looks like min_tree is not used?

alf/experience_replayers/replay_buffer.py
```diff
@@ -41,8 +63,87 @@ def __init__(self,
...
+        with alf.device(self._device):
+            env_ids = self.check_convert_env_ids(env_ids)
+            self.enqueue(batch, env_ids, blocking=blocking)
+            if self._prioritized_sampling:
+                self._initialize_priority(env_ids)
+
+    @property
+    def initial_priority(self):
+        """The initial priority used for newly added experiences."""
+        return torch.max(self._max_tree.summary(), torch.tensor(1.0))
```
le-horizon: comment why max is used as initial priority and why it can be larger than 1?

le-horizon (commented):
LG, mostly nits.

alf/experience_replayers/replay_buffer.py
emailweixu: It can be one. For example, SarsaAlgorithm or supervised training.

alf/experience_replayers/replay_buffer.py
emailweixu: Will remove.

alf/experience_replayers/replay_buffer.py
emailweixu: the max from max tree can be anything, large or small.

alf/experience_replayers/replay_buffer.py
emailweixu: Because the previous zero values are updated in previous add_batch()

alf/experience_replayers/replay_buffer.py
emailweixu: This is not a problem because change_mini_batch_length() will set the weights for the last (batch_length-1) experiences to 0.

alf/experience_replayers/replay_buffer_test.py
emailweixu: Will fix.

alf/experience_replayers/segment_tree.py
emailweixu: nice catch

alf/experience_replayers/segment_tree_test.py
emailweixu: changed.

alf/experience_replayers/replay_buffer.py
hnyu: I mean there is a function that changes mini_batch_length when sampling (not for adding) in this PR. On line 92 every time a batch is added, there is a check for self._mini_batch_length > 1. And self._mini_batch_length == 1 in the beginning. This value won't get changed until the first call of sampling (might be late during training).

hnyu: OK I see.

alf/experience_replayers/replay_buffer.py
hnyu: I wonder that this constant (1.0 currently) can be problem-dependent? It's also based on what kind of weights data we are storing. Maybe make it a configurable constant.

emailweixu: This doesn’t matter unless all the actual weights set by update_priority are all smaller than one, which is very unlikely in my opinion.

emailweixu: This doesn’t matter unless all the actual weights set by update_priority are all smaller than one, which is very unlikely in my opinion.

hnyu: That's what I refer to. The range of weights might be [0,1] as an algorithm can define whatever a weight is. From the probability perspective, a larger constant > 1 definitely favors new data more.

hnyu:
I was re-visiting the Prioritized Replay paper, especially paying attention to the importance correction part of the paper. Their motivation is that they want to make sure that after prioritization, the expectation is still uniform over the buffer (just like uniform sampling). However, as I recall, for off-policy algorithms like DQN, it doesn't matter if every (s,a) pair is visited with the same frequency or not, as long as each pair continues to get updated. 

The ablation study in their paper about the role of importance correction shows:

image

(The uncorrected one's learning rate is even after being divided by 4)

Looking at these results, to me the correction doesn't seem necessary on top of a prioritized replay. As long as we control the max priority to a reasonable range so that the sampling is not heavily biased towards a handful of exps.

This is just for normal one-level off-policy algorithm. In a hierarchical case, we should still correct the TD error with another different importance ratio that accounts for the changing lower-level policy.

alf/experience_replayers/replay_buffer.py
le-horizon: I see.

Is it possible for two alg to share the same replaybuffer, e.g. to save memory?  In that case, if the two share a different batch_length, it would become very inefficient?

Anyways, I think it's fine to submit the current code, and defer these worries into the future when they actually appear with a more concrete context.

emailweixu:
They choose to use beta=0.4, so there's still some correction.

> I was re-visiting the Prioritized Replay paper, especially paying attention to the importance correction part of the paper. Their motivation is that they want to make sure that after prioritization, the expectation is still uniform over the buffer (just like uniform sampling). However, as I recall, for off-policy algorithms like DQN, it doesn't matter if every (s,a) pair is visited with the same frequency or not, as long as each pair continues to get updated.

hnyu (approved):

PR #590, opened 2020-06-02
Use prioritized replay for sarsa, sac and ddpg

### LINKED ISSUE #562: Prioritized Replay

emailweixu:


alf/algorithms/algorithm.py
```diff
@@ -266,7 +272,7 @@ def _set_exp_replayer(self, sample_exp):
                 self._experience_spec)
             self._exp_replayer = SyncUniformExperienceReplayer(
```
hnyu: The name of this replayer should now be changed, no longer uniform

alf/algorithms/algorithm.py
```diff
@@ -1161,22 +1184,24 @@ def train_from_replay_buffer(self, update_global_counter=False):
                 if config.clear_replay_buffer:
                     self._exp_replayer.clear()
                 num_updates = config.num_updates_per_train_iter
+                batch_info = None
             else:
-                experience = self._exp_replayer.replay(
+                experience, batch_info = self._exp_replayer.replay(
```
hnyu, line 1189: Can we directly put batch_info inside experience here?

alf/algorithms/algorithm.py
```diff
@@ -1323,13 +1364,22 @@ def _update(self, experience, weight):
         experience = dist_utils.params_to_distributions(
             experience, self.processed_experience_spec)
 
+        if batch_info is not None:
+            experience = experience._replace(batch_info=batch_info)
         loss_info = self.calc_loss(experience, train_info)
+        if loss_info.priority != ():
+            priority = (loss_info.priority**self._config.priority_replay_alpha
+                        + self._config.priority_replay_eps)
+            self._exp_replayer.update_priority(batch_info.env_ids,
```
hnyu, line 1373: What if the user wants to define his own priority that's not necessarily loss?

alf/algorithms/algorithm.py
```diff
@@ -1323,13 +1364,22 @@ def _update(self, experience, weight):
         experience = dist_utils.params_to_distributions(
             experience, self.processed_experience_spec)
 
+        if batch_info is not None:
+            experience = experience._replace(batch_info=batch_info)
         loss_info = self.calc_loss(experience, train_info)
+        if loss_info.priority != ():
+            priority = (loss_info.priority**self._config.priority_replay_alpha
```
hnyu, line 1371: If we are always storing `priority**alpha` then in the ReplayBuffer's comment, we should state that the `initial_priority` should be as if it's applied with alpha.

alf/algorithms/algorithm.py
```diff
@@ -976,16 +986,29 @@ def update_with_gradient(self, loss_info, valid_masks=None, weight=1.0):
...
         Returns:
             tuple:
             - loss_info (LossInfo): loss information.
             - params (list[(name, Parameter)]): list of parameters being updated.
         """
+        masks = None
+        if (batch_info is not None and batch_info.importance_weights != ()
+                and self._config.priority_replay_beta != 0):
+            masks = batch_info.importance_weights.pow(
+                -self._config.priority_replay_beta).unsqueeze(0)
```
hnyu, line 1000: In ReplayBuffer, when being returned, importance_weights has been multiplied by self._sum_tree.nnz. In the paper, however, this number should be N (the total number of leaves) regardless if there is any 0 value.

alf/algorithms/algorithm.py
```diff
@@ -976,16 +986,29 @@ def update_with_gradient(self, loss_info, valid_masks=None, weight=1.0):
...
+                ``ReplayBuffer.get_batch()``
         Returns:
             tuple:
             - loss_info (LossInfo): loss information.
             - params (list[(name, Parameter)]): list of parameters being updated.
         """
+        masks = None
+        if (batch_info is not None and batch_info.importance_weights != ()
+                and self._config.priority_replay_beta != 0):
+            masks = batch_info.importance_weights.pow(
```
hnyu, line 999: If mini_batch_length>1, then batch_info contains different importance weights along T. However, when replayed using the segment tree, the T timesteps  are sampled altogether using just the weight of the first step? Is it better to propagate the weight of the first step to the remaining T-1 steps?

alf/data_structures.py
```diff
@@ -120,15 +120,20 @@ class Experience(
...
             ],
             default_value=())):
     """An ``Experience`` is a ``TimeStep`` in the context of training an RL algorithm.
-    For the training purpose, it's augmented with three new attributes:
+    For the training purpose, it's augmented with several new attributes:
 
     - action: A (nested) ``Tensor`` for action taken for the current time step.
     - rollout_info: ``AlgStep.info`` from ``rollout_step()``.
     - state: State passed to ``rollout_step()`` to generate ``action``.
+    - batch_info: Its type is alf.experience_replays.replay_buffer.BatchInfo.
```
hnyu: \`` \``

alf/experience_replayers/experience_replay.py
```diff
@@ -59,8 +59,13 @@ def replay(self, sample_batch_size, mini_batch_length):
             mini_batch_length (int): the temporal length of each sample
 
         Output:
-            exp (Experience): each item has the shape (`sample_batch_size`,
-                `mini_batch_length`, ...)
+            tuple:
+                - nested Tensors: The samples. Its shapes are [batch_size, batch_length, ...]
```
hnyu, line 63: nested Tensors -> Experience

alf/experience_replayers/experience_replay.py
```diff
@@ -215,16 +228,34 @@ def replay(self, sample_batch_size, mini_batch_length):
             sample_batch_size (int): number of sequences
             mini_batch_length (int): the length of each sequence
         Returns:
-            Experience: experience batch in batch major (B, T, ...)
+            tuple:
+                - nested Tensors: The samples. Its shapes are [batch_size, batch_length, ...]
```
hnyu, line 232: nested Tensors -> Experience

hnyu (changes requested):

alf/algorithms/algorithm.py
emailweixu: The shape of importance with is [B]

alf/algorithms/algorithm.py
emailweixu: experience with 0 importance is effectively discarded. So nnz is N

alf/algorithms/algorithm.py
emailweixu: Here we only use LossInfo to pass priority. calc_loss can assign whatever value for LossInfo.priority.

alf/data_structures.py
emailweixu: fixed

alf/experience_replayers/experience_replay.py
emailweixu: It doesn't have to be Experience. User can decide to put any kind of data to the replay buffer.

alf/experience_replayers/experience_replay.py
emailweixu: same as above.

alf/algorithms/algorithm.py
emailweixu: ReplayBuffer does not care how priority is calculated.

alf/algorithms/algorithm.py
emailweixu: The shape batch_info is [B], where the shape of experience is [B, T, ...]. In most place of our code, a shape of [B,T,..] is assumed.

alf/algorithms/algorithm.py
hnyu: > ReplayBuffer does not care how priority is calculated.

I mean we have an argument that let the user specify `initial_priority`. When doing that, the user needs to keep in mind that the specified value is after applying alpha. For example, a normal priority is in [0,2], after alpha, the range changes. If the user originally wants to specify initial_priority as 2, then now he needs to specify a value  2**alpha

alf/algorithms/algorithm.py
hnyu: Imagine that we have an extreme case where 99 out of 100 exps are 0. The replay buffer length N is still 100. A uniform replay would still treat the 100 values equally likely. So the importance weight should still use 1/100 (expected probability) instead of 1/1 to divide the actually used probability (computed by priority).

emailweixu: We want E(1/importance_weight) to be 1.

hnyu: \sum p(i) * ((1/N) / p(i)) * TD(i) = \sum 1/N * TD(i) is the unbiased quantity (uniform replay). Because we are sampling according to p(i), then the ratio should be (1/N) / p(i). And we have -beta on top of this, so the returned importance weights should be N p(i)? where N is the replay buffer size. Right now, we have NNZ p(i) for this value.

emailweixu: Things change if you have a 0.

emailweixu: It's no longer unbiased. The reason of having 0 is that we really don't want that sample, hence should not include it into N.

hnyu: OK, if having 0 means absolutely discarding that value then I agree NNZ is better.  I think it's depending on how we define the meaning of having a 0 priority. In the example I gave, using 100 instead of 1 as N would essentially slow down the training. But if 0 means completely discarding, then repeatedly training on that single example with 100% learning rate makes sense.

alf/algorithms/config.py
```diff
@@ -135,7 +135,10 @@ def __init__(self,
                 ``importance_weight**(-priority_replay_beta)``, where ``importance_weight``
                 is from the BatchInfo returned by ``ReplayBuffer.get_batch()``.
                 This is only useful if ``prioritized_sampling`` is enabled for
-                ``ReplayBuffer``.
+                ``ReplayBuffer``. Note that the effect of ``ReplayBuffer.initial_priority``
+                may change with different values of ``priority_replay_beta``.
```
hnyu: should this comment be updated for alpha instead of beta? because what's stored in the buffer is p^alpha (line 1371)

hnyu (approved):

alf/algorithms/algorithm.py
emailweixu: Effectively, 0 means discarding, because it will never be sampled.

PR #593, opened 2020-06-04
fix priority when special step is included

hnyu:
If T>1 and there is a special steptype in it (LAST_STEP or FIRST_STEP), the valid_mask will mask out the priority before summing over T. This will bias towards sampling T that doesn't contain the special steptype, which is not right.

### DIFF
--- a/alf/algorithms/sarsa_algorithm.py
+++ b/alf/algorithms/sarsa_algorithm.py
@@ -440,11 +440,13 @@ class SarsaAlgorithm(OnPolicyAlgorithm):
 
         critic_loss = math_ops.add_n(critic_losses)
 
-        not_first_step = (experience.step_type != StepType.FIRST)
-        critic_loss = critic_loss * not_first_step.to(torch.float32)
+        not_first_step = (experience.step_type != StepType.FIRST).to(
+            torch.float32)
+        critic_loss = critic_loss * not_first_step
         if (experience.batch_info != ()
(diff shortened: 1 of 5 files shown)


alf/algorithms/sarsa_algorithm.py
```diff
@@ -440,11 +440,13 @@ def calc_loss(self, experience, info: SarsaInfo):
...
-        not_first_step = (experience.step_type != StepType.FIRST)
-        critic_loss = critic_loss * not_first_step.to(torch.float32)
+        not_first_step = (experience.step_type != StepType.FIRST).to(
+            torch.float32)
+        critic_loss = critic_loss * not_first_step
         if (experience.batch_info != ()
                 and experience.batch_info.importance_weights != ()):
-            priority = critic_loss.sum(dim=0).sqrt()
+            valid_n = torch.clamp(not_first_step.sum(dim=0), min=1.0)
+            priority = (critic_loss.sum(dim=0) / valid_n).sqrt()
```
emailweixu, line 449: There is still some issue. For mini_batch_length=1, priority will always be 0 if step_type is FIRST_STEP. Hence this kind of samples will not be sampled anymore. Though this does not affect the training of critic, it may affect the training of actor. Since the actor loss will not contain samples at the FIRST_STEP anymore. I am not sure what is best way to fix this.

emailweixu (approved):

PR #594, opened 2020-06-04
add NestMultiply

hnyu:
added a NestCombiner for preprocessing. Will later implement a more general Multiply layer in layers.py

### DIFF
--- a/alf/nest/utils.py
+++ b/alf/nest/utils.py
@@ -15,12 +15,12 @@
 
 import abc
 import gin
-import functools
 
 import torch
 import torch.nn as nn
 
 import alf
+from alf.utils import math_ops
 from . import nest
 from alf.tensor_specs import TensorSpec
 
@@ -130,11 +130,7 @@ class NestMultiply(NestCombiner):
         self._activation = activation
 
     def _combine_flat(self, tensors):
-        # Reduce is faster than ``torch.prod(torch.stack(tensors, dim=0), dim=0)``,
-        # and it doesn't require activation to be in-place as what ``torch.prod``
-        # does.
-        ret = functools.reduce(lambda x, y: x * y, tensors,
-                               torch.ones_like(tensors[0]))
+        ret = math_ops.mul_n(tensors)
         return self._activation(ret)
 
 
--- a/alf/nest/utils.py
+++ b/alf/nest/utils.py
@@ -20,7 +20,6 @@ import torch
 import torch.nn as nn
 
 import alf
-from alf.utils import math_ops
 from . import nest
 from alf.tensor_specs import TensorSpec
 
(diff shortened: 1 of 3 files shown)


alf/nest/utils.py
```diff
@@ -111,6 +112,32 @@ def _combine_flat(self, tensors):
...
+        super(NestMultiply, self).__init__(name)
+        if activation is None:
+            activation = lambda x: x
+        self._activation = activation
+
+    def _combine_flat(self, tensors):
+        # Reduce is faster than ``torch.prod(torch.stack(tensors, dim=0), dim=0)``,
+        # and it doesn't require activation to be in-place as what ``torch.prod``
+        # does.
+        ret = functools.reduce(lambda x, y: x * y, tensors,
```
emailweixu: we have similar math_ops.add_n(). Perhaps we can implement a math_ops.mul_n()

alf/utils/math_ops.py
```diff
@@ -81,49 +82,54 @@ def swish(x):
...
-        the elementwise sum of all the tensors in `inputs`
+        Tensor: the element-wise sum of all the tensors in ``inputs``.
     """
-    assert isinstance(inputs, list)
-    ret = inputs[0]
-    inputs = inputs[1:]
-    for x in inputs:
-        ret = torch.add(ret, x)
-    return ret
+    return functools.reduce(torch.add, inputs, torch.zeros_like(inputs[0]))
```
emailweixu: Are you sure functools.reduce is more efficient?

emailweixu: > Are you sure functools.reduce is more efficient?
Removing initializer might be better:
```
functools.reduce(torch.add, inputs)
```

alf/utils/math_ops.py
```diff
@@ -81,49 +82,54 @@ def swish(x):
...
+    The current solution uses ``functools.reduce``, which is faster than
+    the alternative ``torch.prod(torch.stack(inputs, dim=0), dim=0)``.
+
+    Args:
+        inputs (iterable[Tensor]): an iterable of tensors. It requires that
+            all tensor shapes can be broadcast to the same shape.
+    Returns:
+        Tensor: the element-wise multiplication of all the tensors in ``inputs``.
+    """
+    return functools.reduce(torch.mul, inputs, torch.ones_like(inputs[0]))
```
emailweixu: functools.reduce(torch.mul, inputs)

alf/utils/math_ops.py
hnyu: > Are you sure functools.reduce is more efficient?

Most of the time reduce and for-loop should be similar. It's more concise using reduce.

alf/utils/math_ops.py
emailweixu: No need to use the initilizer. `functools.reduce(torch.mul, inputs)` is better.

hnyu: already updated

emailweixu (approved):

PR #598, opened 2020-06-10
cnest

### LINKED ISSUE #597: Faster nest related functions
c++ implementation with pybind11

### LINKED ISSUE #607: Speed up ALF

hnyu:
My plan is to just implement several key functions (flatten, assert_same_structure, etc.) that are performance critical and C++-preferring, and reuse other python helper functions. 

I haven't written C++ for a long time, and probably some places in the code are not optimized. 

Tested with `sarsa_sac_pendulum.gin` in the branch `sac_profile`.

image

Total time decreased from 5s to 4.4s.


alf/cnest/cnest_test.py
```diff
@@ -0,0 +1,100 @@
...
+
+        t0 = time.time()
+        for i in range(10):
+            alf.nest.extract_fields_from_nest(x)
+        print("nest extract_fields_from_nest: %s" % (time.time() - t0))
+
+        t0 = time.time()
+        for i in range(10):
+            cnest.extract_fields_from_nest(x)
+        print("cnest extract_fields_from_nest: %s" % (time.time() - t0))
```
le-horizon: It may be more prudent to reverse the two for loops, and time again.  If the gains are still there, then it's more likely real.  Otherwise, the timing difference may be due to how python optimizes its execution.

alf/cnest/nest.cpp
```diff
@@ -0,0 +1,180 @@
...
+  }
+  return flat;
+}
+
+py::list Flatten(py::object nest) {
+  auto flat = _Flatten(nest);
+  py::list ret;
+  for (auto x : flat) {
+    ret.append(x);
+  }
```
le-horizon: does it make sense to pass in e.g. a reference of ret to _Flatten directly, instead of having this additional for loop here?

le-horizon (commented):
Nice change @hnyu !
Looks mostly good to me.  Just two suggestions if they make sense.

alf/cnest/nest.cpp
hnyu: I guess I still need this for loop to fill in py::list either way. py::list doesn't provide a way for insert (extend) as what vector does.

alf/cnest/cnest_test.py
hnyu: Updated

emailweixu:
For map_structure(func, ...), if something wrong with func, do we still have the stack context to debug func?

hnyu:
> For map_structure(func, ...), if something wrong with func, do we still have the stack context to debug func?

Yeah because I re-raise the exception after catching it. 

image

emailweixu:
But suppose we run the code using pdb, can we still examine the variables in func to see what's wrong after the exception?

hnyu:
> But suppose we run the code using pdb, can we still examine the variables in func to see what's wrong after the exception?

Just tested with both vscode debug and pdb, it will let me examine variables in func after exception.

setup.py
```diff
@@ -19,6 +19,8 @@
     version='0.0.1',
     install_requires=[
         'atari_py == 0.1.7',
+        'cpplint',
```
emailweixu, line 25: Need to handle the installation of cnest too.

alf/nest/cnest/nest.cpp
```diff
@@ -0,0 +1,494 @@
...
+  }
+  return ret;
+}
+
+py::object MapStructure(const py::function& func, const py::args& nests) {
+  if (py::len(nests) == 0) {
+    throw std::runtime_error("There should be at least one input nest!");
+  }
+  for (unsigned int i = 1; i < nests.size(); i++) {
+    AssertSameStructure(py::cast<py::object>(nests[py::int_(0)]),
```
emailweixu: Will it be more efficient if these checks are performed inside _MapStructure?

alf/nest/cnest/nest.cpp
```diff
@@ -0,0 +1,494 @@
...
+    return func(*nests);
+  }
+  auto first_nest_type = first_nest.get_type();
+  py::object ret;
+  if (py::isinstance<py::list>(first_nest) || IsUnnamedtuple(first_nest)) {
+    py::list results;
+    for (unsigned int i = 0; i < py::len(first_nest); i++) {
+      py::list args;
+      for (const auto& nest : nests) {
+        args.append(py::cast<py::object>(nest[py::int_(i)]));
```
emailweixu: A new python object (int_(i)) will be created every time. Is the overhead negligible? At least int_(i) should be created outside this loop. If it is costly to create python object, it is better to use some caching mechanism: `GetPyInt(i)` to get the python int object with caching. Or use native python API: PyList_GetItem(nest.ptr(), i), which is the most efficient way to get the item.

alf/nest/cnest/nest.cpp
```diff
@@ -0,0 +1,494 @@
...
+
+  py::list ret;
+  for (const auto& field : fields) {
+    if (is_dict) {
+      auto key = py::str(field);
+      if (!nest_dict.contains(key)) {
+        throw std::runtime_error(py_format(
+            "Only support string keys in a dictionary!! Wrong key: {}", key));
+      }
+      ret.append(py::make_tuple(key, nest_dict[key]));
```
emailweixu: use c++ pair instead of python tuple.

alf/nest/cnest/nest.cpp
```diff
@@ -0,0 +1,494 @@
...
+    for (const auto& field : nest.attr("_fields")) {
+      fields.push_back(py::str(field));
+    }
+  }
+  std::sort(fields.begin(), fields.end());
+
+  py::list ret;
+  for (const auto& field : fields) {
+    if (is_dict) {
+      auto key = py::str(field);
```
emailweixu: Reuse `str(field)` created at line 86 if creating python object is costly,

alf/nest/cnest/nest.cpp
```diff
@@ -0,0 +1,494 @@
...
+      fields.push_back(py::str(item.first));
+    }
+  } else {
+    for (const auto& field : nest.attr("_fields")) {
+      fields.push_back(py::str(field));
+    }
+  }
+  std::sort(fields.begin(), fields.end());
+
+  py::list ret;
```
emailweixu: use c++ vector instead

emailweixu (commented):
In general, use c++ data structure whenever possible. Avoid creating python objects if possible.

alf/nest/cnest/nest.cpp
hnyu: If I use c++ structure, I think in the end when pybind converts this object to a python object, it still will convert c++ pair to python tuple (the memory will not be shared)? What's the benefit of not directly creating a python tuple here?

emailweixu: Because _MapStructure can access the  C++ structure more efficiently

hnyu: I tried using vector and c++ make_pair. This basically makes the whole typing much more difficult. Because the inputs have uncertain types of the values, I need to make template for the returned value and it seems a disaster in the end where the code cannot deduce the type automatically.

alf/nest/cnest/nest.cpp
hnyu: I tried using vector. This basically makes the whole typing much more difficult. Because the inputs have uncertain types of the values, I need to make template for the returned value and it seems a disaster in the end where the code cannot deduce the type automatically everywhere I use this function in the code.

However, it will be easy that I return a vector with type py::tuple, if not using c++ pair.

emailweixu: The the element type of the vector can be  `std::pair<string, PyObject*>`. It should be possible to obtain `PyObject*` from the pybind object.

emailweixu: Here is how py::list.append() implemented:
https://github.com/pybind/pybind11/blob/8c0cd94465fbc1dbc34217e5a614edc784f0913e/include/pybind11/pytypes.h#L1276

alf/nest/cnest/nest.cpp
```diff
@@ -31,6 +32,15 @@ std::string py_format(const std::string& format, Args... args) {
   return py::cast<py::str>(py_format(format, args...));
 }
 
+py::int_ GetPyInt(unsigned int i) {
+  // cache py::int_ to avoid creating new objects for repeating i
+  static std::map<unsigned int, py::int_> table;
```
emailweixu: vector<py::int_> is better for the specific purpose of this file because i won't be big.

alf/nest/cnest/nest.cpp
```diff
@@ -219,26 +222,29 @@ py::object _MapStructure(const py::function& func, const py::args& nests) {
...
       results.append(_MapStructure(func, args));
     }
     ret = first_nest_type(results);
   } else {
     py::dict results;
-    std::vector<py::list> fields_and_values;
-    for (const auto& nest : nests) {
-      fields_and_values.push_back(
-          ExtractFieldsFromNest(py::cast<py::object>(nest)));
+    std::vector<field_list> fields_and_values;
```
emailweixu, line 235: avoid allocation and copy:
```
fields_and_values.reserve(nests_.size())
```

alf/nest/cnest/nest.cpp
```diff
@@ -287,35 +281,36 @@ py::object _MapStructureUpTo(const py::object& shallow_nest,
...
       results.append(_MapStructureUpTo(
-          py::cast<py::object>(shallow_nest[py::int_(i)]), func, args));
+          py::cast<py::object>(shallow_nest[GetPyInt(i)]), func, args));
     }
     ret = (shallow_nest.get_type())(results);
   } else {
     py::dict results;
     auto shallow_fields_and_values = ExtractFieldsFromNest(shallow_nest);
-    std::vector<py::list> fields_and_values;
+    std::vector<field_list> fields_and_values;
```
emailweixu, line 297: fields_and_values.reserve(nests_.size())

alf/nest/cnest_test.py
```diff
@@ -0,0 +1,151 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Unittest for nest.cpp"""
+
+import time
+from collections import OrderedDict
+import sys
+sys.path.append("./cnest")
```
emailweixu: cnest is already installed. No need to do this.

alf/nest/nest.py
```diff
@@ -15,9 +15,176 @@
 
 from absl import logging
 
+import sys
+sys.path.append("./cnest")
```
emailweixu: remove

alf/nest/nest_test.py
```diff
@@ -15,97 +15,116 @@
 
 import torch
 
+from absl.testing import parameterized
+import sys
+sys.path.append("./cnest")
```
emailweixu: remove

emailweixu (approved):

PR #599, opened 2020-06-11
first upgrade clang-format to 9.0

hnyu:
otherwise the cnest PR CI won't pass. My desktop doesn't allow installing a version of 3.8 which is the current requirement.

### DIFF
--- a/.ci-cd/Dockerfile.cpu
+++ b/.ci-cd/Dockerfile.cpu
@@ -21,12 +21,13 @@ RUN apt install -y python3.6 python3-pip
 # install code style tools
 RUN pip3 install  pre-commit==1.17.0 \
     cpplint==1.4.4 \
+    clang-format==9.0 \
     pydocstyle==4.0.0 \
     pylint==2.3.1 \
     yapf==0.28.0
 
 
-RUN apt update && apt install -y  git clang-format
+RUN apt update && apt install -y git
 
 RUN apt install -y \
         libsm6  \
--- a/tools/codestyle/clang_format.hook
+++ b/tools/codestyle/clang_format.hook
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -e
 
-readonly VERSION="3.8"
+readonly VERSION="9.0"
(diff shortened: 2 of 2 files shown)


Haichao-Zhang:
If the docker image used by github ci is updated first, then is doesn't matter to separate this upgrade part or not?
In my understanding, merging this pr first won't make the CI pass on cnest. Instead, you might need to update the docker image first and then update this line
https://github.com/HorizonRobotics/alf/blob/c7c8832b9a55ed4f0faccd2eb54162ec4ebdb51b/.github/workflows/test.yml#L9

hnyu:
> If the docker image used by github ci is updated first, then is doesn't matter to separate this upgrade part or not?
> In my understanding, merging this pr first won't make the CI pass on cnest. Instead, you might need to update the docker image first and then update this line
> https://github.com/HorizonRobotics/alf/blob/c7c8832b9a55ed4f0faccd2eb54162ec4ebdb51b/.github/workflows/test.yml#L9

I'm actually confused by our current CI setup everytime. It seems that the Dockerfile doesn't need to exist if we are pulling an alf docker from the docker hub?

hnyu:
> If the docker image used by github ci is updated first, then is doesn't matter to separate this upgrade part or not?
> In my understanding, merging this pr first won't make the CI pass on cnest. Instead, you might need to update the docker image first and then update this line
> https://github.com/HorizonRobotics/alf/blob/c7c8832b9a55ed4f0faccd2eb54162ec4ebdb51b/.github/workflows/test.yml#L9

OK I'm confused every time by this. I thought somehow CI will automatically build an image given the Dockerfile. It's not quite smart or automatic I guess.

Haichao-Zhang:
> OK I'm confused every time by this. I thought somehow CI will automatically build an image given the Dockerfile. It's not quite smart or automatic I guess.

Yes it is a bit mis-leading . Unfortunately, it won't build the image. The Dockerfile here is mainly used for versioning/backing-up purpose.

PR #601, opened 2020-06-19
state repr model and RepresentationAgent

hnyu:
A state representation learning model used by PlaNet and Dreamer. Will report results of using it to train Breakout later. The model is basically a VAE. But because of some detail difference, I didn't use VAE as a component. 

I also created a new Agent class RepresentationAgent to use this representation model, to separate it from our master Agent for now.

### DIFF
--- a/.ci-cd/build.sh
+++ b/.ci-cd/build.sh
@@ -55,6 +55,7 @@ function test() {
         alf.algorithms.merlin_algorithm_test \
         alf.algorithms.mi_estimator_test \
         alf.algorithms.ppo_algorithm_test \
+        alf.algorithms.repr_agent_test \
         alf.algorithms.rl_algorithm_test \
         alf.algorithms.sarsa_algorithm_test \
         alf.algorithms.sac_algorithm_test \
--- a/alf/algorithms/agent.py
+++ b/alf/algorithms/agent.py
@@ -90,14 +90,14 @@ class Agent(OnPolicyAlgorithm):
                 needs to contain ``action_distribution``.
             entropy_target_cls (type): If provided, will be used to dynamically
                 adjust entropy regularization.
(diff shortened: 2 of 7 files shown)


emailweixu:
Why not just add representation learning module to current Agent?

hnyu:
> Why not just add representation learning module to current Agent?

Haven't figured out how to use the learned representation in the current Agent effectively because usually our policy/value networks directly accept raw observations as inputs. The learned latent representation is non-stationary and the benefit to the downstream policy/value networks is unclear at this point. I think we can add it to Agent after verifying its benefit.

It also makes sense to me that besides the master Agent, we have several other smaller-size agents that serve different purposes.

emailweixu:
But this is exactly what you are doing in RepresentationAgent.

> Haven't figured out how to use the learned representation in the current Agent effectively because usually our policy/value networks directly accept raw observations as inputs. The learned latent representation is non-stationary and the benefit to the downstream policy/value networks is unclear at this point. I think we can add it to Agent after verifying its benefit.
> 
> It also makes sense to me that besides the master Agent, we have several other smaller-size agents that serve different purposes.

hnyu:
> But this is exactly what you are doing in RepresentationAgent.
> 
> > Haven't figured out how to use the learned representation in the current Agent effectively because usually our policy/value networks directly accept raw observations as inputs. The learned latent representation is non-stationary and the benefit to the downstream policy/value networks is unclear at this point. I think we can add it to Agent after verifying its benefit.
> > It also makes sense to me that besides the master Agent, we have several other smaller-size agents that serve different purposes.

I was thinking that maybe we should first verify the effectiveness of a representation model and then integrated into our master Agent class, because that change will be more than here.

Right now I'm still debugging the model.

PR #603, opened 2020-06-22
fix infinite recursion bug in preprocessor

hnyu:
```
scope1/A.x=y
scope1/B.z=@scope1/A()
```
Usually we group classes by a common scope name in a gin file. This might be problematic if inside `A`'s init code we also create an instance of `B`. By default that instance will inherit `A`'s scope. So there will be an infinite recursion. 

A concrete example:
```
scope/EmbeddingPreprocessor.input_tensor_spec=%spec
scope/EncodingNetwork.input_preprocessors=@scope/EmbeddingPreprocessor()
```

Generally if A creates B and B also creates A (whether in a gin file or in the code), we need to use different name scopes for the two classes. Or when writing the code, we have to manually clear the scope whenever it's needed. The scope management is quite a dangerous thing in gin, especially there are multiple instances of the same class created with different configurations.

The safest way is to use @ for directly evaluating an instance in a gin file without specifying scopes, or use a completely new scope name for every instance created.

### DIFF
--- a/alf/layers.py
+++ b/alf/layers.py
@@ -822,6 +822,7 @@ class ParallelConvTranspose2D(nn.Module):
         return self._bias
 
 
+@gin.configurable
 class Reshape(nn.Module):
     def __init__(self, shape):
         """A layer for reshape the tensor.
--- a/alf/networks/preprocessors.py
+++ b/alf/networks/preprocessors.py
@@ -74,14 +74,25 @@ class EmbeddingPreprocessor(Network):
             # use nn.Embedding to support a large dictionary
             self._embedding_net = nn.Embedding(N, embedding_dim)
         else:
-            # Only use an MLP for embedding a continuous input
-            self._embedding_net = alf.networks.EncodingNetwork(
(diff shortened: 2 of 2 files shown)


emailweixu:
It doesn't seem right to change the code to accommodate the misuse of gin config. This can potentially happen anywhere in the code. It would be better to fix gin to detect the loop instead.

hnyu:
It might not just be about an infinite recursion. For example, 

```
scope/EmbeddingPreprocessor.input_tensor_spec=%spec
scope/EncodingNetwork.activation=@torch.tanh
```
If inside EmbeddingPreprocessor the activation of the created EncodingNetwork is not provided (meant to use the default torch.relu_), then it will be overwritten by the gin configured torch.tanh. This case is even more dangerous because gin cannot detect it and no error will be thrown. So I think it's user code's responsibility to pay attention to this.

hnyu:
Closed regarding https://github.com/HorizonRobotics/gin-config/pull/4

PR #605, opened 2020-06-25
Options to guarantee to get recent data from ReplayBuffer

emailweixu:
Also default to sample without replacement.


alf/experience_replayers/replay_buffer.py
```diff
@@ -286,49 +331,65 @@ def get_batch(self, batch_size, batch_length):
             result = alf.nest.map_structure(lambda b: b[(out_env_ids, idx)],
                                             self._buffer)
 
-            if self._prioritized_sampling:
-                indices = self._env_id_idx_to_index(env_ids,
-                                                    self.circular(start_pos))
-                avg_weight = self._sum_tree.nnz / self._sum_tree.summary()
-                info = info._replace(
-                    importance_weights=self._sum_tree[indices] * avg_weight)
-
```
le-horizon, line 295: Will this change the existing behavior of prioritized sample?

alf/experience_replayers/replay_buffer.py
```diff
@@ -286,49 +331,65 @@ def get_batch(self, batch_size, batch_length):
...
+            remaining_eids = torch.randperm(self._num_envs)[:remaining]
+            env_ids = torch.cat([env_ids, remaining_eids], dim=0)
 
         r = torch.rand(*env_ids.shape)
-
-        num_positions = self._current_size - batch_length + 1
-        num_positions = num_positions[env_ids]
-        pos = (r * num_positions).to(torch.int64)
-        pos += (self._current_pos - self._current_size)[env_ids]
+        if not self._with_replacement:
```
le-horizon, line 373: Is it intended to default with_replacement to False?
it will change the existing behavior of uniform_sample?

alf/experience_replayers/replay_buffer.py
```diff
@@ -65,6 +72,12 @@ def __init__(self,
             initial_priority (float): initial priority used for new experiences.
                 The actual initial priority used for new experience is the maximum
                 of this value and the current maximum priority of all experiences.
+            recent_data_steps (int): the most recent so many steps of data is
+                considered as recent data for ``get_batch()``.
+            recent_data_ratio (float): ``recent_data_ratio * batch_size`` samples
+                in the batch are sampled from recent data for ``get_batch()``.
+            with_replacement (bool): If True, sample without replacement whenever
```
hnyu: variable name inconsistent with the description. with or without?

alf/experience_replayers/replay_buffer.py
```diff
@@ -65,6 +72,12 @@ def __init__(self,
             initial_priority (float): initial priority used for new experiences.
                 The actual initial priority used for new experience is the maximum
                 of this value and the current maximum priority of all experiences.
+            recent_data_steps (int): the most recent so many steps of data is
+                considered as recent data for ``get_batch()``.
```
hnyu: Need to clarify that this steps is per env

hnyu (changes requested):

alf/experience_replayers/replay_buffer.py
```diff
@@ -270,10 +298,27 @@ def get_batch(self, batch_size, batch_length):
                         non-zero priorities in the buffer.
         """
         with alf.device(self._device):
+            recent_batch_size = 0
+            if self._recent_data_ratio > 0:
+                avg_size = self.total_size / float(
+                    self._num_envs) - batch_length + 1
+                if (avg_size * self._recent_data_ratio >
+                        self._recent_data_steps):
```
le-horizon, line 308: is this if condition for the case where there is enough recent data?
Shouldn't this be
```
if (batch_size / num_envs - batch_length + 1) * recent_data_ratio < recent_data_steps and (
    avg_size > recent_data_steps)
```

alf/experience_replayers/replay_buffer.py
```diff
@@ -270,10 +298,27 @@ def get_batch(self, batch_size, batch_length):
                         non-zero priorities in the buffer.
         """
         with alf.device(self._device):
+            recent_batch_size = 0
+            if self._recent_data_ratio > 0:
```
le-horizon, line 304: add assert recent_data_ratio * batch_size / num_env < recent_data_steps.
Otherwise, there isn't enough recent data, and the configuration is wrong.

alf/experience_replayers/replay_buffer.py
```diff
@@ -270,10 +298,27 @@ def get_batch(self, batch_size, batch_length):
...
+            normal_batch_size = batch_size - recent_batch_size
             if self._prioritized_sampling:
-                info = self._prioritized_sample(batch_size, batch_length)
+                info = self._prioritized_sample(normal_batch_size,
+                                                batch_length)
             else:
-                info = self._uniform_sample(batch_size, batch_length)
+                info = self._uniform_sample(normal_batch_size, batch_length)
+
+            if recent_batch_size > 0:
```
le-horizon, line 323: comment that normal_batch could contain recent data and duplicate with recent_batch, even in the case of w/o replacement sampling?

alf/experience_replayers/replay_buffer.py
```diff
@@ -286,49 +331,65 @@ def get_batch(self, batch_size, batch_length):
...
+
     def _uniform_sample(self, batch_size, batch_length):
         min_size = self._current_size.min()
         assert min_size >= batch_length, (
             "Not all environments have enough data. The smallest data "
             "size is: %s Try storing more data before calling get_batch" %
             min_size)
+        return self._sample(batch_size, batch_length)
 
+    def _sample(self, batch_size, batch_length, recent_data_steps=None):
```
le-horizon: rename this arg to something like sample_from_recent_n_steps, so it's more readable.

alf/experience_replayers/replay_buffer.py
```diff
@@ -286,49 +331,65 @@ def get_batch(self, batch_size, batch_length):
...
+            # that environment.
+            i = torch.arange(batch_size_per_env).unsqueeze(-1).expand(
+                -1, self._num_envs).reshape(-1)
+            if remaining > 0:
+                num_samples_per_env[remaining_eids] += 1
+                remaining_i = batch_size_per_env * torch.ones(
+                    (remaining, ), dtype=torch.int64)
+                i = torch.cat([i, remaining_i])
+
+            r = (i + r) / num_samples_per_env[env_ids]
```
le-horizon, line 388: comment that this isn't really uniform sampling at the global level, but uniform sampling within each evenly spaced region?

alf/experience_replayers/replay_buffer_test.py
```diff
@@ -407,6 +409,30 @@ def test_replay_buffer(self, allow_multiprocess):
...
+        replay_buffer.add_batch(get_batch([0, 1, 2, 3], self.dim, t=1, x=1.0))
+        batch, info = replay_buffer.get_batch(8, 1)
+        self.assertEqual(info.env_ids, torch.tensor([0, 1, 2, 3] * 2))
+
+        for t in range(2, 16):
+            replay_buffer.add_batch(
+                get_batch([0, 1, 2, 3], self.dim, t=t, x=t))
+        batch, info = replay_buffer.get_batch(32, 1)
+        self.assertEqual(info.env_ids[16:], torch.tensor([0, 1, 2, 3] * 4))
+
```
le-horizon, line 440: add a test that recent data is really being sampled out.  Say when sampling 32 with 4 env, all 4 recent steps data are being sampled out.

le-horizon (commented):
Good to have the change, Wei.  A few suggestions if they make sense.

Thanks,
Le

alf/experience_replayers/replay_buffer.py
emailweixu: The behavior is same. It's moved to _prioritzed_sample() and is a little bit more efficient because no need to recompute indices.

alf/experience_replayers/replay_buffer.py
emailweixu: Fixed.

alf/experience_replayers/replay_buffer.py
emailweixu: Fixed

alf/experience_replayers/replay_buffer.py
emailweixu: If not enough recent recent_data, just ignore this part as line 317 does.

alf/experience_replayers/replay_buffer.py
emailweixu: If this condition is False, regular sampling without considering
recent data will get enough samples from recent data. So
we don't need to have a separate step just for sampling from
the recent data.

alf/experience_replayers/replay_buffer.py
emailweixu: Commented.

alf/experience_replayers/replay_buffer.py
emailweixu: Yes. It will change the existing behavior of uniform sample. But the new behavior should be better as the variance is smaller. And in most realistic scenario, the difference should be very small.

alf/experience_replayers/replay_buffer.py
emailweixu: This is uniform sampling because all the regions have same length. Every sample has equal probability of being selected.

alf/experience_replayers/replay_buffer.py
emailweixu: changed

alf/experience_replayers/replay_buffer_test.py
emailweixu: added

alf/experience_replayers/replay_buffer.py
le-horizon: Fair enough.  It is uniform, just not independent.

le-horizon (dismissed):
LG.

hnyu (dismissed):

emailweixu:
@hnyu @le-horizon fixed the unittest. Can you take a look?

hnyu (approved):

le-horizon (approved):
LG, looking forward to trying it on ddpg and HER.

PR #609, opened 2020-07-01
Fix carla speed calculation bug and add success info

emailweixu:
Also resize image for rendering.

### DIFF
--- a/alf/environments/suite_carla.py
+++ b/alf/environments/suite_carla.py
@@ -75,6 +75,10 @@ def is_available():
     return carla is not None
 
 
+MINIMUM_RENDER_WIDTH = 640
+MINIMUM_RENDER_HEIGHT = 240
+
+
 class SensorBase(abc.ABC):
     """Base class for sersors."""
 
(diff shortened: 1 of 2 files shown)


alf/environments/suite_carla.py
```diff
@@ -75,6 +75,10 @@ def is_available():
     return carla is not None
 
 
+MINIMUM_RENDER_WIDTH = 640
+MINIMUM_RENDER_HEIGHT = 240
```
hnyu, line 79: why changing from 480 to 240?

alf/environments/suite_carla.py
```diff
@@ -1023,10 +1035,13 @@ def get_current_time_step(self, current_frame):
                                                 self._actor.get_location())
         v = self._actor.get_velocity()
         obs['speed'] = np.float32(
-            np.float32(v.x * v.x + v.y * v.y + v.z * v.z))
+            np.sqrt(np.float32(v.x * v.x + v.y * v.y + v.z * v.z)))
```
hnyu, line 1038: Is this a serious issue? It seems that in principle before and after are equivalent.

alf/environments/suite_carla.py
emailweixu: The height of text panel is 240. There is no need beyond that.

alf/environments/suite_carla.py
emailweixu: The range would huge without the sqrt. I thought it shouldn't affect the training much. But somehow the training is better after fixing this. Although one training is not enough to draw the conclusion.

hnyu (approved):

PR #606, opened 2020-06-27
ALF performance optimization

emailweixu:
The following list the speed difference after each change. Overall speedup is about 14%. Tiny things add up. 

```
  ncalls  tottime  percall  cumtime  percall filename:lineno(function)
common.update(): use lerp_
    1991    0.848    0.000    1.482    0.001 /home/weixu/code/alf/alf/utils/common.py:175(update)
=>  1991    0.043    0.000    0.701    0.000 /home/weixu/code/alf/alf/utils/common.py:175(update)

utils.get_outer_rank => remote list()
    83766    0.168    0.000    0.200    0.000 /home/weixu/code/alf/alf/nest/utils.py:172(_get_outer_rank)
    83766    0.137    0.000    0.168    0.000 /home/weixu/code/alf/alf/nest/utils.py:172(_get_outer_rank)
remove flatten
    25901    0.027    0.000    0.290    0.000 /home/weixu/code/alf/alf/nest/utils.py:151(get_outer_rank)
    25901    0.022    0.000    0.256    0.000 /home/weixu/code/alf/alf/nest/utils.py:151(get_outer_rank)

AdamTF.step(): lerp_
    5973    0.247    0.000    2.538    0.000 /home/weixu/code/alf/alf/optimizers/adam_tf.py:83(step)
=>  5973    0.237    0.000    2.300    0.000 /home/weixu/code/alf/alf/optimizers/adam_tf.py:83(step)

RingBuffer._enqueue(): precompute circular(current_pos)
    2000    0.239    0.000    1.740    0.001 /home/weixu/code/alf/alf/utils/data_buffer.py:210(_enqueue)
=>  2000    0.246    0.000    1.469    0.001 /home/weixu/code/alf/alf/utils/data_buffer.py:210(_enqueue)

RingBuffer._enqueue(): max => clamp
    2000    0.246    0.000    1.469    0.001 /home/weixu/code/alf/alf/utils/data_buffer.py:210(_enqueue)
    2000    0.239    0.000    1.424    0.001 /home/weixu/code/alf/alf/utils/data_buffer.py:210(_enqueue)

RingBuffer._enqueue(): flattened buffer
    2000    0.239    0.000    1.424    0.001 /home/weixu/code/alf/alf/utils/data_buffer.py:210(_enqueue)
=>  2000    0.278    0.000    1.410    0.001 /home/weixu/code/alf/alf/utils/data_buffer.py:212(_enqueue)

unique under assert and -O
[...]

PR #613, opened 2020-07-02
Fix suite_dmlab_test.py

emailweixu:
TODO: add dmlab to docker image for CI
Also fix a bug for "one_time"replay caused by PR #606

### DIFF
--- a/alf/algorithms/merlin_algorithm.py
+++ b/alf/algorithms/merlin_algorithm.py
@@ -516,7 +516,7 @@ class ResnetEncodingNetwork(alf.networks.Network):
         enc_layers.extend([
             nn.Flatten(),
             alf.layers.FC(
-                input_size=np.prod(shape),
+                input_size=int(np.prod(shape)),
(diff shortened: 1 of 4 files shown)


hnyu:
If I ran test_thread_env(), there is a segfault error in the end. Do you also have this issue?

emailweixu:
> If I ran test_thread_env(), there is a segfault error in the end. Do you also have this issue?

I don't have this issue. All 8 tests can pass.

hnyu:
In train_play_test.py, the test case for dmlab uses the wrong gin file:

```
    @unittest.skip(SKIP_TODO_MESSAGE)
    def test_merlin_dmlab_collect_good_objects(self):
        self._test(
            gin_file='icm_super_mario_intrinsic_only.gin',
            skip_checker=self._skip_if_dmlab_unavailable,
            extra_train_params=ON_POLICY_TRAIN_PARAMS)
```

hnyu (approved):

PR #619, opened 2020-07-06
Wasserstein implicit vi

runjerry:
- Hypernetwork algorithm for supervised training using particle-based VI.
- Related networks and layers that are needed for the hypernetwork.
- Refactor trainer.


alf/bin/supervised_train.py
```diff
@@ -0,0 +1,220 @@
...
+    random_seed = common.set_random_seed(config.random_seed)
+    assert config.epochs > 0, \
+        "Must provide #epochs for training!"
+
+    trainset, testset = create_dataset()
+    input_tensor_spec = TensorSpec(shape=testset.dataset[0][0].shape)
+    output_dim = len(testset.dataset.classes)
+
+    # print algorithm related gin parameters
+    _, inopt_configs = common.get_gin_confg_strs()
```
emailweixu: get_gin_confg_strs => get_gin_config_strs
And since gin config is written to tensorboard at line 187, why do you need this?

alf/bin/supervised_train.py
```diff
@@ -0,0 +1,220 @@
...
+        # We always write summary in the initial `summary_interval` steps
+        # because there might be important changes at the beginning.
+        return (alf.summary.is_summary_enabled()
+                and (epoch_num < config.summary_interval
+                     or epoch_num % config.summary_interval == 0))
+
+    with alf.summary.push_summary_writer(summary_writer):
+        with alf.summary.record_if(_cond):
+            epoch_num = 0
+            print("==> Begin Training")
```
emailweixu: print => logging.info

alf/bin/supervised_train.py
```diff
@@ -0,0 +1,220 @@
...
+            while True:
+                print("-" * 68)
```
emailweixu: logging.info here and other places.

alf/bin/supervised_train.py
```diff
@@ -0,0 +1,220 @@
...
+
+    trainset, testset = getattr(dataset_loader,
+                                'load_{}'.format(dataset_name))(
+                                    train_bs=train_batch_size,
+                                    test_bs=test_batch_size)
+
+    return trainset, testset
+
+
+def train(config: Config):
```
emailweixu: I suggest reuse current policy_trainer.Trainer since a lot of code here is copied from Trainer. Take the part related to RL from Trainer to a subclass RLTrainer.

alf/algorithms/hypernetwork_algorithm.py
```diff
@@ -0,0 +1,479 @@
...
+        net = EncodingNetwork(
+            noise_spec,
+            fc_layer_params=hidden_layers,
+            use_fc_bn=use_fc_bn,
+            last_layer_size=gen_output_dim,
+            last_activation=math_ops.identity,
+            name="Generator")
+
+        if print_network:
+            print("Generated network")
```
emailweixu: logging.info

alf/algorithms/hypernetwork_algorithm.py
```diff
@@ -0,0 +1,479 @@
...
+
+    def predict(self, inputs, particles=None):
+        """Predict ensemble outputs for inputs using the hypernetwork model."""
+
+        params = self.sample_parameters(particles=particles)
+        self._param_net.set_parameters(params)
+        outputs, _ = self._param_net(inputs)
+        return outputs
+
+    def train_iter(self, particles=None, state=None):
```
emailweixu: comment particles and state

alf/algorithms/hypernetwork_algorithm.py
```diff
@@ -0,0 +1,479 @@
...
+        self._train_loader = train_loader
+        self._test_loader = test_loader
+
+    def set_particles(self, particles):
+        """Set the number of particles to sample through one forward
+        pass of the hypernetwork. """
+        self._particles = particles
+
+    @property
+    def particles(self):
```
emailweixu: comment

alf/algorithms/hypernetwork_algorithm.py
```diff
@@ -0,0 +1,479 @@
...
+        return self._particles
+
+    def sample_parameters(self, noise=None, particles=None, training=True):
```
emailweixu: comment arguments and return value

alf/algorithms/hypernetwork_algorithm.py
```diff
@@ -0,0 +1,479 @@
...
+                 input_tensor_spec,
+                 conv_layer_params=None,
+                 fc_layer_params=None,
+                 activation=torch.relu_,
+                 last_layer_size=None,
+                 last_activation=None,
+                 noise_dim=32,
+                 hidden_layers=(64, 64),
+                 use_fc_bn=False,
+                 particles=32,
```
emailweixu: particles => num_particles so there is less confusion when reading the code.

alf/algorithms/hypernetwork_algorithm.py
```diff
@@ -0,0 +1,479 @@
...
+        assert self._train_loader is not None, "Must set data_loader first."
+        with record_time("time/train"):
+            loss = 0.
+            if self._loss_type == 'classification':
+                avg_acc = []
+            for batch_idx, (data, target) in enumerate(self._train_loader):
+                data = data.to(alf.get_default_device())
+                target = target.to(alf.get_default_device())
+                params = None
+                if not self._regenerate_for_each_batch:
```
emailweixu: Why do you still generate parameter for every batch even self._regenerate_for_each_batch?

alf/algorithms/hypernetwork_algorithm.py
```diff
@@ -0,0 +1,479 @@
...
+                                           particles=particles,
+                                           state=state)
+                self.update_with_gradient(alg_step.info)
+                loss += alg_step.info.extra.loss
+                if self._loss_type == 'classification':
+                    avg_acc.append(alg_step.info.extra.extra)
+        if self._loss_type == 'classification':
+            acc = torch.as_tensor(avg_acc)
+            logging.info("Avg acc: {}".format(acc.mean() * 100))
+        logging.info("Cum loss: {}".format(loss))
```
emailweixu: suggest to use tensorboard for logging the losses so that we can view the curves. One function might be useful is summary_utils.summarize_loss()

alf/algorithms/hypernetwork_algorithm.py
```diff
@@ -0,0 +1,479 @@
...
+        target = target.unsqueeze(1).expand(*target.shape[:1], particles,
+                                            *target.shape[1:])
+        loss, extra = self._compute_loss(output, target, loss_func)
+
+        loss_grad = torch.autograd.grad(loss.sum(), params_j)[0]  # [Nj, W]
+        q_i = params_i + torch.rand_like(params_i) * 1e-8
+        q_j = params_j + torch.rand_like(params_j) * 1e-8
+        kappa, kappa_grad = self._rbf_func(q_j, q_i)  # [Nj, Ni], [Nj, Ni, W]
+        Nj = kappa.shape[0]
+        kernel_logp = torch.einsum('ji, jw->iw', kappa, loss_grad) / Nj
```
emailweixu: matmul(kapper.t(), loss_grad) might be faster than einsum()

alf/algorithms/hypernetwork_algorithm.py
```diff
@@ -0,0 +1,479 @@
...
+        params_i, params_j = torch.split(params, particles, dim=0)
+        self._param_net.set_parameters(params_j)
+        output, _ = self._param_net(data)  # [B, N/2, D]
+        target = target.unsqueeze(1).expand(*target.shape[:1], particles,
+                                            *target.shape[1:])
+        loss, extra = self._compute_loss(output, target, loss_func)
+
+        loss_grad = torch.autograd.grad(loss.sum(), params_j)[0]  # [Nj, W]
+        q_i = params_i + torch.rand_like(params_i) * 1e-8
+        q_j = params_j + torch.rand_like(params_j) * 1e-8
```
emailweixu: what's the reason for the above two lines? Should 1e-8 be a configurable parameter?

alf/algorithms/hypernetwork_algorithm.py
```diff
@@ -0,0 +1,479 @@
...
+from alf.networks import EncodingNetwork
+from alf.tensor_specs import TensorSpec
+from alf.utils import common, math_ops
+from alf.utils.summary_utils import record_time
+
+HyperNetworkLossInfo = namedtuple("HyperNetworkLossInfo", ["loss", "extra"])
+
+
+@gin.configurable
+class HyperNetwork(Generator):
```
emailweixu: I feel HyperNetwork should not be a subclass of ``Generator``. The output of ``Generator`` is the result directly from the generator. While the results of ``HyperNetwork`` is the results of the generated network applied on inputs.
I feel it might be better if the HyperNetwork is implemented by using a generator and applying proper likelihood function (which is neural network with some loss in this case).

alf/layers.py
```diff
@@ -262,6 +264,7 @@ def __init__(self,
         # get the argument list with vals
         self._kwargs = copy.deepcopy(locals())
         self._kwargs.pop('self')
+        self._kwargs.pop('use_bn')
```
emailweixu: I suppose you did this because ParallelFC does not support use_bn. But instead of silently ingoring the problem, we should report error.

alf/layers.py
```diff
@@ -277,6 +280,9 @@ def __init__(self,
         self._kernel_init_gain = kernel_init_gain
         self._bias_init_value = bias_init_value
         self._use_bias = use_bias
+        self._use_bn = use_bn
+        if use_bn:
+            self._bn = nn.BatchNorm1d(output_size)
         self.reset_parameters()
 
     def reset_parameters(self):
```
emailweixu: need to update this for bn

emailweixu:
It seems that your unittest take too long to run in CI. Please make it shorter.  Should not be longer than a couple of minutes.

alf/utils/datagen.py
```diff
@@ -52,9 +52,13 @@ def load_mnist(train_bs=100, test_bs=100, num_workers=0):
     return train_loader, test_loader
 
 
-def load_notmnist(train_bs=32, test_bs=100):
+def load_notmnist(train_bs=100, test_bs=100, num_workers=0):
     torch.cuda.manual_seed(1)
```
neale: Could remove the manual seed, as Alf already seeds torch

alf/utils/datagen.py
```diff
@@ -52,9 +52,13 @@ def load_mnist(train_bs=100, test_bs=100, num_workers=0):
...
-    kwargs = {'num_workers': 1, 'pin_memory': True, 'drop_last': False}
+    kwargs = {
+        'num_workers': num_workers,
+        'pin_memory': True,
```
neale: pin_memory should be set to False if default setup does not require users to have a GPU

alf/algorithms/hypernetwork_algorithm.py
```diff
@@ -0,0 +1,430 @@
...
+    target = target.unsqueeze(1).expand(*target.shape[:1], num_particles,
+                                        *target.shape[1:])
+    return loss_func(output, target)
+
+
+@gin.configurable
+class HyperNetwork(Algorithm):
+    """HyperNetwork 
+
+    HyperrNetwork algorithm maintains a generator that generates a set of
```
neale: HyperrNetwork -> HyperNetwork

alf/algorithms/hypernetwork_algorithm.py
```diff
@@ -0,0 +1,430 @@
...
+    It is based on the following work:
+
+    https://github.com/neale/HyperGAN
+
+    Ratzlaff and Fuxin. "HyperGAN: A Generative Model for Diverse, 
+    Performant Neural Networks." International Conference on Machine Learning. 2019.
+
+    Major differences versus the original paper are:
+
+    * A single genrator that generates parameters for all network layers.
```
neale: genrator -> generator

alf/algorithms/hypernetwork_algorithm.py
```diff
@@ -0,0 +1,430 @@
...
+            par_vi (str): types of particle-based methods for variational inference,
+                types are [``svgd``, ``svgd2``, ``svgd3``, ``gfsf``]
+            optimizer (torch.optim.Optimizer): The optimizer for training.
+            logging_network (bool): whether logging the archetectures of networks.
+            logging_training (bool): whether logging loss and acc during training.
+            logging_evaluate (bool): whether logging loss and acc of evaluate.
+            config (TrainerConfig): configuration for training
+            name (str):
+        """
+        super().__init__(train_state_spec=(), name=name)
```
neale: A possibly better option is to use an optimizer here so that the upstream algorithm can detect other components that need to be optimized. Then pass optimizer=None to the generator constructor.

alf/algorithms/hypernetwork_algorithm.py
```diff
@@ -0,0 +1,430 @@
...
+    def _classification_vote(self, output, target):
+        """ensmeble the ooutputs from sampled classifiers."""
+        num_particles = output.shape[1]
+        probs = F.softmax(output, dim=-1)  # [B, N, D]
+        if self._voting == 'soft':
+            pred = probs.mean(1).cpu()  # [B, D]
+            vote = pred.argmax(-1)
+        elif self._voting == 'hard':
+            pred = probs.argmax(-1).cpu()  # [B, N, 1]
+            vote = pred.mode(1)[0]  # [B, 1]
```
neale: torch.mode returns only the first index if there are multiple indices with the same number of "votes". A better option could be to uniformly sample one of the indices for the prediction if there are multiple modes.

alf/algorithms/hypernetwork_algorithm.py
```diff
@@ -0,0 +1,430 @@
...
+        return self._generator.train_step(
+            inputs=None,
+            loss_func=functools.partial(neglogprob, inputs, self._param_net,
+                                        self._loss_type),
+            outputs=params,
+            entropy_regularization=entropy_regularization,
+            state=())
+
+    def evaluate(self, num_particles=None):
+        """Evaluate on a randomly drawn network. """
```
neale: network -> ensemble

alf/utils/datagen.py
```diff
@@ -0,0 +1,183 @@
...
+        trainset, batch_size=train_bs, shuffle=True, num_workers=num_workers)
+
+    test_loader = torch.utils.data.DataLoader(
+        trainset, batch_size=test_bs, shuffle=True, num_workers=num_workers)
+
+    return train_loader, test_loader
+
+
+def load_mnist(train_bs=100, test_bs=100, num_workers=0):
+    torch.cuda.manual_seed(1)
```
neale: Could remove the manual seed, as Alf already seeds torch

alf/utils/datagen.py
```diff
@@ -0,0 +1,183 @@
...
+        datasets.MNIST(
+            path,
+            train=True,
+            download=True,
+            transform=transforms.Compose([
+                transforms.ToTensor(),
+                transforms.Normalize((0.1307, ), (0.3081, ))
+            ])),
+        batch_size=train_bs,
+        shuffle=False,
```
neale: I think the training set should be shuffled.

alf/algorithms/hypernetwork_algorithm.py
runjerry: Good point, updated.

alf/algorithms/generator.py
```diff
@@ -155,37 +173,67 @@ def __init__(self,
...
+            training=training)
         return AlgStep(output=outputs, state=(), info=())
 
-    def train_step(self, inputs, loss_func, batch_size=None, state=None):
+    def train_step(self,
+                   inputs,
+                   loss_func,
+                   outputs=None,
+                   batch_size=None,
+                   entropy_regularization=None,
```
emailweixu, line 247: why do you need this entropy_regularization? The one provided at `__init__` seems enough.

alf/algorithms/generator.py
```diff
@@ -218,35 +270,183 @@ def train_step(self, inputs, loss_func, batch_size=None, state=None):
...
+        return self._kernel_width_averager.get()
+
+    def _rbf_func(self, x, y):
+        """Compute RGF kernel, used by svgd_grad. """
         d = (x - y)**2
-        self._kernel_width_averager.update(torch.mean(d, dim=0))
-        d = torch.mean(d / self._kernel_width_averager.get(), dim=-1)
-        w = torch.exp(-self._kernel_sharpness * d)
+        d = torch.sum(d, -1)
+        h = self._kernel_width(d)
```
emailweixu, line 312: Previously, the kernel_width is a vector. It has width for each dimension.
This is different from the previous implementation. Does it make a difference in terms of convergence?

alf/algorithms/generator.py
```diff
@@ -155,37 +173,67 @@ def __init__(self,
...
+            inputs=inputs,
+            noise=noise,
+            batch_size=batch_size,
+            training=training)
         return AlgStep(output=outputs, state=(), info=())
 
-    def train_step(self, inputs, loss_func, batch_size=None, state=None):
+    def train_step(self,
+                   inputs,
+                   loss_func,
```
emailweixu, line 244: From the code, loss_func can return a tuple. Please update the comment for loss_func

alf/algorithms/generator.py
```diff
@@ -100,26 +108,36 @@ def __init__(self,
...
-                We calcualte the kernel in SVGD as:
-                    :math:`\exp(-kernel\_sharpness * mean(\frac{(x-y)^2}{width}))`
-                where width is the elementwise moving average of :math:`(x-y)^2`
             mi_estimator_cls (type): the class of mutual information estimator
                 for maximizing the mutual information between [noise, inputs]
                 and [outputs, inputs].
+            par_vi (string): ParVI methods, options are
+                [``svgd``, ``svgd2``, ``svgd3``, ``gfsf``],
+                note that for conditional case, i.e., generating from [noise, inputs],
+                only ``svgd`` can be used.
```
emailweixu: comment differences among the choices and which one should be used under which condition.

alf/algorithms/generator.py
```diff
@@ -218,35 +270,183 @@ def train_step(self, inputs, loss_func, batch_size=None, state=None):
...
+        kernel_logp = torch.matmul(kernel_weight.t(),
+                                   loss_grad) / num_particles  # [N, D]
+        grad = kernel_logp - entropy_regularization * kernel_grad.mean(0)
+        loss_propagated = torch.sum(grad.detach() * outputs, dim=-1)
+
+        return loss, loss_propagated
+
+    def _gfsf_grad(self, inputs, outputs, loss_func, entropy_regularization):
+        """Compute particle gradients via GFSF (Stein estimator). """
+        assert inputs is None, "\"gfsf\" does not support conditional generator"
```
emailweixu: Use single quotation to ease the code: '"gfsf" does ... '

alf/algorithms/hypernetwork_algorithm.py
```diff
@@ -0,0 +1,430 @@
...
+    https://github.com/neale/HyperGAN
+
+    Ratzlaff and Fuxin. "HyperGAN: A Generative Model for Diverse, 
+    Performant Neural Networks." International Conference on Machine Learning. 2019.
+
+    Major differences versus the original paper are:
+
+    * A single genrator that generates parameters for all network layers.
+
+    * Remove the mixer and the distriminator.
```
emailweixu: distriminator => discriminator

alf/trainers/policy_trainer.py
```diff
@@ -395,6 +454,128 @@ def _eval(self):
...
+
+    def __init__(self, config: TrainerConfig):
+        """Create a SLTrainer
+
+        Args:
+            config (TrainerConfig): configuration used to construct this trainer
+        """
+        super().__init__(config)
+
+        assert config.num_epochs > 0, \
```
emailweixu: seems num_epochs and num_iterations are same. Should just use the existing num_iterations instead.

alf/trainers/policy_trainer.py
```diff
@@ -395,6 +454,128 @@ def _eval(self):
...
+            if (self._num_epochs and epoch_num >= self._num_epochs):
+                if self._evaluate:
+                    self._algorithm.evaluate()
+                break
+
+            if self._num_epochs and epoch_num >= time_to_checkpoint:
+                self._save_checkpoint()
+                time_to_checkpoint += checkpoint_interval
+
+    def _restore_checkpoint(self):
```
emailweixu, line 546: This is almost same as restore_checkpoint() for RLTrainer. You can create checkpointer in the subclass and call the baseclass `_restore_checkpoint()`

alf/algorithms/hypernetwork_algorithm.py
neale: To elaborate:
compute a tensor of real valued predictions `pred`
```python3
items, indices = pred.unique(return_counts=True)
modes = torch.nonzero(indices==indices.max()).view(-1)  # modes holds a tensor of the most occurring values
idx = modes[torch.randint(len(modes), (1,))} # get a random mode index
rand_mode = modes[idx]  # output mode value
```

alf/trainers/policy_trainer.py
```diff
@@ -395,6 +454,128 @@ def _eval(self):
...
+
+                    common.summarize_gin_config()
+                    alf.summary.text('commandline', ' '.join(sys.argv))
+                    alf.summary.text(
+                        'optimizers',
+                        _markdownify(self._algorithm.get_optimizer_info()))
+                    alf.summary.text('revision', git_utils.get_revision())
+                    alf.summary.text('diff',
+                                     _markdownify(git_utils.get_diff()))
+                    alf.summary.text('seed', str(self._random_seed))
```
emailweixu: put line 530-542 to a function in baseclass and reuse it in RLTrainer and SLTrainer

alf/algorithms/hypernetwork_algorithm.py
```diff
@@ -0,0 +1,430 @@
...
+        self._voting = voting
+        if loss_type == 'classification':
+            self._vote = self._classification_vote
+        elif loss_type == 'regression':
+            self._vote = self._regression_vote
+        else:
+            raise ValueError("Unsupported loss_type: %s" % loss_type)
+
+    def set_data_loader(self, train_loader, test_loader=None):
+        """Set data loadder for training and testing."""
```
emailweixu: comment arguments

alf/trainers/policy_trainer.py
```diff
@@ -395,6 +454,128 @@ def _eval(self):
...
+        checkpoint_interval = math.ceil(
+            self._num_epochs / self._num_checkpoints)
+        time_to_checkpoint = checkpoint_interval
+
+        logging.info("==> Begin Training")
+        while True:
+            logging.info("-" * 68)
+            logging.info("Epoch: {}".format(epoch_num + 1))
+            with record_time("time/train_iter"):
+                train_steps = self._algorithm.train_iter()
```
emailweixu: train_steps is not used.

alf/algorithms/hypernetwork_algorithm.py
```diff
@@ -0,0 +1,430 @@
...
+    def num_particles(self):
+        """number of sampled particles. """
+        return self._num_particles
+
+    def sample_parameters(self, noise=None, num_particles=None, training=True):
+        """Sample parameters for an ensemble of networks. 
+        
+        Args:
+            noise (Tensor): input noise to self._generator.
+            num_particles (int): number of sampled particles.
```
emailweixu: comment num_particles==None, here and at other places

alf/algorithms/hypernetwork_algorithm.py
```diff
@@ -0,0 +1,430 @@
...
+            training (bool): whether or not training self._generator
+
+        Returns:
+            AlgorithmStep from predict_step of self._generator
```
emailweixu: wrong comment.

alf/algorithms/hypernetwork_algorithm.py
```diff
@@ -0,0 +1,430 @@
...
+
+        Args:
+            inputs (Tensor): inputs to the ensemble of networks.
+            params (Tensor): parameters of the ensemble of networks,
+                if None, will resample.
+            num_particles (int): size of sampled ensemble.
+            state: not used.
+
+        Returns:
+            AlgorithmStep: outputs with shape (batch_size, output_dim)
```
emailweixu: AlgStep

emailweixu: output_dim is not defined.

alf/algorithms/hypernetwork_algorithm.py
```diff
@@ -0,0 +1,430 @@
...
+                   state=None):
+        """Perform one batch of training computation.
+
+        Args:
+            inputs (nested Tensor): input training data. 
+            num_particles (int): number of sampled particles. 
+            state: not used
+
+        Returns:
+            AlgorithmStep:
```
emailweixu: AlgStep

alf/algorithms/hypernetwork_algorithm.py
```diff
@@ -0,0 +1,430 @@
...
+                logging.info("Avg acc: {}".format(acc))
+            logging.info("Cum loss: {}".format(loss))
+        self.summarize_train(loss_info, params, cum_loss=loss, avg_acc=acc)
+
+        return batch_idx + 1
+
+    def train_step(self,
+                   inputs,
+                   num_particles=None,
+                   entropy_regularization=None,
```
emailweixu, line 315: why not use the entropy_regularization provided to `__init__`?

alf/algorithms/hypernetwork_algorithm.py
```diff
@@ -0,0 +1,430 @@
...
+                outputs: Tensor with shape (batch_size, dim)
+                info: LossInfo
+        """
+        params = self.sample_parameters(num_particles=num_particles)
+        if entropy_regularization is None:
+            entropy_regularization = self._entropy_regularization
+
+        return self._generator.train_step(
+            inputs=None,
+            loss_func=functools.partial(neglogprob, inputs, self._param_net,
```
emailweixu: If neglogprob is a member function, this can be simplified as `partial(neglogprob, inputs)`

alf/algorithms/hypernetwork_algorithm.py
```diff
@@ -0,0 +1,430 @@
...
+        Returns:
+            AlgorithmStep:
+                outputs: Tensor with shape (batch_size, dim)
+                info: LossInfo
+        """
+        params = self.sample_parameters(num_particles=num_particles)
+        if entropy_regularization is None:
+            entropy_regularization = self._entropy_regularization
+
+        return self._generator.train_step(
```
emailweixu, line 336: why not let `self._generator` to create the parameter?
```
 return self._generator.train_step(
   inputs=None,
   batch_size=num_particles,
   loss_func=...)
```

emailweixu (changes requested):

alf/algorithms/hypernetwork_algorithm.py
runjerry: I have not figured out a way to do this in batch mode without converting back to numpy, so I used a for loop.

alf/algorithms/generator.py
runjerry: I need it for the purpose of shrinking the repulsive term, to work with minibatch training data.

alf/algorithms/generator.py
runjerry: I made this change based on our discussion that even the vectored kernel-width might not be suitable for general cov matrix. And in the Bayesian linear regression test and generator tests, reducing it to a scalar does not seem to slow down the convergence, but I am not sure how it affects more complicated cases.

alf/algorithms/generator.py
runjerry: updated.

alf/algorithms/generator.py
runjerry: added.

alf/algorithms/generator.py
runjerry: updated

alf/algorithms/hypernetwork_algorithm.py
runjerry: corrcted

alf/algorithms/hypernetwork_algorithm.py
runjerry: This parameter depends on the minibatch size for supervised training, which might be altered by the other functions that calls train_step().

alf/algorithms/hypernetwork_algorithm.py
runjerry: updated

alf/algorithms/hypernetwork_algorithm.py
runjerry: I cannot remember the reason I specifically pulled it out from generator, but I agree that it does not need to be done in hypernetwork at all. Updated.

alf/algorithms/hypernetwork_algorithm.py
runjerry: updated

alf/algorithms/hypernetwork_algorithm.py
```diff
@@ -85,7 +68,7 @@ class HyperNetwork(Algorithm):
 
     Major differences versus the original paper are:
 
-    * A single genrator that generates parameters for all network layers.
+    * A single generator that generates parameters for all network layers.
 
     * Remove the mixer and the distriminator.
```
emailweixu: discriminator

alf/algorithms/generator.py
```diff
@@ -218,35 +270,183 @@ def train_step(self, inputs, loss_func, batch_size=None, state=None):
...
+        r"""
+        Compute the rbf kernel and its gradient w.r.t. first entry 
+        :math:`K(x, y), \nabla_x K(x, y)`, used by svgd_grad2.
+
+        Args:
+            x (Tensor): set of N particles, shape (Nx x W), where W is the 
+                dimenseion of each particle
+            y (Tensor): set of N particles, shape (Ny x W), where W is the 
+                dimenseion of each particle
+            h_min (float): minimum kernel bandwidth
```
neale: unused argument h_min

runjerry: removed

alf/algorithms/hypernetwork_algorithm.py
runjerry: corrected

alf/trainers/policy_trainer.py
runjerry: updated

alf/trainers/policy_trainer.py
runjerry: updated.

alf/trainers/policy_trainer.py
runjerry: updated.

alf/algorithms/hypernetwork_algorithm.py
runjerry: updated

alf/trainers/policy_trainer.py
runjerry: updated

alf/algorithms/hypernetwork_algorithm.py
runjerry: updated.

alf/algorithms/hypernetwork_algorithm.py
runjerry: corrected

alf/algorithms/hypernetwork_algorithm.py
runjerry: corrected

alf/algorithms/hypernetwork_algorithm.py
```diff
@@ -230,12 +235,14 @@ def sample_parameters(self, noise=None, num_particles=None, training=True):
         """Sample parameters for an ensemble of networks. 
         
         Args:
-            noise (Tensor): input noise to self._generator.
-            num_particles (int): number of sampled particles.
+            noise (Tensor): input noise to self._generator. Default is None.
+            num_particles (int): number of sampled particles. Default is None.
+                If both noise and num_particles are None, self._num_particles
```
emailweixu: ``num_particles`` provided to the constructor will be used ...

alf/trainers/policy_trainer.py
```diff
@@ -196,17 +196,53 @@ def _close(self):
...
-    def _restore_checkpoint(self):
-        """retore from saved checkpoint."""
-        pass
-
     def _save_checkpoint(self):
         global_step = alf.summary.get_global_counter()
         self._checkpointer.save(global_step=global_step)
 
+    def _restore_checkpoint(self, checkpointer):
+        """retore from saved checkpoint.
```
emailweixu: retore => Restore

alf/trainers/policy_trainer.py
```diff
@@ -196,17 +196,53 @@ def _close(self):
         """Closing operations after training. """
         pass
 
+    def _set_summary(self):
```
emailweixu: _summarize_training_setting might be a better name

alf/trainers/policy_trainer.py
```diff
@@ -98,36 +125,149 @@ def __init__(self, config: TrainerConfig):
...
 
         self._summary_interval = config.summary_interval
         self._summaries_flush_secs = config.summaries_flush_secs
         self._summary_max_queue = config.summary_max_queue
         self._debug_summaries = config.debug_summaries
         self._summarize_grads_and_vars = config.summarize_grads_and_vars
-        alf.summary.should_summarize_output(config.summarize_output)
         self._config = config
 
+        self._random_seed = config.random_seed
```
emailweixu: can be merged with the following line

alf/trainers/policy_trainer.py
```diff
@@ -294,20 +389,7 @@ def _train(self):
...
-                    common.summarize_gin_config()
-                    alf.summary.text('commandline', ' '.join(sys.argv))
-                    alf.summary.text(
-                        'optimizers',
-                        _markdownify(self._algorithm.get_optimizer_info()))
-                    alf.summary.text('revision', git_utils.get_revision())
-                    alf.summary.text('diff',
-                                     _markdownify(git_utils.get_diff()))
-                    alf.summary.text('seed', str(self._random_seed))
+                    self._set_summary()
```
emailweixu: line 387-391 can also be put into this function

alf/algorithms/hypernetwork_algorithm.py
runjerry: I have corrected this following you previous comments, it seems that you are looking at an older commit?

alf/algorithms/hypernetwork_algorithm.py
runjerry: updated.

alf/trainers/policy_trainer.py
runjerry: corrected.

alf/trainers/policy_trainer.py
runjerry: updated.

alf/trainers/policy_trainer.py
runjerry: updated.

alf/trainers/policy_trainer.py
runjerry: This has been addressed following you previous comments. Seems that github is somehow showing you an older commits?

alf/trainers/policy_trainer.py
```diff
@@ -292,22 +386,8 @@ def _train(self):
...
-                    common.summarize_gin_config()
-                    alf.summary.text('commandline', ' '.join(sys.argv))
-                    alf.summary.text(
-                        'optimizers',
-                        _markdownify(self._algorithm.get_optimizer_info()))
-                    alf.summary.text('revision', git_utils.get_revision())
-                    alf.summary.text('diff',
-                                     _markdownify(git_utils.get_diff()))
-                    alf.summary.text('seed', str(self._random_seed))
+                    self._summarize_training_setting()
```
emailweixu: line 386-390 is same as line 529-533. And they can be put into self._summarize_training_setting()

runjerry: updated.

emailweixu (approved):

PR #621, opened 2020-07-06
Observation normalization

le-horizon:
This change
1) makes ObservationNormalizer update stats during training, not unroll.
2) sets default parameters for the best performance of HER.
3) introduces alf.layers.Module to replace nn.Module so that plotting network outputs can be made easier (to distinguish among the three possible modes the networks can be invoked: rollout, replay and eval).
plus some small refactors.

Thanks,
Le


alf/utils/normalizers.py
```diff
@@ -81,6 +82,38 @@ def update(self, tensor):
         self._mean_averager.update(tensor)
         sqr_tensor = alf.nest.map_structure(math_ops.square, tensor)
         self._m2_averager.update(sqr_tensor)
+        if alf.summary.should_summarize_output():
```
hnyu: Why is this summary controlled by summarize_output?

alf/nest/nest.py
```diff
@@ -258,6 +258,37 @@ def extract_any_leaf_from_nest(nest):
                 return ret
 
 
+def all_leaves_with_path_from_nest(nest, path=""):
```
hnyu: This will be inefficient. Should be in C++. A better way is to write a function called map_structure_with_path, and the lambda function is expected to take an extra path argument. It will throw an error if the structure doesn't have fields.

emailweixu:
should condition the calling of summary functions on should_record_summaries() so that there is no waste on preparing the data for summary (e.g. norm(), mean(), etc). ALF already has a lot of overhead. We should be careful to avoid adding more.

le-horizon:
> should condition the calling of summary functions on should_record_summaries() so that there is no waste on preparing the data for summary (e.g. norm(), mean(), etc). ALF already has a lot of overhead. We should be careful to avoid adding more.

Good point.  Forgot to do that for a few places in replay_buffer.  Done.

alf/utils/common.py
```diff
@@ -441,7 +442,7 @@ def forward(self, observation):
         """Normalize a given observation. If during unroll, then first update
         the normalizer. The normalizer won't be updated in other circumstances.
         """
-        if not is_training():
+        if is_training():
```
hnyu: Probably it's better to add an option for the user to select when to update statistics

alf/utils/normalizers.py
```diff
@@ -81,6 +82,38 @@ def update(self, tensor):
...
+            suffix = self.exe_mode_str()
+
+            def _summary(name, val):
+                if val.shape[0] < 30:
```
hnyu: Use a variable to hold this constant 30

alf/layers.py
```diff
@@ -100,8 +109,41 @@ def unflatten(self, tensor):
                      tuple(tensor.shape[1:])))
 
 
+class Module(nn.Module):
+    """Alf's nn.Module wrapper"""
```
hnyu: What's the purpose of having this new Module class?

alf/layers.py
```diff
@@ -20,13 +20,22 @@
...
+import alf
 from alf.initializers import variance_scaling_init
 from alf.nest.utils import get_outer_rank
 from alf.tensor_specs import TensorSpec
 from alf.utils import common
 from alf.utils.math_ops import identity
 
 
+def reduce_along_batch_dims(x, mean, op):
+    spec = TensorSpec.from_tensor(mean)
```
hnyu: from_tensor by default will only treat dim=0 as the batch dim. Then there is no need to use BatchSquash to flatten the tensor x. Generally I feel that this function is not that necessary.

alf/algorithms/agent.py
```diff
@@ -127,6 +127,7 @@ def __init__(self,
         rl_algorithm = rl_algorithm_cls(
             observation_spec=rl_observation_spec,
             action_spec=action_spec,
+            config=config,
```
hnyu: Why do we need config here? usually passing a config means that this algorithm needs to create a replay buffer or set rollout state stuff.

hnyu (changes requested):

alf/algorithms/agent.py
le-horizon: Good catch, this is only meant for updating in unroll, where a separate replay buffer for unroll experience is created in DdpgAlgorithm.

alf/layers.py
le-horizon: This is to replace the nn.Module in alf for all custom code we need to have.  Here, it tracks the mode of rollout vs replay (during training).  It is also responsible for generating the mode string for tensorboard plotting purposes.

alf/utils/normalizers.py
le-horizon: Good point.  Using should_record_summary.

alf/utils/normalizers.py
le-horizon: Added a const: MAX_DIMS_TO_OUTPUT

alf/nest/nest.py
le-horizon: Added comment about efficiency.  This is used for tb summary, so let's leave it as is?  Also for display purposes lists are allowed currently, and assumed to have its index as path, e.g. goal.0, achieved_goal.1 etc..

alf/utils/common.py
le-horizon: For unroll, we need to do it with HER sampling, which means we probably cannot reuse this class.

For the non-her case, we also need to fix the case where test exp should be excluded.

Added a TODO.

alf/layers.py
le-horizon: mean is actually the tensor that has dim=0 as the batch dim, while x can include batch_size, batch_length two batch dims.

This function is modeled after the normalize_along_batch_dims function below it.

le-horizon (commented):
@hnyu thanks for the quick and careful review.
I fixed or answered to all your comments.  PTAL.

Thanks,
Le

alf/algorithms/rl_algorithm.py
```diff
@@ -452,8 +452,10 @@ def unroll(self, unroll_length):
             # to store it in replay buffers
             transformed_time_step = transformed_time_step._replace(
                 untransformed=time_step)
+            self.rollout()
```
emailweixu: We should avoid adding overhead like this. Recursive call can be expensive. Perhaps modify common.mark_training to handle all three modes.

alf/networks/actor_networks.py
```diff
@@ -140,18 +140,20 @@ def forward(self, observation, state=()):
...
+            layer.summary_name = ""
             action = spec_utils.scale_to_spec(action, spec)
 
             if alf.summary.should_summarize_output():
-                a_name = ('summarize_output/' + self.name + '.action_layer.' +
-                          str(i) + '.action.output_norm')
-                if not self.training:
-                    a_name += ".eval"
-                a = action
+                a_name = (
```
emailweixu, line 155: should_summarize_output() will always return True if config.summarize_output is True. But summary is only performed when should_record_summaries() return True. Need to avoid the overhead by conditioning on should_record_summaries() too.

alf/layers.py
```diff
@@ -100,8 +109,41 @@ def unflatten(self, tensor):
...
 
+class Module(nn.Module):
+    """Alf's nn.Module wrapper"""
+
+    def __init__(self):
+        super().__init__()
+        self.rollingout = False
+        self.summary_name = ""
+
+    def rollout(self, mode=True):
```
emailweixu: comment

alf/layers.py
```diff
@@ -100,8 +109,41 @@ def unflatten(self, tensor):
                      tuple(tensor.shape[1:])))
 
 
+class Module(nn.Module):
+    """Alf's nn.Module wrapper"""
+
+    def __init__(self):
+        super().__init__()
+        self.rollingout = False
```
emailweixu: self._rollingout

alf/layers.py
```diff
@@ -100,8 +109,41 @@ def unflatten(self, tensor):
...
+        self.summary_name = ""
```
emailweixu: self._summary_name
and use commented property accessor/setter for it since it's publicly accessible

emailweixu (changes requested):
In general, try not to make existing function slower if possible.

alf/nest/nest.py
hnyu: OK we can support lists. I was thinking of some function like this:

```
map_structure_with_path(lambda t1, t2, path: summary_fn(path, t1+t2), tensor1, tensor2)
```

It basically takes a lambda function with an additional 'path' argument. This path argument will be automatically generated according to the current leaf visited by this map structure function.

The whole purpose of all_leaves_with_path_from_nest is to be used by another loop through all leaves. We can just combine them into one.

alf/utils/common.py
hnyu: Might just set the flag to be one of "train", "unroll", and "eval"?

alf/layers.py
hnyu: I think it's for specific purpose so probably it should be an inner function in whatever calls it and prefixed by '_'.

alf/algorithms/agent.py
hnyu: I'm still confused. If only config is passed but not env, then no replay buffer will be created?

alf/layers.py
hnyu: Generally let's be conservative about changing some very basic module or building block because it will usually bring in efficiency issues or code structure issues. If there is some other way like common.mark_training to achieve the same purpose them we should try to avoid this.

alf/algorithms/rl_algorithm.py
le-horizon: Good point.  The logic is cleaner this way.

alf/layers.py
le-horizon: Good point.

alf/layers.py
le-horizon: Good points.  Done.

alf/layers.py
le-horizon: no longer needed.

alf/layers.py
le-horizon: no longer needed.

alf/layers.py
le-horizon: no longer needed.

alf/networks/actor_networks.py
le-horizon: good point.

alf/utils/common.py
le-horizon: Done.

alf/nest/nest.py
le-horizon: Good point.  Done.

le-horizon (commented):
Thanks for the very good points guys.  Done all, PTAL.

Thanks,
Le

alf/nest/utils.py
```diff
@@ -230,8 +230,11 @@ def _traverse_transform(nested, levels):
             return nested._replace(**{level: new_val})
         elif isinstance(nested, dict):
             new_val = nested.copy()
-            new_val[level] = _traverse_transform(
-                nested=nested[level], levels=levels[1:])
+            _val = _traverse_transform(nested=nested[level], levels=levels[1:])
+            if _val is not None:
```
emailweixu: This is unexpected behavior. What if the caller really want to transform it to None?
And with this behavior, None has different effect for namedtuple and dict. It seems not right.

alf/utils/common.py
```diff
@@ -790,42 +797,107 @@ def detach(nests):
...
         training (bool): True for training, False for unrolling
     """
-    global _training
-    _training = training
+    global _exe_mode
+    _exe_mode = mode
 
 
-def is_training():
+def exe_mode_str():
```
emailweixu: exe_mode_name is a better function name

alf/utils/common.py
```diff
@@ -441,7 +445,10 @@ def forward(self, observation):
         """Normalize a given observation. If during unroll, then first update
         the normalizer. The normalizer won't be updated in other circumstances.
         """
-        if not is_training():
+        # TODO: To update on unroll, we need to fix the case where test exp
+        # should be excluded.
```
emailweixu: test is performed using policy_trainer._step(), which is under eval mode. So there seems no problem.

le-horizon: Done.

alf/nest/utils.py
le-horizon: good point.  reverted.

alf/utils/common.py
le-horizon: Done.

le-horizon (commented):
Addressed all comments, @hnyu and @emailweixu , could you take another look?

Thanks,
Le

emailweixu (approved):

hnyu (approved):

PR #622, opened 2020-07-07
always clip observation in suite_robotics

hnyu:
I'd like to change back to always enforcing observation clipping in suite_robotics. The reason is that this suite actually contains quite some envs to experiment with (Fetch*, HandManipulate*). It's kind of easy to forget adding ObservationClipWrapper in a new gin file when as a fact it should always be included. Also, if we have multiple methods to run on this suite, it's easy to forget adding this wrapper for a new method.

For this purpose, now I also check the environment name passed to suite_robotics to make sure that only Fetch and ShadowHand are supported, which is indeed the original intention of this suite file. ("robotics" refers to OpenAI's robotics envs which only include Fetch and ShadowHand as introduced in their paper).

### DIFF
--- a/alf/environments/suite_robotics.py
+++ b/alf/environments/suite_robotics.py
@@ -135,6 +135,10 @@ def load(environment_name,
     Returns:
         An AlfEnvironment instance.
     """
+    assert (environment_name.startswith("Fetch")
+            or environment_name.startswith("HandManipulate")), (
+                "This suite only supports OpenAI's Fetch and ShadowHand envs!")
+
     _unwrapped_env_checker_.check_and_update(wrap_with_process)
 
     gym_spec = gym.spec(environment_name)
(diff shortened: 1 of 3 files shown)


alf/environments/suite_robotics.py
```diff
@@ -166,6 +170,7 @@ def env_ctor(env_id=None):
             from gym.wrappers import FlattenDictWrapper  # pytype:disable=import-error
             env = FlattenDictWrapper(env, keys)
     env = SuccessWrapper(env, max_episode_steps)
+    env = ObservationClipWrapper(env)
```
le-horizon, line 173: I'm fine with this change, could you at the same time remove the line
suite_robotics.load.gym_env_wrappers=(@ObservationClipWrapper,)
in ddpg_fetchreach.gin?  All the gin files I wrote for robotics env inherits this gin.

Thanks,
Le

le-horizon (commented):
Almost good.  Thanks for the action_l2 change.

alf/environments/suite_robotics.py
hnyu: Removed.

le-horizon (approved):
LG, thanks!

PR #623, opened 2020-07-12
support storing rollout states for SAC

hnyu:


alf/algorithms/sac_algorithm.py
```diff
@@ -423,18 +426,48 @@ def predict_step(self,
...
+                state.critic.critics)
+            _, target_critics_state = self._compute_critics(
+                self._target_critic_networks, time_step.observation, action,
+                state.critic.target_critics)
+            critic_state = SacCriticState(
+                critics=critics_state, target_critics=target_critics_state)
+            if self._act_type == ActionType.Continuous:
+                # During unroll, the computations of ``critics_state`` and
+                # ``actor_state`` are the same.
+                actor_state = critics_state
```
emailweixu, line 458: This doesn't seem correct. SacActionState and SacCriticState are different.

alf/algorithms/algorithm.py
```diff
@@ -1255,17 +1255,22 @@ def _train_experience(self, experience, batch_info, num_updates,
...
 
-        if len(alf.nest.flatten(
-                self.train_state_spec)) > 0 and not self._use_rollout_state:
-            if mini_batch_length == 1:
-                logging.fatal(
-                    "Should use TrainerConfig.use_rollout_state=True "
-                    "for training from a replay buffer when minibatch_length==1, "
-                    "otherwise the initial states are always zeros!")
-            else:
+        if len(alf.nest.flatten(self.train_state_spec)) > 0:
```
Haichao-Zhang, line 1258: It would be useful to explicitly list proper setting for some common examples case here, e.g.  1) with/without RNN 2) mini_batch_length ==1 / > 1, maybe as a table.

alf/algorithms/sac_algorithm.py
hnyu: This is not ActionState. This is the critics state used for training the actor. In the SAC code, actor_state is used in actor_train_step like below:

```
critics, critics_state = self._compute_critics(
                self._critic_networks, exp.observation, action, state)
```

On the other hand, critics_state is used:

```
critics, critics_state = self._compute_critics(
            self._critic_networks, exp.observation, exp.action, state.critics)
```

They are different in train_step. But in rollout_step, they should be the same.

emailweixu: You are right. Though the name is kind of confusing here. I confused critics_state with critic_state.

alf/algorithms/algorithm.py
hnyu: added comments

alf/algorithms/agent.py
```diff
@@ -127,6 +127,7 @@ def __init__(self,
         rl_algorithm = rl_algorithm_cls(
             observation_spec=rl_observation_spec,
             action_spec=action_spec,
+            config=config,  # set use_rollout_state
```
emailweixu: The @use_rollout_property.setter is supposed to handle this. Apparently, it does not have effect because rl_algorithm is not set as the property of self before the base constructor is called.

hnyu: So what's your suggestion? There is a circular dependency here.

hnyu: > The @use_rollout_property.setter is supposed to handle this. Apparently, it does not have effect because rl_algorithm is not set as the property of self before the base constructor is called.

Seems that pytorch won't allow setting a child module before the nn.Module's constructor is called. That's the reason why we always have this kind of pattern creating child modules as local variables first, and them assigning to members after the base constructor.

emailweixu: Previously the setter is called by policy trainer. So there is no problem. Do you think we should remove the setter now?

hnyu: Now I've moved the setter to the end of agent's init. Because several child algorithms might need to inherit this flag. If the root algorithm is not Agent and it has child algorithms, then I guess the user needs to pass the config or set use_rollout_state himself.

emailweixu (approved):

PR #625, opened 2020-07-14
Sparse reward, collision reward for carla

emailweixu:
Also:
1. fixed a bug for CollisionSensor.
2. calculate timeout based route length instead of initial distance to goal.
3. wait longer for starting carla server
4. End epsisode if stuck at collision
5. Log to file
6. sac_carla.gin


alf/environments/suite_carla.py
```diff
@@ -936,13 +996,26 @@ def reset(self):
...
-        self._navigation.set_destination(goal_loc)
+        self._route_length = self._navigation.set_destination(goal_loc)
+
+        self._prev_collision = False  # whether there is collision in the previous frame
+        self._collision = False  # whether there is colliion in the current frame
+        self._collision_loc = None  # the location of the car when it starts to have collition
+
+        # The intermediate goal for sparse reward
+        self._intermediate_goal_index = min(self._sparse_reward_distance,
+                                            self._navigation.num_waypoints - 1)
```
le-horizon, line 1041: If I guess right, then _navigation.num_waypoints == total distance from start to finish?
and _intermediate_goal_index also equals the distance in meters in the route?
Maybe call it intermediate_goal_position to be more accurate?

alf/environments/suite_carla.py
```diff
@@ -803,14 +811,40 @@ def get_current_observation(self, current_frame):
...
+        """
+        return min(self._nearest_index + 1, self._num_waypoints - 1)
+
 
 def _to_numpy_loc(loc: carla.Location):
     return np.array([loc.x, loc.y, loc.z])
 
 
+@gin.configurable(blacklist=['actor', 'alf_world'])
 class Player(object):
```
le-horizon, line 863: This file is getting really long.  Maybe put sensor classes into a file called carla_sensors.py, player class into carla_player.py, and only leave the CarlaServer and CarlaEnv in this file?

alf/environments/suite_carla.py
```diff
@@ -803,14 +811,40 @@ def get_current_observation(self, current_frame):
...
     """Player is a vehicle with some sensors.
```
le-horizon, line 865: Add a comment after the line below, on restarting episode when collision causing player to get stuck.

alf/environments/suite_carla.py
```diff
@@ -852,6 +904,11 @@ def __init__(self,
         self._min_speed = min_speed
         self._delta_seconds = actor.get_world().get_settings(
         ).fixed_delta_seconds
+        self._collision_reward = collision_reward
```
le-horizon: min_collision_reward more appropriate?

alf/environments/suite_carla.py
```diff
@@ -1037,38 +1114,94 @@ def get_current_time_step(self, current_frame):
...
+            # We only report the first collision event among contiguous collision
+            # events.
+            info['collision'] = np.float32(1.0)
+            logging.info("actor=%d frame=%d COLLISION" % (self._actor.id,
+                                                          current_frame))
+            self._collision_loc = curr_loc
+            self._collision_frame = current_frame
+            # We don't want the collision penalty to be too large if the player
+            # cannot even get enough positive moving reward. So we cap the penalty
+            # at ``max(0., self._episode_reward)``
```
le-horizon, line 1170: move this comment to the Args section of the renamed variable?

alf/environments/suite_carla.py
```diff
@@ -1037,38 +1114,94 @@ def get_current_time_step(self, current_frame):
...
+            reward += max(self._collision_reward,
+                          -max(0., self._episode_reward))
 
         if self._fail_frame is None:
             step_type = ds.StepType.FIRST
             max_frames = math.ceil(
-                self._current_distance / self._min_speed / self._delta_seconds)
+                self._route_length / self._min_speed / self._delta_seconds)
             self._fail_frame = current_frame + max_frames
```
le-horizon: Given the change, fail_frame should just be max_frames?

alf/environments/suite_carla.py
```diff
@@ -1037,38 +1114,94 @@ def get_current_time_step(self, current_frame):
...
+            info['collision'] = np.float32(1.0)
+            logging.info("actor=%d frame=%d COLLISION" % (self._actor.id,
+                                                          current_frame))
+            self._collision_loc = curr_loc
+            self._collision_frame = current_frame
+            # We don't want the collision penalty to be too large if the player
+            # cannot even get enough positive moving reward. So we cap the penalty
+            # at ``max(0., self._episode_reward)``
+            reward += max(self._collision_reward,
+                          -max(0., self._episode_reward))
```
le-horizon, line 1173:
```
elif self._collision is False:
    self._collision_loc = None
```

alf/environments/suite_carla.py
```diff
@@ -1037,38 +1114,94 @@ def get_current_time_step(self, current_frame):
...
+                                          self._goal_location) -
+                           np.linalg.norm(curr_loc - self._goal_location))
+            elif self._intermediate_goal_index < current_index:
+                # This means that the car has passed the intermediate goal.
+                # And we give it a reward which is equal to the distance it
+                # travels.
+                intermediate_goal = self._navigation.get_waypoint(
+                    self._intermediate_goal_index)
+                reward += np.linalg.norm(intermediate_goal -
+                                         self._intermediate_start)
```
le-horizon, line 1218: If current_index is the position on the route, then perhaps it's better to use current_pos - _intermediate_start_pos as the reward, instead of the linear distance between the two locations?

alf/examples/sac_carla.gin
```diff
@@ -0,0 +1,65 @@
...
+encoder/EncodingNetwork.activation=%activation
+encoder/EncodingNetwork.fc_layer_params=%fc_layers_params
+encoder/EncodingAlgorithm.encoder_cls=@encoder/EncodingNetwork
+Agent.observation_transformer=@agent/image_scale_transformer
+Agent.representation_learner_cls=@encoder/EncodingAlgorithm
+Agent.optimizer=@Adam()
+agent/image_scale_transformer.min=0.0
+agent/image_scale_transformer.fields=['camera']
+
+#create_environment.num_parallel_environments=16
```
le-horizon: Maybe remove this commented line?

le-horizon (commented):
Nice change Wei.  Mostly nits.

Thanks,
Le

alf/environments/suite_carla.py
```diff
@@ -835,6 +874,19 @@ def __init__(self,
...
             success_distance_thresh (float): success is achieved if the current
                 location is with such distance of the goal
+            collision_reward (float): the reward for collision. Note that this
+                reward is only given once at the first step of contiguous collisions.
+            max_stuck_at_collision_seconds (float): the episode will end and is
+                considerred as failure if the car is stuck at the collision for
+                so many seconds,
+            stuck_at_collision_distance (float): the car is considerred as being
+                stuck at the collision if it is within such distance of the first
+                collision location.
```
hnyu, line 912: What if the car hits some other car with a speed so that the ending stuck position is this distance away from the first collision location?

alf/environments/suite_carla.py
```diff
@@ -835,6 +874,19 @@ def __init__(self,
...
+            sparse_reward (bool): If False, the distance reward is given at every
+                step based on how much it moves along the navigation route. If
+                True, the distance reward is only given after moving ``sparse_reward_distance``.
+            sparse_reward_distance (float): the sparse reward is given after every
+                such distance along the route has been driven.
```
hnyu: I'm actually confused about the two comments here. What does "sparse_reward_distance" exactly do?

alf/environments/suite_carla.py
```diff
@@ -1037,38 +1114,94 @@ def get_current_time_step(self, current_frame):
...
+            obs['collision'] == 0) and not self._is_first_step
+        if self._collision and not self._prev_collision:
+            # We only report the first collision event among contiguous collision
+            # events.
+            info['collision'] = np.float32(1.0)
+            logging.info("actor=%d frame=%d COLLISION" % (self._actor.id,
+                                                          current_frame))
+            self._collision_loc = curr_loc
+            self._collision_frame = current_frame
+            # We don't want the collision penalty to be too large if the player
```
hnyu, line 1168: As this file gets longer and longer, I think it would be best if we have a note documenting the sensor inputs and reward design.

hnyu (changes requested):

alf/environments/suite_carla.py
emailweixu: It might be possible that the waypoints are not evenly spaced in the future. Carla leaderboard does use non-evenly spaced waypoints. So index is a proper name.

alf/environments/suite_carla.py
emailweixu: Let me do that in another PR which only contains the movement of big chunk of code so it is easier to review.

alf/environments/suite_carla.py
emailweixu: added.

alf/environments/suite_carla.py
emailweixu: Right now this is not handled because there are other reasons of getting stuck at some place which cannot be easily distinguished from this. Currently another observed reason for getting stuck is that the waypoints are behind the car. In this situation, the car does not know to turn around or to go backward. But I feel I should just let it try to learn even it is stuck.

alf/environments/suite_carla.py
emailweixu: This means that the the car only get reward after moving ahead so much distance. After it gets such a reward, it has to move another sparse_reward_distance to get reward.

alf/environments/suite_carla.py
emailweixu: done

alf/environments/suite_carla.py
emailweixu: added.

alf/environments/suite_carla.py
emailweixu: updated

alf/environments/suite_carla.py
emailweixu: index may not corresponds to distance if waypoints are not evenly spaced.

alf/examples/sac_carla.gin
emailweixu: removed.

alf/environments/suite_carla.py
le-horizon: _collision_loc is not cleared?

alf/environments/suite_carla.py
le-horizon: If waypoints may not be evenly spaced, then what does this line mean min(sparse_reward_distance, _num_waypoints -1)?

le-horizon (commented):
A couple points left.

alf/environments/suite_carla.py
hnyu: It it makes more sense to name it "sparse_reward_interval".

alf/environments/suite_carla.py
emailweixu: _collision_loc is the location at the first moment of collision. So it should not be cleared.

alf/environments/suite_carla.py
emailweixu: I agree that there is some confusion here. sparse_reward_distance is meant to be distance and intermediate_goal_index is meant to be index. Currently, the distances between two adjacent waypoints are approximately 1 meter (this can not be true for every adjacent pairs because the route length is not guaranteed to be integer number of meters). Updated code to use sampling_resolution to calculating the index.

hnyu (approved):

PR #627, opened 2020-07-15
Birdeye view for carla_spectator

emailweixu:
Solve #626

### DIFF
--- a/alf/environments/carla_spectator.py
+++ b/alf/environments/carla_spectator.py
@@ -62,7 +62,7 @@ def main(_):
 
     width = 640
     height = 320
-    display = pygame.display.set_mode((640, 320),
+    display = pygame.display.set_mode((width, height),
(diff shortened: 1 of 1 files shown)


alf/environments/carla_spectator.py
```diff
@@ -60,9 +60,9 @@ def main(_):
     pygame.init()
     pygame.font.init()
 
-    width = 480
-    height = 240
-    display = pygame.display.set_mode((width, height),
+    width = 640
+    height = 320
+    display = pygame.display.set_mode((640, 320),
```
le-horizon: Should this line still use (width, height)?

le-horizon (dismissed):
LG, Wei.  Please check comment to see if it makes sense.

Thanks,
Le

emailweixu:
@le-horizon need your approval again after addressing your comment

le-horizon (approved):
LG

PR #629, opened 2020-07-16
Support batch normalization

emailweixu:


alf/layers.py
```diff
@@ -336,6 +352,7 @@ def __init__(self,
             output_size (int): output size
             n (int): n independent ``FC`` layers
             activation (torch.nn.functional):
+            use_bn (bool): whehter use Batch Normalization.
```
Haichao-Zhang: whehter -> whether

alf/networks/network.py
```diff
@@ -188,8 +188,11 @@ def output_spec(self):
         can be obtained easily in other ways.
         """
         if self._output_spec is None:
+            training = self.training
+            self.eval()
```
Haichao-Zhang, line 192: Similar switch between ```module.train()``` and ```module.eval()``` mode might be needed once BN is introduced (e.g. in rollout, play..) ?

emailweixu: Done. Though I am not very satisfied how eval and train are implemented. They are using recursive calls to go through all submodules. It's kind of expensive. But it seems there no good way given how pytorch works.

alf/layers.py

Haichao-Zhang (approved):

alf/networks/network.py
Haichao-Zhang: Right. This will add some extra cost. We can try to think of ways to improve it in the future.

PR #630, opened 2020-07-16
Split sensors from suite_carla.py to carla_sensors.py

emailweixu:
Because suite_carla.py is getting longer.

No code change at all except the movement and necessary imports.


le-horizon (approved):
LG

PR #633, opened 2020-07-17
Generating same images for rgb_array mode rendering for suite_carla

### LINKED ISSUE #632: Better rgb_array mode of rendering in suite_carla to include HUD panel in the rendering

emailweixu:
Here is the generated video using alf.bin.play. Previously only a small image without information.
carla


le-horizon (approved):
LG.

PR #641, opened 2020-08-03
DataTransformer

### LINKED ISSUE #645: Avoid saving redundant frame to replay buffer for FrameStack

emailweixu:
The new FrameStacker achieves its purpose by a different transform_experience() function to directly operate on ReplayBuffer.
Also introduced DataTransformer and unify frame stack, reward_shaping_fn and observation_transformer into DataTransformer.


alf/algorithms/data_transformer.py
```diff
@@ -0,0 +1,579 @@
...
+    passing to actual algorithms.
+
+    Most data transformers can subclass from ``SimpleDataTransformer``, which
+    provides a simpler interface.
+    """
+
+    def __init__(self, transformed_observation_spec, state_spec):
+        """
+        Args:
+            transformed_observation_spec (nested TensorSpec): descriging the
```
le-horizon: =>describing

alf/algorithms/data_transformer.py
```diff
@@ -0,0 +1,579 @@
...
+                Note that ``experience.batch_info``, ``experience.replay_buffer``
+                need to be set.
+        Returns:
+            Experience: transformed experience
+        """
+        raise NotImplementedError()
+
+
+class SequentialDataTransformer(DataTransformer):
+    """A data transformer consisting of sequence of data transformers."""
```
le-horizon: "a" sequence of

alf/algorithms/data_transformer.py
```diff
@@ -0,0 +1,579 @@
...
+                    "FrameStacker need to be the "
+                    "first data transformers if it is used.")
+            else:
+                has_non_frame_stacker = True
+            observation_spec = obs_trans.transformed_observation_spec
+            data_transformers.append(obs_trans)
+            state_spec.append(obs_trans.state_spec)
+
+        super().__init__(observation_spec, state_spec)
+        self._stack_size = max_stack_size
```
le-horizon, line 140: if multiple FrameStackers are provided, does it mean the final stack_size should be the multiplications of all the stack_sizes?  If so, then stack_size shouldn't be max_stack_size?

alf/algorithms/data_transformer.py
```diff
@@ -0,0 +1,579 @@
...
+            timestep_or_exp (TimeStep|Experience): data to be transformed
+        Returns:
+            transformed TimeStep of Experience
+        """
+        raise NotImplementedError()
+
+
+class IdentityDataTransformer(SimpleDataTransformer):
+    """A data transformer that keeps the data unchanged."""
+
```
le-horizon, line 399: Maybe also implement __init__(self), so callers of IdentityTransformer don't need to explicitly call with (None) as parameter.

alf/algorithms/data_transformer.py
```diff
@@ -0,0 +1,579 @@
...
+
+@gin.configurable
+class FrameStacker(DataTransformer):
+    def __init__(self,
+                 observation_spec,
+                 stack_size=4,
+                 stack_axis=0,
+                 fields=None):
+        """Create a FrameStacker object.
```
le-horizon, line 167: Add a comment that transform_timestep is expected to be called in an on-policy fashion, while transfor_experience is assumed to be called in an offline fashion.

alf/data_structures.py
```diff
@@ -150,6 +152,10 @@ class Experience(
         This is only used when experiece is passed as an argument for ``Algorithm.calc_loss()``.
         Different from other members, the shape of the tensors in ``batch_info``
         is [B], where B is the batch size.
+    - replay_buffer: The replay buffer where the batch_info generated from
```
le-horizon: From this definition it seems every exp can have this field present, but in reality, this field is only very briefly used when transforming an exp, and cleared right afterwards.  Maybe add a comment explaining that this is how the field is used?

alf/algorithms/algorithm.py
```diff
@@ -270,6 +267,10 @@ def _set_exp_replayer(self, sample_exp):
             alf.nest.extract_fields_from_nest(sample_exp)))
 
         if self._exp_replayer_type == "one_time":
+            assert self._num_earliest_frames_ignored == 0, (
+                "For TrainerConfig, "
+                "whole_replay_buffer_training and clear_replay_buffer cannot both be "
```
hnyu: This error message is not accurate because it's checking for one_time replayer, not directly related to the two flags.

alf/experience_replayers/replay_buffer.py
```diff
@@ -68,6 +69,11 @@ def __init__(self,
             num_environments (int): total number of parallel environments
                 stored in the buffer.
             max_length (int): maximum number of time steps stored in buffer.
+            num_earliest_frames_ignored (int): ignore the earlist so many frames
+                when sample from the buffer. This is typically required when
```
le-horizon: "ignore the earlist so many frames when sample from the buffer." is a bit ambiguous, especially when combined with the espisodic_info comment below.
"ignore the earliest so many frames in the buffer, when sampling."  is more clear.

alf/algorithms/data_transformer.py
```diff
@@ -0,0 +1,579 @@
...
+            positions = convert_device(batch_info.positions)
+
+            prev_positions = torch.arange(self._stack_size -
+                                          1) - self._stack_size + 1
+
+            # [B, stack_size - 1]
+            prev_positions = positions.unsqueeze(
+                -1) + prev_positions.unsqueeze(0)
+            episode_begin_positions = replay_buffer.get_episode_begin_position(
+                positions, env_ids)
```
le-horizon, line 311: Just a comment, no need to address.  Using episodic information can speed up processing in this function, but also incurs cost when storing data.  Simply getting the previous K frames, and find the FIRST steps from them and copy them onto earlier steps could be a bit faster.

le-horizon (commented):
Thanks for the nice and large refactor.  Look almost good.  Mostly nits.

Le

alf/algorithms/data_transformer.py
```diff
@@ -0,0 +1,579 @@
...
+            # repeat the first frame
+            if first_samples.numel() > 0:
+                for t in range(self._stack_size - 1):
+                    # prev_frames[i][t] might be used somewhere else, we should
+                    # not directly modify it.
+                    prev_frames[i][t] = prev_frames[i][t].clone()
+                    prev_frames[i][t][first_samples] = obs[first_samples]
+            if obs.ndim > 1:
+                stacked = torch.cat(prev_frames[i] + [obs], dim=stack_axis)
+            else:
```
hnyu, line 276: What is the case when obs.ndim==1?

alf/algorithms/data_transformer.py
```diff
@@ -0,0 +1,579 @@
...
+            # [B, 1]
+            episode_begin_positions = episode_begin_positions.unsqueeze(-1)
+            # [B, stack_size - 1]
+            prev_positions = torch.max(prev_positions, episode_begin_positions)
+            # [B, 1]
+            env_ids = env_ids.unsqueeze(-1)
+            assert torch.all(
+                prev_positions[:, 0] >= replay_buffer.get_earliest_position(
+                    env_ids)
+            ), ("Some previous posisions are no longer in the replay buffer")
```
hnyu, line 321: What's the possible case of this assertion error?

alf/utils/common.py
```diff
@@ -570,26 +456,48 @@ def set_global_env(env):
 
 
 @gin.configurable
-def get_observation_spec(field=None):
+def get_raw_observation_spec(field=None):
```
hnyu, line 459: Do we still need this function? Typically gin files won't call it.

alf/examples/ppo_rnd_mrevenge.gin
```diff
@@ -7,6 +7,12 @@ include "image_game_model.gin"
 import alf.algorithms.rnd_algorithm
 import alf.tensor_specs
 
+# FrameStacker data transformer does not work with PPO because FrameStacker needs
+# ReplayBuffer, while PPO uses "one_time" experience replay, which does not use
+# ReplayBuffer. So we need to use FrameStack gym wrapper instead.
+suite_gym.load.gym_env_wrappers=(@DMAtariPreprocessing, @FrameStack)
+FrameStack.stack_size=4
```
hnyu: It's easy to make mistakes given the difference between PPO and other algorithms. For one_time replay buffer, is it possible to always maintain extra stack_size-1 frames when clearing the buffer?

hnyu (changes requested):

alf/algorithms/data_transformer.py
emailweixu: If multiple FrameStackers are used, it only make sense if they should operate on different fields. Though this is not enforced.

alf/algorithms/data_transformer.py
emailweixu: This is commented in the base class.

alf/algorithms/data_transformer.py
emailweixu: That is when the spec is a scalar.

alf/algorithms/data_transformer.py
emailweixu: It's not clear how to write efficient code to find the FIRST steps from them and copy them onto earlier steps. Even if it is possible, it might not be faster than using episodic info. And code could be much more complex.

alf/algorithms/data_transformer.py
emailweixu: This can happen if num_ignore_earliest_frames is not correctly set for ReplayBuffer

alf/algorithms/data_transformer.py
emailweixu: added.

alf/examples/ppo_rnd_mrevenge.gin
emailweixu: The current replay buffer does not support that.

alf/algorithms/algorithm.py
emailweixu: I feel it is a more useful message because the concept of "one_time" is not exposed at TrainerConfig.

alf/utils/common.py
emailweixu: Perhaps we don't need it anymore. But it doesn't hurt.

le-horizon (dismissed):
LGTM.

alf/examples/ppo_rnd_mrevenge.gin
hnyu: I mean the OneTimeExperienceReplay class can easily do that. But it seems difficult to integrate it with the current code in FrameStacker.

hnyu (dismissed):

emailweixu:
@hnyu @le-horizon need you to take another look after solving the issue of one_time replayer.

hnyu:
Looks good. Should we remove FrameStack env wrapper in case someone uses it after this change?

emailweixu:
> Looks good. Should we remove FrameStack env wrapper in case someone uses it after this change?

Still keeping it right now in case we need it. Added message about deprecation.

hnyu (approved):

PR #644, opened 2020-08-04
Fix discrete action for SAC

hnyu:
Should check the length of `discrete_action_spec` before it's changed to `action_spec`, otherwise len() will throw errors


hnyu:
Closed because I realize that my local version is different from the master

PR #649, opened 2020-08-12
Checkpoint for ReplayBuffer

### LINKED ISSUE #500: TTP: replay buffer is not checkpointed
There are two problems:
1. ExperienceReplayer needs to be subclass of nn.Module. Current it is subclass of object.
2. Right now. the checkpoint is loaded before training. But experience replayer is created in the first iteration.
3. RingBuffer._buffer and ._flattened_buffer are not stored or loaded during checkpointing.  Override load to construct _buffer from _buffer0...n.

emailweixu:
If there is a checkpoint, we load it after one train_iter() so that replay buffer is created.

### DIFF
--- a/alf/experience_replayers/replay_buffer.py
+++ b/alf/experience_replayers/replay_buffer.py
@@ -26,6 +26,7 @@ from alf.data_structures import namedtuple
 from alf.nest.utils import convert_device
 from alf.utils.common import warning_once
 from alf.utils.data_buffer import atomic, RingBuffer
+from alf.utils import checkpoint_utils
 
 from .segment_tree import SumSegmentTree, MaxSegmentTree, MinSegmentTree
 
@@ -62,6 +63,7 @@ class ReplayBuffer(RingBuffer):
                  keep_episodic_info=None,
                  step_type_field="step_type",
                  postprocess_exp_fn=None,
+                 enable_checkpoint=False,
                  name="ReplayBuffer"):
         """
         Args:
@@ -102,6 +104,7 @@ class ReplayBuffer(RingBuffer):
                     could be populated by gin.
                 Returns:
                     updated ``(batch, batch_info)``.
+            enable_checkpoint (bool): whether checkpointing this replay buffer.
             name (string): name of the replay buffer object.
         """
         super().__init__(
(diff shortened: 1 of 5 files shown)


alf/experience_replayers/replay_buffer.py
```diff
@@ -62,6 +63,7 @@ def __init__(self,
                  keep_episodic_info=None,
                  step_type_field="step_type",
                  postprocess_exp_fn=None,
+                 enable_checkpoint=False,
```
le-horizon, line 66: do we need the ability to turn on or off checkpointing per module?  If this is just for backward compatibility, then the easier way is perhaps to catch any error thrown during loading checkpoint and don't raise it?

le-horizon (dismissed):
Thanks for the elegant change Wei.  I wouldn't have done it so easily.

Only one question, and please also wait for Haichao's review as I'm not familiar with certain parts of checkpointing.

Thanks,
Le

alf/trainers/policy_trainer.py
```diff
@@ -204,6 +204,13 @@ def progress():
 
     def train(self):
         """Perform training."""
+        self._checkpointer = Checkpointer(
+            ckpt_dir=os.path.join(self._train_dir, 'algorithm'),
+            algorithm=self._algorithm,
+            metrics=nn.ModuleList(self._algorithm.get_metrics()),
+            trainer_progress=self._trainer_progress)
+        if self._checkpointer.has_checkpoint():
+            self._algorithm.train_iter()
```
Haichao-Zhang: Might be able to move Line 207-213 into the ```_restore_checkpoint()``` function, since they are all essentially part of this restoring process?
And it might be useful to add a note on the calling of ```self._algorithm.train_iter()``` for constructing any lazily created content related to checkpoint.

Haichao-Zhang (commented):
Looks good. Just one comment.

alf/trainers/policy_trainer.py
emailweixu: good point

alf/experience_replayers/replay_buffer.py
emailweixu: ReplayBuffer can be huge. Disable it by default to reduce disk usage.

Haichao-Zhang (approved):

PR #650, opened 2020-08-13
Do not load ReplayBuffer checkpoint for play

emailweixu:
ReplayBuffer is not created during play. Hence we need to ignore them in order to load checkpoint successfully.

### DIFF
--- a/alf/trainers/policy_trainer.py
+++ b/alf/trainers/policy_trainer.py
@@ -395,7 +395,7 @@ def play(root_dir,
          max_episode_length=0,
          sleep_time_per_step=0.01,
          record_file=None,
-         ingored_parameter_prefixes=['_exp_replayer.']):
+         ignored_parameter_prefixes=['_exp_replayer.']):
     """Play using the latest checkpoint under `train_dir`.
 
     The following example record the play of a trained model to a mp4 video:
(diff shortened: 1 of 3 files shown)


alf/trainers/policy_trainer.py
```diff
@@ -422,13 +423,17 @@ def play(root_dir,
...
+            for skipping loading the checkpoint of ReplayBuffer.
+"""
     root_dir = os.path.expanduser(root_dir)
     train_dir = os.path.join(root_dir, 'train')
 
     ckpt_dir = os.path.join(train_dir, 'algorithm')
     checkpointer = Checkpointer(ckpt_dir=ckpt_dir, algorithm=algorithm)
-    checkpointer.load(checkpoint_step)
+    checkpointer.load(
+        checkpoint_step, ignored_parameter_prefixes=['_exp_replayer.'])
```
Haichao-Zhang: ```ignored_parameter_prefixes=['_exp_replayer.'] -> ignored_parameter_prefixes= ingored_parameter_prefixes ```

Haichao-Zhang (changes requested):

alf/trainers/policy_trainer.py
emailweixu: fixed

Haichao-Zhang (approved):

PR #652, opened 2020-08-14
Scheduler

### LINKED ISSUE #589: Schedulers
During training, some hyper-parameters (e.g learning rate, entropy target, etc.) need to change with iterations. It's tedious to separately implementing the scheduling for each different hyper-parameter. We can use the method in this post to implement schedulers which can be used in the same way as an int or float: https://stackoverflow.com/questions/1117149/does-python-have-conversion-operators

emailweixu:
Support target entropy scheduling and learning rate scheduling

### DIFF
--- a/alf/trainers/policy_trainer.py
+++ b/alf/trainers/policy_trainer.py
@@ -202,6 +202,14 @@ class Trainer(object):
         """
         return Trainer._trainer_progress.progress
 
+    @staticmethod
+    def current_iterations():
+        return Trainer._trainer_progress._iter_num
+
+    @staticmethod
+    def current_env_steps():
+        return Trainer._trainer_progress._env_step
+
     def train(self):
         """Perform training."""
         self._restore_checkpoint()
@@ -230,8 +238,9 @@ class Trainer(object):
                 ps.print_callees()
 
                 logging.info(s.getvalue())
-        finally:
+
             self._save_checkpoint()
+        finally:
             self._close_envs()
 
     def _train(self):
--- a/alf/trainers/policy_trainer.py
+++ b/alf/trainers/policy_trainer.py
@@ -238,8 +238,11 @@ class Trainer(object):
                 ps.print_callees()
 
                 logging.info(s.getvalue())
-
             self._save_checkpoint()
+        except:
+            ans = input("Do you want to save checkpoint? (y/n): ")
+            if ans.lower().startswith('y'):
+                self._save_checkpoint()
         finally:
             self._close_envs()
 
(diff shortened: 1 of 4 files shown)


alf/utils/schedulers.py
```diff
@@ -0,0 +1,154 @@
...
+class LinearScheduler(Scheduler):
+    """The value is linearly changed in each defined region of progress."""
+
+    def __init__(self, progress_type, schedule):
+        """
+        Args:
+            progress_type (str): one of "percent", "iterations", "env_steps"
+            schedule (list[tuple]): each tuple is a pair of (progress, value)
+                which means that if current progress is less than progress,
+                the current value will be value.
```
Haichao-Zhang: the current value -> a linearly interpolated value

alf/trainers/policy_trainer.py
```diff
@@ -230,8 +238,9 @@ def train(self):
                 ps.print_callees()
 
                 logging.info(s.getvalue())
-        finally:
```
hnyu, line 233: Why making this change? Currently it will save the checkpoint if an error occurs, which is good.

alf/utils/schedulers.py
```diff
@@ -0,0 +1,154 @@
...
+        super().__init__(progress_type)
+        self._progresses, self._values = zip(*schedule)
+        self._index = 0
+
+    def __call__(self):
+        progress = self.progress()
+        index = self._index
+        progresses = self._progresses
+        if progress < progresses[index]:
+            while index > 0 and progress < progresses[index - 1]:
```
hnyu: In what situation we need to backstep the index? It seems that the progress is always incrementing

alf/utils/schedulers.py
```diff
@@ -0,0 +1,154 @@
...
+
+@gin.configurable
+class LinearScheduler(Scheduler):
+    """The value is linearly changed in each defined region of progress."""
+
+    def __init__(self, progress_type, schedule):
+        """
+        Args:
+            progress_type (str): one of "percent", "iterations", "env_steps"
+            schedule (list[tuple]): each tuple is a pair of (progress, value)
```
hnyu, line 95: This comment is still for stepscheduler.

hnyu (changes requested):

Haichao-Zhang (dismissed):

alf/utils/schedulers.py
```diff
@@ -0,0 +1,147 @@
...
+                which means that if the current progress between progress[i-1]
+                and progress[i], a linear interpolation between value[i-1] and
+                value[i] will be used. progress[0] must be 0.
```
hnyu: Maybe also add a sentence about if the current progress is greater than progress[-1]? (then it's always the same last value)

hnyu (approved):

PR #655, opened 2020-08-14
VideoRecorder for display pred info

hnyu:
Currently support showing action and action distribution. Value info display to be added. An example is below (The actual frame size is bigger than this):

ezgif com-video-to-gif (1)

Also allow specifying fps to render slow-motion videos.

### DIFF
--- a/alf/utils/video_recorder.py
+++ b/alf/utils/video_recorder.py
@@ -1,0 +1,323 @@
+# Copyright (c) 2020 Horizon Robotics. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import io
+import cv2
+import gin
+import numpy as np
+import matplotlib.pyplot as plt
+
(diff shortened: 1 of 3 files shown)


alf/utils/video_recorder.py
```diff
@@ -0,0 +1,323 @@
...
+                excluding the env frame width
+            value_range (float): if quantities plotted as heatmaps, the values
+                will be clipped according to ``[-value_range, value_range]``.
+            frames_per_sec (fps):
+        """
+        super(VideoRecorder, self).__init__(env=env, **kwargs)
+        self._img_plot_width = img_plot_width
+        self._dist_imgs_per_row = 4
+        self._value_range = value_range
+        self.frames_per_sec = frames_per_sec  # overwrite the base class
```
emailweixu: frames_per_sec should be retrieved from env.metadata['video.frames_per_second']

alf/utils/video_recorder.py
```diff
@@ -0,0 +1,323 @@
...
+                self._encode_ansi_frame(frame)
+            else:
+                self._encode_image_frame(frame)
+
+    def _plot_prob_curve(self, name, probs, xticks=None):
+        if xticks is None:
+            xticks = range(len(probs))
+
+        fig, ax = plt.subplots(figsize=(2, 2))
+        ax.plot(xticks, probs)
```
Haichao-Zhang, line 178: The video on the plot curve part could to be more smooth if the frames are aligned, for example, by using the same axis range for all the plots in the same video/episode.

hnyu: It's difficult to determine the axis range beforehand. It's a density distribution so there is no upper bound on y.

alf/utils/video_recorder.py
hnyu: As I just verified, we can overwrite the env fps to render a slow-motion video by specifying a smaller fps. This might be convenient to analyze the action outputs.

emailweixu (approved):

PR #659, opened 2020-08-20
compare tool can now run

le-horizon:
Hi team,

This is a basic algorithm compare tool that runs two different algorithms on a set of random seeds and compare the behavior of the two.

Setups/random_seeds that cause a difference in AverageReturn in the two algorithms will be saved into an HTML file, with top runs' videos captured.  The HTML table supports reordering by columns as well as searching/filtering, due to DataTables js plugin.

See screenshot:
compare_tool

Currently it depends on play.py to output the two metrics.

Cheers,
Le

### DIFF
--- a/alf/trainers/policy_trainer.py
+++ b/alf/trainers/policy_trainer.py
@@ -482,6 +482,10 @@ def play(root_dir,
     episode_reward = 0.
     episode_length = 0
     episodes = 0
+    metrics = [
+        alf.metrics.AverageReturnMetric(buffer_size=num_episodes),
+        alf.metrics.AverageEpisodeLengthMetric(buffer_size=num_episodes),
+    ]
     while episodes < num_episodes:
         time_step, policy_state, trans_state, info = _step(
             algorithm=algorithm,
@@ -490,7 +494,7 @@ def play(root_dir,
             policy_state=policy_state,
             trans_state=trans_state,
             epsilon_greedy=epsilon_greedy,
-            metrics=[])
+            metrics=metrics)
         episode_length += 1
         if recorder:
             recorder.capture_frame(info)
@@ -506,8 +510,13 @@ def play(root_dir,
             episode_reward = 0.
             episode_length = 0.
             episodes += 1
+            # observe the last step
+            for m in metrics:
+                m(time_step.cpu())
             time_step = env.reset()
 
+    for m in metrics:
+        logging.info("%s: %f", m.name, m.result())
     if recorder:
         recorder.close()
     env.reset()
--- a/alf/bin/compare.py
+++ b/alf/bin/compare.py
@@ -41,7 +41,6 @@ flags.DEFINE_string('root_dir2', None, 'Root directory for algorithm two.')
 flags.DEFINE_string('output_file', None, 'output html file.')
 flags.DEFINE_integer('num_runs', 10, 'Compare on so many runs.')
 flags.DEFINE_integer('start_from', 0, 'Start random seeds from here.')
-flags.DEFINE_integer('num_videos', 2, 'Record videos for so many top diffs.')
 flags.DEFINE_string(
     'common_gin', '', 'Common config for the two sides, '
     'e.g. "--gin_param=\'GoalTask.random_range=5\'"')
@@ -53,23 +52,23 @@ AVG_R_METRIC = "AverageReturn"
 AVG_R_DIFF = AVG_R_METRIC + "_diff"
 
 
-def return_diff(item):
+def _return_diff(item):
     return abs(item[AVG_R_DIFF]) / (max(
         abs(float(item["alg1_" + AVG_R_METRIC])),
         abs(float(item["alg2_" + AVG_R_METRIC]))) + 1.e-5)
 
 
-def return_1_larger(item):
+def _return_1_larger(item):
     return float(item["alg1_" + AVG_R_METRIC]) > float(
         item["alg2_" + AVG_R_METRIC])
 
 
-def file_exists(file):
+def _file_exists(file):
     return (not FLAGS.overwrite and os.path.isfile(file)
             and os.stat(file).st_size > 100)
 
 
-def play_cmd(root_dir, seed):
+def _play_cmd(root_dir, seed):
     return ("cd {root_dir} && "
             "python3 -m alf.bin.play "
             " --random_seed={seed} --num_episodes=1"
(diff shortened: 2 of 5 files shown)


le-horizon:
Btw, this generated html lives on the server/desktop, so I use vglrun to run firefox to view it.  One can also copy over the root_dirs for the two algorithms to local, and rerun the script to re-generate the compare.html.

le-horizon:
@emailweixu and @hnyu the test error is due to calling time_step.cpu() in policy_trainer.py
The nest seems to contain a "set" object, which doesn't have "cpu" attribute.
I'm not familiar with this.  How should we fix?

prev_action=tensor([11]), env_id=tensor([0], dtype=torch.int32), untransformed=(), env_info={'coins': tensor([1]), 'levelHi': tensor([0]), 'levelLo': tensor([0]), 'levels': {(0, 0)}, 'lives': tensor([-1]), 'num_env_frames': tensor([1]), 'retro_episode': {'levels': {(0, 0)}}, 'score': tensor([20]), 'scrolling': tensor([16]), 'time': tensor([390]), 'xscrollHi': tensor([0]), 'xscrollLo': tensor([173])}),). Error message: ''set' object has no attribute 'cpu''

hnyu:
> @emailweixu and @hnyu the test error is due to calling time_step.cpu() in policy_trainer.py
> The nest seems to contain a "set" object, which doesn't have "cpu" attribute.
> I'm not familiar with this. How should we fix?
> 
> prev_action=tensor([11]), env_id=tensor([0], dtype=torch.int32), untransformed=(), env_info={'coins': tensor([1]), 'levelHi': tensor([0]), 'levelLo': tensor([0]), 'levels': {(0, 0)}, 'lives': tensor([-1]), 'num_env_frames': tensor([1]), 'retro_episode': {'levels': {(0, 0)}}, 'score': tensor([20]), 'scrolling': tensor([16]), 'time': tensor([390]), 'xscrollHi': tensor([0]), 'xscrollLo': tensor([173])}),). Error message: ''set' object has no attribute 'cpu''

I saw lots of map_structure failures. Make sure to first run the test locally?

hnyu:
> @emailweixu and @hnyu the test error is due to calling time_step.cpu() in policy_trainer.py
> The nest seems to contain a "set" object, which doesn't have "cpu" attribute.
> I'm not familiar with this. How should we fix?
> 
> prev_action=tensor([11]), env_id=tensor([0], dtype=torch.int32), untransformed=(), env_info={'coins': tensor([1]), 'levelHi': tensor([0]), 'levelLo': tensor([0]), 'levels': {(0, 0)}, 'lives': tensor([-1]), 'num_env_frames': tensor([1]), 'retro_episode': {'levels': {(0, 0)}}, 'score': tensor([20]), 'scrolling': tensor([16]), 'time': tensor([390]), 'xscrollHi': tensor([0]), 'xscrollLo': tensor([173])}),). Error message: ''set' object has no attribute 'cpu''

The time step contains info that's a set. We haven't considered set as a sequence data structure currently.

le-horizon:
@hnyu and @emailweixu,  play.py already uses ThreadEnv, which has Wei's earlier fix (https://github.com/HorizonRobotics/alf/pull/606/files#diff-89963dc6ba191149c7bb4e2758db0603)

But the problem with "levels" fields is still there.

I've cleared the fields in this change.

Let me know what you think.

Thanks,
Le

hnyu:
> @hnyu and @emailweixu, play.py already uses ThreadEnv, which has Wei's earlier fix (https://github.com/HorizonRobotics/alf/pull/606/files#diff-89963dc6ba191149c7bb4e2758db0603)
> 
> But the problem with "levels" fields is still there.
> 
> I've cleared the fields in this change.
> 
> Let me know what you think.
> 
> Thanks,
> Le

It's not about whether you use ThreadEnv or not. It's about calling "x.cpu" when x is a set (levels info). To fix this issue, you only need to check if 'x' is a tensor in TimeStep.cpu(). 

https://github.com/HorizonRobotics/alf/blob/10c46ae08603186232927585fa919816ea3f844a/alf/data_structures.py#L122

Also don't forget to add the same check in cuda().

le-horizon:
> > @hnyu and @emailweixu, play.py already uses ThreadEnv, which has Wei's earlier fix (https://github.com/HorizonRobotics/alf/pull/606/files#diff-89963dc6ba191149c7bb4e2758db0603)
> > But the problem with "levels" fields is still there.
> > I've cleared the fields in this change.
> > Let me know what you think.
> > Thanks,
> > Le
> 
> It's not about whether you use ThreadEnv or not. It's about calling "x.cpu" when x is a set (levels info). To fix this issue, you only need to check if 'x' is a tensor in TimeStep.cpu().
> 
> https://github.com/HorizonRobotics/alf/blob/10c46ae08603186232927585fa919816ea3f844a/alf/data_structures.py#L122
> 
> Also don't forget to add the same check in cuda().

Good point.  But this general fix would slow things down, and doing it just for fixing this Mario specific glitch is probably not worth it?

hnyu:
> Good point. But this general fix would slow things down, and doing it just for fixing this Mario specific glitch is probably not worth it?

I don't think it will slow down things by adding the check. Moreover, it's a more general fix because if in the future we have something other than a tensor in TimeStep (non-tensor info from other env, for example), then this error will occur again. 

```python
r = nest.map_structure(lambda x: x.cpu() if isinstance(x, torch.Tensor) else x, self)
```

le-horizon:
> > Good point. But this general fix would slow things down, and doing it just for fixing this Mario specific glitch is probably not worth it?
> 
> I don't think it will slow down things by adding the check. Moreover, it's a more general fix because if in the future we have something other than a tensor in TimeStep (non-tensor info from other env, for example), then this error will occur again.
> 
> ```python
> r = nest.map_structure(lambda x: x.cpu() if isinstance(x, torch.Tensor) else x, self)
> ```

Ok, done.

alf/trainers/policy_trainer.py
```diff
@@ -506,8 +510,13 @@ def play(root_dir,
             episode_reward = 0.
             episode_length = 0.
             episodes += 1
+            # observe the last step
+            for m in metrics:
```
hnyu, line 514: I think the metrics are already recorded by line 508-509? Is there any reason to record again using metrics?

le-horizon: If you read the code carefully, you'll see, the last steps are not actually observed by the metrics above..

le-horizon: Line 516 overwrites the last time_step of all episodes before it is observed by metrics.

hnyu: It's the first time_step (FIRST) not observed. Time steps >=1 till the LAST will always be counted by line 498 and 505. And for the initial time step it's probably fine because the reward dones't mean anything. BTW, I was referring to line 508-509 not using metrics for counting the reward and length. It seems duplicate with the metrics here.

le-horizon: If you look at _step, it actually observes the input timestep, not the output timestep..

hnyu: I mean line 508-509 after the change:
```python
logging.info("episode_length=%s episode_reward=%s" %
                         (episode_length, episode_reward))
```
is a duplicate of the metrics.

le-horizon: Oh, that one is reported for every episode, and the final number is grand average of all episodes.  It's a useful thing to report both?

hnyu: OK fair enough

hnyu:
Can you prepend any internal functions with '_', so that they will be ignored for docstring generation?

le-horizon:
> Can you prepend any internal functions with '_', so that they will be ignored for docstring generation?

Done.  Please take another look.

Thanks,
Le

hnyu (approved):

PR #661, opened 2020-08-24
on-demand debugging

emailweixu:
on-demand checkpoint and debugging for SLTrainer

### DIFF
--- a/alf/trainers/policy_trainer.py
+++ b/alf/trainers/policy_trainer.py
@@ -112,7 +112,7 @@ class Trainer(object):
     summary, checkpointing, running training iterations, and evaluating periodically.
     """
 
-    _trainer_progress = _TrainerProgress()
+    _trainer_progress = None
 
     def __init__(self, config: TrainerConfig):
         """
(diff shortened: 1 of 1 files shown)


alf/trainers/policy_trainer.py
```diff
@@ -178,13 +186,32 @@ def train(self):
 
                 logging.info(s.getvalue())
             self._save_checkpoint()
-        except:
-            ans = input("Do you want to save checkpoint? (y/n): ")
-            if ans.lower().startswith('y'):
-                self._save_checkpoint()
+            checkpoint_saved = True
         finally:
+            if not checkpoint_saved:
```
emailweixu, line 192: The reason of not using except is that we can debug if anything go wrong.

hnyu: Will finally print out the exception message?

emailweixu: Yes.

hnyu (approved):

PR #665, opened 2020-08-28
Fix histogram_discrete

hnyu:
The bin edges should be 0.5 off the bin centers. The original code will crash if there is only one discrete action.

### DIFF
--- a/alf/utils/summary_utils.py
+++ b/alf/utils/summary_utils.py
@@ -169,9 +169,8 @@ def summarize_loss(loss_info: LossInfo):
     Args:
         loss_info (LossInfo): ``loss_info.extra`` must be a namedtuple
     """
-    if not loss_info.loss:
-        return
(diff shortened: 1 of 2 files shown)


alf/utils/summary_utils.py
```diff
@@ -169,6 +169,8 @@ def summarize_loss(loss_info: LossInfo):
     Args:
         loss_info (LossInfo): ``loss_info.extra`` must be a namedtuple
     """
+    if not loss_info.loss:
```
emailweixu: bool(torch.tensor(0.0)) is False
and should only skip line 174

emailweixu (changes requested):

alf/utils/summary_utils.py
hnyu: fixed

emailweixu (approved):

PR #666, opened 2020-08-29
fix videorecorder import again

hnyu:
My previous PR #658 actually introduced a bug because any file that uses gin must be imported in the beginning, otherwise gin will complain. So this PR will ignore using matplotlib if none is detected.

### DIFF
--- a/alf/bin/train_play_test.py
+++ b/alf/bin/train_play_test.py
@@ -231,6 +231,7 @@ class TrainPlayTest(alf.test.TestCase):
               skip_checker=None,
               extra_train_params=None,
               test_play=True,
+              test_video_recording=False,
               extra_play_params=None,
               test_perf=True,
               test_perf_func=None):
@@ -242,6 +243,7 @@ class TrainPlayTest(alf.test.TestCase):
                 exception when the test is not available.
             extra_train_params (list[str]): extra params used for training.
             test_play (bool): A bool for test play.
+            test_video_recording (bool): whether test video recording for play.
             extra_play_params (list[str]): extra param used for play.
             test_perf (bool): A bool for check performance.
             test_perf_func (Callable): called as test_perf_func(episode_returns, episode_lengths)
@@ -262,7 +264,8 @@ class TrainPlayTest(alf.test.TestCase):
         with tempfile.TemporaryDirectory() as root_dir:
             self._test_train(gin_file, extra_train_params, root_dir)
             if test_play:
-                self._test_play(root_dir, extra_play_params)
+                self._test_play(root_dir, extra_play_params,
+                                test_video_recording)
             if test_perf and test_perf_func:
                 self._test_performance(root_dir, test_perf_func)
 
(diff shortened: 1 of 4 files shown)


alf/utils/video_recorder.py
```diff
@@ -16,7 +16,15 @@
...
-import matplotlib.pyplot as plt
+
+from alf.utils import common
+
+try:
+    import matplotlib.pyplot as plt
+except ImportError:
+    common.warning_once("matplotlib is not installed; prediction info will not"
+                        " be plotted when rendering videos.")
+    plt = None
```
emailweixu, line 25: move this to line 173 so that it will not bother if it is not used.

le-horizon (commented):
Thanks for fixing @hnyu .  Could we also add a unit test to capture this case to protect against future breakages?

Also, I tried running play on a trained model in recording mode (with record_file=something.mp4), no .mp4 file was generated.  Only something.meta.json is there.  Could you check if you can record a video with info passed into capture_frame()?

Here is the error:
python3.7/site-packages/gym/logger.py:30: UserWarning: ^[[33mWARN: Tried to pass invalid video frame, marking as broken: Your frame has shape (960, 1920, 3), but the VideoRecorder is configured for shape (480, 640, 3)

After I remove the info in the call to capture_frame(), video can be recorded properly.

Thanks,
Le

hnyu:
> Thanks for fixing @hnyu . Could we also add a unit test to capture this case to protect against future breakages?
> 
> Also, I tried running play on a trained model in recording mode (with record_file=something.mp4), no .mp4 file was generated. Only something.meta.json is there. Could you check if you can record a video with info passed into capture_frame()?
> 
> Here is the error:
> python3.7/site-packages/gym/logger.py:30: UserWarning: ^[[33mWARN: Tried to pass invalid video frame, marking as broken: Your frame has shape (960, 1920, 3), but the VideoRecorder is configured for shape (480, 640, 3)
> 
> After I remove the info in the call to capture_frame(), video can be recorded properly.
> 
> Thanks,
> Le

Unit test added. I don't have this error as tested. Make sure that you don't call capture_frame() without any argument in other places because that will init a different frame size. Right now, this call should only appear in policy_trainer.py line 666.

alf/bin/train_play_test.py
```diff
@@ -289,18 +292,22 @@ def _test_train(self, gin_file, extra_params, root_dir):
...
             extra_params (list[str]): extra parameters used for play
+            test_video_recording (bool): if True, also test if a video can be
+                recorded.
         """
         cmd = [
             'python3', '-m', 'alf.bin.play',
             '--root_dir=%s' % root_dir, '--num_episodes=1'
         ]
+        if test_video_recording:
+            cmd.append('--record_file=%s/play.mp4' % root_dir)
```
le-horizon, line 310: after run_cmd, need to check
`
file = root_dir + "/play.mp4"
assert os.path.isfile(file) and os.stat(file).st_size > 100`

In my failure, the run didn't fail, but no video either.

Thanks,
Le

le-horizon (commented):
Oh, I see, I did call it without arg at one place after env.reset().  In that case, there is no policy step, should I just use all zero as policy_info?

Thanks,
Le

alf/bin/train_play_test.py
hnyu: I'm not sure what you mean here? I've tested locally that I will get an mp4 file after the play. 

In the current train_play_test.py, if any error happens during the rendering, then the test will fail. No need to check the resulting file.

le-horizon:
> Unit test added. I don't have this error as tested. Make sure that you don't call capture_frame() without any argument in other places because that will init a different frame size. Right now, this call should only appear in policy_trainer.py line 666.

@hnyu Oh, I see, I did call it without arg first, right after env.reset(). In that case, there is no policy step, should I just use all zero as policy_info?

hnyu:
> > Unit test added. I don't have this error as tested. Make sure that you don't call capture_frame() without any argument in other places because that will init a different frame size. Right now, this call should only appear in policy_trainer.py line 666.
> 
> @hnyu Oh, I see, I did call it without arg first, right after env.reset(). In that case, there is no policy step, should I just use all zero as policy_info?

We can just ignore the first frame like the current code already does.

le-horizon (dismissed):
LG, thanks for the fix!

le-horizon (approved):
Ok, LG

PR #668, opened 2020-08-31
Do not use bias when batchnorm is used.

emailweixu:
It's a waste.


alf/layers.py
```diff
@@ -700,7 +704,7 @@ def __init__(self,
             activation (torch.nn.functional):
             strides (int or tuple):
             padding (int or tuple):
-            use_bias (bool):
+            use_bias (bool): If None, will use ``not use_bn``
```
Haichao-Zhang: (bool) -> (bool|None)

alf/layers.py
```diff
@@ -777,7 +783,7 @@ def __init__(self,
...
```
Haichao-Zhang: (bool) -> (bool|None)

alf/layers.py
```diff
@@ -551,7 +553,7 @@ def __init__(self,
             activation (torch.nn.functional):
             strides (int or tuple):
             padding (int or tuple):
-            use_bias (bool): whether use bias
+            use_bias (bool): whether use bias. If None, will use ``not use_bn``
```
Haichao-Zhang: (bool) -> (bool|None)

alf/layers.py
```diff
@@ -472,7 +472,7 @@ def __init__(self,
             activation (torch.nn.functional):
             strides (int or tuple):
             padding (int or tuple):
-            use_bias (bool): whether use bias
+            use_bias (bool|None): whether use bias. If None, will use ``not use_bn``
```
Haichao-Zhang, line 475: Other places can be updated like this one for the type of ``use_bias``

Haichao-Zhang (approved):

PR #670, opened 2020-09-02
Observation normalizer for subset of fields

emailweixu:
Also added summary for observation

### DIFF
--- a/alf/nest/nest.py
+++ b/alf/nest/nest.py
@@ -693,6 +693,58 @@ def get_field(nested, field):
     return _traverse(nested=nested, levels=field.split('.') if field else [])
 
 
+def transform_nest(nested, field, func):
+    """Transform the node of a nested structure indicated by ``field`` using
+    ``func``.
+
+    This function can be used to update our ``namedtuple`` structure conveniently,
+    comparing the following two methods:
+
+        .. code-block:: python
+
+            info = info._replace(rl=info.rl._replace(sac=info.rl.sac * 0.5))
+
+    vs.
+
+        .. code-block:: python
+
+            info = transform_nest(info, 'rl.sac', lambda x: x * 0.5)
+
+    The second method is usually shorter, more intuitive, and less error-prone
+    when ``field`` is a long string.
+
+    Args:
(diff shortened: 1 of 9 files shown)


alf/nest/nest.py
```diff
@@ -691,3 +691,39 @@ def _traverse(nested, levels):
...
+            return nested._replace(**{level: new_val})
+        elif isinstance(nested, dict):
+            new_val = nested.copy()
+            new_val[level] = _traverse(nested=nested[level], levels=levels[1:])
+            return new_val
+        else:
+            raise TypeError("If value is a nest, it must be either " +
+                            "a dict or namedtuple!")
+
+    return _traverse(nested=nested, levels=field.split('.') if field else [])
```
le-horizon: Can this be done using transform_nest, e.g.
`result = alf.nest.utils.transform_nest(result, field, lambda _: new_value)`?

le-horizon (commented):
Almost good, only one comment.

alf/nest/nest.py
emailweixu: Sure it can. But not feeling well to use more complex API to do the simpler task.

hnyu: Right. But I think that to reduce maintenance cost, this can just be a wrapper instead that calls transform_nest

le-horizon: If there isn't much efficiency gain, then I'd vote for less code.  Having set_field func call transform_nest is probably a good compromise.

alf/nest/nest.py
```diff
@@ -693,37 +693,73 @@ def _traverse(nested, levels):
     return _traverse(nested=nested, levels=field.split('.') if field else [])
 
 
-def set_field(nested, field, new_value):
-    """Set the field in nested to ``new_value``.
+def transform_nest(nested, field, func):
```
emailweixu, line 696: This is moved from nest/utils.py without any change.

le-horizon (dismissed):
LG.  I'm surprised there are only so few calls to transform_nest.  I would imagine more places need to be updated.

hnyu (approved):

PR #672, opened 2020-09-02
Action observation wrapper

emailweixu:
A wrapper that add prev_action into observation.
Also allow suite_carla.load() to use env wrappers.


le-horizon (approved):
LG

PR #676, opened 2020-09-04
dynamic action repetition algorithm

hnyu:
Tested on FetchPickAndPlace and FetchSlide. Seems better than what's reported here (grid-searched) https://alf.readthedocs.io/en/latest/notes/fetch_env_sac_actrepeat_notes.html

I rendered a video on FetchPickAndPlace and the learned repetition duration looks reasonable (First row is continuous action where the first 3 dims are xyz movement and the last is close/open gripper. Red means -1 and blue means 1. The second row is how many steps remaining for this action in one-hot).

Keeping stationary:
Screen Shot 2020-09-03 at 8 05 30 PM
Screen Shot 2020-09-03 at 8 05 40 PM
Screen Shot 2020-09-03 at 8 05 53 PM

Lifting: 
Screen Shot 2020-09-03 at 8 03 54 PM
Screen Shot 2020-09-03 at 8 04 28 PM
Screen Shot 2020-09-03 at 8 04 40 PM


alf/algorithms/sac_algorithm.py
```diff
@@ -512,24 +540,34 @@ def _compute_critics(self, critic_net, observation, action, critics_state):
...
-            critics = self._select_q_value(action[0], critics)
-            action = action[1]
-
-        target_q_value = critics.min(dim=1)[0]
+            discrete_act_dist = action_distribution[0]
+            discrete_entropy = discrete_act_dist.entropy()
+            # critics is already after min over replicas
+            #assert discrete_act_dist.probs.shape == critics.shape, (
+            #    "Mismatched shapes {} vs. {}".format(discrete_act_dist.probs.shape,
+            #                                         critics.shape))
```
emailweixu: remove debug code

alf/algorithms/dynamic_action_repeat_agent.py
```diff
@@ -0,0 +1,260 @@
...
+            config=config,
+            optimizer=optimizer,
+            debug_summaries=debug_summaries,
+            name=name)
+
+        self._repr_learner = repr_learner
+        self._rl = rl
+        self._original_observe_for_replay = self.observe_for_replay
+        # Do not observe data at every time step; customized observing
+        self.observe_for_replay = lambda exp: None
```
emailweixu: I feel it's better to use usual way of ovirriding:
```python
def observe_for_replay(self, exp):
  pass
```
The original observe_for_replay can be accessed using `super()._observe_for_replay()`

alf/algorithms/rl_algorithm.py
```diff
@@ -187,6 +187,17 @@ def action_spec(self):
         """Return the action spec."""
         return self._action_spec
 
+    @property
+    def train_action_spec(self):
```
emailweixu: We should add an interface only if we cannot do it reasonable in a different way.
In this case, I feel it's better to do the logging by the customized algorithm instead of add a new interface function.

alf/algorithms/dynamic_action_repeat_agent.py
```diff
@@ -0,0 +1,267 @@
...
+from alf.algorithms.sac_algorithm import SacAlgorithm
+from alf.algorithms.config import TrainerConfig
+from alf.data_structures import TimeStep, Experience, namedtuple, AlgStep
+from alf.data_structures import make_experience, LossInfo
+from alf.tensor_specs import BoundedTensorSpec, TensorSpec
+from alf.utils.conditional_ops import conditional_update
+from alf.utils import common
+
+ActionRepeatState = namedtuple(
+    "PeriodicActionState",
```
Haichao-Zhang: Maybe better to follow the convention to have the same name here, either ``PeriodicActionState`` or ``ActionRepeatState``?

alf/algorithms/sac_algorithm.py
```diff
@@ -543,45 +578,50 @@ def actor_loss_fn(dqda, action):
...
-
-        target_critic = target_critics.min(dim=1)[0].reshape(log_pi.shape) - \
-                         (torch.exp(self._log_alpha) * log_pi)
+            discrete_act_dist = action_distribution[0]
+            target_critics = torch.sum(
+                discrete_act_dist.probs * target_critics, dim=-1)
 
-        target_critic = target_critic.detach()
+        target_critic = target_critics.reshape(exp.reward.shape)
+        entropy_reward = nest.map_structure(lambda la, lp: torch.exp(la) * lp,
```
Haichao-Zhang: The name ``entropy_reward`` might be mis-leading. Might be better to move the negative sign from line 624 to here?

alf/examples/sac_half_cheetah.gin
```diff
@@ -0,0 +1,3 @@
+include 'sac_humanoid.gin'
+
```
Haichao-Zhang: There is ``sarsa_half_cheetah.gin`` for reproducing sac performance.

alf/algorithms/rl_algorithm.py
hnyu: I think the issue is that summarize_train is always called. And it will assume the action spec is the original action spec in https://github.com/HorizonRobotics/alf/blob/pytorch/alf/algorithms/rl_algorithm.py#L262 , which causes a crash. The only way to do this is to replace experience.action with the actual action before this. But the current framework only retrieves train_info from train_step and provides no way to change the action.

alf/algorithms/dynamic_action_repeat_agent.py
```diff
@@ -253,6 +256,10 @@ def train_step(self, rl_exp: Experience, state: ActionRepeatState):
 
     def calc_loss(self, rl_experience, rl_info):
         """Calculate the loss for training ``self._rl``."""
+        with alf.summary.scope(self._name):
+            # summarize_reward will check ``_debug_summaries`` and
+            # ``should_record_summaries()``
+            self.summarize_reward("training_reward", rl_experience.reward)
```
emailweixu: Why don't you want to check  ``_debug_summaries`` and ``should_record_summaries()``?
Even you don't do it here, ``should_record_summaries()`` will always be checked in summary_ops.py

alf/algorithms/rl_algorithm.py
```diff
@@ -272,9 +261,7 @@ def summarize_train(self, experience, train_info, loss_info, params):
 
         if self._debug_summaries:
             summary_utils.summarize_action(experience.action,
-                                           self.train_action_spec)
-            if not self.is_on_policy():
-                self.summarize_reward("training_reward", experience.reward)
+                                           self._action_spec)
```
emailweixu, line 264: You simply changed to self._action_spec from self.train_action_spec. Is the spec still correct for experience.action?

alf/algorithms/dynamic_action_repeat_agent.py
hnyu: summarize_reward already does the check. So there is no need to add more code here.

alf/algorithms/rl_algorithm.py
hnyu: There is no easy solution here because we can't change experience.action.

alf/algorithms/sac_algorithm.py
```diff
@@ -620,8 +620,8 @@ def _critic_train_step(self, exp: Experience, state: SacCriticState,
         target_critic = target_critics.reshape(exp.reward.shape)
         entropy_reward = nest.map_structure(lambda la, lp: torch.exp(la) * lp,
                                             self._log_alpha, log_pi)
-        entropy_reward = sum(nest.flatten(entropy_reward))
-        target_critic = (target_critic - entropy_reward).detach()
+        entropy_reward = -sum(nest.flatten(entropy_reward))
```
Haichao-Zhang: It might be better to move the negative sign up one more step into the lambda function in line 621,
Or can change the name of ``entropy_reward`` in Line 621 ?

Haichao-Zhang (approved):
I have no further comments

PR #683, opened 2020-09-17
Support multidimensional reward for SAC

emailweixu:


alf/algorithms/sac_algorithm.py
```diff
@@ -230,10 +236,27 @@ def __init__(self,
...
+        if reward_dim > 1:
+            assert not use_entropy_reward, (
+                "use_entropy_reward=True is not supported for multidimensional reward"
+            )
+            assert self._act_type == ActionType.Continuous, (
+                "Only continuous action is supported for multidimensional reward"
+            )
+
+        self._reward_weights = None
+        if reward_weights:
```
hnyu, line 255: should assert len(reward_weights) == reward_dim

alf/algorithms/sac_algorithm.py
```diff
@@ -182,6 +184,10 @@ def __init__(self,
             q_network (Callable): is used to construct QNetwork for estimating ``Q(s,a)``
                 given that the action is discrete. Its output spec must be consistent with
                 the discrete action in ``action_spec``.
+            reward_weights (list[float]): this is only used when the reward is
```
Haichao-Zhang: reward_weights (None|list[float])

alf/algorithms/sac_algorithm.py
```diff
@@ -182,6 +184,10 @@ def __init__(self,
...
+                multidimensional. In that case, the weighted sum of the q values
+                is used for training the actor.
```
Haichao-Zhang: In that case, the weighted sum of the q values is used for training the actor if reward_weights is not None.
Otherwise, the sum of the q values is used.

hnyu (approved):

Haichao-Zhang (approved):

PR #684, opened 2020-09-18
CEM planner for MBRL

Haichao-Zhang:


alf/algorithms/planning_algorithm.py
```diff
@@ -269,3 +275,107 @@ def _calc_cost_for_action_sequence(self, time_step: TimeStep, state,
 
     def after_update(self, training_info):
         pass
+
+
+@gin.configurable
+class CEMPlanAlgorithm(RandomShootingAlgorithm):
+    """CEM-based planning method.
```
emailweixu, line 241: Add a reference and/or briefly describe the algorithm.

alf/optimizers/traj_optimizers.py
```diff
@@ -89,3 +91,115 @@ def obtain_solution(self, time_step: TimeStep, state):
         # solutions [B, pop_size, sol_dim] -> [B, sol_dim]
         solution = solutions[(torch.arange(batch_size), min_ind)]
         return solution
+
+
+class CEMOptimizer(TrajOptimizer):
```
emailweixu, line 83: Describe what this optimizer do.

alf/optimizers/traj_optimizers.py
```diff
@@ -89,3 +91,115 @@ def obtain_solution(self, time_step: TimeStep, state):
...
+
+        i = 0
+        pop_mean = init_mean
+        pop_var = init_var
+
+        # [B, population, solution_dim]
+        trunc_normal_samples = torch.zeros(batch_size, self._population_size,
+                                           self._solution_dim)
+
+        while i < self._max_iter_num and pop_var.view(
```
emailweixu: `pop_var.view(-1).max()` => `pop_var.max()`

alf/algorithms/planning_algorithm.py
emailweixu, line 241: need to note that this is not thread-safe due to the use of self._prev_solution

alf/algorithms/planning_algorithm.py
```diff
@@ -269,3 +275,107 @@ def _calc_cost_for_action_sequence(self, time_step: TimeStep, state,
...
+                time_step.step_type == StepType.FIRST)
+
+        self._plan_optimizer.set_cost(self._calc_cost_for_action_sequence)
+
+        init_mean = self._prev_solution.unsqueeze(1)
+        opt_action = self._plan_optimizer.obtain_solution(
+            time_step,
+            state,
+            init_mean=init_mean,
+            init_var=torch.ones_like(init_mean) * 1)
```
emailweixu: This seems to assume that the actions are in certain range? Should make this proportional to the action bound.

alf/optimizers/traj_optimizers.py
```diff
@@ -89,3 +91,115 @@ def obtain_solution(self, time_step: TimeStep, state):
...
+        batch_size = init_obs.shape[0]
+
+        if init_mean is None:
+            # [B, 1, solution_dim]
+            init_mean = torch.zeros(batch_size, 1, self._solution_dim)
+        else:
+            assert init_mean.shape == (batch_size, 1, self._solution_dim)
+
+        if init_var is None:
+            init_var = torch.ones(batch_size, 1, self._solution_dim)
```
emailweixu: Instead of being fixed, it should be derived from `upper_bound` and `lower_bound`

alf/optimizers/traj_optimizers.py
```diff
@@ -89,3 +91,115 @@ def obtain_solution(self, time_step: TimeStep, state):
...
+                trunc_normal_samples)
+            samples = trunc_normal_samples * torch.sqrt(
+                constrained_var) + pop_mean
+
+            costs = self.cost_function(time_step, state, samples.clone())
+
+            # select elite set from the population
+            ind = torch.topk(-costs, self._elite_size)[1]
+            # samples: [batch, population, horizon]
+            elites = samples.index_select(1, ind.view(-1))
```
emailweixu: This is not correct when batch_size is not 1.
```
elites = samples[torch.arange(batch_size).unsqueeze(-1), ind]
```

alf/optimizers/traj_optimizers.py
```diff
@@ -89,3 +91,115 @@ def obtain_solution(self, time_step: TimeStep, state):
...
+                 min_var=1e-5):
+        """Creates a CEM Optimizer
+
+        Args:
+            solution_dim (int): the dimensionality of the problem space
+            population_size (int): the number of candidate solutions to be
+                sampled at every iteration
+            upper_bound (float|Tensor): upper bounds for elements in solution
+            lower_bound (float|Tensor): lower bounds for elements in solution
+            elite_size (int): the number of elites selected in each round
```
emailweixu: Need to explain how elites are used.

alf/algorithms/planning_algorithm.py
```diff
@@ -269,3 +275,107 @@ def _calc_cost_for_action_sequence(self, time_step: TimeStep, state,
...
+
+        # [B, pop_size=1, horizon * action_dim] -> [B, horizon, action_dim]
+        opt_action = torch.reshape(
+            opt_action,
+            [opt_action.shape[0], self._planning_horizon, self._num_actions])
+
+        # [B, horizon, action_dim]
+        self._prev_solution = torch.cat(
+            (opt_action[:, 1:], torch.zeros_like(opt_action[:, 0:1])), 1)
```
emailweixu: 0 might be out of action bound. Perhaps use mean? `opt_action.mean(dim=(0, 1), keepdim=True).expand(batch_size, -1, -1)`

alf/optimizers/traj_optimizers.py
```diff
@@ -89,3 +91,115 @@ def obtain_solution(self, time_step: TimeStep, state):
...
+            trunc_normal_samples = alf.initializers.trunc_normal_(
+                trunc_normal_samples)
+            samples = trunc_normal_samples * torch.sqrt(
+                constrained_var) + pop_mean
+
+            costs = self.cost_function(time_step, state, samples.clone())
+
+            # select elite set from the population
+            ind = torch.topk(-costs, self._elite_size)[1]
+            # samples: [batch, population, horizon]
```
emailweixu: [batch, population, solution_dim]

alf/optimizers/traj_optimizers.py
```diff
@@ -89,3 +91,115 @@ def obtain_solution(self, time_step: TimeStep, state):
...
+            distance_to_ub = self._upper_bound - pop_mean
+
+            # compute the constrained var based on the computed distance
+            constrained_var = torch.min(
+                torch.min((distance_to_lb / 2.0)**2, (distance_to_ub / 2.0)
+                          **2), pop_var)
+            constrained_var = constrained_var.clamp(min=self._min_var)
+
+            trunc_normal_samples = alf.initializers.trunc_normal_(
+                trunc_normal_samples)
```
emailweixu: It's better to modify `trunc_normal_()` to handle vector form of mean, std, a, b.

emailweixu (changes requested):

alf/algorithms/planning_algorithm.py
Haichao-Zhang: Added comments and reference

alf/optimizers/traj_optimizers.py
Haichao-Zhang: Added description and reference.

alf/optimizers/traj_optimizers.py
Haichao-Zhang: changed

alf/algorithms/planning_algorithm.py
Haichao-Zhang: Changed to adaptive to action bound.

alf/optimizers/traj_optimizers.py
Haichao-Zhang: Changed to be adaptive to action bound.

alf/optimizers/traj_optimizers.py
Haichao-Zhang: Corrected.

alf/optimizers/traj_optimizers.py
Haichao-Zhang: Added explanation.

alf/algorithms/planning_algorithm.py
Haichao-Zhang: Switched to use mean.

alf/optimizers/traj_optimizers.py
Haichao-Zhang: Updated.

alf/optimizers/traj_optimizers.py
Haichao-Zhang: Leave ```trunc_normal_``` unchanged for now as it is a pytorch function (copied from a higher version). Will think about proper ways to deal with this issue.

alf/optimizers/traj_optimizers.py
```diff
@@ -89,3 +91,133 @@ def obtain_solution(self, time_step: TimeStep, state):
...
+
+        while i < self._max_iter_num and pop_var.max() > self._epsilon:
+            # compute the distance to lower and upper bound
+            distance_to_lb = pop_mean - self._lower_bound
+            distance_to_ub = self._upper_bound - pop_mean
+
+            # compute the constrained var based on the computed distance
+            constrained_var = torch.min(
+                torch.min((distance_to_lb / 2.0)**2, (distance_to_ub / 2.0)
+                          **2), pop_var)
```
emailweixu: This will make the bound at least 2 standard deviation away from the mean. It can make the convergence slow if the desired solution is close the bound.

alf/algorithms/planning_algorithm.py
```diff
@@ -269,3 +282,126 @@ def _calc_cost_for_action_sequence(self, time_step: TimeStep, state,
...
+    trajectory by minimizing a given cost function. The optimized action
+    trajectory is termed as a 'plan' which can be used by other components
+    such as a MPC-based controller. This has been used by some MBRL works
+    such as `Deep Reinforcement Learning in a Handful of Trials using
+    Probabilistic Dynamics Models <https://arxiv.org/abs/1805.12114>`_
+
+    To speedup, when possible, we have used the plan obtained at the previous
+    time step to initialize the the mean of the plan distribution at the current
+    time step, after proper shifting and padding. This, however, makes the
+    current implementation no longer thread-safe.
```
emailweixu: Perhaps we can use state to handle this to make it thread-safe

Haichao-Zhang: Good point. Will try to switch to state.

alf/optimizers/traj_optimizers.py
Haichao-Zhang: Changed to using Normal as generation distribution. Performs similar to TruncNormal but computation speed is much faster.

alf/algorithms/planning_algorithm.py
Haichao-Zhang: Now switched to use state for passing the previous plan.

alf/algorithms/planning_algorithm.py
```diff
@@ -269,3 +283,116 @@ def _calc_cost_for_action_sequence(self, time_step: TimeStep, state,
...
+                                               "before planning")
+
+        assert self._dynamics_func is not None, ("specify dynamics function "
+                                                 "before planning")
+
+        batch_size = time_step.observation.shape[0]
+
+        self._plan_optimizer.set_cost(self._calc_cost_for_action_sequence)
+
+        prev_plan = state.planner.prev_plan
```
emailweixu: one problem of this is that the first prev_plan is all zero. Will get a bad init_mean from it.

Haichao-Zhang: Now add back the self-defined solution resetting which is more adaptive to solution bounds.

alf/algorithms/planning_algorithm.py
```diff
@@ -269,3 +283,130 @@ def _calc_cost_for_action_sequence(self, time_step: TimeStep, state,
...
+        # [B, horizon, action_dim] -> [B, horizon*action_dim]
+        prev_solution = prev_plan.reshape(prev_plan.shape[0], -1)
+
+        if self._default_solution is None:
+            self._default_solution = torch.ones(batch_size, self._solution_size) \
+                    * (self._upper_bound + self._lower_bound) / 2.
+
+        # reset to default solution at the beginning of an episode
+        prev_solution = common.reset_state_if_necessary(
```
emailweixu: The following is thread-safe:
```python
prev_solution = prev_solution.clone()
prev_solution[time_step.step_type==StepType.FIRST] = torch.full(self._solution_size, (self._upper_bound + self._lower_bound) / 2.)
```

alf/algorithms/mbrl_algorithm.py
```diff
@@ -174,8 +174,8 @@ def _calc_step_reward(self, obs, action, state: MbrlState):
     def _predict_with_planning(self, time_step: TimeStep, state,
                                epsilon_greedy):
         # full state in
-        action = self._planner_module.generate_plan(time_step, state,
-                                                    epsilon_greedy)
+        action, state = self._planner_module.generate_plan(
+            time_step, state, epsilon_greedy)
```
emailweixu: pass in  `state.planner` instead as planner_module should not know about the structure of `state` 
and generate_plan => predict_plan

Haichao-Zhang:
> Should pass `state.planner` to `generate_plan()` instead of `state`.
> Also feel it is better to change generate_plan to predict_step to be consistent with other places.

In the current design, ```generate_plan``` takes full ```MbrlState``` as input state. The reason is that
the cost function to be optimized over in the planner requires state information beyond `state.planner` in order to be calculated. For example, the  `state.dynamics` is needed in order to compute the next predicted observation.

Changed the name of function `generate_plan` to `predict_plan `.

alf/algorithms/mbrl_algorithm.py
Haichao-Zhang: I agree that planner_module should not know about the structure of state. Currently there are still some flaws in the interface design.

Haichao-Zhang: Now changed name to predict_plan

alf/algorithms/planning_algorithm.py
Haichao-Zhang: Corrected

alf/algorithms/mbrl_algorithm.py
```diff
@@ -174,8 +174,8 @@ def _calc_step_reward(self, obs, action, state: MbrlState):
     def _predict_with_planning(self, time_step: TimeStep, state,
                                epsilon_greedy):
         # full state in
-        action = self._planner_module.generate_plan(time_step, state,
-                                                    epsilon_greedy)
+        action, state = self._planner_module.predict_plan(
+            time_step, state, epsilon_greedy)
```
emailweixu: For the planner module, `state.planner` is different from `state.dynamics` or `state.reward`. `state.dynamics` and `state.reward` are more like an input. While `state.planner` is indeed a state.

Haichao-Zhang: I have refactored the APIs for resolving the issue.

alf/algorithms/planning_algorithm.py
```diff
@@ -107,24 +108,18 @@ def train_step(self, time_step: TimeStep, state: PlannerState):
...
-        """Set the dynamics function for planning
-        Args:
-            dynamics_func (Callable): reward function to be used for planning.
-            dynamics_func takes (time_step, state) as input and returns
-            next_time_step (TimeStep) and the next_state
-        """
-        self._dynamics_func = dynamics_func
-
-    def generate_plan(self, time_step: TimeStep, state, epsilon_greedy):
+    def predict_plan(self, time_step: TimeStep, state, epsilon_greedy):
```
emailweixu: comment for `state` should be updated.

Haichao-Zhang: Comments updated.

emailweixu (approved):

PR #685, opened 2020-09-22
fix  collisionsensor

hnyu:
For CollisionSensor and RadarSensor, get_current_observation will clear stored data and cache the most recent one. This commit is independent of the branch carla_working

Will update the result of the testing run later.


hnyu:
The test run so far looks: 

image

This is with the PR code before the lock was added this morning.

emailweixu (approved):

PR #686, opened 2020-09-23
Play with vector reward

Haichao-Zhang:


alf/trainers/policy_trainer.py
```diff
@@ -668,6 +668,12 @@ def play(root_dir,
             env.render(mode='human')
             time.sleep(sleep_time_per_step)
 
+        if time_step.reward.numel() > 1:
+            # use sum of individual rewards when a vector reward is provided
+            time_step_reward = time_step.reward.sum()
+        else:
+            time_step_reward = time_step.reward
```
emailweixu: `time_step_reward` is not used.
I suggest still use a vector to summarize the episode reward since different dimension have different meaning.

Haichao-Zhang: Now keep all elements of the reward if it's a vector.

emailweixu (approved):

PR #687, opened 2020-09-24
vector reward for carla

emailweixu:
Also
1. fix AverageReturnMetric for vector reward
2. do not create unwrapped_env if not necessary to reduce resource usage
3. use batch normalization at the end of each preprocessor


alf/environments/suite_carla.py
```diff
@@ -460,35 +537,49 @@ def get_current_time_step(self, current_frame):
...
                     self._sparse_reward_index_interval,
                     self._navigation.num_waypoints - 1)
         else:
             goal0 = obs['navigation'][2]  # This is about 10m ahead
-            reward += (np.linalg.norm(prev_loc - goal0) -
-                       np.linalg.norm(curr_loc - goal0))
+            distance_reward = (np.linalg.norm(prev_loc - goal0) -
+                               np.linalg.norm(curr_loc - goal0))
+
+        reward_vector[Player.REWARD_DISTANCE] = distance_reward
```
le-horizon, line 562: shouldn't this happen after the if statement?

alf/environments/suite_carla.py
```diff
@@ -555,15 +652,16 @@ def render(self, mode):
...
+                (3.6 * obs['velocity']).tolist()),
             'Acceleration: (%4.1f, %4.1f, %4.1f)' % tuple(
                 obs['imu'][0:3].tolist()),
             'Compass: %5.1f' % math.degrees(float(obs['imu'][6])),
             'Throttle: %4.2f' % self._control.throttle,
             'Brake:    %4.2f' % self._control.brake,
             'Steer:    %4.2f' % self._control.steer,
             'Reverse:  %4s' % self._control.reverse,
-            'Reward: %.1f' % float(self._current_time_step.reward),
+            'Reward: %.1f' % float(self._current_time_step.reward.sum()),
```
Haichao-Zhang: Maybe better to display the full vector, which will take up more space on the rendered image.
Or we can simply show the overall reward?

alf/metrics/metrics.py
```diff
@@ -237,23 +237,34 @@ class AverageReturnMetric(AverageEpisodicSumMetric):
     def __init__(self,
                  name='AverageReturn',
                  prefix='Metrics',
+                 reward_shape=(),
                  dtype=torch.float32,
+                 reward_weights=None,
```
le-horizon: reward_weights not used in this class?

le-horizon (commented):
Almost good.  Two quick comments.

Thanks,
Le

alf/environments/suite_carla.py
emailweixu: I intentionally put it before the if statement because the unmodified distance reward is easier to predict.

alf/metrics/metrics.py
emailweixu: removed.

alf/environments/suite_carla.py
emailweixu: done

Haichao-Zhang (approved):

le-horizon (approved):
LG.  For reward, the full reward no longer equals the sum of individual ones?  It's probably just a small difference, so fine to me.

emailweixu:
> LG. For reward, the full reward no longer equals the sum of individual ones? It's probably just a small difference, so fine to me.

Right, And this depends on the environment.

PR #689, opened 2020-09-25
PredictiveRepresentationLearner

emailweixu:


alf/algorithms/predictive_representation_learner.py
```diff
@@ -0,0 +1,360 @@
...
+import gin
+import torch
+
+import alf
+from alf.algorithms.algorithm import Algorithm
+from alf.data_structures import AlgStep, TimeStep, Experience, LossInfo, namedtuple
+from alf.experience_replayers.replay_buffer import BatchInfo, ReplayBuffer
+from alf.nest.utils import convert_device
+from alf.networks import Network, LSTMEncodingNetwork
+from alf.utils.losses import element_wise_squared_loss, element_wise_huber_loss
```
Haichao-Zhang: ```element_wise_squared_loss``` and ```huber_loss``` are not used

alf/algorithms/predictive_representation_learner.py
```diff
@@ -0,0 +1,360 @@
...
+
+@gin.configurable
+class PredictiveRepresentationLearner(Algorithm):
+    """Learn representation based on the prediction of future values.
+
+    ``PredictiveRepresentationLearner`` contains 3 ``Network``s:
+    * encoding_net: it is used to encode the raw observation to a latent vector.
+    * dynamics_net: it is used to generate the future latent states from the
+        current latent state.
+    * decoder: it is used to decode the target values and calcuate the loss.
```
Haichao-Zhang: Might be better to make it clear here that decoder is not purely a ``Network`` but an ``Algorithm``.

alf/algorithms/predictive_representation_learner.py
emailweixu: removed

alf/algorithms/predictive_representation_learner.py
emailweixu: updated

Haichao-Zhang (approved):

PR #690, opened 2020-09-25
Option to not substract mean for normalizer and ObservationNormalizer

emailweixu:


alf/utils/normalizers.py
```diff
@@ -107,19 +113,26 @@ def _summary(name, val):
                         alf.summary.scalar(name + ".min." + suffix, val.min())
                         alf.summary.scalar(name + ".max." + suffix, val.max())
 
-            def _summarize_all(t, m, m2, path):
+            def _summarize_all(t, m2, path, m=None):
                 if path:
```
hnyu, line 117: I believe that 'path' will always be the last argument?

alf/utils/normalizers.py
```diff
@@ -107,19 +113,26 @@ def _summary(name, val):
...
+                if m is not None:
+                    _summary(path + "mean", m)
+                    _summary(path + "var", m2 - math_ops.square(m))
+                else:
+                    _summary(path + "var", m2)
+
+            if self._mean_averager:
+                alf.nest.py_map_structure_with_path(_summarize_all, tensor,
+                                                    self._mean_averager.get(),
+                                                    self._m2_averager.get())
```
hnyu: mean and m2 order is not correct

hnyu (changes requested):

alf/utils/normalizers.py
le-horizon: This is a bit messy, maybe we should have py_map_structure_with_path take a function that has "path" as the first argument:
 _summarize_all(path, t, m2, m=None)

alf/utils/normalizers.py
```diff
@@ -137,18 +150,27 @@ def normalize(self, tensor, clip_value=-1.0):
...
             if clip_value > 0:
                 t = torch.clamp(t, -clip_value, clip_value)
             return t
 
-        return alf.nest.map_structure(_normalize, self._mean_averager.get(),
-                                      self._m2_averager.get(), tensor)
+        if self._mean_averager:
+            return alf.nest.map_structure(_normalize,
+                                          self._mean_averager.get(),
+                                          self._m2_averager.get(), tensor)
```
le-horizon: wrong order here?  Maybe have (t, m2, m=None)?

le-horizon (commented):
Almost good.

alf/utils/normalizers.py
emailweixu: good point.

alf/utils/normalizers.py
emailweixu: fixed

alf/utils/normalizers.py
```diff
@@ -60,6 +61,7 @@ def __init__(self,
             auto_update (bool): If True, automatically update mean and variance
               for each call to `normalize()`. Otherwise, the user needs to call
               `update()`
+            zero_mean (bool): whether to make the normalized value be zero-meane
```
hnyu: zero-mean; should comment that if not zero-mean, then the values are divided by m2 instead of var

alf/utils/normalizers.py
```diff
@@ -107,19 +113,26 @@ def _summary(name, val):
...
-
-            alf.nest.py_map_structure_with_path(_summarize_all, tensor,
-                                                self._mean_averager.get(),
-                                                self._m2_averager.get())
+                         _reduce_along_batch_dims(t, m2, torch.max))
+                if m is not None:
+                    _summary(path + "mean", m)
+                    _summary(path + "var", m2 - math_ops.square(m))
+                else:
+                    _summary(path + "var", m2)
```
hnyu: This quantity is no longer variance. Second moment should be better

alf/utils/normalizers.py
```diff
@@ -169,13 +191,15 @@ def __init__(self,
             auto_update (bool): If True, automatically update mean and variance
               for each call to `normalize()`. Otherwise, the user needs to call
               `update()`
+            zero_mean (bool): whether to make the normalized value be zero-meane
```
hnyu: zero-mean

alf/utils/normalizers.py
```diff
@@ -218,13 +245,15 @@ def __init__(self,
...
```
hnyu: zero-mean

alf/utils/normalizers.py
```diff
@@ -269,6 +301,7 @@ def __init__(self,
...
```
hnyu: zero-mean

hnyu (changes requested):

le-horizon (approved):
LG, thanks.

hnyu (approved):

PR #691, opened 2020-09-25
carla single reward wrapper

hnyu:


alf/environments/suite_carla.py
```diff
@@ -1187,6 +1188,35 @@ def _reset(self):
         return self._get_current_time_step()
 
 
+@gin.configurable
+class CarlaSingleRewardWrapper(AlfEnvironmentBaseWrapper):
```
emailweixu: Suggest to change it to "ScalarRewardWrapper", which accept a weight vector and output the weighted reward.

hnyu: updated

alf/environments/suite_carla.py
```diff
@@ -1187,6 +1188,43 @@ def _reset(self):
...
 
 
+@gin.configurable
+class CarlaScalarRewardWrapper(AlfEnvironmentBaseWrapper):
+    """A wrapper that averages all rewards by a weight vector."""
+
+    def __init__(self, env, reward_weights=None):
+        """
+        Args:
+            env (CarlaEnvironment): A CarlaEnvironment instance to wrap.
```
emailweixu: It doesn't have to a CarlaEnvironment

alf/environments/suite_carla.py
```diff
@@ -1187,6 +1188,43 @@ def _reset(self):
...
+            reward_weights (list[float] | tuple[float]): a list/tuple of weights
+                for the rewards; if None, then the overall reward will be used.
+        """
+        super(CarlaScalarRewardWrapper, self).__init__(env)
+        if reward_weights is None:
+            reward_weights = [0.] * Player.REWARD_DIMENSION
+            reward_weights[Player.REWARD_OVERALL] = 1.
+        assert (isinstance(reward_weights, (list, tuple))
+                and len(reward_weights) == Player.REWARD_DIMENSION)
```
emailweixu: you can use `env.reward_spec()` to get the spec for reward so it's not tied to CarlaEnvironment.

alf/environments/suite_carla.py
hnyu: If not a CarlaEnvironment, by default we don't know what the weight vector size is and which dim is the overall reward.

hnyu: > It doesn't have to a CarlaEnvironment

If not a CarlaEnvironment, by default we don't know which dim is the overall reward.

Or do you mean it can be a wrapper that wraps a CarlaEnvironment?

alf/environments/alf_wrappers.py
```diff
@@ -410,3 +410,47 @@ def _add_action(self, time_step):
...
+        rewards_n = reward_spec.shape[0]
+        if reward_weights is None:
+            reward_weights = [1.] + [0.] * (rewards_n - 1)
+        assert (isinstance(reward_weights, (list, tuple))
+                and len(reward_weights) == rewards_n)
+        self._reward_weights = torch.tensor(reward_weights)
+
+    def _step(self, action):
+        time_step = self._env._step(action)
+        reward = torch.sum(time_step.reward * self._reward_weights, dim=-1)
```
emailweixu: use torch.tensordot instead.
```python
torch.tensordot(time_step.reward, self._reward_weights, dims=1)
```

alf/environments/alf_wrappers.py
```diff
@@ -436,17 +436,20 @@ def __init__(self, env, reward_weights=None):
             reward_weights = [1.] + [0.] * (rewards_n - 1)
         assert (isinstance(reward_weights, (list, tuple))
                 and len(reward_weights) == rewards_n)
-        self._reward_weights = torch.tensor(reward_weights)
+        self._reward_weights = torch.tensor(reward_weights).unsqueeze(0)
```
emailweixu: no need to unsqueeze

emailweixu (approved):

PR #694, opened 2020-09-29
fix video recorder error and image size issue

hnyu:
Sometimes the pyplot fig won't have the same image aspect ratio. This will cause video frame size error.


emailweixu (approved):

le-horizon (commented):
LG.

Thanks,
Le

PR #696, opened 2020-09-30
Support LayerNorm for FC and Normalizer as layer

emailweixu:


alf/layers.py
```diff
@@ -309,6 +328,10 @@ def forward(self, inputs):
             y = inputs.matmul(self._weight.t())
             if self._use_bias:
                 y += self._bias
+        if self._use_ln:
```
hnyu, line 331: What if self._use_ln and self._use_bn are both True? Why it's the current order being applied?

alf/utils/normalizers.py
```diff
@@ -171,6 +173,12 @@ def _normalize(m2, t, m=None):
             return alf.nest.map_structure(_normalize, self._m2_averager.get(),
                                           tensor)
 
+    def forward(self, input):
+        if self.training:
```
hnyu, line 177: Will this self.training flag automatically be set?

hnyu (changes requested):
nn.LayerNorm is still different from ALF Normalizer because it uses mean and std from the current input batch, instead of historical data.

alf/layers.py
emailweixu: In general, they should not be applied at the same time.

alf/utils/normalizers.py
emailweixu: Yes.

emailweixu (commented):
Right.

> nn.LayerNorm is still different from ALF Normalizer because it uses mean and std from the current input batch, instead of historical data.

alf/layers.py
hnyu: So maybe add a check for exclusion?

emailweixu: But I guess if someone really want to use both, that's ok.

hnyu (approved):

PR #700, opened 2020-10-03
Latent space MBRL

Haichao-Zhang:
Training curve on pendulum:
mbrl_latent_pendulum


alf/algorithms/mbrl_algorithm.py
```diff
@@ -94,9 +96,12 @@ def __init__(self,
 
         """
         train_state_spec = MbrlState(
-            dynamics=dynamics_module.train_state_spec,
-            reward=reward_module.train_state_spec,
-            planner=planner_module.train_state_spec)
+            dynamics=dynamics_module.train_state_spec \
```
emailweixu: no need to have `\`

alf/algorithms/predictive_representation_learner.py
```diff
@@ -265,6 +266,44 @@ def rollout_step(self, time_step: TimeStep, state):
         latent, state = self._encoding_net(time_step.observation, state)
         return AlgStep(output=latent, state=state)
 
+    def predict_multi_step(self, init_rep, actions, state=None):
+        """Perform multi-step predictions based on the initial latent
+            representation and actions sequences.
+        Args:
```
emailweixu, line 277: Comment that the target of the current step is not predicted.

alf/algorithms/predictive_representation_learner.py
```diff
@@ -265,6 +266,44 @@ def rollout_step(self, time_step: TimeStep, state):
...
+        # do not store latent of the current time step, only predict future
+        sim_latents = []
+
+        if self._latent_to_dstate_fc is not None:
+            dstate = self._latent_to_dstate_fc(latent)
+            dstate = dstate.split(self._dynamics_state_dims, dim=1)
+            dstate = alf.nest.pack_sequence_as(self._dynamics_net.state_spec,
+                                               dstate)
+        else:
+            dstate = state
```
emailweixu: share code with line 315-323

alf/algorithms/predictive_representation_learner.py
```diff
@@ -265,6 +266,44 @@ def rollout_step(self, time_step: TimeStep, state):
...
+
+        for i in range(num_unroll_steps):
+            sim_latent, dstate = self._dynamics_net(actions[:, i, ...], dstate)
+            sim_latents.append(sim_latent)
+
+        sim_latent = torch.cat(sim_latents, dim=0)
+        prediction = self._decoder.train_step(sim_latent).info
```
emailweixu: Implement `predict_step()` for decoder and use it.

alf/algorithms/predictive_representation_learner.py
```diff
@@ -347,9 +386,13 @@ def preprocess_experience(self, experience: Experience):
             # [B, T, unroll_steps+1]
             positions = torch.min(positions, episode_end_positions)
 
-            # [B, T, unroll_steps+1]
+            # [B, T, unroll_steps+1, ...], for the scalar case, the shape is
+            # [B, T, unroll_steps+1, 1]
```
emailweixu: I believe the the shape is still [B, T, unroll_steps+1] for scalar case.

alf/algorithms/predictive_representation_learner.py
```diff
@@ -347,9 +386,13 @@ def preprocess_experience(self, experience: Experience):
...
             target = replay_buffer.get_field(self._target_fields, env_ids,
                                              positions)
+            # scalar target case
+            if target.dim() == 3:
+                target = target.unsqueeze(3)
```
emailweixu: Instead of modifying the shape here, I think the right way is to change `SimpleDecoder` to handle shapes different from vector.

alf/algorithms/mbrl_algorithm.py
```diff
@@ -275,3 +280,132 @@ def calc_loss(self, experience, training_info: MbrlInfo):
...
+
+    def _predict_with_planning(self, time_step: TimeStep, state,
+                               epsilon_greedy):
+        action, state = self._planner_module.predict_plan(
+            time_step, state, epsilon_greedy)
+
+        return AlgStep(output=action, state=state, info=MbrlInfo())
+
+    def train_step(self, exp: Experience, state: MbrlState):
+        return AlgStep(output=(), state=state, info=MbrlInfo())
```
emailweixu, line 411: you can return `AlgStep(output=(), state=state, info=LossInfo())` and omit `calc_loss()`

alf/algorithms/mbrl_algorithm.py
Haichao-Zhang: removed

alf/algorithms/predictive_representation_learner.py
Haichao-Zhang: Now made it consistent within PredictiveRepresentationLearner and always include the current step. Any special post processing such as removing the current step is left for the caller.

alf/algorithms/predictive_representation_learner.py
Haichao-Zhang: Refactored with a function ``_multi_step_latent_rollout``.

alf/algorithms/predictive_representation_learner.py
Haichao-Zhang: Done

alf/algorithms/predictive_representation_learner.py
Haichao-Zhang: My comment was incorrect. I intended to use ``[B, T, unroll_steps+1, 1]`` do denote the desired shape.

alf/algorithms/predictive_representation_learner.py
Haichao-Zhang: Now handling it in decoder.

alf/algorithms/mbrl_algorithm.py
Haichao-Zhang: The dummy ``train_step`` and ``calc_cost`` are used to overwrite those functions from the base ``MbrlAlgorithm``. Comments added.

alf/algorithms/predictive_representation_learner.py
```diff
@@ -266,67 +272,72 @@ def rollout_step(self, time_step: TimeStep, state):
...
-        else:
-            dstate = state
-
-        for i in range(num_unroll_steps):
-            sim_latent, dstate = self._dynamics_net(actions[:, i, ...], dstate)
-            sim_latents.append(sim_latent)
+        sim_latents = self._multi_step_latent_rollout(
+            init_latent, num_unroll_steps, actions, state)
 
         sim_latent = torch.cat(sim_latents, dim=0)
```
emailweixu: This cat() can also be put into _multi_step_latent_rollout()

alf/algorithms/predictive_representation_learner.py
```diff
@@ -386,13 +397,9 @@ def preprocess_experience(self, experience: Experience):
             # [B, T, unroll_steps+1]
             positions = torch.min(positions, episode_end_positions)
 
-            # [B, T, unroll_steps+1, ...], for the scalar case, the shape is
-            # [B, T, unroll_steps+1, 1]
+            # [B, T, unroll_steps+1]
```
emailweixu: `[B, T, unroll_steps+1, ...]`. The shape of target may not be a scalar.

alf/algorithms/predictive_representation_learner.py
```diff
@@ -97,6 +98,11 @@ def train_step(self, repr, state=()):
         return AlgStep(
             output=predicted_reward, state=state, info=predicted_reward)
 
+    def predict_step(self, repr, state=()):
+        predicted_reward = self._decoder_net(repr)[0]
```
emailweixu: predicted_target. Also please update line 97 and line 99

alf/algorithms/predictive_representation_learner.py
```diff
@@ -108,6 +114,7 @@ def calc_loss(self, target, predicted, mask=None):
         Returns:
             LossInfo
         """
+        target = target.view_as(predicted)
```
emailweixu: It's shape should be changed at predict_step and train_step so that the output shape matches the target shape.

alf/algorithms/predictive_representation_learner.py
Haichao-Zhang: Moved cat() into the function.

alf/algorithms/predictive_representation_learner.py
Haichao-Zhang: Updated

alf/algorithms/predictive_representation_learner.py
Haichao-Zhang: Changed

alf/algorithms/predictive_representation_learner.py
Haichao-Zhang: Now changing the shape at train_step and predict_step.

alf/algorithms/predictive_representation_learner.py
```diff
@@ -94,14 +94,22 @@ def get_target_fields(self):
...
     def train_step(self, repr, state=()):
-        predicted_reward = self._decoder_net(repr)[0]
+        predicted_target = self._decoder_net(repr)[0]
+        # when the predicted target is a scalar, remove the
+        # singleton dimension to match the shape of target
+        if predicted_target.ndim == 2 and predicted_target.shape[1] == 1:
+            predicted_target = predicted_target.squeeze(1)
         return AlgStep(
-            output=predicted_reward, state=state, info=predicted_reward)
+            output=predicted_target, state=state, info=predicted_target)
```
emailweixu, line 99: Sorry, I just realized that the actual shape of the target is not known here. So this still has problem if the rank of the target is greater than 1 or if the target is 1 dimensional vector. One dimensional vector may be quite common.  Perhaps we should just make sure the shape of decoder_net output is correct?

Haichao-Zhang: As the ```preprocess_experience``` has the convention that the ``target`` tensor won't have a singular last dimension, so here probably we can simply use ```predicted_target = predicted_target.squeeze(-1)```?

emailweixu: preprocess_experience does not change the shape of the target. target can be scalar, 1d vector, regular vector, matrix, etc.

Haichao-Zhang: Now use an optional ```output_tensor_spec``` for reshaping the network output.

alf/networks/encoding_networks.py
```diff
@@ -516,6 +517,15 @@ def __init__(self,
...
                 None.
+            output_tensor_spec (None|TensorSpec): spec for the output. If None,
+                the output tensor spec will be assumed as
+                ``TensorSpec((output_size, ))``, where ``output_size`` is
+                inferred from network output. Otherwise, the output tensor
+                spec will be ``output_tensor_spec`` and the network output
+                will be reshaped according to ``output_tensor_spec``.
+                Note that ``output_tensor_spec`` is only used for reshaping
+                the network outputs for interpretation purpose and is not used
+                for specifying any network layers.
```
emailweixu, line 528: There's some redundancy for setting both `output_tensor_spec` and `last_layer_size`. Do you think reasonable to allow the user not to set last_layer_size when output_tensor_spec is provided?

Haichao-Zhang: I agree the suggested way could reduce some redundancy.  It might be a good choice in some cases such as the CriticNetwork, but might not be as clear in more general cases, as there will be multiple ways to specify the same network structure, which might introduce some confusions.
Also, there might be cases where the net is specified without last layer but only want to re-interprete the output.

I think still use our original way of specifying the networks, and only resort to output_tensor_spec for interpreting the output when necessary might be less confusing. I might miss something but those are my current thought.

emailweixu (approved):

PR #701, opened 2020-10-05
add a script for curve plotting

hnyu:
Currently only support plotting one figure at at time. For two different MeanCurves, you can either align their x-axes or not. 

This is useful if you know that different x-axes are in fact equivalent and synced, regardless of their original values. For example, SAC+actrepeat vs. SAC
test1

This is plotting whatever the x-axes are
test2

### DIFF
--- a/alf/utils/plot_tb_curves.py
+++ b/alf/utils/plot_tb_curves.py
@@ -16,6 +16,8 @@ from tensorboard.backend.event_processing.event_accumulator import EventAccumula
 import numpy as np
 import os
 import glob
+from scipy.interpolate import interp1d
+from scipy.signal import savgol_filter
 
 import matplotlib
 import matplotlib.pyplot as plt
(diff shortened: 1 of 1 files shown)


alf/utils/plot_tb_curves.py
```diff
@@ -0,0 +1,240 @@
...
+
+
+class MeanCurveReader(object):
+    """Read and compute a MeanCurve from one or multiple TB event files.
+    """
+    _SIZE_GUIDANCE = {
+        'compressedHistograms': 10,
+        'images': 0,
+        'scalars':
+            100,  # sampled points will even distribute over the training time
```
le-horizon: "evenly"

alf/utils/plot_tb_curves.py
```diff
@@ -0,0 +1,240 @@
...
+        plt.close(self._fig)
+
+
+def _get_curve_path(dir):
+    return os.path.join(HOME, "tensorboard_curves", dir)
+
+
+if __name__ == "__main__":
+    """A plotting example."""
+    event_files = glob.glob(_get_curve_path("sac_kickball_gs/*/eval"))
```
le-horizon: comment that the event_files are usually multiple runs of the same configuration with different random seeds.

le-horizon (commented):
Nice tool!  LG, only couple nits.

le-horizon (approved):
LG!

PR #707, opened 2020-10-09
support reward normalization for DynamicActionRepeat

hnyu:


alf/algorithms/dynamic_action_repeat_agent.py
```diff
@@ -77,8 +80,12 @@ def __init__(self,
                 calculate the representation from the original observation as
                 the observation for downstream algorithms such as ``rl_algorithm``.
                 We assume that the representation is trained by ``rl_algorithm``.
+            reward_normalizer (SimpleDataTransformer): if not None, environment
```
emailweixu: reward_normalizer -> reward_normalizer_ctor

alf/algorithms/dynamic_action_repeat_agent.py
```diff
@@ -188,14 +202,36 @@ def _generate_new_action(time_step, state):
...
+            k=state.k + 1)
+
+        def _update_sample_rewards(time_step, state):
+            return state._replace(sample_rewards=time_step.reward)
+
+        if self._reward_normalizer is not None:
+            # The probability of a reward at step k being kept till K steps is:
+            # 1/k * k/(k+1) * .. * (K-1)/K = 1/K. This provides enough randomness
+            # to make the normalizer unbiased.
+            state = conditional_update(
```
emailweixu: Use torch.where, which is simpler for this case
```python
state = state._replace(sample_rewards=torch.where(
    torch.rand_like(state.sample_rewards) < 1. / state.k.to(torch.float32), 
    time_step.reward,
    state.sample_rewards)
```

alf/algorithms/dynamic_action_repeat_agent.py
```diff
@@ -188,14 +202,36 @@ def _generate_new_action(time_step, state):
 
     def rollout_step(self, time_step: TimeStep, state: ActionRepeatState):
         switch_action = self._should_switch_action(time_step, state)
+
+        # state.k is the current step index over K steps
         state = state._replace(
-            rl_reward=state.rl_reward + state.rl_discount * time_step.reward,
-            rl_discount=state.rl_discount * time_step.discount * self._gamma)
+            rl_reward=state.rl_reward + torch.pow(
```
emailweixu, line 208: `rl_reward=state.rl_reward + state.rl_discount * time_step.reward`

alf/algorithms/dynamic_action_repeat_agent.py
```diff
@@ -279,3 +319,39 @@ def summarize_train(self, experience, train_info, loss_info, params):
...
+            # compute current variance
+            m = normalizer._mean_averager.get()
+            m2 = normalizer._m2_averager.get()
+            var = torch.relu(m2 - m**2)
+
+            # compute accumulated mean over ``repeats`` steps
+            acc_mean = ((1 - torch.pow(self._gamma, repeats.to(torch.float32)))
+                        / (1 - self._gamma) * m)
+
+            reward -= acc_mean
```
emailweixu, line 340: why not do this using normalize_along_batch_dims?

alf/utils/tensor_utils.py
```diff
@@ -45,6 +45,18 @@ def reverse_cumsum(x, dim):
     return torch.flip(torch.cumsum(torch.flip(x, [dim]), dim), [dim])
 
 
+def reverse_cumprod(x, dim):
```
emailweixu, line 48: not used?

alf/examples/dyna_actrepeat_sac_bipedalwalker.gin
```diff
@@ -22,4 +21,6 @@ SacAlgorithm.target_entropy=(
 
 TrainerConfig.algorithm_ctor=@DynamicActionRepeatAgent
 TrainerConfig.debug_summaries=True
-TrainerConfig.unroll_length=2
+
+RewardNormalizer.clip_value=1.0
```
emailweixu: With 1.0, a lot of reward can be clipped.

emailweixu (changes requested):

alf/algorithms/dynamic_action_repeat_agent.py
hnyu: This will not necessarily be correct if time_step.discount is between 0 and 1 because rl_reward (t+1) already has one extra multiplied discount for training the previous time step (t).

alf/algorithms/dynamic_action_repeat_agent.py
hnyu: acc_mean has a batch size > 1 in this case, which is not allowed by normalize_along_batch_dims (spec check will fail between mean and var)

alf/utils/tensor_utils.py
hnyu: Just keep it there for future use

alf/algorithms/dynamic_action_repeat_agent.py
hnyu: If time_step.discount is betwee 0 and 1, I think this place will be somewhat inconsistent with line 342 because there acc_mean only considers gamma^k.

alf/examples/dyna_actrepeat_sac_bipedalwalker.gin
hnyu: changed to 5.0. The new curve looks similar to the current one.

emailweixu (approved):

PR #713, opened 2020-10-23
Play with defer mode

Haichao-Zhang:
Defer mode in play enables displaying information from the future in the current frame.
This enables easier perception of temporal pattern of different quantities (e.g. degree of smoothness of actions),
and also makes it easier to compare the predicted future quantities (e.g. reward) v.s. the ground-truth.

Some examples are shown as below:
defer_mode   pendulum    defer_mode_2 carla

### DIFF
--- a/alf/trainers/policy_trainer.py
+++ b/alf/trainers/policy_trainer.py
@@ -29,6 +29,7 @@ import alf
 from alf.algorithms.algorithm import Algorithm
 from alf.algorithms.config import TrainerConfig
 from alf.algorithms.data_transformer import create_data_transformer
+from alf.data_structures import StepType
 from alf.environments.utils import create_environment
 from alf.nest import map_structure
 from alf.tensor_specs import TensorSpec
@@ -634,8 +635,9 @@ def play(root_dir,
             instead of shown on the screen.
         future_steps (int): whether to encode some information from future steps
             into the current frame. If future_steps is larger than zero,
-            then the episodic information will be cached and the encoding of
-            them to video frames is deferred to the end of an episode.
+            then the related information (e.g. observation, reward, action etc.)
+            will be cached and the encoding of them to video frames is deferred
+            to the time when ``future_steps`` of future frames are available.
             This defer mode is potentially useful to display for each frame
             some information that expands beyond a single time step to the future.
             Currently this mode only support offline rendering, i.e. rendering
@@ -658,7 +660,8 @@ def play(root_dir,
 
     recorder = None
     if record_file is not None:
-        recorder = VideoRecorder(env, path=record_file)
+        recorder = VideoRecorder(
+            env, future_steps=future_steps, path=record_file)
     else:
         # pybullet_envs need to render() before reset() to enable mode='human'
         env.render(mode='human')
@@ -677,14 +680,6 @@ def play(root_dir,
         alf.metrics.AverageEpisodeLengthMetric(buffer_size=num_episodes),
     ]
 
-    if future_steps > 0:
-        defer_mode = True
-        observations = []
-        rewards = []
-        actions = []
-    else:
-        defer_mode = False
-
     while episodes < num_episodes:
 
         # transform time step
@@ -700,17 +695,17 @@ def play(root_dir,
             epsilon_greedy=epsilon_greedy,
             metrics=metrics)
         policy_state = policy_step.state
-        action = policy_step.output
-        info = policy_step.info
-
-        if defer_mode:
-            observations.append(transformed_time_step.observation)
-            rewards.append(transformed_time_step.reward)
-            actions.append(action)
 
         episode_length += 1
+
+        is_last_step = time_step.is_last() or (episode_length >=
+                                               max_episode_length > 0)
+
         if recorder:
-            recorder.capture_frame(info, encode_frame=not defer_mode)
+            recorder.capture_frame(
+                transformed_time_step._replace(step_type=StepType.LAST) \
+                        if is_last_step else transformed_time_step,
+                policy_step)
         else:
             env.render(mode='human')
             time.sleep(sleep_time_per_step)
@@ -719,34 +714,9 @@ def play(root_dir,
 
         episode_reward += time_step_reward
 
-        # start defer encoding
-        if defer_mode and (episode_length > future_steps):
-            _record_future_info(
-                recorder,
-                observations,
-                actions,
-                rewards,
-                future_steps=future_steps)
-            observations.pop(0)
-            actions.pop(0)
-            rewards.pop(0)
-
-        if time_step.is_last() or episode_length >= max_episode_length > 0:
+        if is_last_step:
             logging.info("episode_length=%s episode_reward=%s" %
                          (episode_length, episode_reward))
-
-            # defer encoding all the rest till the end
-            if defer_mode:
-                _record_future_info(
-                    recorder,
-                    observations,
-                    actions,
-                    rewards,
-                    future_steps=future_steps,
-                    encode_all=True)
-                observations = []
-                actions = []
-                rewards = []
 
             episode_reward = 0.
             episode_length = 0.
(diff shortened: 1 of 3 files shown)


alf/trainers/policy_trainer.py
```diff
@@ -465,7 +465,7 @@ def _eval(self):
             self._eval_env.batch_size)
         episodes = 0
         while episodes < self._num_eval_episodes:
-            time_step, policy_state, trans_state, _ = _step(
+            time_step, policy_state, trans_state, _, _ = _step(
```
hnyu: Can just let _step() return policy_step to reduce number of return arguments

alf/trainers/policy_trainer.py
```diff
@@ -662,18 +674,38 @@ def play(root_dir,
             buffer_size=num_episodes, reward_shape=env.reward_spec().shape),
         alf.metrics.AverageEpisodeLengthMetric(buffer_size=num_episodes),
     ]
+
+    if future_steps > 0:
+        defer_mode = True
+        observations = []
+        rewards = []
+        actions = []
```
hnyu: Storing all quantities seems memory inefficient if one episode is too long. I think an alternative is to render t-K:t at step t (looking back but pretending to look into the future). In this way, you only need to maintain a window of length K at most and can also support online rendering.

hnyu (changes requested):

alf/trainers/policy_trainer.py
le-horizon: It would be great to have this, probably can also simplify code a bit.

alf/trainers/policy_trainer.py
```diff
@@ -698,3 +741,101 @@ def play(root_dir,
...
+
+    for t in range(episode_length):
+        H = min(num_steps, episode_length - t - 1)  # future steps
+        if H > 0:
+            t_actions = actions[t:t + H]
+            t_rewards = rewards[t + 1:t + 1 + H]
+
+            if info_func is not None:
+                predictions = info_func(observations[t], t_actions)
+                assert predtions.ndim == 1 or predictions.shape[1] == 1, \
```
le-horizon: predtions->predictions

alf/utils/video_recorder.py
```diff
@@ -315,6 +364,35 @@ def _plot_value(self, value):
...
+                          name="value_curve",
+                          xticks=None):
+        """Generate the value curve for elements in values.
+        Args:
+            values (list[np.array]): each element from the list corresponding
+                to one curve in the generated figure.
+            legends (list[np.array]): name for each element from values
+            size (int): the size of the figure
+            name (str):
+            xticks:
```
le-horizon: missing comment?

le-horizon (commented):
Main point agrees with Haonan's.  A couple nits also.

Thanks for the improvement!
Le

alf/trainers/policy_trainer.py
Haichao-Zhang: done

alf/trainers/policy_trainer.py
Haichao-Zhang: done

alf/trainers/policy_trainer.py
Haichao-Zhang: corrected

alf/utils/video_recorder.py
Haichao-Zhang: comment added

alf/trainers/policy_trainer.py
```diff
@@ -714,16 +718,31 @@ def play(root_dir,
...
             if defer_mode:
-                _record_episodic_info(
+                _record_future_info(
                     recorder,
                     observations,
                     actions,
                     rewards,
-                    future_steps=future_steps)
+                    future_steps=future_steps,
+                    encode_all=True)
```
hnyu: Why do you need this encode_all?

alf/trainers/policy_trainer.py
```diff
@@ -773,14 +793,23 @@ def _record_episodic_info(recorder,
...
     rewards = torch.cat(rewards, dim=0)
-    episode_length = rewards.shape[0]
-    assert future_steps < episode_length, (
-        "future steps should be smaller than "
-        "the episode length {a}, but got {b}".format(
-            a=episode_length, b=future_steps))
+    episode_buffer_length = rewards.shape[0]
+
+    if not encode_all:
+        assert future_steps < episode_buffer_length, (
```
hnyu: This assertion is unnecessary?

alf/utils/video_recorder.py
```diff
@@ -315,6 +364,42 @@ def _plot_value(self, value):
         steps."""
         return
 
+    def _plot_value_curve(self,
```
hnyu: This seems duplicate with the existing _plot_prob_curve (although it doesn't have to be probability)

alf/utils/video_recorder.py
```diff
@@ -179,13 +184,57 @@ def capture_frame(self, pred_info=None):
...
             if self.ansi_mode:
                 self._encode_ansi_frame(frame)
             else:
                 self._encode_image_frame(frame)
+        # clear frame buffer after encoding all
+        self._frame_buffer = []
 
-            assert not self.broken, (
-                "The output file is broken! Check warning messages.")
+    def encode_frames_in_buffer_with_external(self, set_of_external_frames):
```
hnyu: I feel that the code is bloated by separating this from the original encoding code. I think potentially an external frame or whatever additional info can be first concatenated with the original frame and then call the encoder. BTW, encode_frames_in_buffer has never been called right now.

alf/trainers/policy_trainer.py
```diff
@@ -698,3 +760,108 @@ def play(root_dir,
     if recorder:
         recorder.close()
     env.reset()
+
+
+def _record_future_info(recorder,
```
hnyu: This function doesn't have to be separate from the existing. All the new stuff can be put into VideoRecorder class's capture_frame() (by letting it accept a policy_step) without bloating policy_trainer.py code

hnyu (changes requested):

hnyu:
The overall policy_trainer.py code is too bloated right now. I think most buffering and rendering logic should be in VideoRecorder instead.

alf/trainers/policy_trainer.py
Haichao-Zhang: Need to encode what's left in the episode buffer at the end fo an episode.

alf/trainers/policy_trainer.py
Haichao-Zhang: Removed

alf/utils/video_recorder.py
Haichao-Zhang: Merged

alf/utils/video_recorder.py
Haichao-Zhang: Refactored and moved into VideoRecorder.
Removed encode_frames_in_buffer

alf/trainers/policy_trainer.py
Haichao-Zhang: Refactored and moved into VideoRecorder.

alf/trainers/policy_trainer.py
```diff
@@ -663,18 +679,33 @@ def play(root_dir,
...
+
+        is_last_step = time_step.is_last() or (episode_length >=
+                                               max_episode_length > 0)
+
         if recorder:
-            recorder.capture_frame(info)
+            recorder.capture_frame(
+                transformed_time_step._replace(step_type=StepType.LAST) \
+                        if is_last_step else transformed_time_step,
+                policy_step)
```
le-horizon: is replacing LAST steptype needed for "human" rendering mode as well?

alf/utils/video_recorder.py
```diff
@@ -179,23 +221,199 @@ def capture_frame(self, pred_info=None):
...
+                    linewidth=5)
+                action_curve_set.append(action_curve)
+
+            self._observation_buffer.pop(0)
+            self._reward_buffer.pop(0)
+            self._action_buffer.pop(0)
+
+        # encode all frames
+        self._encode_frames_in_buffer_with_external(
+            [reward_curve_set, action_curve_set, predictive_curve_set])
```
le-horizon, line 420: Currently the code accumulates all frames and dump all of them at the end.  This will make later extending it to "human" playing mode difficult?  Maybe cache n frames, and render the current time - n frame at current time?  This would mean moving the accumulators to policy_trainer.play() and only feed the curves for one timestep to record or render.  I think this is what Haonan meant by "an external frame or whatever additional info can be first concatenated with the original frame and then call the encoder".  Does that make sense?

le-horizon (commented):
Hi Haichao and Haonan,

I think it's important to keep playing and recording behavior the same.  This way, recorded videos can be played in "human" mode for debugging if needed.  It's Ok to implement the change for recording mode at first, but we should leave the code extensible so it'll be easy to do it for the other human playing mode.

What do you think?

Thanks,
Le

alf/trainers/policy_trainer.py
hnyu: > Need to encode what's left in the episode buffer at the end fo an episode.

I think what's left in the buffer is always the length of future_steps. So it will always be encoded at the previous step.

alf/trainers/policy_trainer.py
```diff
@@ -663,18 +679,33 @@ def play(root_dir,
             buffer_size=num_episodes, reward_shape=env.reward_spec().shape),
         alf.metrics.AverageEpisodeLengthMetric(buffer_size=num_episodes),
     ]
+
     while episodes < num_episodes:
-        time_step, policy_state, trans_state, info = _step(
+
+        # transform time step
+        transformed_time_step, trans_state = algorithm.transform_timestep(
```
hnyu, line 686: _step will do transform_step inside. Do you really need transformed_time_step to be recorded?

alf/trainers/policy_trainer.py
```diff
@@ -663,18 +679,33 @@ def play(root_dir,
...
         episode_length += 1
+
+        is_last_step = time_step.is_last() or (episode_length >=
+                                               max_episode_length > 0)
+
         if recorder:
-            recorder.capture_frame(info)
+            recorder.capture_frame(
+                transformed_time_step._replace(step_type=StepType.LAST) \
+                        if is_last_step else transformed_time_step,
```
hnyu: Would be better to just pass in is_last_step as an argument to capture_frame instead of modifying the step_type (in a hacky way).

alf/utils/video_recorder.py
```diff
@@ -140,16 +152,46 @@ def __init__(self,
         if frames_per_sec is not None:
             self.frames_per_sec = frames_per_sec  # overwrite the base class
 
-    def capture_frame(self, pred_info=None):
+        self._future_steps = future_steps
+        self._frame_buffer = []
+        self._observation_buffer = []
+        self._reward_buffer = []
+        self._action_buffer = []
```
hnyu: Organizing these variables into a namedtuple would be better:
```
RecorderBuffers = namedtuple("RecorderBuffers", ["frame", "observation", "reward", "action"], default_value=[])

File: alf/utils/video_recorder.py [hunk 15]
```diff
@@ -179,23 +221,199 @@ def capture_frame(self, pred_info=None):
...
+
+        Args:
+            info_func (None|callable): a callable for calculating some customized
+                information (e.g. predicted future reward) based on the observation
+                at each time step and action sequences from the current time step
+                to the next ``future_steps`` steps (if available). It is called
+                as ``pred_info=info_func(current_observation, action_sequences)``.
+                Currently only support displaying scalar predictive information
+                returned from info_func.
+            encode_all (bool): whether to encode all the steps in the episode
```
hnyu, line 257: encode_all is not necessary because after self._future_steps < episode_buffer_length is satisfied, it will always be satisfied till the end.

alf/utils/video_recorder.py
Haichao-Zhang: The current code only keeps future_steps + 1 frames.

alf/utils/video_recorder.py
```diff
@@ -179,23 +221,199 @@ def capture_frame(self, pred_info=None):
...
+            self._observation_buffer.pop(0)
+            self._reward_buffer.pop(0)
+            self._action_buffer.pop(0)
+
+        # encode all frames
+        self._encode_frames_in_buffer_with_external(
+            [reward_curve_set, action_curve_set, predictive_curve_set])
+
+    def _encode_frames_in_buffer_with_external(self, set_of_external_frames):
+        """ Encode jointly internal and external frames
```
hnyu: The word "internal" and "external" are unintuitive. Better choices should be considered.

alf/utils/video_recorder.py
```diff
@@ -179,23 +221,199 @@ def capture_frame(self, pred_info=None):
...
+                        "only support displaying scalar predictive information"
+                    predictions = predictions.view(-1).detach().cpu().numpy()
+
+                    pred_curve = self._plot_value_curve(
+                        "prediction", [predictions],
+                        legends=["Prediction"],
+                        fig_size=6,
+                        height=300,
+                        width=300,
+                        linewidth=5)
```
hnyu: Maybe these common constant argument numbers should be variables for reuse.

hnyu (changes requested):

alf/trainers/policy_trainer.py
Haichao-Zhang: No. Actually ```human``` mode currently needs non-of the inputs to ```capture_frame```.
But it can also be extend.

alf/utils/video_recorder.py
Haichao-Zhang: Maybe there are some confusions. I will improve the code further to make it more clear.

alf/trainers/policy_trainer.py
le-horizon: Agree with Haonan, let's pass in is_last_step to capture_frame().

alf/utils/video_recorder.py
le-horizon: Ah, I see, this function has two modes.
Anyways, what I worry about is recording causing randomness to be different from human playing, and making reproducing behavior and debugging difficult.  I think it's best to keep the two as similar as possible.  You don't have to do the "human" mode in this PR, but the code should accommodate such a future change.

le-horizon (commented):
Two comments if they make sense.

alf/trainers/policy_trainer.py
Haichao-Zhang: Done

alf/utils/video_recorder.py
Haichao-Zhang: Improved

alf/utils/video_recorder.py
Haichao-Zhang: Updated with more descriptive names

alf/utils/video_recorder.py
Haichao-Zhang: Done.

alf/utils/video_recorder.py
Haichao-Zhang: Good point. I think the current code can be extend to support direct viewing on screen easily.
Might implement that in another PR.

alf/trainers/policy_trainer.py
Haichao-Zhang: Done

alf/utils/video_recorder.py
```diff
@@ -98,6 +98,65 @@ def _generate_img_matrix(imgs, cols, img_width):
...
+                the corresponding fields of the buffer.
+            n (int): the number of elements to pop
+        """
+        for field in common.as_list(fields):
+            del self._field_to_buffer_mapping[field][:n]
+
+
 @gin.configurable(
     whitelist=['img_plot_width', 'value_range', 'frames_per_sec'])
 class VideoRecorder(GymVideoRecorder):
```
hnyu, line 162: Maybe now the docstring of this class should be updated accordingly.

hnyu (approved):

le-horizon (approved):
LG, and thanks for the very useful change!

PR #716, opened 2020-10-28
fix metric logging error and support specified device in convert device

Haichao-Zhang:
1) fixed the error when logging a vector-valued metric
2) add an optional argument to the ``conver_device`` function for specifying target device
3) this is a PR created from a forked branch which tests CI support for external PRs at the same time

### DIFF
--- a/alf/nest/utils.py
+++ b/alf/nest/utils.py
@@ -192,8 +192,22 @@ def get_outer_rank(tensors, specs):
     return outer_rank
 
 
-def convert_device(nests):
-    """Convert the device of the tensors in nests to default device."""
+def convert_device(nests, device=None):
+    """Convert the device of the tensors in nests to the specified
+        or the default device.
+    Args:
+        nests (nested Tensors): Nested list/tuple/dict of Tensors.
+        device (None|str): the target device, should either be `cuda` or `cpu`.
+            If None, then the default device will be used as the target device.
+    Returns:
+        nests (nested Tensors): Nested list/tuple/dict of Tensors after device
+            conversion.
+
+    Raises:
+        NotImplementedError: If the target device is not one of
+            None, `cpu` or `cuda`.
+
+    """
 
     def _convert_cuda(tensor):
         if tensor.device.type != 'cuda':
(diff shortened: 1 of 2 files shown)


alf/nest/utils.py
```diff
@@ -207,10 +221,14 @@ def _convert_cpu(tensor):
...
-    d = alf.get_default_device()
+    if device is None:
+        d = alf.get_default_device()
+    else:
+        d = device
+
     if d == 'cpu':
         return nest.map_structure(_convert_cpu, nests)
-    elif d == 'cuda':
+    elif d == 'cuda' and torch.cuda.is_available():
```
hnyu: cuda.is_available() should be inside this branch and raise error otherwise

alf/trainers/policy_trainer.py
```diff
@@ -694,7 +695,8 @@ def play(root_dir,
             time_step = env.reset()
 
     for m in metrics:
-        logging.info("%s: %f", m.name, m.result())
+        logging.info("%s: %s", m.name,
+                     map_structure(lambda x: x.numpy().item(), m.result()))
```
hnyu: Do we need item()? It could be an np.array for each element?

hnyu (changes requested):

alf/nest/utils.py
Haichao-Zhang: Done

alf/trainers/policy_trainer.py
Haichao-Zhang: ``item()`` is needed, as the ``AverageReturnMetric`` returns a list of ndim-0 tensors (one for each return/reward dimension) in the case of play. Without ``item()``, we will get something like:
```
 AverageReturn: [array(0., dtype=float32), array(0., dtype=float32), array(0., dtype=float32), array(0., dtype=float32), array(0., dtype=float32)]
```
By adding item, we get a cleaner log message like:
```
AverageReturn: [0.0, 0.0, 0.0, 0.0, 0.0]
```



Added support to the case of np.array for each element, although currently there is no such metrics in play.

hnyu (approved):

PR #717, opened 2020-11-04
warning_once bug fix

Haichao-Zhang:
Previously the ``warning_once`` function cannot show all the warning_once messages generated at different places. For example:

```
while condition is True:
   warning_once("msg1")
   some other computation
   warning_once("msg2")
``` 

``"msg1"`` can be correctly displayed once, but ``"msg2"`` is never shown.

### DIFF
--- a/alf/utils/common.py
+++ b/alf/utils/common.py
@@ -625,11 +625,18 @@ def write_gin_configs(root_dir, gin_file):
 def warning_once(msg, *args):
     """Generate warning message once.
 
+    Note that the current implementation resembles that of the ``log_every_n()```
+    function in ``logging`` but reduces the calling stack by one to ensure
+    the multiple warning once messages generated at difference places can be
+    displayed correctly.
+
     Args:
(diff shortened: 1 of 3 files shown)


alf/utils/common.py
```diff
@@ -625,11 +625,18 @@ def write_gin_configs(root_dir, gin_file):
 def warning_once(msg, *args):
     """Generate warning message once.
 
+    Note that the current implementation resembles that of the ``log_every_n()```
+    function in ``logging`` but reduces the calling stack by one to ensure
+    the multiple warning once messages generated at difference places can be
+    displayed correctly.
```
emailweixu, line 631: Does this mean that log_every_n has a bug?

Haichao-Zhang: Currently  log_every_n can only get the caller one level up, and its count of n is based on this retrieved caller.
Therefore, when we use common.warning_once,  the caller of log_every_n is all the same.

emailweixu (approved):

PR #720, opened 2020-11-18
hardcode gym floating types

hnyu:
Higher Gym versions have some issues (bugs) of handling floating sensors where they will have float64 for at least Mujoco locomotion tasks. The observation spec limits are [-inf, inf] but somehow they forgot to cast it to np.float32 when creating. This raises errors later in PyTorch.


hnyu:
Seems this will break float64 ALF test cases. maybe an observation wrapper should be used instead.

hnyu:
Seems this will break float64 ALF test cases. maybe an observation wrapper should be used instead.

PR #721, opened 2020-11-19
Fix red light detection for suite_carla

emailweixu:


alf/environments/carla_sensors.py
```diff
@@ -955,6 +957,15 @@ def _get_traffic_light_waypoints(self, traffic_light):
                     wpx = next_wp
                 else:
                     break
+            if wpx.transform.location.distance(
+                    area_loc) >= self.RED_LIGHT_ENFORCE_DISTANCE - 2:
```
hnyu, line 961: What is special about -2 here?

alf/environments/carla_sensors.py
```diff
@@ -640,6 +640,10 @@ def _to_numpy_loc(loc: carla.Location):
     return np.array([loc.x, loc.y, loc.z], dtype=np.float)
 
 
+def _to_carla_loc(loc):
```
hnyu, line 643: This function seems not used anywhere.

hnyu (changes requested):

alf/environments/carla_sensors.py
emailweixu: distance between the center of the car and its tail

alf/environments/carla_sensors.py
emailweixu: I used it when I wrote debugging code. It doesn't hurt to keep it.

hnyu (approved):

PR #722, opened 2020-11-20
normalizer with unit std

hnyu:
Add an option for using normalizers with zero mean only (unit variance). Additionally support target updater for single nn.Parameter

### DIFF
--- a/alf/algorithms/data_transformer.py
+++ b/alf/algorithms/data_transformer.py
@@ -605,14 +605,6 @@ class RewardNormalizer(SimpleDataTransformer):
                 or (self._update_mode == "rollout" and common.is_rollout())):
             norm.update(timestep_or_exp.reward)
 
-            with alf.summary.scope("RewardNormalizer"):
-                if norm._mean_averager:
-                    alf.summary.scalar("normalizer/mean",
-                                       norm._mean_averager.get())
-                if norm._m2_averager:
-                    alf.summary.scalar("normalizer/m2",
-                                       norm._m2_averager.get())
-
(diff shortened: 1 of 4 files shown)


alf/algorithms/data_transformer.py
```diff
@@ -600,11 +600,21 @@ def __init__(self,
         self._update_mode = update_mode
 
     def _transform(self, timestep_or_exp):
+        norm = self._normalizer
         if ((self._update_mode == "replay" and common.is_replay())
                 or (self._update_mode == "rollout" and common.is_rollout())):
-            self._normalizer.update(timestep_or_exp.reward)
+            norm.update(timestep_or_exp.reward)
+
+            with alf.summary.scope("RewardNormalizer"):
```
emailweixu: This part should be a member function of Normalizer as it accesses private members.

hnyu: Deleted. Just realized that the normalizer already has a summary function inside

hnyu: > This part should be a member function of Normalizer as it accesses private members.

Deleted. Just realized that the normalizer already has a summary function inside

emailweixu: Make sure you are aware that normalizer won't summarize if debug_summaries is not True

hnyu: > Make sure you are aware that normalizer won't summarize if debug_summaries is not True

Yeah I've passed a flag in the gin file

emailweixu (approved):

PR #723, opened 2020-11-21
Support multi-decoder in predictive learner

Haichao-Zhang:
Add support to use multiple decoders in predictive learner.

### DIFF
--- a/alf/algorithms/predictive_representation_learner.py
+++ b/alf/algorithms/predictive_representation_learner.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 """PredictiveRepresentationLearner."""
 
+from functools import partial
 import gin
 import torch
 
@@ -23,7 +24,7 @@ from alf.experience_replayers.replay_buffer import BatchInfo, ReplayBuffer
 from alf.nest import nest
 from alf.nest.utils import convert_device
 from alf.networks import Network, LSTMEncodingNetwork
-from alf.utils import dist_utils, spec_utils, tensor_utils
+from alf.utils import common, dist_utils, spec_utils, tensor_utils
 from alf.utils.normalizers import AdaptiveNormalizer
 from alf.utils.summary_utils import safe_mean_hist_summary, safe_mean_summary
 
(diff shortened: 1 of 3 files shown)


alf/algorithms/predictive_representation_learner.py
```diff
@@ -72,18 +74,25 @@ def __init__(self,
...
+                ``loss(y_pred, y_true)``. Note that it should not reduce to a
+                scalar. It should at least keep the batch dimension in the
+                returned loss.
             loss_weight (float): weight for the loss.
             optimizer (Optimzer|None): if provided, it will be used to optimize
                 the parameter of decoder_net
             normalize_target (bool): whether to normalize target.
                 Note that the effect of this is to change the loss. The predicted
                 value itself is not normalized.
+            append_target_field_to_name (bool): whether append target field to
```
emailweixu, line 87: In this case, why cant the user pass "name.target_field" directly as the argument? Or perhaps default this argument to True to make it a little bit easier to config?

alf/algorithms/predictive_representation_learner.py
```diff
@@ -308,8 +345,33 @@ def predict_multi_step(self, init_latent, actions, state=None):
...
+    def get_decoder(self, target_field):
+        """Get the decoder which predicts the target specified by ``target_name``.
+        Args:
+            target_field (str): the name of the prediction quantity corresponding
+                to the decoder
+        Returns:
+            decoder (Algorithm)
+        """
+
+        decoder_ind = [
```
emailweixu: `common.as_list(self._target_fields).index(target_field)`

alf/algorithms/predictive_representation_learner.py
```diff
@@ -347,34 +409,47 @@ def _multi_step_latent_rollout(self, init_latent, num_unroll_steps,
...
 
     @torch.no_grad()
     def preprocess_experience(self, experience: Experience):
         """Fill experience.rollout_info with PredictiveRepresentationLearnerInfo
 
-        Note that the shape of experience is [B, T, ...]
+        Note that the shape of experience is [B, T, ...].
+
+        The target is a Tensor with the shape of [B, T, unroll_steps+1, ...]
+        when there is only one decoder. When multiple decorders are used,
```
emailweixu: This is not very accurate. Even with one decode, the target could be a nest because a field can correspond to a nest.

alf/algorithms/predictive_representation_learner.py
Haichao-Zhang: Changed as suggested.

alf/algorithms/predictive_representation_learner.py
Haichao-Zhang: Changed default value to True. 
The reason for adding this argument is that ``target_field`` is not yet available when we are constructing the decoders in predictive representation learner.

alf/algorithms/predictive_representation_learner.py
Haichao-Zhang: Updated the comments.

emailweixu (approved):

PR #725, opened 2020-11-24
several gym wrappers

hnyu:
Add an action normalization wrapper that always maps continuous actions to [-1, 1].

### DIFF
--- a/alf/environments/gym_wrappers.py
+++ b/alf/environments/gym_wrappers.py
@@ -642,11 +642,11 @@ class NormalizedAction(ContinuousActionMapping):
 
 
 @gin.configurable
-class NoDoneEnv(gym.Wrapper):
+class NonEpisodicEnv(gym.Wrapper):
     """Make a gym environment non-episodic by always setting ``done=False``."""
 
     def __init__(self, env):
-        gym.Wrapper.__init__(self, env)
+        super().__init__(env)
 
     def step(self, action):
         ob, reward, done, info = self.env.step(action)
--- a/alf/environments/gym_wrappers.py
+++ b/alf/environments/gym_wrappers.py
@@ -588,3 +588,65 @@ class ContinuousActionClip(gym.ActionWrapper):
         action = alf.nest.map_structure_up_to(
             action, _clip_action, self.action_space, action, self.bounds)
         return action
+
+
+@gin.configurable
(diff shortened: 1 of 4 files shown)


alf/environments/gym_wrappers.py
```diff
@@ -588,3 +588,65 @@ def _clip_action(space, action, bounds):
...
+class NormalizedAction(ContinuousActionMapping):
+    """Normalize actions to ``[-1, 1]``. This normalized action space is
+    friendly to algorithms that computes action entropy, e.g., SAC."""
+
+    def __init__(self, env):
+        super().__init__(env, low=-1., high=1.)
+
+
+@gin.configurable
+class NoDoneEnv(gym.Wrapper):
```
le-horizon: Would you consider renaming it to NonEpisodicEnv?

le-horizon (dismissed):
LG, only one suggestion.

alf/environments/suite_gym.py
```diff
@@ -76,6 +76,7 @@ def wrap_env(gym_env,
              max_episode_steps=0,
              gym_env_wrappers=(),
              time_limit_wrapper=alf_wrappers.TimeLimit,
+             normalize_action=True,
```
Haichao-Zhang, line 79: Might be better to make the default value to be False, so that the default behavior is the same as the original env?

le-horizon:
The fix looks good too.

Thanks,
Le

alf/environments/suite_gym.py
hnyu: What do you mean by default behavior?

hnyu: > Might be better to make the default value to be False, so that the default behavior is the same as the original env?

What do you mean by default behavior? So far all the continuous actions are by default [-1, 1], except that CARLA might have a [0, 1] action, but I think this change won't make much a difference. 

On the other hand, it's *really* easy for someone to overlook this normalization for some special envs that have non-standard action ranges (e.g., [-20, 20]), unless he remembers to look at the entropy. So defaulting to True is reasonable.

alf/environments/gym_wrappers.py
hnyu: > Would you consider renaming it to NonEpisodicEnv?

Updated

Haichao-Zhang (approved):

PR #726, opened 2020-12-02
Change default epsilon_greedy to 0

hnyu:
It's easy to forget this hyperparameter when training a job. For almost all experiments, this should be 0.

### DIFF
--- a/alf/trainers/policy_trainer.py
+++ b/alf/trainers/policy_trainer.py
@@ -392,7 +392,7 @@ class RLTrainer(Trainer):
         if self._num_iterations:
             time_to_checkpoint = self._trainer_progress._iter_num + checkpoint_interval
         else:
-            time_to_checkpoint = self._trainer_progress._num_env_steps + checkpoint_interval
+            time_to_checkpoint = self._trainer_progress._env_steps + checkpoint_interval
 
         while True:
             t0 = time.time()
@@ -594,11 +594,12 @@ def _step(algorithm, env, time_step, policy_state, trans_state, epsilon_greedy,
     return next_time_step, policy_step, trans_state
 
 
+@common.mark_eval
 def play(root_dir,
          env,
          algorithm,
          checkpoint_step="latest",
-         epsilon_greedy=0.1,
+         epsilon_greedy=0.,
          num_episodes=10,
          max_episode_length=0,
          sleep_time_per_step=0.01,
--- a/alf/trainers/policy_trainer.py
+++ b/alf/trainers/policy_trainer.py
@@ -719,8 +719,8 @@ def play(root_dir,
         logging.info(
             "%s: %s", m.name,
             map_structure(
-                lambda x: x.numpy().item() if x.ndim == 0 else x.numpy(),
-                m.result()))
+                lambda x: x.cpu().numpy().item()
+                if x.ndim == 0 else x.cpu().numpy(), m.result()))
     if recorder:
         recorder.close()
     env.reset()
--- a/alf/algorithms/config.py
+++ b/alf/algorithms/config.py
@@ -36,7 +36,7 @@ class TrainerConfig(object):
                  load_checkpoint_strict=True,
                  evaluate=False,
                  eval_interval=10,
-                 epsilon_greedy=0.1,
+                 epsilon_greedy=0.,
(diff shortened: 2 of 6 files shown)


emailweixu:
I thought a nonzero epsilon_greedy is to prevent infinite evaluation due to deterministic behavior (e.g. breakout). Perhaps you need to explicitly set epsilon_greedy for those tasks.

hnyu:
> I thought a nonzero epsilon_greedy is to prevent infinite evaluation due to deterministic behavior (e.g. breakout). Perhaps you need to explicitly set epsilon_greedy for those tasks.

So far, infinite loop only exists in Atari games. But it's easy to find out this kind of issue.

emailweixu:
> > I thought a nonzero epsilon_greedy is to prevent infinite evaluation due to deterministic behavior (e.g. breakout). Perhaps you need to explicitly set epsilon_greedy for those tasks.
> 
> So far, infinite loop only exists in Atari games. But it's easy to find out this kind of issue.

Actually, I feel it will be hard to find out because there won't be any error message.

hnyu:
> > > I thought a nonzero epsilon_greedy is to prevent infinite evaluation due to deterministic behavior (e.g. breakout). Perhaps you need to explicitly set epsilon_greedy for those tasks.
> > 
> > 
> > So far, infinite loop only exists in Atari games. But it's easy to find out this kind of issue.
> 
> Actually, I feel it will be hard to find out because there won't be any error message.

Both won't produce any error message. Using the default epsilon_greedy=0.1 but thinking it's 0 (at most papers set) is confusing. For an infinite loop, the episode length will be very large during evaluation and the training probably just stops proceeding; for playing, it's very easy to find out the loop.

emailweixu:
> > > > I thought a nonzero epsilon_greedy is to prevent infinite evaluation due to deterministic behavior (e.g. breakout). Perhaps you need to explicitly set epsilon_greedy for those tasks.
> > > 
> > > 
> > > So far, infinite loop only exists in Atari games. But it's easy to find out this kind of issue.
> > 
> > 
> > Actually, I feel it will be hard to find out because there won't be any error message.
> 
> Both won't produce any error message. Using the default epsilon_greedy=0.1 but thinking it's 0 (at most papers set) is confusing. For an infinite loop, the episode length will be very large during evaluation and the training probably just stops proceeding; for playing, it's very easy to find out the loop.

So changing the gin file for atari to explicitly setting epsilon_greedy=0.0 seems a reasonable solution, as we know it is a problem for atari.

hnyu:
> > > > > I thought a nonzero epsilon_greedy is to prevent infinite evaluation due to deterministic behavior (e.g. breakout). Perhaps you need to explicitly set epsilon_greedy for those tasks.
> > > > 
> > > > 
> > > > So far, infinite loop only exists in Atari games. But it's easy to find out this kind of issue.
> > > 
> > > 
> > > Actually, I feel it will be hard to find out because there won't be any error message.
> > 
> > 
> > Both won't produce any error message. Using the default epsilon_greedy=0.1 but thinking it's 0 (at most papers set) is confusing. For an infinite loop, the episode length will be very large during evaluation and the training probably just stops proceeding; for playing, it's very easy to find out the loop.
> 
> So changing the gin file for atari to explicitly setting epsilon_greedy=0.0 seems a reasonable solution, as we know it is a problem for atari.

updated

alf/trainers/policy_trainer.py
```diff
@@ -718,8 +719,8 @@ def play(root_dir,
         logging.info(
             "%s: %s", m.name,
             map_structure(
-                lambda x: x.numpy().item() if x.ndim == 0 else x.numpy(),
-                m.result()))
+                lambda x: x.cpu().numpy().item()
+                if x.ndim == 0 else x.cpu().numpy(), m.result()))
```
emailweixu, line 723: It's strange that how the previous code can work.

emailweixu (approved):

alf/trainers/policy_trainer.py
hnyu: > It's strange that how the previous code can work.

If play.py is called in CPU mode, then it won't be any problem. So I guess on CI it'll work but locally there will be problem.

PR #729, opened 2020-12-03
Support changing time of the day for suite_carla

emailweixu:


alf/environments/suite_carla.py
```diff
@@ -31,6 +31,7 @@
     cd ..
     ./ImportAssert.sh
     easy_install PythonAPI/carla/dist/carla-0.9.9-py3.7-linux-x86_64.egg
+    pip install networkx==2.2
```
le-horizon, line 34: where is this needed?

le-horizon (commented):
Almost LG, just one question.

Thanks,
Le

alf/environments/suite_carla.py
emailweixu: carla needs it.

alf/environments/suite_carla.py
```diff
@@ -888,6 +904,9 @@ def __init__(self,
         self._num_walkers = num_walkers
         self._percentage_walkers_running = percentage_walkers_running
         self._percentage_walkers_crossing = percentage_walkers_crossing
+        self._day_length = day_length
+        self._time_of_the_day = 0.5 * day_length
```
hnyu, line 908: Maybe better called "daytime_length"?

emailweixu: daytime_length is a little ambiguous, as it may mean the contrary of night time.

hnyu: >  as it may mean the contrary of night time.

I think that's what it should mean, no?

emailweixu: day_length means the length of the full cycle of one day.

hnyu: Right, I was talking about the variable name "self._time_of_the_day" which seems the same with "self._day_length". But I think it really means the daytime length

emailweixu: time_of_the_day means the current time. It is initialized at the middle of the day. It's constantly changed by update_time_of_the_day().

le-horizon (approved):
LG

alf/environments/suite_carla.py
hnyu: I see. I had a misunderstanding

PR #731, opened 2020-12-03
add embedding summary

hnyu:
Add a summary function for embedding visualization. Can be used to visualize latent vectors. If being called multiple times, there will be multiple figures rendered. So it's recommended to use a DataBuffer to collect latent vectors and then summarize once for all.  

After the embedding is visualized, you have to reissue the TB command to see it updated. Just refreshing the webpage won't work.

image

### DIFF
--- a/alf/summary/summary_ops.py
+++ b/alf/summary/summary_ops.py
@@ -167,7 +167,7 @@ def embedding(name, data, step=None, class_labels=None, label_imgs=None):
     The visualized embeddings can be seen in the "PROJECTOR" page of Tensorboard.
 
     Note: if this function is called multiple times, on the page there will be
-    multiple visualizations, each for every step.
+    multiple visualizations, each for every call.
 
     Args:
         name (str): data identifier
--- a/alf/summary/summary_ops.py
+++ b/alf/summary/summary_ops.py
@@ -19,6 +19,17 @@ import torch
 from torch.utils.tensorboard import SummaryWriter
 from typing import Callable
 
+try:
+    # If tensorflow has been installed, pytorch might use tensorflow's
+    # tensorboard. In this case, gfile needs to be redirected if embedding
+    # projector is to be used.
+    # https://github.com/pytorch/pytorch/issues/30966#issuecomment-582747929
+    import tensorflow as tf
+    import tensorboard as tb
+    tf.io.gfile = tb.compat.tensorflow_stub.io.gfile
+except:
+    pass
+
 _summary_enabled = False
 
 _summarize_output = False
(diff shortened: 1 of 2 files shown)


alf/summary/summary_ops.py
```diff
@@ -19,6 +19,17 @@
...
 from typing import Callable
 
+try:
+    # If tensorflow has been installed, pytorch might use tensorflow's
+    # tensorboard. In this case, gfile needs to be redirected if embedding
+    # projector is to be used.
+    # https://github.com/pytorch/pytorch/issues/30966#issuecomment-582747929
+    import tensorflow as tf
+    import tensorboard as tb
+    tf.io.gfile = tb.compat.tensorflow_stub.io.gfile
```
emailweixu, line 29: Someone in that thread said that this fix "did slow down program execution by a huge factor if you have many summarywriter calls"

hnyu: I think that it doesn't make sense to call many times. In that case, there will be a long list of figures, each runs visualization on a different batch of tensors. For most applications, summarizing several times throughout the training (for comparison) will be good enough.

emailweixu: But this fix seems affecting all the summarywriter calls.

hnyu: Not sure about if that person meant only for embedding summary or all summaries, because the gfile error only occurs when calling add_embedding(). I can do a test

hnyu: Seems that only add_embedding uses gfile for file I/O. Other summary calls use TB's FileWriter. 

https://github.com/pytorch/pytorch/blob/22a34bcf4e5eaa348f0117c414c3dd760ec64b13/torch/utils/tensorboard/writer.py#L798

I just tested and this change won't affect the efficiency if add_embedding is not called many times.

alf/utils/data_buffer.py
```diff
@@ -506,6 +506,9 @@ def get_batch_by_indices(self, indices):
                                             self._derived_buffer)
         return convert_device(result)
 
+    def is_full(self):
+        return (self.current_size == self._capacity).cpu().numpy()
+
```
le-horizon, line 511: Please note that this will return an array of 1 element?

le-horizon (commented):
One small nit.

Thanks,
Le

alf/summary/summary_ops.py
```diff
@@ -147,6 +158,37 @@ def histogram(name, data, step=None, bins=None, walltime=None, max_bins=None):
...
+@_summary_wrapper
+def embedding(name, data, step=None, class_labels=None, label_imgs=None):
+    """Add embeddings to summary. The potentially high-dimensional embeddings
+    will be projected down to either 2D or 3D for visualization, with several
+    projection techniques to choose from in Tensorboard.
+
+    The visualized embeddings can be seen in the "PROJECTOR" page of Tensorboard.
+
+    Note: if this function is called multiple times, on the page there will be
+    multiple visualizations, each for every step.
```
Haichao-Zhang: each for every step --> each for every call?

alf/utils/data_buffer.py
Haichao-Zhang: This function is not used?

hnyu: > Please note that this will return an array of 1 element?

It's not an array. self.current_size is a function that uses self._current_size[0]

hnyu: > This function is not used?

not used here. But to summarize embeddings, potentially we need a buffer to store many of them and summarize when it's full.

alf/summary/summary_ops.py
hnyu: updated

alf/utils/data_buffer.py
le-horizon: > > Please note that this will return an array of 1 element?
> 
> It's not an array. self.current_size is a function that uses self._current_size[0]

Try tensor([0]).numpy(), it returns an array(0).

le-horizon:
> Add a summary function for embedding visualization. Can be used to visualize latent vectors. If being called multiple times, there will be multiple figures rendered. So it's recommended to use a DataBuffer to collect latent vectors and then summarize once for all.
> 
> After the embedding is visualized, you have to reissue the TB command to see it updated. Just refreshing the webpage won't work.
> 
> image

Would be helpful to have some example commands on how this is done.

Thanks,
Le

alf/utils/data_buffer.py
hnyu: > > > Please note that this will return an array of 1 element?
> > 
> > 
> > It's not an array. self.current_size is a function that uses self._current_size[0]
> 
> Try tensor([0]).numpy(), it returns an array(0).

I mean self.current_size is torch.tensor(0) (self._current_size[0]) instead of torch.tensor([0]).

hnyu:
> > Add a summary function for embedding visualization. Can be used to visualize latent vectors. If being called multiple times, there will be multiple figures rendered. So it's recommended to use a DataBuffer to collect latent vectors and then summarize once for all.
> > After the embedding is visualized, you have to reissue the TB command to see it updated. Just refreshing the webpage won't work.
> > image
> 
> Would be helpful to have some example commands on how this is done.
> 
> Thanks,
> Le

Typically this is very similar to other ALF summary calls like alf.summary.scalar or alf.summary.histogram. It's compatible with our current ALF summary logic. You can just need to call this function in the right place, passing in a tensor matrix. See alf.summary.embedding docstring for details.

alf/utils/data_buffer.py
le-horizon: I see, scalar tensors can be automatically converted into float and bool.  LG.

le-horizon (approved):
LGTM.

Thanks,
Le

Haichao-Zhang (approved):

PR #732, opened 2020-12-14
let TB record code files

hnyu:
Sometimes I think it's much better to just let TB record whole files instead of recording GIT differences:
1) The cluster doesn't have .git because copying it will take a lot of time
2) We might want to record files that are not version controlled 

Example of recording code files:
image

### DIFF
--- a/alf/trainers/policy_trainer.py
+++ b/alf/trainers/policy_trainer.py
@@ -241,6 +241,20 @@ class Trainer(object):
             alf.summary.text('diff', _markdownify(git_utils.get_diff()))
             alf.summary.text('seed', str(self._random_seed))
 
+            if self._config.code_snapshots is not None:
+                for f in self._config.code_snapshots:
+                    path = os.path.join(
+                        os.path.abspath(os.path.dirname(__file__)), "..", f)
+                    if not os.path.isfile(path):
+                        common.warning_once(
+                            "The code file '%s' for summary is invalid" % path)
+                        continue
+                    with open(path, 'r') as fin:
+                        code = fin.read()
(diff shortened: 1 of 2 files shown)


emailweixu (approved):

alf/trainers/policy_trainer.py
```diff
@@ -241,6 +241,20 @@ def _markdownify(paragraph):
...
+                        os.path.abspath(os.path.dirname(__file__)), "..", f)
+                    if not os.path.isfile(path):
+                        common.warning_once(
+                            "The code file '%s' for summary is invalid" % path)
+                        continue
+                    with open(path, 'r') as fin:
+                        code = fin.read()
+                        # adding "<pre>" will make TB show raw text instead of MD
+                        alf.summary.text('code/%s' % f,
+                                         "<pre>" + code + "</pre>")
```
le-horizon, line 256: Just for your consideration, this snippet of js code shows python syntax highlighting:
https://stackoverflow.com/questions/62029757/nicely-formatted-python-code-in-html-files

le-horizon (approved):
LG, Haonan.  Just one suggestion.

Cheers,
Le

alf/trainers/policy_trainer.py
hnyu: > Just for your consideration, this snippet of js code shows python syntax highlighting:
> https://stackoverflow.com/questions/62029757/nicely-formatted-python-code-in-html-files

Not sure if tensorboard allows injecting js in its text tab?

hnyu: Just tried TB always interprets the text as markdown.

PR #734, opened 2020-12-16
support multi-dim reward in Gym and time_step refactor

hnyu:
random_alf_environment_test.py had some bugs but not detected because of using CPU on github CI.

PR #735, opened 2020-12-16
add suite_safety_gym

hnyu:
As with babyai, safety gym might have some conflicting python dependencies with our current ALF ones. So automatic installation is not updated in setup.py yet. It's recommended to follow Safety Gym's installation instructions in a new virtual python environment. To use this suite, either set `unconstrained=True` to enable a traditional unconstrained environment, or use PR #734 for multi-dim reward learning.

### DIFF
--- a/alf/environments/suite_safety_gym.py
+++ b/alf/environments/suite_safety_gym.py
@@ -45,6 +45,7 @@ except ImportError:
 
 import functools
 import numpy as np
+import copy
 import gym
 
 import gin
@@ -81,9 +82,10 @@ class CompleteEnvInfo(gym.Wrapper):
                 'cost_gremlins', 'cost_vases_displace', 'cost_vases_velocity',
                 'cost_hazards'
             ]
+        self._default_env_info = self._generate_default_env_info()
 
     def _is_level0_env(self, env_name):
-        return "0-" in env_name
+        return "0-v" in env_name
 
     def _generate_default_env_info(self):
         env_info = {}
@@ -95,7 +97,7 @@ class CompleteEnvInfo(gym.Wrapper):
         return env_info
 
     def step(self, action):
-        env_info = self._generate_default_env_info()
+        env_info = copy.copy(self._default_env_info)
         obs, reward, done, info = self.env.step(action)
         env_info.update(info)
         return obs, reward, done, env_info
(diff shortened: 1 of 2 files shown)


alf/environments/suite_safety_gym.py
```diff
@@ -0,0 +1,201 @@
...
+        alf_env_wrappers: Iterable with references to wrapper classes to use on
+            the torch environment.
+
+    Returns:
+        An AlfEnvironment instance.
+    """
+
+    # We can directly make the env here because none of the safety gym tasks
+    # is registered with a ``max_episode_steps`` argument (the
+    # ``gym.wrappers.time_limit.TimeLimit``) won't be applied. But each task
```
emailweixu: "won't be applied" should be inside the ")"

alf/environments/suite_safety_gym.py
```diff
@@ -0,0 +1,201 @@
...
+        if not self._is_level0_env(env_name):
+            # for level 1 and 2 envs, there are constraints cost info
+            self._env_info_keys += [
+                'cost_vases_contact', 'cost_pillars', 'cost_buttons',
+                'cost_gremlins', 'cost_vases_displace', 'cost_vases_velocity',
+                'cost_hazards'
+            ]
+
+    def _is_level0_env(self, env_name):
+        return "0-" in env_name
```
emailweixu: use "0-v" to be safer.

alf/environments/suite_safety_gym.py
```diff
@@ -0,0 +1,201 @@
...
+        env_info = {}
+        for key in self._env_info_keys:
+            if key == "goal_met":
+                env_info[key] = False
+            else:
+                env_info[key] = np.float32(0.)
+        return env_info
+
+    def step(self, action):
+        env_info = self._generate_default_env_info()
```
emailweixu: default_env_info can be generated only once at `__init__`

alf/environments/suite_safety_gym.py
```diff
@@ -0,0 +1,201 @@
...
+        obs, reward, done, info = self.env.step(action)
+        env_info.update(info)
+        return obs, reward, done, env_info
+
+
+class VectorReward(gym.Wrapper):
+    """This wrapper makes the env returns a reward vector of length 3. The three
+    dimensions are:
+
+    1. distance-improvement reward indicating the delta smaller distance to the goal
```
emailweixu: The original reward from safety gym seems more complicated than what you said here (https://github.com/openai/safety-gym/blob/f31042f2f9ee61b9034dd6a416955972911544f5/safety_gym/envs/engine.py#L1306)

hnyu: They have lots of options for rewards. Some for debugging purposes or customized environments. But in the public configured environments, only the distance delta (first three options) is used
https://github.com/openai/safety-gym/blob/f31042f2f9ee61b9034dd6a416955972911544f5/safety_gym/envs/suite.py#L90

emailweixu (approved):

PR #736, opened 2020-12-18
Transformer

emailweixu:


alf/layers.py
```diff
@@ -1092,3 +1092,214 @@ def calc_output_shape(self, input_shape):
         x = torch.zeros(1, *input_shape)
         y = self.forward(x)
         return y.shape[1:]
+
+
+def _masked_softmax(logits, mask, dim=-1):
+    if mask is not None:
+        p = logits.exp() * mask
+        return p / p.sum(dim=dim, keepdim=True)
```
hnyu: Need to check if the sum is zero or not

alf/layers.py
```diff
@@ -1092,3 +1092,214 @@ def calc_output_shape(self, input_shape):
         x = torch.zeros(1, *input_shape)
         y = self.forward(x)
         return y.shape[1:]
+
+
+def _masked_softmax(logits, mask, dim=-1):
+    if mask is not None:
+        p = logits.exp() * mask
```
Haichao-Zhang: Is it possible that ``mask`` has all zero values along ``dim``? If possible, the normalization  in line 1100 will have some issue.

alf/layers.py
```diff
@@ -1092,3 +1092,214 @@ def calc_output_shape(self, input_shape):
...
+                 d_model,
+                 num_heads,
+                 memory_size,
+                 d_k=None,
+                 d_v=None,
+                 relative_positional_encoding=True):
+        """
+        Args:
+            d_model (int): dimension of the model, same as d_model in [1]
+            num_heads (int): the number of attenion heads
```
Haichao-Zhang: attenion -> attention

alf/layers.py
```diff
@@ -1092,3 +1092,214 @@ def calc_output_shape(self, input_shape):
...
+        y = y.transpose(0, 1).reshape(b, h, m, n)
+        return y
+
+    def forward(self, memory, query=None, mask=None):
+        """
+        Args:
+            memory (Tensor): The shape is [B, N, d_model]
+            query (Tensor): The shape [B, d_model] or [B, M, d_model]. If None,
+                will use memory as query
+            mask (Tensor|None): [B, N] or [B, M, N]
```
Haichao-Zhang: Is mask used to exclude certain memory positions? Maybe can add comments explaining the purpose of mask or modify the name?

alf/layers_test.py
```diff
@@ -269,6 +271,126 @@ def test_harr_basis_correctness(self, input_size, output_size):
...
+                    relative_positional_encoding=task_type == 2))
+        layers.append(alf.layers.FC(d_model, 1))
+        model = nn.Sequential(*layers)
+
+        def _get_batch_content_based(batch_size):
+            """
+            A simple memory task:
+            In the first half of the sequence x[n] (i, j < max_len // 2):
+                one of x[n, i, 0] is set to 1, which means type 0 value is x[n, i, 3]
+                one of x[n, j, 1] is set to 1, which means type 1 value is x[n, j, 3]
```
hnyu: There is no description for dimension [:, :, 2]? To be clearer, "one of x[n, j, 1] (over j) is set to 1..."

hnyu (changes requested):
I didn't check many technical details

alf/layers.py
```diff
@@ -1092,3 +1092,214 @@ def calc_output_shape(self, input_shape):
...
+        # [M * N, B * H]
+        y = x_flattened[index]
+        y = y.transpose(0, 1).reshape(b, h, m, n)
+        return y
+
+    def forward(self, memory, query=None, mask=None):
+        """
+        Args:
+            memory (Tensor): The shape is [B, N, d_model]
+            query (Tensor): The shape [B, d_model] or [B, M, d_model]. If None,
```
hnyu, line 1217: Would be better if the meaning of ``M`` is explained. From my understanding, it's the number of queries per memory?

alf/layers.py
emailweixu: That shouldn't happen

alf/layers.py
emailweixu: fixed.

alf/layers.py
```diff
@@ -1092,3 +1092,214 @@ def calc_output_shape(self, input_shape):
...
+        q = torch.matmul(query_flattened, self._q_proj).reshape(
+            batch_size, m, num_heads, d_k).transpose(1, 2)
+        # [B, H, N, d_k]
+        k = torch.matmul(memory_flattened, self._k_proj).reshape(
+            batch_size, n, num_heads, d_k).transpose(1, 2)
+        # [B, H, N, d_v]
+        v = torch.matmul(memory_flattened, self._v_proj).reshape(
+            batch_size, n, num_heads, d_v).transpose(1, 2)
+        # [B, H, M, N]
+        logits = torch.matmul(q + self._qk_bias, k.transpose(2, 3))
```
hnyu, line 1286: Curious why here k is not added with a bias like q before computing the logits?

alf/layers_test.py
emailweixu: added.

alf/layers.py

alf/layers.py
```diff
@@ -1092,3 +1092,214 @@ def calc_output_shape(self, input_shape):
...
+        positional_encoding = self._positional_encoding
+        if n < self._memory_size:
+            d = self._memory_size - n
+            positional_encoding = positional_encoding[:, d:-d]
```
hnyu: The encoding length after truncating is self._memory_size - 2d? Do you instead want to have d/2:-d/2? And what's the benefit of truncating around the center vs. truncating the tail?

alf/layers.py
```diff
@@ -1092,3 +1092,214 @@ def calc_output_shape(self, input_shape):
...
+        self._attention_scale = 1. / d_k**0.5
+        self._mlp = torch.nn.Sequential(
+            FC(d_model, d_model, torch.relu_), FC(d_model, d_model))
+        self._norm1 = torch.nn.LayerNorm(d_model)
+        self._norm2 = torch.nn.LayerNorm(d_model)
+
+        l = 2 * memory_size - 1 if relative_positional_encoding else memory_size
+        self._positional_encoding = nn.Parameter(torch.Tensor(d_k, l))
+        self._qk_bias = nn.Parameter(torch.Tensor(1, num_heads, 1, d_k))
+        self._qp_bias = nn.Parameter(torch.Tensor(1, num_heads, 1, d_k))
```
hnyu, line 1188: Comment is needed on the two biases

emailweixu: added

alf/layers.py
emailweixu: added

alf/layers.py
emailweixu: None of the papers included bias for k. In fact, the original transformer paper does not have bias for q either. The two biases for q is introduced in [3] for better handling of relative positional encoding.

alf/layers.py
emailweixu: There was a bug here. This truncation only works for relative positional encoding, where the center encoding is for representing being at the same position.
Code updated for differentiate relative/non-relative positional encoding.

alf/layers.py
```diff
@@ -1092,3 +1092,246 @@ def calc_output_shape(self, input_shape):
...
+        """
+        y[i, j, :] <= x[n - 1 + i - j, :] for 0<=i<m, 0<=j<n
+        Args:
+            x: [2 * N - 1, d]
+        Returns:
+            [M, N, d]
+        """
+        n = (x.shape[0] + 1) // 2
+        # [M, N], index[i, j] = n - 1 + i - j
+        index = n - 1 + torch.arange(m).unsqueeze(-1) - torch.arange(n)
```
hnyu, line 1213: In this relative positional encoding case, it seems that the query length should be checked so that it's no greater than the memory size. Also, it only makes sense to use relative encoding when the query is the memory itself? Because otherwise we can't say that a position offset +-1 in the query corresponds to an offset +-1 in the memory.

emailweixu: There is a check at line 1270.
You are right, the relative encoding only make sense if query and memory are derived from same sequence. query and memory can still be different though.

hnyu (approved):

PR #737, opened 2020-12-23
Add NestSelectiveConcat

Haichao-Zhang:


alf/nest/utils.py
```diff
@@ -86,6 +86,34 @@ def _combine_flat(self, tensors):
...
+@gin.configurable
+class NestSelectiveConcat(NestCombiner):
+    def __init__(self, ind, dim=-1, name="NestSelectiveConcat"):
+        """A combiner for selecting from the tensors in a nest and then
+        concatenating them along a specified axis. It assumes that all the
+        selected tensors have the same tensor spec. Can be used as a
+        preprocessing combiner in ``EncodingNetwork``.
+
+        Args:
+            ind (int|list[ind]): the index or a list of indices indicating
```
hnyu: A NestCombiner assumes a general nested structure, not necessarily a flat list. I think your `ind` should also assumes a matched structure. 

Also it would be better to just combine this class with NestConcat with an optional argument `ind`.

Haichao-Zhang: Change and also merged with NestConcat.

alf/nest/utils.py
```diff
@@ -70,20 +70,38 @@ def __call__(self, nested):
...
+            nest_mask (nest|None): nest structured mask indicating which of the
+                tensors in the nest to be selected or not, indicated by a
+                value of True/False (1/0). Note that the structure of the mask
+                should be the same as the nest of data to apply this operator on.
+                If is None, then all the tensors from the nest will be selected.
             dim (int): the dim along which the tensors are concatenated
             name (str):
         """
         super(NestConcat, self).__init__(name)
+        self._flat_mask = nest.flatten(nest_mask) if nest_mask else nest_mask
```
hnyu, line 90: Should better use "is not None" for checking because if nest_mask=() then later it will fail

alf/nest/utils.py
```diff
@@ -70,20 +70,38 @@ def __call__(self, nested):
...
     def _combine_flat(self, tensors):
-        return torch.cat(tensors, dim=self._dim)
+        if self._flat_mask is not None:
+            assert len(self._flat_mask) == len(tensors), (
+                "incompatible structures "
+                "between mask and data nest")
+            selected_tensors = []
+            for i, mask_value in enumerate(self._flat_mask):
+                if mask_value:
+                    selected_tensors.append(tensors[i])
```
hnyu, line 101: `selected_tensors = [t for val, t in zip(self._flat_mask, tensors) if val]` One line

hnyu (changes requested):

hnyu (approved):

PR #738, opened 2020-12-23
simplify sac mixed dqda

hnyu:
For mixed type, if the inner discrete distribution is always optimal given the current action a, then it can be proved that dqda (of the outer layer) is just the sum of individual dqda (w.r.t. q values of different discrete actions in the inner layer), weighted by the optimal (detached) discrete probabilities. So line 599-604 (before change) gives the same gradient with line 599 (after change).

### DIFF
--- a/alf/algorithms/sac_algorithm.py
+++ b/alf/algorithms/sac_algorithm.py
@@ -589,23 +589,19 @@ class SacAlgorithm(OffPolicyAlgorithm):
                     critics = torch.tensordot(
                         critics, self._reward_weights, dims=1)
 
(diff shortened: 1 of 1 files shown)


Haichao-Zhang:
What if the inner discrete distribution is not optimal given the current action a?

hnyu:
> What if the inner discrete distribution is not optimal given the current action a?

It's guaranteed by the closed form exp(Q/alpha) in the mixed type case.

Haichao-Zhang (approved):

emailweixu (approved):

PR #747, opened 2020-12-31
Suite highway environment

Haichao-Zhang:
Added suite for highway environment.

### DIFF
--- a/alf/environments/suite_highway.py
+++ b/alf/environments/suite_highway.py
@@ -1,0 +1,144 @@
+# Copyright (c) 2020 Horizon Robotics. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Suite for loading highway environments.
+    Installation:
+    pip install git+https://github.com/eleurent/highway-env
+"""
+
+import collections
+import gin
+import gym
+import gym.spaces
+
+from alf.environments import suite_gym, alf_wrappers, gym_wrappers, process_environment
+
+try:
+    import highway_env
+except ImportError:
+    highway_env = None
+
+
+def is_available():
+    return highway_env is not None
+
+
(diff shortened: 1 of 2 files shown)


alf/environments/suite_highway.py
```diff
@@ -0,0 +1,144 @@
...
+    import highway_env
+except ImportError:
+    highway_env = None
+
+
+def is_available():
+    return highway_env is not None
+
+
+class VectorizeObservation(gym_wrappers.BaseObservationWrapper):
```
hnyu: The class name can be better called `FlattenObservation`; missing docstring for this class

alf/environments/suite_highway.py
```diff
@@ -0,0 +1,144 @@
...
+                    "sorted"
+            },
+            "action": {
+                "type": "ContinuousAction"
+            }
+        }
+        env_config = default_env_config
+
+    gym_env.configure(env_config)
+    _ = gym_env.reset()
```
hnyu: gym_env.reset() without returning

alf/environments/suite_highway.py
```diff
@@ -0,0 +1,144 @@
...
+    Returns:
+        An AlfEnvironment instance.
+    """
+    assert environment_name in {
+        "highway-v0", "merge-v0", "roundabout-v0", "intersection-v0",
+        "parking-v0"
+    }, "wrong highway environment name"
+
+    gym_spec = gym.spec(environment_name)
+    gym_env = gym.make(environment_name)
```
hnyu: Does the env itself use gym.make to wrap with max_steps?

hnyu (changes requested):

alf/environments/suite_highway.py
Haichao-Zhang: Changed class name and added doc string

alf/environments/suite_highway.py
Haichao-Zhang: updated

alf/environments/suite_highway.py
Haichao-Zhang: The original env does not differentiate terminations due to max_steps or others.
Now correctly handle max_steps in a way similar to suite_safety_gym.

hnyu: I mean by calling "gym.make", it probably already wrap an incorrect time_limit wrapper. Should use "gym_spec.make" instead. 
https://alf.readthedocs.io/en/latest/notes/knowledge_base.html#about-timestep-discount

hnyu (approved):

alf/environments/suite_highway.py
Haichao-Zhang: Right, I changed to use ``gym.spec.make`` instead of ``gym.make``in the second commit because of this.

PR #748, opened 2021-01-06
Add option to append blank frames when recording video file

Haichao-Zhang:
For some environments, it is difficult to tell the episode boundary in the recorded video.
Add option to append blank frames at episode end when recording video files to make it easier.

### DIFF
--- a/alf/utils/video_recorder.py
+++ b/alf/utils/video_recorder.py
@@ -177,6 +177,7 @@ class VideoRecorder(GymVideoRecorder):
                  value_range=1.,
                  frames_per_sec=None,
                  future_steps=0,
+                 append_blank_frames=0,
                  **kwargs):
         """
         Args:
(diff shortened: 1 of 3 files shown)


alf/utils/video_recorder.py
```diff
@@ -299,6 +306,15 @@ def capture_frame(self,
                 else:
                     self._encode_image_frame(frame)
 
+            if self._append_blank_frames > 0 and is_last_step:
+                if self._blank_frame is None:
+                    self._blank_frame = np.zeros_like(self.last_frame)
```
hnyu: self._last_frame

alf/utils/video_recorder.py
```diff
@@ -299,6 +306,15 @@ def capture_frame(self,
...
+                for _ in range(self._append_blank_frames):
+                    if self.ansi_mode:
```
hnyu: This branch appears multiple times in this file. Can use a function to wrap it.

hnyu (changes requested):

alf/utils/video_recorder.py
Haichao-Zhang: done

alf/utils/video_recorder.py
Haichao-Zhang: changed

hnyu (approved):

PR #749, opened 2021-01-08
PR Carla change weather

Haichao-Zhang:
Currently we used a **fixed** weather setting for each run:
```
cloudiness=80.0, precipitation=0.0, 
precipitation_deposits=60.0, wind_intensity=50.0, 
fog_density=20.0, fog_distance=0.0,
sun_azimuth_angle=250.0, sun_altitude_angle=20.0
```

While the sun angles can be updated by ``day_length`` related part, other weather parameters remain unchanged.
The variation of the visual data is still limited.

This PR enables weather randomization and can introduce more diversity in the experienced visual data, as shown below:

ezgif com-gif-maker (3)

ezgif com-gif-maker (2)

### DIFF
--- a/alf/environments/suite_carla.py
+++ b/alf/environments/suite_carla.py
@@ -37,7 +37,7 @@ Make sure you are using python3.7
 
 """
 
-from collections import OrderedDict
+from collections import OrderedDict, namedtuple
 from absl import logging
 import gin
 import math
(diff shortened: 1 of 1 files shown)


hnyu:
The randomization seems only done once for the entire training process?

Haichao-Zhang:
> The randomization seems only done once for the entire training process?

I have changed it to a second way, which is do randomization each time when the env is reset.
This might be less physically realistic but might be  good for training.

hnyu (dismissed):

hnyu (dismissed):

emailweixu (changes requested):
Instead of using a fixed random weather at the beginning, it's better to change the weather during simulation. Otherwise the test will performance gap compared to training. CARLA has an example of changing weahter PythonAPI/examples/dynamic_weather.py  for reference.

hnyu:
> Instead of using a fixed random weather at the beginning, it's better to change the weather during simulation. Otherwise the test will performance gap compared to training. CARLA has an example of changing weahter PythonAPI/examples/dynamic_weather.py for reference.

I think now whenever reset() is called, a random weather will be picked. Do you mean changing the weather at every step?

emailweixu:
> > Instead of using a fixed random weather at the beginning, it's better to change the weather during simulation. Otherwise the test will performance gap compared to training. CARLA has an example of changing weahter PythonAPI/examples/dynamic_weather.py for reference.
> 
> I think now whenever reset() is called, a random weather will be picked. Do you mean changing the weather at every step?

reset() of a batch AlfEnvironment is only called once or twice in the beginning. It will not be called during training.

Haichao-Zhang:
> > > Instead of using a fixed random weather at the beginning, it's better to change the weather during simulation. Otherwise the test will performance gap compared to training. CARLA has an example of changing weahter PythonAPI/examples/dynamic_weather.py for reference.
> > 
> > 
> > I think now whenever reset() is called, a random weather will be picked. Do you mean changing the weather at every step?
> 
> reset() of a batch AlfEnvironment is only called once or twice in the beginning. It will not be called during training.

I miss-understood the ``_reset()`` function. Will try to handle it correctly.

Haichao-Zhang:
> Instead of using a fixed random weather at the beginning, it's better to change the weather during simulation. Otherwise the test will performance gap compared to training. CARLA has an example of changing weahter PythonAPI/examples/dynamic_weather.py for reference.

The weather is now updated each time during simulation.

alf/environments/suite_carla.py
```diff
@@ -1076,6 +1118,56 @@ def _spawn_vehicles(self):
...
+        """
+
+        # sample new weather parameter
+        if self._weather_length_count == 0:
+            # the actual lasting time of the new weather
+            self._actual_weather_length = max(
+                0.25, np.random.rand()) * self._max_weather_length
+            p = np.random.uniform(0, 100, len(WeatherParameters()))
+            weather = self._world.get_weather()
+            wp = WeatherParameters(*p)
```
emailweixu: not used.

alf/environments/suite_carla.py
```diff
@@ -1076,6 +1118,56 @@ def _spawn_vehicles(self):
...
+            prev_weather_parameter = [
+                weather.cloudiness, weather.precipitation,
+                weather.precipitation_deposits, weather.wind_intensity,
+                weather.fog_density, weather.fog_distance
+            ]
+
+            trans_steps = max(
+                1, self._actual_weather_length * self._weather_transition_ratio
+                / self._step_time)
+            self._dp = WeatherParameters(*(
```
emailweixu: The requires the order at line 1141 is exactly same as the order at line 882, which is error prone.

emailweixu (changes requested):

alf/environments/suite_carla.py
Haichao-Zhang: removed

alf/environments/suite_carla.py
Haichao-Zhang: Good point. Improved the implementation and now we don't have to maintain the correspondence.

alf/environments/suite_carla.py
```diff
@@ -131,6 +131,118 @@ def _to_numpy_loc(loc: carla.Location):
...
+        self._fields = None
+
+    def get_weather_fields(self):
+        """ Get the list of configurable weather fields
+
+        Returns:
+            A list of strings, each as the name of a configurable field
+        """
+        if self._fields is None:
+            self._fields = [
```
emailweixu: since get_weather_fields is always used, it's better to move this to `__init__` to avoid the if statement later.

Haichao-Zhang: Now moved it to ``__init__``

emailweixu (approved):

PR #753, opened 2021-01-14
option for not rendering prediction info

hnyu:
Currently VideoRecorder decides whether to render prediction info (action, action_distribution) based on if any info is returned by predict_step(), and the rendering is usually triggered. Rendering pred info takes a very long time, so an option is added for convenience (default to not rendering pred info).


emailweixu (approved):
I also noticed this.

PR #754, opened 2021-01-14
BatchEnsemble for FC

emailweixu:


alf/layers.py
```diff
@@ -424,6 +424,143 @@ def make_parallel(self, n):
...
+                 use_ln=False,
+                 kernel_initializer=None,
+                 kernel_init_gain=1.0,
+                 bias_init_range=0.):
+        """
+        Args:
+            input_size (int): input size
+            output_size (int): output size
+            ensemble_size (int): ensemble size
+            output_ensemble_ids (bool): If True, the forward() funciton will return
```
Haichao-Zhang: function

alf/layers.py
```diff
@@ -424,6 +424,143 @@ def make_parallel(self, n):
...
+        self._ensemble_bias = nn.Parameter(
+            torch.Tensor(ensemble_size, output_size))
+        self._use_ensemble_bias = use_bias
+        self._ensemble_size = ensemble_size
+        self._output_ensemble_ids = output_ensemble_ids
+        self._bias_init_range = bias_init_range
+        super().__init__(
+            input_size,
+            output_size,
+            activation=identity,
```
Haichao-Zhang: activation=identity --> activation=activation ?

alf/layers.py
```diff
@@ -424,6 +424,143 @@ def make_parallel(self, n):
...
+            kernel_init_gain=kernel_init_gain)
+
+    def reset_parameters(self):
+        """Reinitialize parameters."""
+        super().reset_parameters()
+        torch.randint(
+            2, size=self._r.shape, dtype=torch.float32, out=self._r.data)
+        torch.randint(
+            2, size=self._s.shape, dtype=torch.float32, out=self._s.data)
+        self._r.data.mul_(2)
```
Haichao-Zhang, line 502: Since ```torch.randint``` is already in [0, 2], do we still need ```mul_(2)```? 
If yes, it can be merged with randint anyway.

Same for ```self._s```.

It might be useful to comment on the initialization of the tuples of params for the low rank part somewhere, e.g. its range.

===
Update to the previous comment:  ```torch.randint(0, 2)``` samples from 0/1 and the initialization is correct.

alf/layers.py
```diff
@@ -424,6 +424,143 @@ def make_parallel(self, n):
...
+    def forward(self, inputs):
+        """Forward computation.
+
+        Args:
+            inputs (Tensor|tuple): if a Tensor, its shape should be ``[batch_size, input_size]`` or
+                ``[batch_size, ..., input_size]``. And a random ensemble id will be
+                generated for each sample in the batch. If a tuple, it should
+                contain two tensors. The first one is the data tensor with shape
+                ``[batch_size, input_size]`` or ``[batch_size, ..., input_size]``.
+                The second one is emsemble ids indicating which ensemble member each
```
Haichao-Zhang: emsemble --> ensemble

Haichao-Zhang (changes requested):

alf/layers.py
```diff
@@ -424,6 +424,143 @@ def make_parallel(self, n):
...
+class FCBatchEnsemble(FC):
+    r"""The BatchEnsemble for FC layer.
+
+    BatchEnsemble is proposed in `Wen et. al. BatchEnsemble: An Alternative Approach
+    to Efficient Ensemble and Lifelong Learning <https://arxiv.org/abs/2002.06715>`_
+
+    In a nutshell, a tuple of vector :math:`(r_k, s_k)` is maintained for ensemble
+    member k in addition to the original FC weight matrix w. For input x, the
+    result for ensemble member k is calculated as :math:`(W \circ (s_k r_k^T)) x`.
+    This can be more efficiently calculated as :math:`(W (x \circ r_k)) \circ s_k`
```
hnyu: Should add an explanation that every forward will use a random k if no ensemble id is provided? Otherwise one might think that there will be K outputs for each input.

alf/layers.py
```diff
@@ -424,6 +424,143 @@ def make_parallel(self, n):
...
+        """
+        Args:
+            input_size (int): input size
+            output_size (int): output size
+            ensemble_size (int): ensemble size
+            output_ensemble_ids (bool): If True, the forward() funciton will return
+                a tuple of (result, ensemble_ids). If False, the forward() funciton
+                will return result only.
+            activation (Callable): activation function
+            use_bias (bool): whether use bias
```
runjerry, line 463: bias -> ensemble_bias ?

Haichao-Zhang (approved):

PR #755, opened 2021-01-15
relu_mlp updates

runjerry:
* fix a bug in direct Jacobian computation when input batch_size is 1

* add direct vector-Jacobian product computation

### DIFF
--- a/alf/networks/relu_mlp.py
+++ b/alf/networks/relu_mlp.py
@@ -122,7 +122,7 @@ class ReluMLP(Network):
         for fc in self._fc_layers:
             z = fc(z)
         if ndim == 1:
-            z = z.squeeze()
+            z = z.squeeze(0)
         if requires_jac:
             z = (z, self._compute_jac())
         elif requires_jac_diag:
@@ -139,7 +139,7 @@ class ReluMLP(Network):
         self.forward(inputs)
         J = self._compute_jac()
         if inputs.ndim == 1:
-            J = J.squeeze()
+            J = J.squeeze(0)
 
         return J
 
@@ -164,7 +164,7 @@ class ReluMLP(Network):
         self.forward(inputs)
         J_diag = self._compute_jac_diag()
         if inputs.ndim == 1:
-            J_diag = J_diag.squeeze()
+            J_diag = J_diag.squeeze(0)
 
         return J_diag
 
@@ -231,6 +231,6 @@ class ReluMLP(Network):
             J = torch.matmul(J * mask, fc.weight)
 
         if ndim == 1:
-            J = J.squeeze()
+            J = J.squeeze(0)
 
         return J  # [B, n_in] or [n_in]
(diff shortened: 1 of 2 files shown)


alf/networks/relu_mlp.py
```diff
@@ -179,3 +178,45 @@ def _compute_jac_diag(self):
...
+    def _compute_vjp(self, vec):
+        """Compute vector-Jacobian product. """
+
+        if vec.ndim == 1:
+            vec = vec.unsqueeze(0)
+
+        J = torch.matmul(vec, self._fc_layers[-1].weight)
+        for fc in reversed(self._fc_layers[0:-1]):
+            mask = (fc.hidden_neurons > 0).float()
+            J = torch.einsum('ba, ba, aj->bj', J, mask, fc.weight)
```
emailweixu: I don't have very high confidence on the performance of einsum. matmul is generally highly optimized. And hence ``matmul(J * mask, fc.weight)`` might be faster.

alf/networks/relu_mlp.py
```diff
@@ -112,9 +112,10 @@ def forward(self,
             requires_jac (bool): whether outputs input-output Jacobian.
             requires_jac_diag (bool): whetheer outputs diagonals of Jacobian.
         """
-        inputs = inputs.squeeze()
-        assert inputs.shape[-1] == self._input_size, \
-            ("inputs should has shape {}!".format(self._input_size))
+        if inputs.ndim == 1:
```
emailweixu: The shape of inputs and output are not consistent when ndim==1. Should squeeze the output if ndim==1

alf/networks/relu_mlp.py
```diff
@@ -179,3 +178,45 @@ def _compute_jac_diag(self):
...
+
+        if vec.ndim == 1:
+            vec = vec.unsqueeze(0)
+
+        J = torch.matmul(vec, self._fc_layers[-1].weight)
+        for fc in reversed(self._fc_layers[0:-1]):
+            mask = (fc.hidden_neurons > 0).float()
+            J = torch.einsum('ba, ba, aj->bj', J, mask, fc.weight)
+
+        return J  # [B, n_in]
```
emailweixu: input and output shape inconsistency.

alf/networks/relu_mlp.py
runjerry: updated.

alf/networks/relu_mlp.py
runjerry: updated.

alf/networks/relu_mlp.py
runjerry: updated.

alf/networks/relu_mlp.py
```diff
@@ -134,8 +137,11 @@ def compute_jac(self, inputs):
             ("inputs should has shape {}!".format(self._input_size))
 
         self.forward(inputs)
+        J = self._compute_jac()
+        if inputs.ndim == 1:
+            J = J.squeeze()
```
emailweixu: To be safe, use J.squeeze(0) in case there are some other singleton dimensions,

runjerry: Updated here and other places.

emailweixu (approved):

PR #756, opened 2021-01-15
Add average discounted episodic return metric

Haichao-Zhang:
This metric computes the average discounted episodic return. 
It can be viewed as an approximate visualization of ```E_{s} V(s)```.

### DIFF
--- a/alf/algorithms/rl_algorithm.py
+++ b/alf/algorithms/rl_algorithm.py
@@ -147,6 +147,10 @@ class RLAlgorithm(Algorithm):
                     batch_size=env.batch_size,
                     buffer_size=metric_buf_size,
                     reward_shape=env.reward_spec().shape),
+                alf.metrics.AverageDiscountedEpisodicReturnMetric(
+                    batch_size=env.batch_size,
+                    buffer_size=metric_buf_size,
+                    reward_shape=env.reward_spec().shape),
                 alf.metrics.AverageEpisodeLengthMetric(
                     batch_size=env.batch_size, buffer_size=metric_buf_size),
                 alf.metrics.AverageEnvInfoMetric(
--- a/alf/algorithms/rl_algorithm.py
+++ b/alf/algorithms/rl_algorithm.py
@@ -147,16 +147,16 @@ class RLAlgorithm(Algorithm):
                     batch_size=env.batch_size,
                     buffer_size=metric_buf_size,
                     reward_shape=env.reward_spec().shape),
-                alf.metrics.AverageDiscountedEpisodicReturnMetric(
-                    batch_size=env.batch_size,
-                    buffer_size=metric_buf_size,
-                    reward_shape=env.reward_spec().shape),
                 alf.metrics.AverageEpisodeLengthMetric(
                     batch_size=env.batch_size, buffer_size=metric_buf_size),
                 alf.metrics.AverageEnvInfoMetric(
                     example_env_info=env.reset().env_info,
                     batch_size=env.batch_size,
-                    buffer_size=metric_buf_size)
+                    buffer_size=metric_buf_size),
+                alf.metrics.AverageDiscountedReturnMetric(
+                    batch_size=env.batch_size,
+                    buffer_size=metric_buf_size,
+                    reward_shape=env.reward_spec().shape)
             ]
 
         self._original_rollout_step = self.rollout_step
--- a/alf/metrics/metrics.py
+++ b/alf/metrics/metrics.py
@@ -230,7 +230,7 @@ class AverageEpisodicSumMetric(metric.StepMetric):
 
     def _extract_and_process_acc_value(self, acc, last_episode_indices):
         """Extract the final accumulated value and perform some optional
-        customizable processings.
+        customizable processing.
         Args:
             acc (Tensor): batched tensor representing an accumulator
             last_episode_indices (Tensor): indices representing the location
@@ -290,13 +290,16 @@ class AverageDiscountedReturnMetric(AverageEpisodicSumMetric):
 
     .. math::
         \begin{array}{ll}
-            R &= r_1 + (1 + gamma) r_2 + + (1 + gamma + gamma^2) r_3 + \cdots \\
+            R &= r_1 + (1 + gamma) r_2 + (1 + gamma + gamma^2) r_3 + \cdots \\
             &= \frac{1}{L}\sum_{l=1}^L \sum_{k=0}^{l-1} \gamma^k r_l,
         \end{array}
-    where :math:`\gamma` is the reward discount.
+    where :math:`\gamma` is the reward discount, and :math:`r_1` denotes the
+    reward due to the first action, which is received at the second time step.
+    :math:`L` equals to the episode length - 1.
 
-    Note that if the last step is not due time limit, the discounted return
-    calculated form the formula above is unbiased. If the last step is due
+
+    Note that if the last step is not due to time limit, the discounted return
+    calculated form the formula above is unbiased. If the last step is due to
     time limit, it is a biased estimate and its expectation is lower than the
     ground-truth one.
     """
@@ -360,8 +363,8 @@ class AverageDiscountedReturnMetric(AverageEpisodicSumMetric):
         is_first = time_step.is_first()
 
         # update discount for the next time step
-        self._accumulated_discount = self._accumulated_discount + \
-                            self._discount**self._current_step
+        self._accumulated_discount = (
+            self._discount * self._accumulated_discount + 1)
         self._accumulated_discount = torch.where(
             is_first, torch.zeros_like(self._accumulated_discount),
             self._accumulated_discount)
--- a/alf/metrics/metrics.py
+++ b/alf/metrics/metrics.py
@@ -16,6 +16,7 @@
 Converted to PyTorch from the TF version.
 https://github.com/tensorflow/agents/blob/master/tf_agents/metrics/tf_metrics.py
 """
+import gin
 import torch
 
 import alf
@@ -164,17 +165,6 @@ class AverageEpisodicSumMetric(metric.StepMetric):
         """Extract metrics from the time step. The return can be a nest."""
         raise NotImplementedError()
 
-    def _after_update(self, time_step):
-        """Some calculations done after the main calculation required for
-        accumulator and buffer update.
-        This function can be overridden by subclasses' ``_after_update()``
-        to perform some class specific calculations.
-        Args:
-            time_step (alf.data_structures.TimeStep): batched tensor
-
-        """
-        pass
-
     def _initialize(self, example_metric_value):
         counter = [0]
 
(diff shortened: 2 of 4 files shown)


alf/metrics/metrics.py
```diff
@@ -266,6 +279,68 @@ def _extract_metric_values(self, time_step):
...
+
+        super(AverageDiscountedEpisodicReturnMetric, self).__init__(
+            name=name,
+            dtype=dtype,
+            prefix=prefix,
+            batch_size=batch_size,
+            buffer_size=buffer_size,
+            example_metric_value=example_metric_value)
+
+    def _extract_metric_values(self, time_step):
```
emailweixu, line 334: So this metric will only calculate the discounted return at the beginning of an episode. The values for RL algorithms usually means the average discounted return over all states. So that are not directly comparable.

alf/metrics/metrics.py
```diff
@@ -266,6 +279,68 @@ def _extract_metric_values(self, time_step):
...
+        if time_step.reward.ndim == ndim:
+            return time_step.reward * self._accumulated_discount
+        else:
+            reward = time_step.reward.reshape(*time_step.step_type.shape, -1)
+            return [
+                reward[..., i] * self._accumulated_discount
+                for i in range(reward.shape[-1])
+            ]
+
+    def _after_update(self, time_step):
```
emailweixu: I don't see the need to introduce another hook. Why don't you do this inside _extract_metric_values()?

Haichao-Zhang: Originally to make different functionalities more isolated. I can merge them.

alf/metrics/metrics.py
Haichao-Zhang: That is true. The current metric is episodic and not directly comparable to the ``values`` curve.
I'm also trying to implementing a version by averaging over the states encountered during rollout.

le-horizon: You'll probably need to record the value computed during loss computation, store it into some member variable of the algorithm, and then plot it (alf.summary.scalar/histogram) in summarize_rollout().

alf/algorithms/rl_algorithm.py
```diff
@@ -147,6 +147,10 @@ def __init__(self,
                     batch_size=env.batch_size,
                     buffer_size=metric_buf_size,
                     reward_shape=env.reward_spec().shape),
+                alf.metrics.AverageDiscountedEpisodicReturnMetric(
+                    batch_size=env.batch_size,
+                    buffer_size=metric_buf_size,
+                    reward_shape=env.reward_spec().shape),
```
le-horizon: maybe put at the end, after EnvInfo metric?

alf/metrics/metrics.py
```diff
@@ -266,6 +279,68 @@ def _extract_metric_values(self, time_step):
...
 
+class AverageDiscountedEpisodicReturnMetric(AverageEpisodicSumMetric):
+    """Metric for computing the average discounted episodic return."""
+
+    def __init__(self,
+                 name='AverageDiscountedEpisodicReturnMetric',
+                 prefix='Metrics',
+                 reward_shape=(),
+                 dtype=torch.float32,
+                 discount=0.99,
```
le-horizon, line 312: it may be better to get this discount from algorithm's loss, if possible, so they are consistent?

alf/metrics/metrics_test.py
```diff
@@ -68,12 +69,17 @@ def _create_trajectories(self):
...
          ('testAverageEpisodeLengthGraph', AverageEpisodeLengthMetric, 6, 2.0),
          ('testAverageEnvInfoMetric', AverageEnvInfoMetric, 6,
           dict(x=torch.as_tensor(5.), y=torch.as_tensor(1.5)))])
     def testMetric(self, metric_class, num_trajectories, expected_result):
         trajectories = self._create_trajectories()
-        if metric_class in [AverageReturnMetric, AverageEpisodeLengthMetric]:
+        if metric_class in [
+                AverageReturnMetric, AverageDiscountedEpisodicReturnMetric,
+                AverageEpisodeLengthMetric
+        ]:
```
le-horizon: I'd be happier if a test for multi dim reward could be added, since this change handles multi dim reward as well.

alf/trainers/policy_trainer.py
```diff
@@ -358,6 +358,9 @@ def __init__(self, config: TrainerConfig):
                 alf.metrics.AverageReturnMetric(
                     buffer_size=self._num_eval_episodes,
                     reward_shape=self._eval_env.reward_spec().shape),
+                alf.metrics.AverageDiscountedEpisodicReturnMetric(
+                    buffer_size=self._num_eval_episodes,
+                    reward_shape=self._eval_env.reward_spec().shape),
```
le-horizon: same here, move to last?

le-horizon (commented):
Hi Haichao,

I don't fully understand the need for this change.  If we are trying to figure out whether value overestimates, then we should maybe look at rollout value or training losses for value?
Other than that, this change looks good in general.  A few suggestions inlined.

Thanks,
Le

alf/metrics/metrics.py
Haichao-Zhang: Removed this function.

alf/metrics/metrics.py
Haichao-Zhang: > So this metric will only calculate the discounted return at the beginning of an episode. The values for RL algorithms usually means the average discounted return over all states. So that are not directly comparable.

Now changed the metric to be the average of all the states in rollout.

Haichao-Zhang: > You'll probably need to record the value computed during loss computation, store it into some member variable of the algorithm, and then plot it (alf.summary.scalar/histogram) in summarize_rollout().

Calculating the value during rollout is probably the best to be comparable to the discounted return. This is not included currently as it will have some additional computational cost. 
Currently, we may use the the value from the loss function as a coarse estimation.

alf/algorithms/rl_algorithm.py
Haichao-Zhang: Done

alf/metrics/metrics.py
Haichao-Zhang: It is not straightforward to retrieve the discount value. Now make it configurable so we can set it as we wish.

alf/metrics/metrics_test.py
Haichao-Zhang: Added

alf/trainers/policy_trainer.py
Haichao-Zhang: Done

Haichao-Zhang:
> Hi Haichao,
> 
> I don't fully understand the need for this change. If we are trying to figure out whether value overestimates, then we should maybe look at rollout value or training losses for value?
> Other than that, this change looks good in general. A few suggestions inlined.
> 
> Thanks,
> Le

This PR is summarize the discounted return (based on rollouts), which is what the value function tries to predict.

You are right, calculating the value during rollout is probably the best to be comparable to the discounted return in order to determine the degree to which we have overestimation. 
This is not included currently as it will have some additional computational cost.
Currently, we may use the the value from the loss function as a coarse estimation.

alf/metrics/metrics.py
```diff
@@ -224,18 +214,32 @@ def _update_accumulator_(acc, val):
...
+                lambda buf, acc: buf.append(
+                    self._extract_and_process_acc_value(
+                        acc, last_episode_indices)), self._buffer,
+                self._accumulator)
 
         return time_step
 
+    def _extract_and_process_acc_value(self, acc, last_episode_indices):
+        """Extract the final accumulated value and perform some optional
+        customizable processings.
```
le-horizon: processings => processing

alf/metrics/metrics.py
le-horizon: Ah, you are right, my suggestion won't work for rollout loss.

le-horizon (dismissed):
Thanks for the improvements and quick turnaround, Haichao.  LGTM!

Le

alf/metrics/metrics.py
```diff
@@ -266,6 +283,107 @@ def _extract_metric_values(self, time_step):
...
+    It is calculated according to the following formula:
+
+    .. math::
+        \begin{array}{ll}
+            R &= r_1 + (1 + gamma) r_2 + + (1 + gamma + gamma^2) r_3 + \cdots \\
+            &= \frac{1}{L}\sum_{l=1}^L \sum_{k=0}^{l-1} \gamma^k r_l,
+        \end{array}
+    where :math:`\gamma` is the reward discount.
+
+    Note that if the last step is not due time limit, the discounted return
```
emailweixu: due => due to

alf/metrics/metrics.py
```diff
@@ -266,6 +283,107 @@ def _extract_metric_values(self, time_step):
...
+    calculated form the formula above is unbiased. If the last step is due
```
emailweixu: due to

alf/metrics/metrics.py
```diff
@@ -266,6 +283,107 @@ def _extract_metric_values(self, time_step):
...
+        second step. Otherwise, is it updated by multiplying with
+        ``self._discount``.
+        Args:
+            time_step (alf.data_structures.TimeStep): batched tensor
+        """
+        is_first = time_step.is_first()
+
+        # update discount for the next time step
+        self._accumulated_discount = self._accumulated_discount + \
+                            self._discount**self._current_step
```
emailweixu: A slightly faster version:
```python
self._accumulated_discount = self._discount * self._accumulated_discount + 1
```

alf/metrics/metrics.py
```diff
@@ -266,6 +283,107 @@ def _extract_metric_values(self, time_step):
...
+        self._accumulated_discount = torch.where(
```
emailweixu, line 372: For FIRST, this is 0, which means that reward for FIRST step is ignored (which is correct, BTW). Because of this, at line 383, should we divide acc using (self._current_step[last_episode_indices] - 1) instead?

Haichao-Zhang: The ```_current_step``` is initialized to be zero for FIRST step and is increased by one each time step after that.
Therefore, at the episode end, ```self._current_step[last_episode_indices]```  equals to episode_length-1.
So it actually performs the correct calculation as you suggested.

I added more comments on this.

alf/metrics/metrics.py
Haichao-Zhang: fixed

alf/metrics/metrics.py
Haichao-Zhang: fixed

alf/metrics/metrics.py
Haichao-Zhang: fixed

alf/metrics/metrics.py
Haichao-Zhang: Changed

alf/metrics/metrics.py
```diff
@@ -266,6 +283,114 @@ def _extract_metric_values(self, time_step):
...
 
 
+@gin.configurable
+class AverageDiscountedReturnMetric(AverageEpisodicSumMetric):
+    """Metric for computing the average discounted episodic return.
+    It is calculated according to the following formula:
+
+    .. math::
+        \begin{array}{ll}
+            R &= r_1 + (1 + gamma) r_2 + (1 + gamma + gamma^2) r_3 + \cdots \\
```
hnyu: gamma needs backslash; just seen from this first line, it's not the discounted return; an average is needed

alf/metrics/metrics.py
```diff
@@ -266,6 +283,114 @@ def _extract_metric_values(self, time_step):
...
+            &= \frac{1}{L}\sum_{l=1}^L \sum_{k=0}^{l-1} \gamma^k r_l,
```
hnyu, line 294: Do we want to average over the ending steps or the beginning steps? If this quantity is to be compared to the value estimation, then we want to average over the beginning steps. In the episodic setting:

\frac{1}{L}\sum_{l=1}^L \sum_{k=l}^{L} \gamma^{k-l} r_k,

alf/metrics/metrics.py
```diff
@@ -266,6 +283,114 @@ def _extract_metric_values(self, time_step):
...
+        \end{array}
+    where :math:`\gamma` is the reward discount, and :math:`r_1` denotes the
+    reward due to the first action, which is received at the second time step.
+    :math:`L` equals to the episode length - 1.
+
+
+    Note that if the last step is not due to time limit, the discounted return
+    calculated form the formula above is unbiased. If the last step is due to
+    time limit, it is a biased estimate and its expectation is lower than the
```
hnyu, line 303: If the last time step is TimeLimit, then I think a better metric would be still using the last step's value estimation for the remaining sum.

alf/metrics/metrics.py
Haichao-Zhang: Fixed.

alf/metrics/metrics.py
Haichao-Zhang: Agree. In this case, the metric is not accurate.

hnyu: For the last several steps, this bias is very severe. If you're going to compare the average metric with the predicted discounted value, then probably the comparison is not informative. Imagine that the predicted value is around 100 at every step in the non-episodic setting, but the steps towards the time limit only have accumulated rewards of about 1.

emailweixu (approved):

PR #758, opened 2021-01-19
fix reward clipping in dynamic_action_repeat_agent

hnyu:
Bug introduced in #707 . The current clipping is performed on any accumulated reward. Because the number of accumulation steps can vary, this clipping value is not ideal. To have similar semantics with reward clipping (on individual rewards) in other typical algorithms, now the clip value is multiplied with `repeats`.  Although this fix is still not equivalent to first clipping individual rewards and then accumulating them, it's done for code simplicity. This fix improves the algorithm's performance.

### DIFF
--- a/alf/algorithms/dynamic_action_repeat_agent.py
+++ b/alf/algorithms/dynamic_action_repeat_agent.py
@@ -23,6 +23,7 @@ from alf.algorithms.algorithm import Algorithm
 from alf.algorithms.off_policy_algorithm import OffPolicyAlgorithm
 from alf.algorithms.sac_algorithm import SacAlgorithm
 from alf.algorithms.config import TrainerConfig
+from alf.algorithms.data_transformer import RewardNormalizer
 from alf.data_structures import TimeStep, Experience, namedtuple, AlgStep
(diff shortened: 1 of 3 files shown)


alf/algorithms/dynamic_action_repeat_agent.py
```diff
@@ -346,7 +345,10 @@ def preprocess_experience(self, rl_exp):
 
             clip = self._reward_normalizer._clip_value
             if clip > 0:
-                reward = torch.clamp(reward, -clip, clip)
+                # The clip value is for single-step rewards, so we need to multiply
+                # it with the repeated steps.
+                clip = clip * repeats
+                reward = torch.max(torch.min(clip, reward), -clip)
```
emailweixu, line 354: why not keep using clamp()?

hnyu: > why not keep using clamp()?

clamp only supports scalar min and max as I know. After multiplying `repeats`, `clip` is a tensor.

emailweixu (approved):

PR #763, opened 2021-01-27
PID controller fix

Haichao-Zhang:


emailweixu (approved):

alf/environments/carla_controller.py
```diff
@@ -63,7 +63,7 @@ def step(self, current, target):
         else:
             derivative = 0.
         self._prev_error = error
-        return self._K_P * error + self._K_I * self._integral + self._K_P * derivative
+        return self._K_P * error + self._K_I * self._integral + self._K_D * derivative
```
hnyu, line 66: Does this affect the performance of using PID controller in CARLA?

Haichao-Zhang: I'm running experiments using the fix one.

PR #764, opened 2021-01-27
Alf config utilities

### LINKED ISSUE #761: Config using python instead of gin
Thera are several disadvantage of gin:
1. gin has its own grammar. Which require some learning.
2. does not support full syntax of python, which make it difficult to do simple things such as "a + b" in gin
3. Error reporting is poor.

So we want to make it possible to do all configuration using python.

emailweixu:
I only changed some gin.configurable decorations in the code to alf.configurable
in this change so that the example ppo_babyai.py can run successfully.

TODO: modify grid_search.py, train_play_test.py to support alf config

### DIFF
--- a/alf/bin/train.py
+++ b/alf/bin/train.py
@@ -47,7 +47,6 @@ from absl import flags
 from absl import logging
 import gin
 import os
-import runpy
 import torch
 
 from alf.utils import common
--- a/alf/bin/train.py
+++ b/alf/bin/train.py
@@ -73,6 +73,8 @@ def train_eval(ml_type, root_dir):
         root_dir (str): directory for saving summary and checkpoints
     """
     trainer_conf = policy_trainer.TrainerConfig(root_dir=root_dir)
+    trainer_conf.random_seed = common.set_random_seed(trainer_conf.random_seed)
+
     if ml_type == 'rl':
         trainer = policy_trainer.RLTrainer(trainer_conf)
     elif ml_type == 'sl':
--- a/alf/bin/train.py
+++ b/alf/bin/train.py
@@ -47,6 +47,7 @@ from absl import flags
 from absl import logging
 import gin
 import os
+import runpy
 import torch
 
 from alf.utils import common
@@ -57,7 +58,9 @@ flags.DEFINE_string('ml_type', 'rl', 'type of the learning task')
 flags.DEFINE_string('root_dir', os.getenv('TEST_UNDECLARED_OUTPUTS_DIR'),
                     'Root directory for writing logs/summaries/checkpoints.')
 flags.DEFINE_multi_string('gin_file', None, 'Paths to the gin-config files.')
+flags.DEFINE_string('conf', None, 'Path to the alf config file.')
 flags.DEFINE_multi_string('gin_param', None, 'Gin binding parameters.')
+flags.DEFINE_multi_string('conf_param', None, 'Config binding parameters.')
 
 FLAGS = flags.FLAGS
 
@@ -82,9 +85,9 @@ def train_eval(ml_type, root_dir):
 
 
 def main(_):
-    gin_file = common.get_gin_file()
     FLAGS.alsologtostderr = True
-    gin.parse_config_files_and_bindings(gin_file, FLAGS.gin_param)
+    conf_file = common.get_conf_file()
+    common.parse_conf_file(conf_file)
     train_eval(FLAGS.ml_type, FLAGS.root_dir)
 
 
--- a/alf/bin/train.py
+++ b/alf/bin/train.py
@@ -57,9 +57,9 @@ from alf.trainers import policy_trainer
 flags.DEFINE_string('ml_type', 'rl', 'type of the learning task')
 flags.DEFINE_string('root_dir', os.getenv('TEST_UNDECLARED_OUTPUTS_DIR'),
                     'Root directory for writing logs/summaries/checkpoints.')
-flags.DEFINE_multi_string('gin_file', None, 'Paths to the gin-config files.')
-flags.DEFINE_string('conf', None, 'Path to the alf config file.')
+flags.DEFINE_string('gin_file', None, 'Path to the gin-config file.')
 flags.DEFINE_multi_string('gin_param', None, 'Gin binding parameters.')
+flags.DEFINE_string('conf', None, 'Path to the alf config file.')
 flags.DEFINE_multi_string('conf_param', None, 'Config binding parameters.')
 
 FLAGS = flags.FLAGS
--- a/alf/config_helpers.py
+++ b/alf/config_helpers.py
@@ -1,0 +1,111 @@
+# Copyright (c) 2021 Horizon Robotics and ALF Contributors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Helper functions for config alf training.
+
+The main motivation is to give the access of observation_spec and action_spec,
+which are necessary for config some models. observation_spec and action_spec are
+only available after the environment is craeted. So we create an environment
+based TrainerConfig in the module.
+"""
+
+from alf.environments.utils import create_environment
+from alf.algorithms.config import TrainerConfig
+from alf.algorithms.data_transformer import create_data_transformer
+
+__all__ = [
+    'get_raw_observation_spec', 'get_observation_spec', 'get_action_spec',
+    'get_env'
+]
+
+_env = None
+_transformed_observation_spec = None
+
+
+def get_raw_observation_spec(field=None):
+    """Get the ``TensorSpec`` of observations provided by the global environment.
+
+    Note: you need to finish all the config for environments and
+    TrainerConfig.random_seed before using this function.
+
+    Args:
+        field (str): a multi-step path denoted by "A.B.C".
+    Returns:
+        nested TensorSpec: a spec that describes the observation.
+    """
+    env = get_env()
+    specs = env.observation_spec()
+    if field:
+        for f in field.split('.'):
+            specs = specs[f]
+    return specs
+
+
+def get_observation_spec(field=None):
+    """Get the spec of observation transformed by data transformers.
+
+    The data transformers are specified by ``TrainerConfig.data_transformer_ctor``.
+
+    Note: you need to finish all the config for environments and
+    TrainerConfig.data_transformer_ctor before using this function.
+
+    Args:
+        field (str): a multi-step path denoted by "A.B.C".
+    Returns:
+        nested TensorSpec: a spec that describes the observation.
+    """
+    global _transformed_observation_spec
+    if _transformed_observation_spec is None:
+        config = TrainerConfig(root_dir='')
+        env = get_env()
+        data_transformer = create_data_transformer(
+            config.data_transformer_ctor, env.observation_spec())
+        observation_spec = data_transformer.transformed_observation_spec
+        _transformed_observation_spec = observation_spec
+
+    specs = _transformed_observation_spec
+    if field:
+        for f in field.split('.'):
+            specs = specs[f]
+    return specs
+
+
+def get_action_spec():
+    """Get the specs of the tensors expected by ``step(action)`` of the
+    environment.
+
+    Note: you need to finish all the config for environments and
+    TrainerConfig.random_seed before using this function.
+
+    Returns:
+        nested TensorSpec: a spec that describes the shape and dtype of each tensor
+        expected by ``step()``.
+    """
+    env = get_env()
+    return env.action_spec()
+
+
+def get_env():
+    """Get the training environment.
+
+    Note: you need to finish all the config for environments and
+    TrainerConfig.random_seed before using this function.
+
+    Returns:
+        AlfEnvironment
+    """
+    global _env
+    if _env is None:
+        trainer_config = TrainerConfig(root_dir='')
+        _env = create_environment(seed=trainer_config.random_seed)
+    return _env
--- a/alf/config_helpers.py
+++ b/alf/config_helpers.py
@@ -24,8 +24,8 @@ from alf.algorithms.config import TrainerConfig
 from alf.algorithms.data_transformer import create_data_transformer
 
 __all__ = [
-    'get_raw_observation_spec', 'get_observation_spec', 'get_action_spec',
-    'get_env'
+    'close_env', 'get_raw_observation_spec', 'get_observation_spec',
+    'get_action_spec', 'get_env'
 ]
 
 _env = None
@@ -109,3 +109,16 @@ def get_env():
         trainer_config = TrainerConfig(root_dir='')
         _env = create_environment(seed=trainer_config.random_seed)
     return _env
+
+
+def close_env():
+    """Close the global environment.
+
+    This function will be automatically called by ``RLTrainer``.
+    """
+    global _env
+    global _transformed_observation_spec
+    if _env is not None:
+        _env.close()
+    _env = None
+    _transformed_observation_spec = None
--- a/alf/config_helpers.py
+++ b/alf/config_helpers.py
@@ -22,6 +22,7 @@ based TrainerConfig in this module.
 from alf.environments.utils import create_environment
 from alf.algorithms.config import TrainerConfig
 from alf.algorithms.data_transformer import create_data_transformer
+from alf.utils.common import set_random_seed
 
 __all__ = [
     'close_env', 'get_raw_observation_spec', 'get_observation_spec',
@@ -106,6 +107,8 @@ def get_env():
     global _env
     if _env is None:
         trainer_config = TrainerConfig(root_dir='')
+        trainer_config.random_seed = set_random_seed(
+            trainer_config.random_seed)
         _env = create_environment(seed=trainer_config.random_seed)
     return _env
 
--- a/alf/config_helpers.py
+++ b/alf/config_helpers.py
@@ -70,8 +70,7 @@ def get_observation_spec(field=None):
         env = get_env()
         data_transformer = create_data_transformer(
             config.data_transformer_ctor, env.observation_spec())
-        observation_spec = data_transformer.transformed_observation_spec
-        _transformed_observation_spec = observation_spec
+        _transformed_observation_spec = data_transformer.transformed_observation_spec
 
     specs = _transformed_observation_spec
     if field:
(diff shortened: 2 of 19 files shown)


alf/utils/common.py
```diff
@@ -429,6 +431,100 @@ def get_gin_file():
         return ''
 
 
+ALF_CONFIG_FILE = 'alf_config.py'
+
+
+def get_conf_file():
+    """Get the configuration file.
+
+    If ``FLAGS.conf`` is not set, find alf_config.py for configured.gin under
```
Haichao-Zhang: ```alf_config.py for configured.gin```, ```for```  should be ```or```?

alf/utils/common.py
```diff
@@ -429,6 +431,105 @@ def get_gin_file():
...
+    if conf_file.endswith(".gin"):
+        gin_params = flags.FLAGS.gin_param
+        gin.parse_config_files_and_bindings([conf_file], gin_params)
+    else:
+        conf_params = flags.FLAGS.conf_param
+        if conf_params:
+            for conf_param in conf_params:
+                pos = conf_param.find('=')
+                if pos == -1:
+                    raise ValueError("conf_params")
```
le-horizon: also include the faulty line in the error?

alf/config_util.py
```diff
@@ -0,0 +1,479 @@
...
+
+            def func(self, a, b):
+                ...
+
+    We can config in the following ways:
+
+    .. code-block::
+
+        alf.config('cool_func', cool_arg2='new_value', cool_arg2='another_value)
+        alf.config('Test.func', b=3)
```
le-horizon: Test.func definition is missing?

alf/bin/train.py
```diff
@@ -47,6 +47,7 @@
 from absl import logging
 import gin
 import os
+import runpy
```
hnyu: not used

alf/config_helpers.py
```diff
@@ -0,0 +1,124 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Helper functions for config alf training.
+
+The main motivation is to give the access of observation_spec and action_spec,
+which are necessary for config some models. observation_spec and action_spec are
+only available after the environment is craeted. So we create an environment
```
hnyu: created

alf/config_helpers.py
```diff
@@ -0,0 +1,124 @@
...
+    Returns:
+        nested TensorSpec: a spec that describes the observation.
+    """
+    global _transformed_observation_spec
+    if _transformed_observation_spec is None:
+        config = TrainerConfig(root_dir='')
+        env = get_env()
+        data_transformer = create_data_transformer(
+            config.data_transformer_ctor, env.observation_spec())
+        observation_spec = data_transformer.transformed_observation_spec
```
hnyu: observation_spec is not used?

alf/config_util.py
```diff
@@ -0,0 +1,480 @@
...
+            if param.default is not inspect.Parameter.empty:
+                config.set_default_value(param.default)
+
+    return configs
+
+
+def _add_to_conf_tree(module_path, func_name, para_name, node):
+    """
+    Args:
+        module_path (list[str]):
```
hnyu: missing description for the arguments.

alf/config_util.py
```diff
@@ -0,0 +1,480 @@
...
+        node (_Config):
+    """
+
+    tree = _CONF_TREE
+    path = module_path + func_name.split('.') + [para_name]
+    names = []
+    for name in reversed(path[1:]):
+        if not isinstance(tree, dict):
+            raise ValueError("'%s' conflicts with existing config name '%s' "
```
hnyu: This error message is confusing. Can be `"%s path is not valid; only found a partial path up to %s" % ('.'join(path), '.'.join(reversed(names)))`

alf/config_util.py
```diff
@@ -0,0 +1,480 @@
...
+            ...
+
+    In this case, the function is associated with the name
+    'my_function' in the global configuration, and both param1 and param2 are
+    configurable.
+
+    The decorator can be supplied with parameters to specify the configurable name
+    or supply a whitelist/blacklist:
+
+        @alf.configurable('my_func', whitelist='param2')
```
hnyu: whitelist=['param2']

alf/config_util.py
```diff
@@ -0,0 +1,480 @@
...
+
+        @alf.configurable
+        class MyClass(object):
+            def __init__(self, param1, param2='a default value'):
+            ...
+
+    In this case, the name of the configurable is 'MyClass', and both `param1`
+    and `param2` are configurable.
+
+    Note: currently, to maitain the compatibility with gin-config, all the
```
hnyu: maintain

alf/config_util.py
```diff
@@ -0,0 +1,480 @@
...
+        functions decorated using alf.configurable are auomatically configurable
```
hnyu: automatically

alf/examples/ppo_babyai.py
```diff
@@ -0,0 +1,140 @@
...
+)
+
+observation_spec = alf.get_observation_spec()
+action_spec = alf.get_action_spec()
+mission_spec = observation_spec['observation']['mission']
+vocab_size = mission_spec.maximum + 1
+
+encoding_dim = 128
+activation = torch.relu_
+fc_layers_params = (256, 256)
```
hnyu, line 62: Will these global variables won't interfere with variables of the same names in python code?

alf/config_helpers.py
```diff
@@ -0,0 +1,124 @@
...
+    """
+    env = get_env()
+    return env.action_spec()
+
+
+def get_env():
+    """Get the training environment.
+
+    Note: you need to finish all the config for environments and
+    TrainerConfig.random_seed before using this function.
```
le-horizon, line 101: Is there a simple way to ensure this in code?  e.g. assert TrainerConfig is properly set?  Or is the assert in _Config.get_effective_value() already doing that?

alf/config_util.py
```diff
@@ -0,0 +1,480 @@
...
+        if not isinstance(v, dict):
+            leaves.append((k, v))
+        else:
+            leaves.extend(
+                [(name + '.' + k, node) for name, node in _get_all_leaves(v)])
+    return leaves
+
+
+class _Config(object):
+    """Object prepresenting one configurable value."""
```
le-horizon: ->representing

alf/config_helpers.py
```diff
@@ -0,0 +1,124 @@
...
+    Returns:
+        nested TensorSpec: a spec that describes the shape and dtype of each tensor
+        expected by ``step()``.
+    """
+    env = get_env()
+    return env.action_spec()
+
+
+def get_env():
+    """Get the training environment.
```
Haichao-Zhang: global environment?

alf/config_util.py
```diff
@@ -0,0 +1,480 @@
...
+    'get_operative_configs', 'get_inoperative_configs'
+]
+
+
+def config(prefix_or_dict, replacing_existing_config=False, **kwargs):
+    """Set the values for the configs with given name as suffix.
+
+    Example:
+
+    Asumme we have the following decorated functions and classes:
```
Haichao-Zhang: Asumme  -> Assume

alf/config_util.py
```diff
@@ -0,0 +1,480 @@
...
+                ...
+
+            def func(self, a, b):
+                ...
+
+    We can config in the following ways:
+
+    .. code-block::
+
+        alf.config('cool_func', cool_arg2='new_value', cool_arg2='another_value)
```
Haichao-Zhang: 'another_value -> 'another_value'

alf/config_util.py
```diff
@@ -0,0 +1,480 @@
...
+        alf.config('Test.func', b=3)
```
Haichao-Zhang: ``Test.func``  missing in the code-block?

alf/config_util.py
```diff
@@ -0,0 +1,480 @@
...
+        alf.config('func', b=3)     # 'Test.func' can be uniquely identified by 'func'
```
Haichao-Zhang: Test.func

alf/config_util.py
```diff
@@ -0,0 +1,480 @@
...
+    tree = _CONF_TREE
+    path = config_name.split('.')
+    for name in reversed(path):
+        if not isinstance(tree, dict) or name not in tree:
+            raise ValueError("Cannot find config name %s" % config_name)
+        tree = tree[name]
+
+    leaves = _get_all_leaves(tree)
+    if len(leaves) > 1:
+        # only show atmost 3 ambiguous choices
```
Haichao-Zhang: atmost -> at most

alf/config_util_test.py
```diff
@@ -0,0 +1,148 @@
...
+        alf.config1('test_func.c', 15)
+        self.assertEqual(test_func(10, 20), (10, 20, 15))
+
+        alf.config("FancyTest", arg=3)
+        self.assertEqual(test_func4(), 3)
+
+        # Test duplicated config
+        with self.assertLogs() as ctx:
+            alf.config1('test_func2.c', 15)
+            alf.config1('test_func2.c', 16)
```
Haichao-Zhang, line 157: This seems to be different from the behavior of gin, where the later config will overwrite the prior one?
By using replacing_existing_config flag, it should be safer.

Haichao-Zhang (changes requested):

alf/bin/train.py
emailweixu: removed

alf/config_helpers.py
emailweixu: Feel it should not be past tense.

alf/config_helpers.py
emailweixu: merged with the next line

alf/config_helpers.py
emailweixu: added.

alf/config_util.py
emailweixu: fixed.

alf/config_util.py
emailweixu: fixed

alf/config_util.py
emailweixu: changed to Worker.func

alf/config_util.py
emailweixu: changed to Worker.func

alf/config_util.py
emailweixu: added

alf/config_util.py
emailweixu: I feel the current one seems more appropriate. For example, suppose we want to add "algorithm.vae.test.b", but "vae.test.b" has already been added. The message will be:

'algorithm.vae.test.b' conflicts with existing config name 'vae.test.b'

which seems to a correct description of the problem.

alf/config_util.py
emailweixu: fixed

alf/config_util.py
emailweixu: fixed

alf/config_util_test.py
emailweixu: Right. The reason for this is to be able to correctly implement '--conf_param' commandline flag. In order for --conf_param flag to work, we have to set it before parsing conf file because the conf file may depends on the value configured through conf_param. Hence we cannot allow replacing_existing_config in the conf file.

And it seems to me it is hard to say replacing_existing_config=True is safer than False. Because there is no way to tell which config the user really want. Anyway, a warning will be generated.

alf/utils/common.py
emailweixu: fixed

Haichao-Zhang (dismissed):

alf/config_helpers.py
hnyu: I mean the typo craeted -> created

alf/config_util_test.py
```diff
@@ -0,0 +1,151 @@
...
+        self.assertEqual(obj2(), (8, 11, 8))
+
+        # Test subclass using constructor from base
+        obj3 = Test3(1, 2)
+        self.assertEqual(obj3(), (0, 0, 7))
+
+        logging.info("get_operative_configs(): \n%s" % pprint.pformat(
+            alf.get_operative_configs()))
+        logging.info("get_inoperative_configs(): \n%s" % pprint.pformat(
+            alf.get_inoperative_configs()))
```
le-horizon: add asserts for the set of operative config names and inoperative config names equal the correct lists of names.

alf/config_util.py
hnyu: I'm actually not sure about the intended usage here. Suppose we have a function called "B.a", and another function "A.B.a". So basically if I add "A.B.a" first and then "B.a", there will be no error. But I can't do in the reversed order?

le-horizon (commented):
Thanks for the nice and clean change and awesome improvement to configs!
I like the clean format of the new config a lot.

Some nitpicks inlined.

Also, I guess the following aren't supported yet?
 config scope,
 include another config file.

Thanks,
Le

alf/config_util_test.py
le-horizon: Can command line --conf_param override existing configs in the conf file?
My personal habit is to have the conf file being self sufficient (so it can be used standalone), and have conf_param provide ways of customizing configs based on the conf file (e.g. running one off experiments tweaking some configs).
Does that make sense?

alf/config_helpers.py
emailweixu: If a value is used but configured later, there will be an exception in config1()
```python
raise ValueError(
            "Config '%s' has already been used. You should config "
            "its value before using it." % config_name)
```

emailweixu:
> Also, I guess the following aren't supported yet?
> config scope,

no config scope. Because we are using python now, it's quite easy to achieve the same effects.

> include another config file.

you can always import another python module

alf/config_util_test.py
emailweixu: Right. --conf_param override existing configs from the conf file.

alf/utils/common.py
emailweixu: updated.

le-horizon:
> > Also, I guess the following aren't supported yet?
> > config scope,
> 
> no config scope. Because we are using python now, it's quite easy to achieve the same effects.

Oh, I see, just provide different args to the objects, e.g.
```
a = Class(arg=a); b = Class(arg=b)
```
> 
> > include another config file.
> 
> you can always import another python module

Make sense.

alf/config_util_test.py
emailweixu: added.

alf/config_util.py
emailweixu: fixed

alf/config_util.py
emailweixu: fixed

alf/config_util_test.py
le-horizon: Then I don't quite understand why some confs in the conf file may depend on --conf_param values?

le-horizon (dismissed):
Haven't closely examined the config_util.py.  The rest LGTM.

Le

Haichao-Zhang (dismissed):

alf/config_helpers.py
```diff
@@ -106,6 +107,8 @@ def get_env():
     global _env
     if _env is None:
         trainer_config = TrainerConfig(root_dir='')
+        trainer_config.random_seed = set_random_seed(
+            trainer_config.random_seed)
```
hnyu: If the random seed is set here, we need to guarantee that no other env is created before the first call of get_env(). This might be problematic. I think a better way is still call set_random_seed at the entry of the main function.

alf/config_util.py
```diff
@@ -448,6 +471,15 @@ def __init__(self, param1, param2='a default value'):
...
     and `param2` are configurable.
 
+    The full name of a configurable value is MODULE_PATH.FUNC_NAME.ARG_NAME. It
+    can be referred using any suffixes as long as there is no ambiguity. For
+    example, assuming there are two configurable values "abc.def.func.a" and
+    "xyz.uvw.func.a", you can use "abc.def.func.a", "def.func.a", "xyz.uvw.func.a"
+    or "uvw.func.a" to refer these two configurable values. You cannot use
+    "func.a" because of the ambiguity. Because of this, you cannot have a config
+    name which is the strict suffix of another config name. For example,
+    "A.Test.arg" and "Test.arg" cannot both be defined.
```
hnyu: What should we do if there is a global function called "a" we want to configure, but there is also another class function "A.a"?

hnyu: I think it would be good to add a section in ALF notes explaining ALF config name scheme on RTD once this PR is merged.

emailweixu: You can supply a different name:
```python
@configurable("global_a")
def a(arg):
   ...
```
Now the config name is "global_a.arg". This usage is described in the docstring.

alf/config_helpers.py
emailweixu: moved to entry

hnyu (dismissed):

alf/config_util.py
```diff
@@ -478,7 +478,21 @@ def __init__(self, param1, param2='a default value'):
     or "uvw.func.a" to refer these two configurable values. You cannot use
     "func.a" because of the ambiguity. Because of this, you cannot have a config
     name which is the strict suffix of another config name. For example,
-    "A.Test.arg" and "Test.arg" cannot both be defined.
+    "A.Test.arg" and "Test.arg" cannot both be defined. You can supply a different
+    name for the function to avoid conflict:
+
+    .. code-block:: pytyon
```
Haichao-Zhang: pytyon -> python

alf/config_util.py
```diff
@@ -478,7 +478,21 @@ def __init__(self, param1, param2='a default value'):
...
+        ``@alf.configurable("NewTest")
+        def Test(arg):
+            ...
+
+    or
+
+    .. code-block:: pytyon
```
Haichao-Zhang: pytyon -> python

alf/config_util.py
```diff
@@ -478,7 +478,21 @@ def __init__(self, param1, param2='a default value'):
     or "uvw.func.a" to refer these two configurable values. You cannot use
     "func.a" because of the ambiguity. Because of this, you cannot have a config
     name which is the strict suffix of another config name. For example,
-    "A.Test.arg" and "Test.arg" cannot both be defined.
+    "A.Test.arg" and "Test.arg" cannot both be defined. You can supply a different
+    name for the function to avoid conflict:
+
+    .. code-block:: pytyon
+        ``@alf.configurable("NewTest")
```
Haichao-Zhang: `` can be removed?

alf/config_util.py
```diff
@@ -478,7 +478,21 @@ def __init__(self, param1, param2='a default value'):
...
+        def Test(arg):
+            ...
+
+    or
+
+    .. code-block:: pytyon
+        ``@alf.configurable("B.Test")
```
Haichao-Zhang: ``

Haichao-Zhang (commented):
Some minor issues.

Haichao-Zhang (approved):

PR #765, opened 2021-01-27
char mode for suite BabyAI

emailweixu:
The ppo_babyai.py example in #764 needs this.

### DIFF
--- a/alf/environments/suite_babyai.py
+++ b/alf/environments/suite_babyai.py
@@ -35,7 +35,7 @@ def is_available():
 def load(environment_name,
          env_id=None,
          max_instruction_length=80,
-         one_token_per_step=False,
+         mode='sent',
          discount=1.0,
          max_episode_steps=None,
          gym_env_wrappers=(),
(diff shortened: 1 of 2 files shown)


alf/environments/suite_babyai.py
```diff
@@ -186,15 +187,19 @@ def __init__(self,
...
         """Convert instruction string to a numpy array."""
+        if self._mode == 'char':
+            tokens = np.array([ord(c) for c in instruction])
+            return tokens
+
         tokens = self._word_pattern.findall(instruction.lower())
-        instr = np.array([self._vocab.get(token, 0) for token in tokens])
-        if np.amin(instr) == 0:
+        tokens = np.array([self._vocab.get(token, 0) for token in tokens])
+        if np.amin(tokens) == 0:
```
le-horizon: This is an unsafe rename -- collides with the line below which also references "tokens" as the words before being translated into word_ids.
Let's name it token_ids or something like that?

le-horizon (commented):
Looks almost good, just one thing to fix.

Thanks,
Le

alf/environments/suite_babyai.py
```diff
@@ -172,8 +171,10 @@ def __init__(self,
             'mission':
                 gym.spaces.MultiDiscrete([vocab_size] * max_instruction_length)
         }
-        if one_token_per_step:
+        if mode == 'word':
             obs_space['mission'] = gym.spaces.Discrete(vocab_size)
+        elif mode == 'char':
+            obs_space['mission'] = gym.spaces.Discrete(128)
```
Haichao-Zhang, line 179: Might be able to add a comment on using UTF-8 here or in the ```Arg:``` section?

alf/environments/suite_babyai.py
emailweixu: fixed.

alf/environments/suite_babyai.py
emailweixu: Added

Haichao-Zhang (approved):

PR #766, opened 2021-01-28
Use "is ()" to test empty tuple

emailweixu:
"x == ()" is often problematic since if x is a Tensor, it will crash.


le-horizon (dismissed):
Good cleanup!

Do we also need to translate all "!= ()" into "is not ()"?

Le

emailweixu:
> Good cleanup!
> 
> Do we also need to translate all "!= ()" into "is not ()"?
> 
> Le

Good point

le-horizon (approved):
LG!

PR #767, opened 2021-01-29
Simple Carla controller

Haichao-Zhang:
This PR implements a simple controller for Carla.
It simply uses the navigation signal plus the PID controller. 

A success run without other vehicles on the street is shown below:

ezgif com-gif-maker (8)

### DIFF
--- a/alf/algorithms/handcrafted_algorithm.py
+++ b/alf/algorithms/handcrafted_algorithm.py
@@ -183,8 +183,7 @@ class SimpleCarlaAlgorithm(HandcraftedAlgorithm):
 
         # here we adjust the speed based on the distance to goal
         action[distance_to_goal > self._distance_to_decelerate, 0] = 1
-        # logical_and is introduced in torch 1.5.0, use * instead
-        ind = (distance_to_goal > self._distance_to_stop) * (
+        ind = (distance_to_goal > self._distance_to_stop) & (
             distance_to_goal <= self._distance_to_decelerate)
         action[ind, 0] = distance_to_goal[ind] / self._distance_to_decelerate
         action[distance_to_goal <= self._distance_to_stop, 0] = 0
--- a/alf/algorithms/handcrafted_algorithm.py
+++ b/alf/algorithms/handcrafted_algorithm.py
@@ -189,7 +189,7 @@ class SimpleCarlaAlgorithm(HandcraftedAlgorithm):
         action[ind, 0] = distance_to_goal[ind] / self._distance_to_decelerate
         action[distance_to_goal <= self._distance_to_stop, 0] = 0
 
-        # direction is computated based on the waypoint
+        # direction is computed based on the waypoint
         action[:, 1] = direction
 
         return action
--- a/alf/algorithms/handcrafted_algorithm.py
+++ b/alf/algorithms/handcrafted_algorithm.py
@@ -1,0 +1,150 @@
+# Copyright (c) 2021 Horizon Robotics and ALF Contributors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


alf/algorithms/handcrafted_algorithm.py
```diff
@@ -0,0 +1,150 @@
...
+        """A naive hand-crafted policy for Carla environment.
+
+        Args:
+            observation (nested Tensor): input observation that is compatible
+                with observation_spec
+        Returns
+            Tensor: action computed based on the observation
+        """
+        nav = alf.nest.get_field(observation, 'observation.navigation')
+        goal = alf.nest.get_field(observation, 'observation.goal')
```
emailweixu, line 163: perhaps add a comment about the data in nav to make the code more understandable.

alf/algorithms/handcrafted_algorithm.py
```diff
@@ -0,0 +1,150 @@
...
+
+        waypoints = nav
+        if waypoints.shape[-1] > 1:
+            wp_vector = waypoints[:, 1]
+        else:
+            wp_vector = waypoints[:, 0]
```
emailweixu, line 168: The shape of waypoints is [B, k, 3], where B is the batch size, k is the number of way points.
Should the if statement be `if waypoints.shape[1] > 1`?

alf/algorithms/handcrafted_algorithm.py
```diff
@@ -0,0 +1,150 @@
...
+import alf
+from alf.algorithms.config import TrainerConfig
+from alf.algorithms.off_policy_algorithm import OffPolicyAlgorithm
+from alf.data_structures import AlgStep, Experience, LossInfo, TimeStep
+from alf.tensor_specs import BoundedTensorSpec
+
+
+@gin.configurable
+class HandcraftedAlgorithm(OffPolicyAlgorithm):
+    """An Algorithm with handcrafted computational logic.
```
emailweixu: Comment that a concrete algorithm should subclass from this and implement _policy_func

emailweixu (changes requested):

alf/algorithms/handcrafted_algorithm.py
Haichao-Zhang: Comments added

alf/algorithms/handcrafted_algorithm.py
Haichao-Zhang: Bug fixed.

alf/algorithms/handcrafted_algorithm.py
Haichao-Zhang: Added comments on the data in nav, goal etc to provide more context to the code.

alf/algorithms/handcrafted_algorithm.py
```diff
@@ -0,0 +1,195 @@
...
+        # reverse: values greater than 0.5 corrsponding to going backward.
+
+        action = torch.zeros(waypoints.shape[0], self._action_spec.shape[0])
+
+        distance_to_goal = torch.norm(goal, dim=1)
+
+        # here we adjust the speed based on the distance to goal
+        action[distance_to_goal > self._distance_to_decelerate, 0] = 1
+        # logical_and is introduced in torch 1.5.0, use * instead
+        ind = (distance_to_goal > self._distance_to_stop) * (
```
emailweixu: FYI: & can be used

emailweixu (dismissed):

alf/algorithms/handcrafted_algorithm.py
Haichao-Zhang: Good suggestion. I have removed the comment and changed to &

emailweixu (approved):

PR #770, opened 2021-01-29
Fix max_episode_length related issue for metrics in play

Haichao-Zhang:
When using ```alf.bin.play```, depends on the value of ```max_episode_length```, the episodic information might not be recorded correctly by the metrics if the step type of the last step is not ```LAST```.

This issue will cause mis-leading metric numbers (e.g. lower average return) for interpreting the algorithm performance.

### DIFF
--- a/alf/trainers/policy_trainer.py
+++ b/alf/trainers/policy_trainer.py
@@ -753,7 +753,7 @@ def play(root_dir,
             # change the step_type to LAST before being observed by metrics
             # to ensure the episodic information will be updated correctly
             time_step = time_step._replace(
-                step_type=torch.ones_like(time_step.step_type) * StepType.LAST)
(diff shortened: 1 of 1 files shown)


alf/trainers/policy_trainer.py
```diff
@@ -749,6 +750,10 @@ def play(root_dir,
             episode_reward = 0.
             episode_length = 0.
             episodes += 1
+            # change the step_type to LAST before being observed by metrics
+            # to ensure the episodic information will be updated correctly
+            time_step = time_step._replace(
+                step_type=torch.ones_like(time_step.step_type) * StepType.LAST)
```
emailweixu: use torch.full()

Haichao-Zhang: Changed

emailweixu (approved):

PR #771, opened 2021-01-29
Better handling of --conf_param

emailweixu:
1. The config from commandline is immutable.
2. Allow changing a mutable config.
3. Configs from the --conf_param are recorded.

### DIFF
--- a/alf/config_util.py
+++ b/alf/config_util.py
@@ -71,7 +71,8 @@ def config(prefix_or_dict, mutable=True, **kwargs):
         mutable (bool): whether the config can be changed later. If the user
             tries to change an existing immutable config, the change will be
             ignored and a warning will be generated. You can always change a
-            mutable config.
+            mutable config. ``ValueError`` will be raised if trying to set a new
+                immutable value to an existing immutable value.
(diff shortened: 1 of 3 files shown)


alf/config_util.py
```diff
@@ -180,15 +188,16 @@ def is_used(self):
...
 
     Args:
         config_name (str): name of the config
         value (any): value of the config
-        replacing_existing_config (bool): whether to replace with the provided
-            new value if the value has been configured before. A warning will be
-            generated when trying to set a value which has been set previously.
+        mutable (bool): whether the config can be changed later. If the user
+            tries to change an existing immutable config, the change will be
+            ignored and a warning will be generated. You can always change a
```
Haichao-Zhang, line 200: Can add comments on the raised ValueError when trying to change immutable to another immutable. Similar for ```config()```

le-horizon (dismissed):
LG.  Thanks for the quick improvements!

Le

alf/config_util.py
emailweixu: added.

Haichao-Zhang (approved):

PR #774, opened 2021-01-30
Avoid creating TrainerConfig during parse_conf_file()

emailweixu:
Before this change, get_observation() will create a TrainerConfig(). This makes the configurable
values of TrainerConfig to be all labelled as used. And it will not be allowed to change after get_observation().

Now we use get_config() to get the value of TrainerConfig.data_transformer_ctor
to avoid creating TrainerConfig object. So the configs of other TrainerConfig values
can still be changed after get_observation_spec().

Also changed sac_carla.gin to use alf-config

### DIFF
--- a/alf/bin/train.py
+++ b/alf/bin/train.py
@@ -85,11 +85,6 @@ def train_eval(ml_type, root_dir):
 
 def main(_):
     FLAGS.alsologtostderr = True
-    random_seed = alf.get_config('TrainerConfig.random_seed')
-    alf.config(
-        'TrainerConfig',
-        random_seed=common.set_random_seed(random_seed),
-        raise_if_used=False)
     conf_file = common.get_conf_file()
     try:
         common.parse_conf_file(conf_file)
--- a/alf/bin/train.py
+++ b/alf/bin/train.py
@@ -73,8 +73,6 @@ def train_eval(ml_type, root_dir):
         root_dir (str): directory for saving summary and checkpoints
     """
     trainer_conf = policy_trainer.TrainerConfig(root_dir=root_dir)
-    trainer_conf.random_seed = common.set_random_seed(trainer_conf.random_seed)
-
     if ml_type == 'rl':
         trainer = policy_trainer.RLTrainer(trainer_conf)
     elif ml_type == 'sl':
@@ -87,9 +85,17 @@ def train_eval(ml_type, root_dir):
 
 def main(_):
     FLAGS.alsologtostderr = True
+    random_seed = alf.get_config('TrainerConfig.random_seed')
+    alf.config(
+        'TrainerConfig',
+        random_seed=common.set_random_seed(random_seed),
+        raise_if_used=False)
     conf_file = common.get_conf_file()
-    common.parse_conf_file(conf_file)
-    train_eval(FLAGS.ml_type, FLAGS.root_dir)
+    try:
+        common.parse_conf_file(conf_file)
+        train_eval(FLAGS.ml_type, FLAGS.root_dir)
+    finally:
+        alf.close_env()
 
 
 if __name__ == '__main__':
--- a/alf/config_helpers.py
+++ b/alf/config_helpers.py
@@ -19,9 +19,10 @@ only available after the environment is created. So we create an environment
 based TrainerConfig in this module.
 """
 
-from alf.environments.utils import create_environment
 from alf.algorithms.config import TrainerConfig
 from alf.algorithms.data_transformer import create_data_transformer
+from alf.config_util import get_config
+from alf.environments.utils import create_environment
 
 __all__ = [
     'close_env', 'get_raw_observation_spec', 'get_observation_spec',
@@ -66,10 +67,11 @@ def get_observation_spec(field=None):
     """
     global _transformed_observation_spec
     if _transformed_observation_spec is None:
-        config = TrainerConfig(root_dir='')
+        data_transformer_ctor = get_config(
+            'TrainerConfig.data_transformer_ctor')
         env = get_env()
-        data_transformer = create_data_transformer(
-            config.data_transformer_ctor, env.observation_spec())
+        data_transformer = create_data_transformer(data_transformer_ctor,
+                                                   env.observation_spec())
         _transformed_observation_spec = data_transformer.transformed_observation_spec
 
     specs = _transformed_observation_spec
@@ -105,8 +107,8 @@ def get_env():
     """
     global _env
     if _env is None:
-        trainer_config = TrainerConfig(root_dir='')
-        _env = create_environment(seed=trainer_config.random_seed)
+        random_seed = get_config('TrainerConfig.random_seed')
+        _env = create_environment(seed=random_seed)
     return _env
 
 
--- a/alf/config_helpers.py
+++ b/alf/config_helpers.py
@@ -23,7 +23,7 @@ import runpy
 
 from alf.algorithms.config import TrainerConfig
 from alf.algorithms.data_transformer import create_data_transformer
-from alf.config_util import config1, get_config_value
+from alf.config_util import config1, get_config_value, pre_config, validate_pre_configs
 from alf.environments.utils import create_environment
 from alf.utils.common import set_random_seed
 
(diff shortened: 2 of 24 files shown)


alf/environments/suite_carla.py
```diff
@@ -1027,7 +1027,7 @@ def __init__(self,
         Args:
             batch_size (int): the number of learning vehicles.
             map_name (str): the name of the map (e.g. "Town01")
-            vehicle_filter (str): the filter for getting vehicle blueprints.
+            vehicle_filter (str): the filter for getting ego vehicle blueprints.
```
hnyu: I think 'ego' is a little confusing here. Might just explain that this filter is used for vehicles that will be trained to drive. And the other vehicles will use a catch-all filter.

alf/bin/train.py
```diff
@@ -87,9 +85,17 @@ def train_eval(ml_type, root_dir):
...
+    random_seed = alf.get_config('TrainerConfig.random_seed')
+    alf.config(
+        'TrainerConfig',
+        random_seed=common.set_random_seed(random_seed),
+        raise_if_used=False)
     conf_file = common.get_conf_file()
-    common.parse_conf_file(conf_file)
-    train_eval(FLAGS.ml_type, FLAGS.root_dir)
+    try:
+        common.parse_conf_file(conf_file)
```
le-horizon, line 90: will it raise error if parse_conf_file tries to set random seed again?

alf/bin/train.py
```diff
@@ -87,9 +85,17 @@ def train_eval(ml_type, root_dir):
 
 def main(_):
     FLAGS.alsologtostderr = True
+    random_seed = alf.get_config('TrainerConfig.random_seed')
```
le-horizon: will this only read from the command line config?  What's the reason for making this change: first get random_seed, then set it and set config, then parse conf file?  It seems complicated and prone to error?

le-horizon (commented):
Some values need to be configured before parse_alf_config, so the config and initialization of alf seems intertwined, and a bit hard to follow.  We may need to create some documentation for this, in case we forget in the future?

alf/config_util.py
```diff
@@ -189,20 +194,8 @@ def is_used(self):
...
-        mutable (bool): whether the config can be changed later. If the user
-            tries to change an existing immutable config, the change will be
-            ignored and a warning will be generated. You can always change a
-            mutable config. ``ValueError`` will be raised if trying to set a new
-            immutable value to an existing immutable value.
-
-
-    """
+def _get_config_node(config_name):
+    """Get the _Config corresponding to config_name."""
```
Haichao-Zhang: Does ```_Config``` has any special meaning here? Maybe can change to ```config node```.

alf/config_util.py
```diff
@@ -224,7 +217,26 @@ def config1(config_name, value, mutable=True):
...
+            tries to change an existing immutable config, the change will be
+            ignored and a warning will be generated. You can always change a
+            mutable config. ``ValueError`` will be raised if trying to set a new
+            immutable value to an existing immutable value.
+        raise_if_used (bool): If True, ValueError will be raised if trying to
+            config a value which has already been used.
+    """
+    config_node = _get_config_node(config_name)
+
+    if raise_if_used and config_node.is_used():
```
Haichao-Zhang, line 240: When raise_if_used is False,  maybe we can generate a warning message

alf/config_util.py
```diff
@@ -250,6 +262,28 @@ def config1(config_name, value, mutable=True):
         config_node.set_mutable(mutable)
 
 
+def get_config(config_name):
```
Haichao-Zhang: Maybe we can change the name of this function to ```get_config_value```?

alf/bin/train.py
emailweixu: moved all random_seed ralated code to get_env

alf/config_util.py
emailweixu: _Config is the class name

alf/config_util.py
emailweixu: With the current way using raise_if_used=True inside config_helper, the config is always used, and a warning message will be quite confusing.

alf/config_util.py
emailweixu: changed

alf/environments/suite_carla.py
emailweixu: Updated.

alf/bin/train.py
emailweixu: moved to config_helpers.get_env() and commented there.

alf/examples/sac_carla_conf.py
```diff
@@ -0,0 +1,206 @@
...
+        mini_batch_length=4,
+        data_transformer_ctor=[ImageScaleTransformer, ObservationNormalizer])
+
+alf.config(
+    'suite_carla.Player',
+    # Not yet able to successfully train with sparse reward.
+    sparse_reward=False,
+    # Cannot train well with allow_negative_distance_reward=False yet.
+    allow_negative_distance_reward=True,
+    # Currently, even a small penalty such as one make the training much worse
```
le-horizon: make->makes

alf/examples/ac_cart_pole_conf.py
```diff
@@ -0,0 +1,53 @@
...
+# reward scaling
+alf.config('TrainerConfig', data_transformer_ctor=RewardScaling)
+alf.config('RewardScaling', scale=0.01)
+
+# algorithm config
+alf.config('ActorDistributionNetwork', fc_layer_params=(100, ))
+alf.config('ValueNetwork', fc_layer_params=(100, ))
+alf.config(
+    'ActorCriticAlgorithm',
+    optimizer=alf.optimizers.Adam(lr=1e-3, gradient_clipping=10.0))
```
le-horizon, line 33: It's fine as is.  Does it make sense to have lr configured using a same named variable in TrainerConfig?

alf/config_helpers.py
```diff
@@ -94,19 +99,71 @@ def get_action_spec():
...
     global _env
     if _env is None:
-        trainer_config = TrainerConfig(root_dir='')
-        _env = create_environment(seed=trainer_config.random_seed)
+        if _is_parsing:
+            random_seed = get_config_value('TrainerConfig.random_seed')
+        else:
+            # We construct a TrainerConfig object here so that the value
+            # configured through gin-config can be properly retrieved.
+            train_config = TrainerConfig(root_dir='')
```
le-horizon, line 159: I'm confused.  This config_helper.py is only used by alf config, not gin-config?  Why do we need to care about gin-config here?

alf/examples/sac_carla_conf.py
```diff
@@ -0,0 +1,206 @@
...
+    continuous_projection_net_ctor=proj_net)
+
+env = alf.get_env()
+
+critic_network_cls = partial(
+    alf.networks.CriticNetwork,
+    joint_fc_layer_params=fc_layers_params,
+    activation=activation,
+    use_fc_bn=use_batch_normalization,
+    output_tensor_spec=env.reward_spec())
```
le-horizon, line 115: maybe have a common.get_reward_spec() function, and hide alf.get_env() inside it?

alf/bin/train.py
le-horizon: Looks nice!

le-horizon (commented):
Accidentally submitted the partial review.
Also took a look at the new commit.

It's much cleaner now, only a few nits/suggestions if they make sense.

Thanks,
Le

alf/examples/sac_carla_conf.py
emailweixu: removed this comment as it is not accurate anymore

alf/examples/sac_carla_conf.py
emailweixu: There are many other env member function. It is not a good idea to do this for all of them. The reason of having get_observation_spec() is that it has additional handling of data_transformer.

alf/config_helpers.py
emailweixu: Because get_env() will be used by policy_trainer to get the global env even when gin-config is used for configuration.

alf/examples/ac_cart_pole_conf.py
emailweixu: There could be multiple optimizers.

le-horizon (approved):
LG.

Haichao-Zhang (approved):

PR #777, opened 2021-02-02
better curve drawing and tb smoothing

hnyu:
This PR improves the current curve drawing by: 

1. Avoid using TB's SIZE_GUIDANCE["scalars"] because when showing training curves we have far more points before the first summary_interval (if summary_interval is very big). Letting TB take a few points will result in very low curve resolution after the first summary_interval because TB will evenly downsample the points. Now we just load all points and only downsample after fitting them to a function.
2. Add TB's smoothing function which is an EMA style smoothing.
3. Some minor visual adjustment for the plotting function.

### DIFF
--- a/alf/utils/plot_tb_curves.py
+++ b/alf/utils/plot_tb_curves.py
@@ -38,13 +38,6 @@ class MeanCurveReader(object):
     """Read and compute a ``MeanCurve`` from one or multiple TB event files. A
     ``MeanCurveReader`` is suitable for one method on one task with multiple runs.
     """
-    _SIZE_GUIDANCE = {
-        'compressedHistograms': 10,
-        'images': 0,
-        'scalars':
-            100,  # sampled points will evenly distribute over the training time
-        'histograms': 1
-    }
 
     def _get_metric_name(self):
         raise NotImplementedError()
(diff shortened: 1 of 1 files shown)


alf/utils/plot_tb_curves.py
```diff
@@ -393,27 +432,40 @@ def _clip_y(y):
             return np.clip(y, y_clipping[0],
                            y_clipping[1]) if y_clipping else y
 
+        if not isinstance(linestyle, list):
+            linestyle = [linestyle] * len(mean_curves)
```
Haichao-Zhang, line 438: Why need to do this conversion to the style string? Is there a use case where user can specify a list of style strings that has the same length as the curve?

hnyu: > Why need to do this conversion to the style string? Is there a use case where user can specify a list of style strings that has the same length as the curve?

If you want to mix dashed line styles with solid lines styles for examples (See my paper's curves)

Haichao-Zhang: This needs to be documented somewhere. Currently the docstr ``linestyle (str)`` is not updated.

hnyu: updated

Haichao-Zhang (approved):

PR #779, opened 2021-02-04
Fix pre_config() for existing configs

emailweixu:
Need to check whether a config is already defined or not in pre_config().


le-horizon (dismissed):
LG.  Maybe in the future we need to add some unit tests for cases like this.

Thanks,
Le

le-horizon (approved):
LG, Thanks!

Le

PR #782, opened 2021-02-08
Switch the order of wrapping for gin-config and alf-config

emailweixu:
The following examples shows the reason to  switch:

```python

@alf.configurable
def func():
   ....

func.abc=1   # (*)
```
In gin config file, @func is the function without attribute abc, which is not the right behavior.  This is because gin wraps func first, then alf wraps the gin-wrapped function. Line (*) changes the function returned by alf, not the function wrapped by gin.


le-horizon (approved):
Good to catch this.  Would be best to have a small test case to capture this error.

Thanks,
Le

PR #783, opened 2021-02-08
use temporally_independent_train_step for sac_carla_conf.py

emailweixu:


alf/examples/sac_carla_conf.py
```diff
@@ -187,14 +176,33 @@
...
     value_net_cls = partial(
         alf.networks.ValueNetwork,
         fc_layer_params=(256, ),
         output_tensor_spec=env.reward_spec())
 
     alf.config('Agent', rl_algorithm_cls=TasacValueAlgorithm)
+    alf.config(
+        'TrainerConfig',
+        use_rollout_state=True,
+        temporally_independent_train_step=True)
```
hnyu, line 204: Is this the reason why the current code is much slower than before?

alf/examples/sac_carla_conf.py
```diff
@@ -187,14 +176,33 @@
...
+
     alf.config(
         'TasacValueAlgorithm',
         actor_network_cls=actor_network_cls,
```
hnyu, line 208: Line 214 "always use 1-step TD" should have been removed

alf/examples/sac_carla_conf.py
emailweixu: yes

alf/examples/sac_carla_conf.py
emailweixu: removed

hnyu (approved):

PR #785, opened 2021-02-11
Causal Conv1d

Haichao-Zhang:
This PR implements CausalConv1D.
This might be useful for standard autoregressive scenarios where RNN is typically used.

### DIFF
--- a/alf/layers.py
+++ b/alf/layers.py
@@ -732,12 +732,12 @@ class CausalConv1D(nn.Module):
 
         Note that the main difference of causal conv v.s. standard conv is that
         each temporal element in the convolutional output is causal w.r.t.
-        the temporal elements from input. For example, for a length :math:`L`
-        sequence :math:`x` with the shape of :math:`[B, C, L]`, and
-        :math:`y = causal_conv(x)`, where the shape of :math:`y` is
-        :math:`[B, C', L]`, by causal we mean :math:`y[..., l]` only depends on
-        :math:`X[..., :l]` (i.e. the past), and there is no dependency on
-        :math:`X[..., l:]` (i.e. future) as in the standard non-causal
+        the temporal elements from input. For example, for a length `L`
+        sequence `x` with the shape of `[B, C, L]`, and
+        `y = causal_conv(x)`, where the shape of `y` is
+        `[B, C', L]`, by causal we mean `y[..., l]` only depends on
+        `X[..., :l]` (i.e. the past), and there is no dependency on
+        `X[..., l:]` (i.e. future) as in the standard non-causal
         convolution.
 
         This can implemented by using an asymmetric padding, which in effect
--- a/alf/layers.py
+++ b/alf/layers.py
@@ -718,7 +718,7 @@ class CausalConv1D(nn.Module):
                  out_channels,
                  kernel_size,
                  dilation=1,
-                 hide_current=True,
+                 hide_current=False,
                  activation=torch.relu_,
                  use_bias=None,
                  use_bn=False,
--- a/alf/layers.py
+++ b/alf/layers.py
@@ -732,12 +732,12 @@ class CausalConv1D(nn.Module):
 
         Note that the main difference of causal conv v.s. standard conv is that
         each temporal element in the convolutional output is causal w.r.t.
-        the temporal elements from input. For example, for a length `L`
-        sequence `x` with the shape of `[B, C, L]`, and
-        `y = causal_conv(x)`, where the shape of `y` is
-        `[B, C', L]`, by causal we mean `y[..., l]` only depends on
-        `X[..., :l]` (i.e. the past), and there is no dependency on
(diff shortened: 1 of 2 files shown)


alf/layers.py
```diff
@@ -705,6 +706,118 @@ def bias(self):
         return self._bias
 
 
+@alf.configurable
+class CausalConv1D(nn.Module):
```
hnyu, line 710: it seems that this is only dilated conv1d? why do we need to create a new layer class named as "CausalConv1D"? Would be better to explicitly explain the difference between a dilated conv1d and this class in the docstring.

Haichao-Zhang: > it seems that this is only dilated conv1d? why do we need to create a new layer class named as "CausalConv1D"? Would be better to explicitly explain the difference between a dilated conv1d and this class in the docstring.

Not really. dilation is just an optional choice. Even without dilation, it is still different from conv1d, in that it is causal. Will add more explanations in comments.

Haichao-Zhang: Now added more explanations.

alf/layers.py
```diff
@@ -705,6 +706,137 @@ def bias(self):
...
+        initialization. An auto gain calculation might depend on the activation
+        following the causal conv1d layer.
+
+        Note that the main difference of causal conv v.s. standard conv is that
+        each temporal element in the convolutional output is causal w.r.t.
+        the temporal elements from input. For example, for a length L sequence
+        x with the shape of [B, C, L], and y = causal_conv(x), where the shape
+        of y is [B, C', L], by causal we mean y[..., l] only depends on
+        X[..., :l] (i.e. the past), and there is no dependency on X[..., :l]
+        (i.e. future) as in the standard non-causal convolution.
```
hnyu: Use correct formatting for these math notations.

alf/layers.py
```diff
@@ -729,13 +730,27 @@ def __init__(self,
...
+        shift the input to the right (future) according to kernel size.
+
         Args:
             in_channels (int): channels of the input
             out_channels (int): channels of the output
             kernel_size (int): size of the kernel
             dilation (int): controls the spacing between the kernel points.
                 Please refer to here for a visual illustration:
                 https://github.com/vdumoulin/conv_arithmetic/blob/master/README.md
+            hide_current (bool): whether to hide the current by shifting the
```
hnyu, line 751: If the current info in the first layer is missing, how will the later layers get it?

alf/layers.py
```diff
@@ -705,6 +706,137 @@ def bias(self):
...
+        shift the input to the right (future) according to kernel size.
+
+        Args:
+            in_channels (int): channels of the input
+            out_channels (int): channels of the output
+            kernel_size (int): size of the kernel
+            dilation (int): controls the spacing between the kernel points.
+                Please refer to here for a visual illustration:
+                https://github.com/vdumoulin/conv_arithmetic/blob/master/README.md
+            hide_current (bool): whether to hide the current by shifting the
```
hnyu, line 753: The need of hide_current seems better addressed by manipulating the input before feeding it to this layer?

alf/layers.py
Haichao-Zhang: Fixed

alf/layers.py
Haichao-Zhang: We can choose either way to do this. Here we just provide an option to the user.

alf/layers.py
```diff
@@ -732,11 +732,13 @@ def __init__(self,
...
-        the temporal elements from input. For example, for a length L sequence
-        x with the shape of [B, C, L], and y = causal_conv(x), where the shape
-        of y is [B, C', L], by causal we mean y[..., l] only depends on
-        X[..., :l] (i.e. the past), and there is no dependency on X[..., :l]
-        (i.e. future) as in the standard non-causal convolution.
+        the temporal elements from input. For example, for a length :math:`L`
+        sequence :math:`x` with the shape of :math:`[B, C, L]`, and
+        :math:`y = causal_conv(x)`, where the shape of :math:`y` is
+        :math:`[B, C', L]`, by causal we mean :math:`y[..., l]` only depends on
+        :math:`X[..., :l]` (i.e. the past), and there is no dependency on
```
hnyu: this is actually python syntax, not latex syntax

Haichao-Zhang: changed

alf/layers.py
```diff
@@ -705,6 +706,139 @@ def bias(self):
...
+            self._bn = nn.BatchNorm1d(out_channels)
+        else:
+            self._bn = None
+
+        self.reset_parameters()
+
+    def reset_parameters(self):
+        """Initialize the parameters."""
+        if self._kernel_initializer is None:
+            variance_scaling_init(
```
hnyu, line 804: One last minor potential issue: variance_scaling_init calculates the gain by assuming the input is N(0,1). Now since we know we are always padding the input with zeros, this auto calculation might be problematic.

hnyu (approved):

PR #786, opened 2021-02-11
Convert speed PID defaults from km/h to m/s

Haichao-Zhang:
Convert speed PID defaults from km/h to m/s to be consistent with other parts of code.

### DIFF
--- a/alf/environments/carla_controller.py
+++ b/alf/environments/carla_controller.py
@@ -93,10 +93,10 @@ class VehicleController(object):
         Note that the max_speed and gain parameters for speed are originally
         specified for speed in the unit of km/h. Since here we use m/s, we have
         converted them as follows as our default values:
-            :math:`max_speed = (20 km/h) / 3.6 =  5.56 m/s`
-            :math:`s_P = (1.0 h/km) * 3.6 = 3.6 s/m`
-            :math:`s_I = (0.05 h/km) * 3.6 = 0.18 s/m`
-            :math:`s_D = (0 h/km) * 3.6 = 0 s/m`
+            max_speed = (20 km/h) / 3.6 =  5.56 m/s
+            s_P = (1.0 h/km) * 3.6 = 3.6 s/m
+            s_I = (0.05 h/km) * 3.6 = 0.18 s/m
+            s_D = (0 h/km) * 3.6 = 0 s/m
 
         Args:
             vehicle (carla.Actor): the actor for vehicle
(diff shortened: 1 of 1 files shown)


alf/environments/carla_controller.py
```diff
@@ -78,29 +78,40 @@ class VehicleController(object):
...
                  d_I=0.07,
                  d_D=0.2):
         """
-        The defaults are from https://github.com/carla-simulator/carla/blob/master/PythonAPI/carla/agents/navigation/local_planner.py
+        The defaults are from https://github.com/carla-simulator/carla/blob/master/PythonAPI/carla/agents/navigation/local_planner.py.
+        Note that the max_speed and gain parameters for speed are originally
+        specified for speed in the unit of km/h. Since here we use m/s, we have
+        converted them as follows as our default values:
+            max_speed = 20 / 3.6 = 5.56
+            s_P = 1.0 / 3.6 = 0.28
```
emailweixu: I feel should be "* 3.6". Suppose error is 1km/h, which is 1/3.6 m/s. We want to get the same control value, PID coefficient should be multiplied by 3.6.

alf/environments/carla_controller.py
```diff
@@ -78,29 +78,40 @@ class VehicleController(object):
...
+            max_speed (float): maximal speed in m/s. Default to 5.6 m/s which
+                is about 20 km/h.
             max_throttle (float): maximal throttle
             max_steering (float): maximal steering
             max_brake (float): maximal brake
-            s_P (float): coefficient of the proportional term for the speed controller
-            s_I (float): coefficient of the integral term for the speed controller
-            s_D (float): coefficient of the derivative term for the speed controller
+            s_P (float): coefficient of the proportional term for the speed
+                controller, with the unit of speed as m/s
```
emailweixu: the unit should be s/m. The unit for error is s/m, the result should be a control value, which can be considered unit free.

alf/environments/carla_controller.py
Haichao-Zhang: You're right. I initially used ``*3.6`` but later changed to ``/3.6`` due to some erroneous thoughts.
The unit free perspective is very helpful for resolving my confusing thoughts. 
I have corrected it.

alf/environments/carla_controller.py
Haichao-Zhang: I originally commented that the unit of the speed should be m/s to highlight the context of the gain coefficient.
Now changed it to its own unit s/m, which is more direct.

emailweixu (approved):

PR #787, opened 2021-02-12
TemporalPool

emailweixu:
A network useful for building temporal CNN

### DIFF
--- a/alf/networks/networks.py
+++ b/alf/networks/networks.py
@@ -132,6 +132,12 @@ class TemporalStack(Network):
     avg:   [0, 0],   [0, 1.5], [0, 1.5], [1.5, 3.5], [1.5, 3.5]
     max:   [0, 0],   [0, 2],   [0, 2],   [2, 4],     [2, 4]
 
+    Note that for 'avg' and 'max', the result is zero for the first ``pooling_size - 1``
+    steps because it needs ``pooling_size`` input to calculate the result. After
+    that, the output changes every ``pooling_size`` steps as the new pooling result
+    available. On the other hand, for 'skip', the first input is immediately
+    reflected in the output because it is a valid way of skipping.
+
     Example:
 
     .. code-block:: python
@@ -140,13 +146,26 @@ class TemporalStack(Network):
         cnn = alf.networks.Sequential([
             alf.networks.TemporalStack(256, 3, 1),
             torch.nn.Flatten(),
-            alf.layers.FC(768, 256, activation=torch.relu_)
+            alf.layers.FC(768, 256, activation=torch.relu_),
             alf.networks.TemporalStack(256, 3, 2),
             torch.nn.Flatten(),
-            alf.layers.FC(768, 256, activation=torch.relu_)
+            alf.layers.FC(768, 256, activation=torch.relu_),
             alf.networks.TemporalStack(256, 3, 4),
             torch.nn.Flatten(),
             alf.layers.FC(768, 256, activation=torch.relu_)])
+
+
+    Note that the output of the above network changes every 4 steps, which may make
+    the response too slow for many tasks. So a practical way of using ``TemporalStack``
+    is to combine it with ``Residue`` so that the output will not lag:
+
+    .. code-block:: python
+
+        block = alf.networks.Residue(
+            alf.networks.Sequential([
+                alf.networks.TemporalStack(256, 3, 2),
+                torch.nn.Flatten(),
+                alf.layers.FC(768, 256, activation=torch.relu_)]))
 
     """
 
--- a/alf/networks/networks.py
+++ b/alf/networks/networks.py
@@ -15,10 +15,13 @@
 
 import torch
 import torch.nn as nn
+import typing
 
 import alf
-
+from alf.utils.common import expand_dims_as
 from .network import Network, wrap_as_network
+
+__all__ = ['LSTMCell', 'GRUCell', 'Residue', 'TemporalStack']
 
 
 class LSTMCell(Network):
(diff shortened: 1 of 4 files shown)


alf/networks/networks.py
```diff
@@ -115,3 +118,122 @@ def __init__(self, block, input_tensor_spec=None, activation=torch.relu_):
...
+class TemporalStack(Network):
+    """Stack features temporally.
+
+    Suppose input_size=(), stack_size=2, pooling_size=2, the following table
+    shows the output of different mode for an input sequence of 1,2,3,4,5 (ignoring
+    batch dimension)
+
+           1,        2,        3,        4,          5
+    skip:  [0, 1],   [0, 1],   [1, 3],   [1, 3],     [3, 5]
+    avg:   [0, 0],   [0, 1.5], [0, 1.5], [1.5, 3.5], [1.5, 3.5]
```
hnyu, line 132: should the third output be [1.5, 1.5] instead? (shifting happens after every pooling_size to be consistent with 'skip' mode, and use a padding of the same value). This updating rule is well defined or defined by ourselves?

alf/networks/networks.py
```diff
@@ -115,3 +118,122 @@ def __init__(self, block, input_tensor_spec=None, activation=torch.relu_):
     def forward(self, x, state=()):
         y, state = self._block(x, state)
         return self._activation(x + y), state
+
+
+class TemporalStack(Network):
+    """Stack features temporally.
+
+    Suppose input_size=(), stack_size=2, pooling_size=2, the following table
```
hnyu, line 126: I feel like "stack" is not the proper word here. With pooling_size =K, a stack_size=2 is actually stacking a max of 2*K inputs. Maybe change to another word or explain in a clearer way?

alf/networks/networks.py
```diff
@@ -115,3 +118,122 @@ def __init__(self, block, input_tensor_spec=None, activation=torch.relu_):
...
+
+    def __init__(self,
+                 input_size,
+                 stack_size,
+                 pooling_size=1,
+                 dtype=torch.float32,
+                 mode='skip'):
+        """
+        Args:
+            input_size (int|tuple[int]): shape of the input
```
Haichao-Zhang, line 161: It might be useful to comment on which dim is assumed to be the temporal dimension, especially when input_size is tuple (multi-dimensional input).

alf/networks/networks.py
```diff
@@ -115,3 +118,122 @@ def __init__(self, block, input_tensor_spec=None, activation=torch.relu_):
...
+        input_tensor_spec = alf.TensorSpec(input_size, dtype=dtype)
+        super().__init__(input_tensor_spec)
+        self._pooling_size = pooling_size
+        if pooling_size == 1:
+            self._state_spec = alf.TensorSpec((stack_size - 1, ) + input_size,
+                                              dtype)
+        elif mode == 'skip':
+            self._pool_func = self._skip_pool
+            pool_state_spec = ()
+            self._update_step = 1
```
Haichao-Zhang, line 210: What is the reason for setting ``_update_step`` for 'skip' to be different from the other modes?
This might be useful for keeping information from the first element, but its computational logic is a bit different from others. And it seems this is related to the question raised on the illustrative example as well.
We can at least explain this somewhere.

alf/networks/networks.py
emailweixu: For 'skip', the output changes at step n*pooling_size.
For 'avg' and 'max', the output changes at step n * pooling_size + pooling_size - 1, this is because the 'avg' or 'max' value can only be obtained every pooling_size steps.

alf/networks/networks.py
emailweixu: If stack_size is 2, the dimension of the output is always 2 * input_dimention. And it's a stack of 2 results of pooling operations.

alf/networks/networks.py
hnyu: Yeah I know how different modes update, but then there is inconsistency in their outputs. As for 'avg' and 'max' can only be obtained every pooling_size steps, we can always pad the last value with the second last value when shifting. Suppose now the pooling size is large=5, then for the current 'max', it will be something like:

```
[0,0], [0,0], [0,0], [0,0], [0,5], |   [0,5], [0,5],[0,5],[0,5],[5,10], | ....
```
Another way of updating is 
```
[0,0], [0,0],[0,0],[0,0],[0,5] | [5,5], [5,5], [5,5], [5,5], [5,10], | ....
```
The second output will have less delay in the first entry of the output window, and is more consistent with 'skip'. I think it's depending on what kind of desired output we want.

alf/networks/networks.py
hnyu: My point is that "temporalstack" might suggest stacking in the temporal dimension. for example, [a_t, a_{t+1},...]. For our code here, we are moving along a temporal dimension but are stacking in the feature dimension.

emailweixu: Changed to TemopralPool

alf/networks/networks.py
emailweixu: This will make the shift unnatural. It needs to have a special handling at step pooling_size. And in terms of information contained in the output, they contains same information.

alf/networks/networks_test.py
```diff
@@ -0,0 +1,163 @@
...
+
+import torch
+
+import alf
+
+from alf.networks import TemporalPool
+
+
+class NestworksTest(alf.test.TestCase):
+    def test_temporal_stack_skip(self):
```
Haichao-Zhang: One minor point is that we can probably change ``test_temporal_stack_max`` to ``test_temporal_pool_max``.
Same for the other tests.

hnyu (dismissed):

alf/networks/networks.py
hnyu: I just felt that the first entry of each output in this case will have a delay of at most `pooling_size - 1`. But yeah, the two sequences have the same information at `N*pooling_size`.

Haichao-Zhang (approved):

PR #788, opened 2021-02-17
Fix the respawn location of vehicles

emailweixu:
When a Player is reset, a new location will be chosen for it. The new location needs
to be far enough from other vechicles and walkers.

There are two problems in previous code:
1. The location of other vehicles are not updated. We fix this by clearing the location cache in AlfWorld.
2. The walkers are not considered.

### DIFF
--- a/alf/environments/suite_carla.py
+++ b/alf/environments/suite_carla.py
@@ -1320,7 +1320,6 @@ class CarlaEnvironment(AlfEnvironment):
                 continue
             walker = self._world.get_actor(response.actor_id)
             self._walkers.append({"walker": walker})
-            self._alf_world.add_actor(walker)
             walker_speeds2.append(walker_speed)
 
         walker_speeds = walker_speeds2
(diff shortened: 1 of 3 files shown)


alf/environments/suite_carla.py
```diff
@@ -1320,7 +1320,9 @@ def _spawn_walkers(self):
                 continue
             walker = self._world.get_actor(response.actor_id)
             self._walkers.append({"walker": walker})
+            self._alf_world.add_actor(walker)
```
Haichao-Zhang: It seems that we have already added the walkers in the existing code?

https://github.com/HorizonRobotics/alf/blob/24ab970d61307c66027a7b140ea0606a65b32ee2/alf/environments/suite_carla.py#L1357

Haichao-Zhang (approved):

alf/environments/suite_carla.py
emailweixu: good catch!

PR #790, opened 2021-02-20
Fix gym_wrappers for nested action space

emailweixu:
Previous, alf.nest.map_structure was used gym space, which may be gym.spaces.Dict or gym.spaces.Tuple and cannot be correctly handled.

### DIFF
--- a/alf/environments/gym_wrappers.py
+++ b/alf/environments/gym_wrappers.py
@@ -554,6 +554,18 @@ class DMAtariPreprocessing(gym.Wrapper):
         return np.expand_dims(int_image, axis=2)
 
 
+def _gym_space_to_nested_space(space):
+    """Change gym Space to a nest which can be handled by alf.nest functions."""
+
+    if isinstance(space, gym.spaces.Dict):
+        return dict((k, _gym_space_to_nested_space(s))
+                    for k, s in space.spaces.items())
+    elif isinstance(space, gym.spaces.Tuple):
+        return tuple(_gym_space_to_nested_space(s) for s in space.spaces)
+    else:
+        return space
+
+
 @gin.configurable
 class ContinuousActionClip(gym.ActionWrapper):
     """Clip continuous actions according to the action space."""
@@ -573,7 +585,10 @@ class ContinuousActionClip(gym.ActionWrapper):
             else:
                 return min_v, max_v
 
-        self.bounds = alf.nest.map_structure(_space_bounds, self.action_space)
+        self._nested_action_space = _gym_space_to_nested_space(
+            self.action_space)
+        self.bounds = alf.nest.map_structure(_space_bounds,
+                                             self._nested_action_space)
 
     def action(self, action):
         def _clip_action(space, action, bounds):
@@ -585,8 +600,9 @@ class ContinuousActionClip(gym.ActionWrapper):
                 action = np.clip(action, bounds[0], bounds[1])
             return action
 
-        action = alf.nest.map_structure_up_to(
-            action, _clip_action, self.action_space, action, self.bounds)
+        action = alf.nest.map_structure_up_to(action, _clip_action,
+                                              self._nested_action_space,
+                                              action, self.bounds)
         return action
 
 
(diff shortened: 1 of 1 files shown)


alf/environments/gym_wrappers.py
```diff
@@ -610,12 +626,14 @@ def _space_bounds(space):
                 assert np.all(np.isfinite(space.high))
                 return (space.low, space.high)
 
-        self._bounds = alf.nest.map_structure(_space_bounds, self.action_space)
-        self.action_space = alf.nest.map_structure(
+        nested_action_space = _gym_space_to_nested_space(self.action_space)
+        self._bounds = alf.nest.map_structure(_space_bounds,
+                                              nested_action_space)
+        self._nested_action_space = alf.nest.map_structure(
```
hnyu, line 650: We also need to change self.action_space to `[low, high]`. Otherwise it's not correct after using this wrapper.

hnyu (changes requested):

alf/environments/gym_wrappers.py
```diff
@@ -573,7 +585,10 @@ def _space_bounds(space):
             else:
                 return min_v, max_v
 
-        self.bounds = alf.nest.map_structure(_space_bounds, self.action_space)
```
hnyu, line 576: Should have commented in docstring that `self.action_space` is not really its valid action space as we can always accept any action and clip it to the desired range. For now we just keep it the same with the action space of the underlying `self.env`.

emailweixu: commented

alf/environments/gym_wrappers.py
emailweixu: fixed

alf/environments/gym_wrappers.py
```diff
@@ -554,9 +554,39 @@ def _pool_and_resize(self):
...
+
+def _nested_space_to_gym_space(space):
+    """Change nested space to gym Space"""
+
+    if isinstance(space, (dict, OrderedDict)):
+        spaces = dict(
+            (k, _nested_space_to_gym_space(s)) for k, s in space.items())
+        return gym.spaces.Dict(spaces)
+    elif isinstance(space, tuple):
+        spaces = tuple(_gym_space_to_nested_space(s) for s in space)
```
hnyu: Should be _nested_space_to_gym_space?

alf/environments/gym_wrappers.py
```diff
@@ -554,9 +554,39 @@ def _pool_and_resize(self):
...
+        return space
+
+
 @gin.configurable
 class ContinuousActionClip(gym.ActionWrapper):
-    """Clip continuous actions according to the action space."""
+    """Clip continuous actions according to the action space.
+
+    Note that any action outside of the bounds specified by action_space will be
+    clipped to the bounds before passing to the underline environment.
```
hnyu: underline -> underlying

alf/environments/gym_wrappers.py
hnyu: If self.action_space still uses Gym's Dict and Tuple, will the Network class correctly handle it?

hnyu:
It seems that for both wrappers, the input `action` still uses ALF's nest structure, because when we convert gym space to tensor spec, we will convert gym.spaces.Tuple and gym.spaces.Dict to tuple and dict in `tensor_spec_from_gym_space` . So the action returned by a network is alf nest. We need to convert `action` to their original data structure before inputting to the underlying env? I think in a general case, a separate wrapper `GymSpaceToNest` is needed, and move the handling in `tensor_spec_from_gym_space` to this wrapper.

Or perhaps tuple and dict can be converted to gym.spaces.Tuple and gym.spaces.Dict by Gym?

emailweixu:
> It seems that for both wrappers, the input `action` still uses ALF's nest structure, because when we convert gym space to tensor spec, we will convert gym.spaces.Tuple and gym.spaces.Dict to tuple and dict in `tensor_spec_from_gym_space` . So the action returned by a network is alf nest. We need to convert `action` to their original data structure before inputting to the underlying env? I think in a general case, a separate wrapper `GymSpaceToNest` is needed, and move the handling in `tensor_spec_from_gym_space` to this wrapper.
> 
> Or perhaps tuple and dict can be converted to gym.spaces.Tuple and gym.spaces.Dict by Gym?

There is no dedicated data structure of actions or observations. They simply use python dict or tuple.

hnyu (approved):

PR #794, opened 2021-02-26
Avoid re-config seed if it is already configured in config_helpers.get_env()

emailweixu:
Otherwise there will be a warning message which can cause confusion.


le-horizon (approved):
LG.

PR #797, opened 2021-03-01
rename 'misc' to 'music' to make terminology consistent in the updated version of the accepted paper

ruizhaogit:
Hi @emailweixu and @hnyu ,

I rename 'misc' to 'music' to make terminology consistent in the updated version of the accepted paper

Best,
Rui

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -111,3 +111,6 @@ venv.bak/
 
 # ~
 *~
+
+# .DS_Store
+*.DS_Store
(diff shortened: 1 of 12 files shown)


hnyu:
Note that we've switched from TF to PyTorch for quite a long time (since early last year) https://github.com/HorizonRobotics/alf/projects/1 . So it's likely that the current MISC code can't run successfully anymore (won't be correctly called by the policy_trainer code). It's best that you could update MISC code to PyTorch. But if you need to make it available as soon as possible, then I suggest forking your own version, and update the name based on the last TF version you used for the experiments.

ruizhaogit:
My fork version is avaliable at https://github.com/ruizhaogit/alf.

PR #799, opened 2021-03-03
Support separate entropy target for nested actions

emailweixu:


alf/utils/dist_utils.py
```diff
@@ -680,6 +680,8 @@ def entropy_with_fallback(distributions):
     Args:
         distributions (nested Distribution): A possibly batched tuple of
             distributions.
+        return_sum (bool): if return_sum, return the total entropy. If not return_sum,
```
hnyu: if True ... if not True

alf/algorithms/entropy_target_algorithm.py
```diff
@@ -32,7 +33,18 @@
...
+                           skip_free_stage=False,
+                           max_entropy=None,
+                           target_entropy=None,
+                           very_slow_update_rate=0.001,
+                           slow_update_rate=0.01,
+                           fast_update_rate=np.log(2),
+                           min_alpha=1e-4,
+                           average_window=2,
+                           debug_summaries=False,
+                           name="EntropyTargetAlgorithm"):
```
hnyu, line 47: Should add an explanation that _NestedEntropyTargetAlgorithm will be created for nested action_spec.

alf/algorithms/entropy_target_algorithm.py
```diff
@@ -32,7 +33,18 @@
 
 
 @alf.configurable
-class EntropyTargetAlgorithm(Algorithm):
+def EntropyTargetAlgorithm(action_spec,
```
le-horizon: this is no longer an Algorithm, maybe call it entropy_target_algorithm_ctor?

le-horizon (commented):
Almost LG.  Just one name nit.

Thanks,
Le

alf/algorithms/entropy_target_algorithm.py
le-horizon: > this is no longer an Algorithm, maybe call it entropy_target_algorithm_ctor?

@emailweixu other way of creating a separate NestedEntropyTargetAlgorithm class, keeping the current EntropyTargetAlgorithm unchanged also looks good to me.

hnyu (approved):

PR #800, opened 2021-03-05
More net building utilities

emailweixu:
Also
* change alf.nn.Sequential to follow the interface of torch.nn.Sequentil
* Improved PredictiveRepresentationLearner for supporting post_processor and non-replay buffer based training when num_unrolll_steps==0

### DIFF
--- a/alf/algorithms/predictive_representation_learner.py
+++ b/alf/algorithms/predictive_representation_learner.py
@@ -22,7 +22,7 @@ from alf.data_structures import AlgStep, TimeStep, Experience, LossInfo, namedtu
 from alf.experience_replayers.replay_buffer import BatchInfo, ReplayBuffer
 from alf.nest import nest
 from alf.nest.utils import convert_device
-from alf.networks import Network, LSTMEncodingNetwork
+from alf.networks import Network, LSTMEncodingNetwork, wrap_as_network
 from alf.utils import common, dist_utils, spec_utils, tensor_utils
 from alf.utils.normalizers import AdaptiveNormalizer
 from alf.utils.summary_utils import safe_mean_hist_summary, safe_mean_summary
@@ -139,16 +139,21 @@ class SimpleDecoder(Algorithm):
             target = self._target_normalizer.normalize(target)
             predicted = self._target_normalizer.normalize(predicted)
 
-        loss = self._loss(predicted, target)
+        b = predicted.shape[0] * predicted.shape[1]
+        loss = self._loss(
+            predicted.reshape(b, *predicted.shape[2:]),
+            target.reshape(b, *target.shape[2:]))
+        loss = loss.reshape(*predicted.shape[:2], *loss.shape[1:])
         if self._debug_summaries and alf.summary.should_record_summaries():
             with alf.summary.scope(self._name):
 
                 def _summarize1(pred, tgt, loss, mask, suffix):
-                    alf.summary.scalar(
-                        "explained_variance" + suffix,
-                        tensor_utils.explained_variance(pred, tgt, mask))
-                    safe_mean_hist_summary('predict' + suffix, pred, mask)
-                    safe_mean_hist_summary('target' + suffix, tgt, mask)
+                    if pred.shape == tgt.shape:
+                        alf.summary.scalar(
+                            "explained_variance" + suffix,
+                            tensor_utils.explained_variance(pred, tgt, mask))
+                        safe_mean_hist_summary('predict' + suffix, pred, mask)
+                        safe_mean_hist_summary('target' + suffix, tgt, mask)
                     safe_mean_summary("loss" + suffix, loss, mask)
 
                 def _summarize(pred, tgt, loss, mask, suffix):
@@ -201,6 +206,7 @@ class PredictiveRepresentationLearner(Algorithm):
                  decoder_ctor,
                  encoding_net_ctor,
                  dynamics_net_ctor,
+                 postprocessor=None,
                  encoding_optimizer=None,
                  dynamics_optimizer=None,
                  debug_summaries=False,
@@ -210,6 +216,8 @@ class PredictiveRepresentationLearner(Algorithm):
             observation_spec (nested TensorSpec): describing the observation.
             action_spec (nested BoundedTensorSpec): describing the action.
             num_unroll_steps (int): the number of future steps to predict.
+                if ``num_unroll_steps`` of 0 means no future prediction and hence
+                ``dynamics_net_ctor`` is ignored.
             decoder_ctor (Callable|[Callable]): each individual constructor is
                 called as ``decoder_ctor(observation)`` to
                 construct the decoder algorithm. It should follow the ``Algorithm``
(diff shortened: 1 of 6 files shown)


alf/networks/containers.py
```diff
@@ -42,19 +41,25 @@ class Sequential(Network):
...
         new_state = (new_state1, new_state2)
 
-    """
-
-    def __init__(self, modules, input_tensor_spec=None, name="Sequential"):
-        """
-
         Args:
-            modules (list[nn.Module]): list of ``torch.nn.Module`` or ``alf.nn.Network``.
+            modules (nn.Module): list of ``torch.nn.Module`` or ``alf.nn.Network``.
```
Haichao-Zhang, line 45: The comment in (), i.e. ``(nn.Module) `` should be updated?

alf/algorithms/predictive_representation_learner.py
```diff
@@ -210,6 +216,8 @@ def __init__(self,
             observation_spec (nested TensorSpec): describing the observation.
             action_spec (nested BoundedTensorSpec): describing the action.
             num_unroll_steps (int): the number of future steps to predict.
+                if ``num_unroll_steps`` of 0 means no future prediction and hence
```
Haichao-Zhang: can remove ``if``

alf/layers.py
```diff
@@ -2122,3 +2122,67 @@ def __init__(self):
...
+
+class Sequential(nn.Sequential):
+    """A thin wrapper over nn.Sequential to allow stateless alf.networks.Network."""
+
+    def __init__(self, *args):
+        super().__init__(*args)
+        for m in self:
+            if isinstance(m, alf.networks.Network):
+                assert not alf.nest.flatten(m.state_spec), (
+                    "Network element of layers.Sequential should stateless. "
```
Haichao-Zhang: should be

alf/algorithms/predictive_representation_learner.py
```diff
@@ -230,6 +238,9 @@ def __init__(self,
                 dynamics net. Otherwise, a linear projection will be used to
                 convert the current latent represenation to the initial state for
                 the dynamics net.
+            postprocessor (None|Callable): If provided, will be called as
+                ``postprocessor(latent)`` to get the actual representation,
+                where ``latent`` is the output from encoding_net.
```
Haichao-Zhang, line 246: Is it useful to provide an option to specify postprocessor optimizer, to be structurally similar to ``encoding_net`` and ``dynamics net``? Might not be necessary in most of the cases when view postprocessor as part of the downstream networks, but might be useful in some cases.

alf/algorithms/predictive_representation_learner.py
```diff
@@ -139,16 +139,21 @@ def calc_loss(self, target, predicted, mask=None):
             target = self._target_normalizer.normalize(target)
             predicted = self._target_normalizer.normalize(predicted)
 
-        loss = self._loss(predicted, target)
+        b = predicted.shape[0] * predicted.shape[1]
+        loss = self._loss(
+            predicted.reshape(b, *predicted.shape[2:]),
+            target.reshape(b, *target.shape[2:]))
+        loss = loss.reshape(*predicted.shape[:2], *loss.shape[1:])
```
hnyu, line 148: Why bothering reshaping here? I believe the loss class can handle arbitrary shapes.

alf/layers.py
```diff
@@ -2122,3 +2122,67 @@ def __init__(self):
...
+        super().__init__()
+        has_network = any(
+            alf.nest.flatten(
+                alf.nest.map_structure(
+                    lambda m: isinstance(m, alf.networks.Network), modules)))
+        assert not has_network, (
+            "modules should not contain alf.networks.Network. "
+            "Try alf.networks.Branch instead.")
+
+        self._networks = modules
```
hnyu, line 2159: It's confusing that the variable name is "networks" but in fact it can't contain any network

alf/networks/containers.py
```diff
@@ -42,19 +41,25 @@ class Sequential(Network):
...
             input_tensor_spec (TensorSpec): the tensor spec of the input. It must
                 be specified if it cannot be inferred from ``modules[0]``.
             name (str):
-        """
+    """
+    # The reason that we use a wrapper function for _Sequential is that Network
+    # does not allow *args for __init__() (see _NetworkMeta.__new__()). And we
+    # want to use *modules here to make the interface consistant with
```
hnyu: consistent

alf/networks/containers.py
```diff
@@ -14,17 +14,16 @@
...
 import alf
 from alf.nest import (flatten, flatten_up_to, map_structure,
                       map_structure_up_to, pack_sequence_as)
+from alf.utils.spec_utils import is_same_spec
 from .network import Network, get_input_tensor_spec, wrap_as_network
 
 
-@gin.configurable
-class Sequential(Network):
+def Sequential(*modules, input_tensor_spec=None, name="Sequential"):
```
hnyu, line 26: Should still keep "configurable"?

alf/networks/containers.py
```diff
@@ -312,3 +318,118 @@ def copy(self, name=None):
...
+        state = (block_state, echo_output)
+
+    """
+
+    def __init__(self, block, input_tensor_spec=None):
+        """
+        Args:
+            block (Network): the module for performing the actual computation
+            input_tensor_spec (nested TensorSpec): If provided, it must match
+                the block.input_tensor_spec[0] or input_tensor_spec['input']
```
hnyu: \``block.input_tensor_spec['input']\``

hnyu (changes requested):

alf/algorithms/predictive_representation_learner.py
emailweixu: CrossEntropyLoss does not behave as we expected. The shape of predicted is [T, B, N] and the shape of target is [T, B], where N is the number of categories. For this kind of multidimensional tensor, CrossEntropyLoss expect shapes like [T, N, B] and [T, B]

alf/algorithms/predictive_representation_learner.py
emailweixu: added.

alf/algorithms/predictive_representation_learner.py
emailweixu: removed

alf/layers.py
emailweixu: self._modules is already used nn.Module

alf/layers.py

alf/networks/containers.py
emailweixu: As far as I know, none of existing gin files uses this. New configs should be written in python. So there is no need for it to be configurable.

alf/networks/containers.py
emailweixu: ``nn.Module`` in the parenthesis is the type of each module.

Haichao-Zhang (approved):

hnyu (approved):

PR #801, opened 2021-03-05
Pr batch_ensemble optimizer

runjerry:
- add ensemble_group attributes in FCBatchEnsemble 

- override add_param_group for alf.optimizers to handle params with ensemble_group

- add gfsf_grad computation in step() of alf.optimizers for params with ensemble_group


runjerry:
updated to PR #804

PR #803, opened 2021-03-06
some speedup

runjerry:
* Change some einsum in generator to elementwise product

* speedup generator, hypernetwork, particle_vi, and
functional_particle_vi tests


emailweixu:
> * Change some einsum in generator to elementwise product

Does the change speedup?

In my tests, if the einsum is something like 'ij,ijk->jk', the change does speedup. For something like 'bj,bj->b', it does not, so I kept the original einsum.

> * speedup generator, hypernetwork, particle_vi, and
>   functional_particle_vi tests

emailweixu (approved):

PR #805, opened 2021-03-09
Fix size inconsistency issue in Carla rendering

Haichao-Zhang:
Currently the way to compute the size for rendering surface and scaled camera image are inconsistent.
This will cause part of the camera image missing in the rendered video frame.

This is not noticed as currently the missing part is small in our typical setting (e.g. ```CameraSensor.image_size_x=200```,
```CameraSensor.image_size_y=100```,  ```MINIMUM_RENDER_WIDTH = 640```,
```MINIMUM_RENDER_HEIGHT = 240``` ).

However, in other cases, the missing part could be quite large. An extreme example is shown below:
image

### DIFF
--- a/alf/environments/carla_sensors.py
+++ b/alf/environments/carla_sensors.py
@@ -567,11 +567,8 @@ class CameraSensor(SensorBase):
             height, width = self._image.shape[1:3]
             image = np.transpose(self._image, (2, 1, 0))
             if width < MINIMUM_RENDER_WIDTH:
-                scaling_factor = max(
-                    float(MINIMUM_RENDER_HEIGHT) / height,
-                    float(MINIMUM_RENDER_WIDTH) / width)
-                height = int(height * scaling_factor)
-                width = int(width * scaling_factor)
+                height = max(height, MINIMUM_RENDER_HEIGHT)
+                width = max(width, MINIMUM_RENDER_WIDTH)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: alf/environments/carla_sensors.py [hunk 1]
```diff
@@ -567,10 +567,14 @@ def render(self, display):
             height, width = self._image.shape[1:3]
             image = np.transpose(self._image, (2, 1, 0))
             if width < MINIMUM_RENDER_WIDTH:
-                height = height * MINIMUM_RENDER_WIDTH // width
+                scaling_factor = max(
+                    float(MINIMUM_RENDER_HEIGHT) / height,
```
emailweixu: if height is very small, scaling_factor will be very big. and width will be very big.

Haichao-Zhang: I was thinking to keep the aspect ratio unchanged. But it might not be good for visualization in some cases as you suggested.
Now the rendering surface part has been rolled back. And the camera image scaling part is updated to be consistent.

emailweixu: Why can't we keep the aspect ratio?

Haichao-Zhang: Updated. Now still keep the aspect ratio and added upper bound for scaled sizes.

emailweixu (approved):

PR #806, opened 2021-03-10
first min within reward dim then average

hnyu:
If there are multi-dim rewards, then when training the actor, it makes more sense if we first take min over replicas within each reward dim separately, and then average across reward dims.

### DIFF
--- a/alf/algorithms/ddpg_algorithm.py
+++ b/alf/algorithms/ddpg_algorithm.py
@@ -252,7 +252,11 @@ class DdpgAlgorithm(OffPolicyAlgorithm):
         target_q_values, target_critic_states = self._target_critic_networks(
             (exp.observation, target_action), state=state.target_critics)
 
-        target_q_values = target_q_values.min(dim=1)[0]
+        if target_q_values.ndim == 3 and self._reward_weights is not None:
+            sign = self._reward_weights.sign()
+            target_q_values = (target_q_values * sign).min(dim=1)[0] * sign
+        else:
+            target_q_values = target_q_values.min(dim=1)[0]
 
         q_values, critic_states = self._critic_networks(
             (exp.observation, exp.action), state=state.critics)
--- a/alf/algorithms/ddpg_algorithm.py
+++ b/alf/algorithms/ddpg_algorithm.py
@@ -170,8 +170,6 @@ class DdpgAlgorithm(OffPolicyAlgorithm):
         if reward_weights:
             self._reward_weights = torch.tensor(
                 reward_weights, dtype=torch.float32)
-            assert torch.all(self._reward_weights >= 0.), (
-                "All reward weights must be non-negative!")
 
         self._target_actor_network = actor_network.copy(
             name='target_actor_networks')
@@ -254,10 +252,7 @@ class DdpgAlgorithm(OffPolicyAlgorithm):
         target_q_values, target_critic_states = self._target_critic_networks(
             (exp.observation, target_action), state=state.target_critics)
 
-        if self._num_critic_replicas > 1:
-            target_q_values = target_q_values.min(dim=1)[0]
-        else:
-            target_q_values = target_q_values.squeeze(dim=1)
+        target_q_values = target_q_values.min(dim=1)[0]
 
         q_values, critic_states = self._critic_networks(
             (exp.observation, exp.action), state=state.critics)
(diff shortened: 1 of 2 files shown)


emailweixu:
what if reward weight is negative?

hnyu:
> what if reward weight is negative?

I think we should enforce reward weight to be positive because otherwise taking min over replicas doesn't always mean conservative (prevent overestimate of Q). We can always change the polarity of a reward if a negative weight is needed.

Wei: By definition, weighted sum is the desired reward (no matter whether the weight is positive or not). So min after the weighted sum is always conservative.

hnyu:
> By definition, weighted sum is the desired reward (no matter whether the weight is positive or not). So min after the weighted sum is always conservative.

Taking min after weighted sum basically has no guarantee of taking min/max over individual reward dims?

For example, 
min([3,1], [5,2]) with a weight of (1, -1). Min after weighted sum will be 2, which is implicitly choose min on the second reward. However, to be conservative, it should choose max because of -1 weight.

emailweixu:
> > By definition, weighted sum is the desired reward (no matter whether the weight is positive or not). So min after the weighted sum is always conservative.
> 
> Taking min after weighted sum basically has no guarantee of taking min/max over individual reward dims?
> 
> For example,
> min([3,1], [5,2]) with a weight of (1, -1). Min after weighted sum will be 2, which is implicitly choose min on the second reward. However, to be conservative, it should choose max because of -1 weight.

Right, so in this case, min before the weighted sum is not right.

hnyu:
> > > By definition, weighted sum is the desired reward (no matter whether the weight is positive or not). So min after the weighted sum is always conservative.
> > 
> > 
> > Taking min after weighted sum basically has no guarantee of taking min/max over individual reward dims?
> > For example,
> > min([3,1], [5,2]) with a weight of (1, -1). Min after weighted sum will be 2, which is implicitly choose min on the second reward. However, to be conservative, it should choose max because of -1 weight.
> 
> Right, so in this case, min before the weighted sum is not right.

In this case, the final result is the same for 1) min before weighted sum and 2) min after the weighted sum. Both will return (2=3-1). What we really want is 3-2=1. So I think the key is to enforce weights to be positive and take min before weighted sum.

emailweixu:
> > > > By definition, weighted sum is the desired reward (no matter whether the weight is positive or not). So min after the weighted sum is always conservative.
> > > 
> > > 
> > > Taking min after weighted sum basically has no guarantee of taking min/max over individual reward dims?
> > > For example,
> > > min([3,1], [5,2]) with a weight of (1, -1). Min after weighted sum will be 2, which is implicitly choose min on the second reward. However, to be conservative, it should choose max because of -1 weight.
> > 
> > 
> > Right, so in this case, min before the weighted sum is not right.
> 
> In this case, the final result is the same for 1) min before weighted sum and 2) min after the weighted sum. Both will return (2=3-1). What we really want is 3-2=1. So I think the key is to enforce weights to be positive and take min before weighted sum.

Negative weights can be handled correctly by first doing elelmentwise product, then min, and finnally sum.

hnyu:
> Negative weights can be handled correctly by first doing elelmentwise product, then min, and finnally sum.

Yeah I also thought of doing this

emailweixu (commented):
Perhaps also change ddpg?

hnyu:
> Perhaps also change ddpg?

ddpg is already changed

alf/algorithms/ddpg_algorithm.py
```diff
@@ -252,10 +252,7 @@ def _critic_train_step(self, exp: Experience, state: DdpgCriticState):
         target_q_values, target_critic_states = self._target_critic_networks(
             (exp.observation, target_action), state=state.target_critics)
 
-        if self._num_critic_replicas > 1:
-            target_q_values = target_q_values.min(dim=1)[0]
-        else:
-            target_q_values = target_q_values.squeeze(dim=1)
+        target_q_values = target_q_values.min(dim=1)[0]
```
emailweixu:
```python
sign = target_q_values.sign()
target_q_values = sign * (sign * target_q_values).min(dim=1)[0]
```
and target_critic for sac

hnyu: updated

emailweixu (approved):

PR #808, opened 2021-03-12
reward_weight_algorithm interface and LagrangianAlgorithm

hnyu:
A general LagrangianAlgorithm that tunes reward weights by comparing rollout immediate rewards to predefined reward thresholds. Can be used by both off-policy and on-policy algorithms.

Also added reward_spec as an argument for RLAlgorithm.


alf/algorithms/rl_algorithm.py
```diff
@@ -87,12 +90,17 @@ def __init__(self,
             action_spec (nested BoundedTensorSpec): representing the actions.
             train_state_spec (nested TensorSpec): for the network state of
                 ``train_step()``.
+            reward_spec (TensorSpec): a rank-1 or rank-0 tensor spec representing
+                the reward(s).
```
emailweixu, line 94: need to assert reward_spec is rank-1 or rank-0

alf/algorithms/lagrangian_algorithm_test.py
```diff
@@ -0,0 +1,67 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from functools import partial
+import pprint
+import torch
+
+import alf
+import alf.data_structures as ds
+from alf.algorithms.lagrangian_algorithm import LagrangianAlgorithm
```
emailweixu: missing lagrangian_algorithm.py

alf/algorithms/rl_algorithm.py
hnyu: added

alf/algorithms/lagrangian_algorithm.py
```diff
@@ -0,0 +1,134 @@
...
+    training iteration.
+
+    Generally speaking, for each reward dimension, the algorithm compares the
+    expected undiscounted episodic return to a predefined threshold, and if the
+    return is greater than the threshold (requirement satisfied) then it decreases
+    the weight; otherwise it increases the weight.
+
+    Usually the threshold is defined for a fixed time length, so equivalently,
+    for simplicity, the algorithm can instead compares individual rewards per
+    step to an averaged threshold.
```
emailweixu: To make it less confusing, you don't need to mention "expected undiscounted episodic return". Just the desired threshold for reward per step.

alf/algorithms/lagrangian_algorithm.py
```diff
@@ -0,0 +1,134 @@
...
+from alf.data_structures import namedtuple, AlgStep
+from alf.nest import nest
+from alf.tensor_specs import TensorSpec
+from alf.utils import tensor_utils
+
+LagInfo = namedtuple("LagInfo", ["rollout_reward"], default_value=())
+
+
+@alf.configurable(blacklist=["reward_spec"])
+class LagrangianAlgorithm(Algorithm):
```
emailweixu: "Lagrangian" seems too general here.  This works only on reward weight.

alf/algorithms/ddpg_algorithm.py
```diff
@@ -252,8 +253,8 @@ def _critic_train_step(self, exp: Experience, state: DdpgCriticState):
         target_q_values, target_critic_states = self._target_critic_networks(
             (exp.observation, target_action), state=state.target_critics)
 
-        if target_q_values.ndim == 3 and self._reward_weights is not None:
-            sign = self._reward_weights.sign()
+        if self.has_multidim_reward():
+            sign = self.reward_weights.sign()
```
Haichao-Zhang, line 257: Is it ``self._reward_weights ``?

alf/examples/sac_lag_cargoal1.gin
```diff
@@ -0,0 +1,64 @@
+include 'sac.gin'
```
emailweixu: Let's switch to python config from now on.

alf/algorithms/ddpg_algorithm.py
hnyu: There is a property defined in RLAlgorithm base class.

alf/algorithms/lagrangian_algorithm.py
hnyu: I think the reason why I mention "expected undiscounted episodic return" here is that this is exactly what this algorithm is doing, as long as we only have a single lambda for each reward dim. We are using the expectation of immediate rewards (equal to the return with a scale) to tune that single lambda, and the desired threshold is actually not set for  every step, but for an expectation. 

Another complete different way is still comparing immediate reward against threshold per step, but having a different lambda every step conditioned on the state. In this case, we are really putting a constraint on every step.

emailweixu: What I means is that you only need to talk about the desired expected immediate reward, not about episodic.

alf/algorithms/agent.py
```diff
@@ -356,11 +379,15 @@ def after_train_iter(self, experience, train_info: AgentInfo = None):
         """
         algorithms = [
             self._rl_algorithm, self._representation_learner,
-            self._goal_generator
+            self._goal_generator, self._reward_weight_algorithm
         ]
         algorithms = list(filter(lambda a: a is not None, algorithms))
         self._agent_helper.after_train_iter(algorithms, experience, train_info)
 
+        if self._reward_weight_algorithm:
```
Haichao-Zhang, line 389: Can document the expected behavior for the case where a ``reward_weight_algorithm`` is used and a ``reward_weights`` is also specified for the ``rl_algorithm``.

hnyu: added documentation

alf/examples/sac_lag_cargoal1.gin
hnyu: updated

alf/algorithms/lagrangian_algorithm.py
hnyu: changed

alf/algorithms/lagrangian_algorithm.py
hnyu: updated

alf/algorithms/lagrangian_reward_weight_algorithm.py
```diff
@@ -0,0 +1,132 @@
...
+        loss.backward()
+        self._optimizer.step()
+
+        if self._debug_summaries:
+            with alf.summary.scope(self._name):
+                alf.summary.scalar("cost", loss)
+                for i in range(len(self._reward_thresholds)):
+                    alf.summary.scalar("reward_threshold/%d" % i,
+                                       self._reward_thresholds[i])
+                    alf.summary.scalar("lambda/%d" % i, self.reward_weights[i])
```
emailweixu: reward_weights here will be calculated multiple time by the property function. It's better to calculate reward_weights once it is updated at line 124.

Haichao-Zhang (commented):
LGTM. I have no further comments.

alf/algorithms/lagrangian_reward_weight_algorithm.py
hnyu: changed

alf/algorithms/lagrangian_reward_weight_algorithm.py
```diff
@@ -100,8 +101,9 @@ def _inv_softplus(self, tensor):
 
     @property
     def reward_weights(self):
-        """Return the reward weights."""
-        return F.softplus(self._lambdas)
+        """Return the detached reward weights. These weights are expected not to
+        be changed by external code."""
```
emailweixu, line 105: I guess you can you detach().clone() to do this so that it's guaranteed it won't be changed by external code.

emailweixu (dismissed):

emailweixu (approved):

PR #810, opened 2021-03-15
Option to return all embeddings from TransformerNetwork

emailweixu:


alf/networks/transformer_networks.py
```diff
@@ -193,4 +201,9 @@ def forward(self, inputs, state=()):
                 query = new_query
```
hnyu, line 203: Here memory actually writes the old core, but in the docstring comment the memory writes the new core. This behavior is also different from centralized memory which always writes the new core.

alf/networks/transformer_networks.py
```diff
@@ -161,7 +166,10 @@ def forward(self, inputs, state=()):
                 at ``__init__()``
             state (nested Tensor): states
         Returns:
-            Tensor: shape is [B, core_size * d_model]
+            - Tensor: shape is [B, core_size * d_model] if ``return_core_only``,
+                    and [B, core_size + input_size, d_model] if not ``return_core_only``,
+                    where input_size is the number of embeddings from (processed) input
```
hnyu: \``input_size\``

alf/networks/transformer_networks.py
emailweixu: good catch. It's quite strange why I did this.

alf/networks/transformer_networks.py
emailweixu: fixed

alf/networks/transformer_networks.py
hnyu: But I think we should write the old one? because the new one will be passed as part of input next

emailweixu: Good point. Perhaps that was what I thought originally.

hnyu: Should also change the docstring?

emailweixu: Already changed.

hnyu: Also for centralized memory, we should also put old core in memory?

emailweixu: I think putting new core into memory makes more sense for centralized memory.

hnyu (approved):

PR #811, opened 2021-03-16
train play test for alf conf files

### LINKED ISSUE #812: Support alf-config in train_play_test

hnyu:
Also added examples_dir in cmd() during play because sometimes there is a dependency file in examples. In the future we should probably write code to trace all dependency files under examples and copy them to root_dir.


emailweixu:
> The reason why I removed `mutable=False` in play.py is that if we use pre_config to specify a random_seed during training, play will report errors.

This is problematic. Because if the user does use --conf_param to set the random_seed during training, then the random_seed set for the play will be ignored.

A better way of fixing it is to give pre_config a special status so that later config to the same value will be ignored so that command line config will always take precedence of config file.

EDIT:
Actually, current behavior of pre_config is already this. The problem is the pre_config inserted at the beginning by write_config.
Perhaps we can just give a warning in this case (i.e. warning for a new immutable config even if there is already an immutable config)?

> 
> Also added examples_dir in cmd() during play because sometimes there is a dependency file in examples. In the future we should probably write code to trace all dependency files under examples and copy them to root_dir.

hnyu:
> EDIT:
> Actually, current behavior of pre_config is already this. The problem is the pre_config inserted at the beginning by write_config.
> Perhaps we can just give a warning in this case (i.e. warning for a new immutable config even if there is already an immutable config)?
> 

Yeah I also noticed that this is due to inserted pre_config (command-line configs) in the written config file. So to solve this we will have to define another level of overwriting priority (training vs play). 

I think we can also just assume the overwriting priority as (immutable > mutable, mutable >= mutable, immutable != immutable), where >= means "can overwrite."

emailweixu:
> > EDIT:
> > Actually, current behavior of pre_config is already this. The problem is the pre_config inserted at the beginning by write_config.
> > Perhaps we can just give a warning in this case (i.e. warning for a new immutable config even if there is already an immutable config)?
> 
> Yeah I also noticed that this is due to inserted pre_config (command-line configs) in the written config file. So to solve this we will have to define another level of overwriting priority (training vs play).
> 
> I think we can also just assume the overwriting priority as (immutable > mutable, mutable >= mutable, immutable >= immutable), where >= means "can overwrite."

If we don't introduce another level of priority (which I prefer because of simplicity), we need immutable not to be overridden by future immutable.

hnyu:
> > > EDIT:
> > > Actually, current behavior of pre_config is already this. The problem is the pre_config inserted at the beginning by write_config.
> > > Perhaps we can just give a warning in this case (i.e. warning for a new immutable config even if there is already an immutable config)?
> > 
> > 
> > Yeah I also noticed that this is due to inserted pre_config (command-line configs) in the written config file. So to solve this we will have to define another level of overwriting priority (training vs play).
> > I think we can also just assume the overwriting priority as (immutable > mutable, mutable >= mutable, immutable >= immutable), where >= means "can overwrite."
> 
> If we don't introduce another level of priority (which I prefer because of simplicity), we need immutable not to be overridden by future immutable.

The issue above is that in saved config file there is a training command-line random seed (e.g., 0). And during play we should ignore that value (either a random seed is generated, or the user specify his random seed). So in this case, future immutable can overwrite existing immutable.

.ci-cd/build.sh
```diff
@@ -70,6 +70,7 @@ function test() {
         alf.algorithms.trac_algorithm_test \
         alf.algorithms.vae_test \
         alf.bin.train_play_test \
+        alf.bin.train_play_test_alfconf.py \
```
emailweixu: File name should end with "_test.py" so that `python -m unittest discover -s alf -p "*_test.py" -v` can run all tests. It seems to me most contents of train_play_test and train_play_test_alfconf are same. Perhaps we can just modify train_play_test instead of writing a new one?

hnyu: merged

alf/config_util.py
```diff
@@ -251,13 +251,15 @@ def config1(config_name, value, mutable=True, raise_if_used=True):
             config_node.set_mutable(mutable)
         else:
             if not mutable:
-                raise ValueError(
+                logging.warning(
+                    "The config '%s' has been configured to an immutable value of"
+                    " %s. It is replaced by a new immutable value of %s" %
```
emailweixu: It will not be replaced by ...
Also need to update docstring

alf/config_util.py
```diff
@@ -311,7 +313,11 @@ def validate_pre_configs():
     """Validate that all the configs set through ``pre_config()`` are correctly bound."""
 
     if _PRE_CONFIGS:
-        raise ValueError("Cannot find config name %s" % _PRE_CONFIGS[0][0])
+        raise ValueError((
+            "A pre-config '%s' was not handled, either because its config name "
+            +
+            "was not found, or there was some conflict with existing configured "
```
emailweixu: "conflict with existing configured" does not cause this.

hnyu: This valueerror was exactly what I saw for the issue causing this PR. When a pre_config cannot be configured, _PRE_CONFIG is not empty.  (Line 286-288) Maybe should just say "configuring is not successful"

alf/config_util.py
hnyu: Right, I was confused with the order of configuration when play.

hnyu:
Now the random seed behavior is to just ignore training random seed no matter a play random seed is provided or not. This is achieved by first configuring GFLAGS random_seed as immutable during play.

emailweixu (approved):

PR #813, opened 2021-03-17
support multi-dim gamma in td loss

hnyu:


alf/algorithms/td_loss.py
```diff
@@ -94,24 +96,33 @@ def forward(self, experience, value, target_value):
         Returns:
             LossInfo: with the ``extra`` field same as ``loss``.
         """
+        if experience.reward.ndim == 3:
+            gamma = self._gamma
+            if gamma.ndim == 0:  # scalar
+                gamma = gamma.unsqueeze(0)
+            discounts = torch.ger(experience.discount.reshape(-1), gamma)
+            discounts = discounts.reshape(experience.discount.shape + (-1, ))
```
emailweixu:
```python
# [T, B, d] or [T, B, 1]
discounts = discount.unsqueeze(-1) * gamma
```

emailweixu (changes requested):

alf/algorithms/td_loss.py
hnyu: I didn't know [..., 1] * [D] is broadcastable.

emailweixu (approved):

PR #818, opened 2021-03-18
support alf conf in grid_search

hnyu:
Issue #817 

Also store an ALF snapshot before searching to make sure file consistency.

### DIFF
--- a/alf/bin/grid_search.py
+++ b/alf/bin/grid_search.py
@@ -25,12 +25,15 @@ To run grid search on DDPG for training gym `Pendulum`:
     --gin_param='create_environment.num_parallel_environments=8' \
     --alsologtostderr
 
+For using ALF conf, replace "--gin_file" with "--conf" and "--gin_param" with
+"--conf_param".
 """
 
 from absl import app
 from absl import flags
 from absl import logging
 from collections import Iterable
+import copy
 import gin
 import itertools
 import json
@@ -39,10 +42,13 @@ import os
 # `pathos.multiprocessing` provides a consistent interface with std lib `multiprocessing`
 # and it's more flexible
 from pathos import multiprocessing
+import pathlib
 import random
 import time
 import torch
 import traceback
+import subprocess
+import sys
 
 import alf
 from alf.bin.train import train_eval
@@ -50,6 +56,10 @@ from alf.utils import common
 
 flags.DEFINE_string('search_config', None,
                     'Path to the grid search config file.')
+flags.DEFINE_bool(
+    'snapshot_gridsearch_activated', False,
+    'Whether a snapshot has been generated for grid search. (ONLY '
+    'change this flag manually if you know what you are doing!')
 
 FLAGS = flags.FLAGS
 
@@ -89,10 +99,10 @@ class GridSearchConfig(object):
         repeats (int): each parameter combination will be repeated for so many
             times, with different random seeds.
         parameters (dict): a ``dict(param_name=param_value,)`` of the configured
-            search space. Each key ``param_name`` is a gin configurable argument
+            search space. Each key ``param_name`` is a gin/alf configurable argument
             string and the paired ``param_value`` must be an iterable python object
             or a ``str`` that can be evaluated to an iterable object. When
-            ``parameters`` is empty, the original gin conf won't be changed.
+            ``parameters`` is empty, the original conf file won't be changed.
 
     See ``alf/examples/ddpg_grid_search.json`` for an example.
     """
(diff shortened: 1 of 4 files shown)


alf/bin/grid_search.py
```diff
@@ -258,10 +260,11 @@ def run(self):
...
 
-        # parsing gin configuration here to make all jobs have same copy
-        #   of base configuration (gin file may be changed occasionally)
-        gin_file = common.get_gin_file()
-        gin.parse_config_files_and_bindings(gin_file, FLAGS.gin_param)
+        # parsing gin/alf configuration here to make all jobs have same copy
+        #   of base configuration (conf file may be changed occasionally)
+        conf_file = common.get_conf_file()
+        # Do not init env seed here because this will init cuda in the main proc
+        common.parse_conf_file(conf_file, get_env=False)
```
emailweixu: Some config need env in order to correctly run (e.g. get obervation_spec).

hnyu: I just moved the last step (`get_env()`) of the original `parse_config_file()`  to separate worker process below. The order is not changed. 

`alf.parse_config` was only called once in the current code and its last step is to call `get_env()`. This call can be merged with gin in `common.parse_config_file()`

hnyu: I see what you mean. In our new alf conf there are places that directly call get_env() in the conf file. Unfortunately this will init cuda device in the main process. I think there might be some way to delay evaluating any env spec (not evaluating them until actually using them).

emailweixu: Why does the conf file needs to be parsed in the main process?

hnyu: If there are multiple jobs in the queue, some jobs will get launched after hours. During that period of time, the conf file might have been changed on disk.

hnyu: I can add an option of "init_random_seed" when calling `get_env()`. The calls in conf files will not use this option.

alf/bin/grid_search.py
```diff
@@ -302,11 +308,22 @@ def _worker(self, root_dir, parameters, device_queue):
...
+                with gin.unlock_config():
+                    gin.parse_config(
+                        ['%s=%s' % (k, v) for k, v in parameters.items()])
+                    gin.parse_config(
+                        "TrainerConfig.confirm_checkpoint_upon_crash=False")
+            else:
+                # re-bind alf conf params
+                alf.config(parameters)
+                alf.config(
+                    'TrainerConfig', confirm_checkpoint_upon_crash=False)
```
emailweixu: This won't work. If conf file is parsed (executed), changing config later is not guaranteed to have effect. Conf file has to run after these config()s (and they need to be changed to preconfig()s). One possibility is to make a copy the conf file in the main process and parse them in the worker later.

emailweixu: Copying also has its own problems. Because conf is a python executable, it can import any kind files. It's hard to track those files.

alf/utils/common.py
```diff
@@ -1089,3 +1090,33 @@ def rsync(src, target, includes):
...
+                             "alf/nest/cnest")  # path to archived cnest.so
+    python_path = os.environ.get("PYTHONPATH", "")
+    python_path = ":".join([alf_repo, alf_cnest, python_path])
+    env_vars = copy.copy(os.environ)
+    env_vars.update({"PYTHONPATH": python_path})
+    return env_vars
+
+
+def format_specified_flags():
+    """Format all user specified FLAGS into a string."""
```
emailweixu: The return value is not a string. It is a list.

alf/utils/common.py
```diff
@@ -1089,3 +1090,33 @@ def rsync(src, target, includes):
...
+    opts = []
+    for attr, flag in flags.FLAGS.__flags.items():
```
emailweixu: `__flags` is prefixed with`__`. It's supposed to be private, not to be used by users. All the command args are in `sys.argv`

alf/bin/grid_search.py
```diff
@@ -315,11 +345,60 @@ def _worker(self, root_dir, parameters, device_queue):
...
+    flags = common.format_specified_flags()
+    # remove the conf file option since we will retrieve it from ``root_dir``
+    flags = [f for f in flags if not ('--conf=' in f or '--gin_file' in f)]
+    flags.append('--snapshot_gridsearch_activated')
+
+    args = ['python', '-m', 'alf.bin.grid_search'] + flags
+
+    try:
+        print(
+            "=== Grid searching using an ALF snapshot at '%s' ===" % alf_repo)
```
emailweixu, line 393: logging.info

alf/utils/common.py
hnyu: removed this function; just use sys.argv[1:]

emailweixu (approved):

PR #819, opened 2021-03-18
Fix summary name for inoperative config

emailweixu:


hnyu (approved):
Also fixed in PR #818

PR #822, opened 2021-03-20
play alf snapshot

hnyu:
This PR enables playing a trained model from past to be consistent with the historical code that trained the model. Basically before training we stores ALF code (*py files) to ``root_dir`` and later when playing we modify the python path to point to that stored repo. 

If this PR gets approved, I will continue working on PR #818 by incorporating a similar trick in grid search.

Tested on simple train/play cases; this PR doesn't break existing trained model.

For some reason, I can't store code snapshot on the CI server. So added an option --nostore_snapshot for train.py and enable it when running train_play_test.py.

### DIFF
--- a/alf/bin/play.py
+++ b/alf/bin/play.py
@@ -18,7 +18,6 @@ You can visualize playing of the trained model by running:
 cd ${PROJECT}/alf/examples;
 python -m alf.bin.play \
   --root_dir=~/tmp/cart_pole \
-  --gin_file=ac_cart_pole.gin \
   --alsologtostderr
 ```
 
@@ -27,12 +26,14 @@ python -m alf.bin.play \
 from absl import app
 from absl import flags
 from absl import logging
+import copy
 import gin
 import os
+import subprocess
+import sys
 
 import torch
 
-from alf.algorithms.algorithm import Algorithm
 from alf.algorithms.data_transformer import create_data_transformer
 from alf.environments.utils import create_environment
 from alf.trainers import policy_trainer
@@ -88,6 +89,7 @@ def main(_):
     alf.config('create_environment', nonparallel=True)
     alf.config('TrainerConfig', mutable=False, random_seed=seed)
     conf_file = common.get_conf_file()
+    assert conf_file is not None, "Conf file not found! Check your root_dir"
     try:
         common.parse_conf_file(conf_file)
     except Exception as e:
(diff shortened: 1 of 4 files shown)


alf/bin/play.py
```diff
@@ -133,9 +135,57 @@ def main(_):
...
+
+
 if __name__ == '__main__':
-    logging.set_verbosity(logging.INFO)
     flags.mark_flag_as_required('root_dir')
-    if torch.cuda.is_available():
-        alf.set_default_device("cuda")
-    app.run(main)
+    snapshot_play = int(os.environ.get("ALF_SNAPSHOT_RUN", "0"))
+    if not snapshot_play:
```
Haichao-Zhang: ```if not snapshot_play``` -> ```if snapshot_play``` ?

hnyu: By default the first layer of play has no env var "ALF_SNAPSHOT_RUN" defined. This special env var is defined later

Haichao-Zhang: The name is a bit mis-leading. May be can be improved, e.g. to something like ``snapshot_play_launched``?

hnyu: updated

alf/bin/play.py
```diff
@@ -133,9 +135,58 @@ def main(_):
...
 
+def launch_snapshot_play(_):
+    """This play function uses historical ALF snapshot for playing a trained
+    model, consistent with the code snapshot that trains the model.
+
+    In the newer version of ``train.py``, a ALF snapshot is saved to ``root_dir``
+    right before the training begins. So this function prepends ``root_dir`` to
+    ``PYTHONPATH`` to allow using the snapshot ALF repo in that place.
+
+    Note that for any old training ``root_dir``, this won't change the ALF repo
```
le-horizon: "old" => such,
old can mean an older version of the alf repo that's older than this commit.

alf/bin/play.py
```diff
@@ -133,9 +135,58 @@ def main(_):
...
+    version play uses, and also doesn't break anything.
```
le-horizon: doesn't break anything => doesn't change the shell environment variables?  Be more specific?

alf/bin/play.py
```diff
@@ -133,9 +135,58 @@ def main(_):
...
+    print("^^^^^^^^^ End of ALF snapshot play ^^^^^^^^^^^")
+
+
 if __name__ == '__main__':
-    logging.set_verbosity(logging.INFO)
     flags.mark_flag_as_required('root_dir')
-    if torch.cuda.is_available():
-        alf.set_default_device("cuda")
-    app.run(main)
+    snapshot_play_activated = int(os.environ.get("ALF_SNAPSHOT_RUN", "0"))
```
le-horizon: relying on an env variable is very hidden, and can be difficult to debug.  Maybe just add a flag, e.g. enable_snapshot_run, and remember to remove the flag in the launch_snapshot_play function before sending to subprocess.check_call.

alf/bin/train.py
```diff
@@ -91,6 +94,14 @@ def main(_):
     root_dir = os.path.expanduser(FLAGS.root_dir)
     os.makedirs(root_dir, exist_ok=True)
     logging.get_absl_handler().use_absl_log_file(log_dir=root_dir)
+
+    if FLAGS.store_snapshot:
+        # ../<ALF_REPO>/alf/bin/train.py
+        common.set_alf_root(
+            str(pathlib.Path(__file__).parent.parent.parent.absolute()))
```
le-horizon: if _alf_root isn't used elsewhere, then just remove this global variable and instead pass it `str(pathlib.Path(__file__).parent.parent.parent.absolute())` in as a second argument to the following generate_alf_root_snapshot function?

le-horizon (commented):
Thanks for the nice improvement, Haonan!

Some suggestions if they make sense.

Thanks,
Le

alf/bin/play.py
runjerry: I assume that "newer" means the version updated by this PR, then it could be confusing after the PR is merged.

runjerry: I assume that "newer" means the version updated by this PR, then it could be confusing after the PR is merged.

alf/bin/train.py
hnyu: I just added _alf_root for potential future use.

alf/bin/play.py
hnyu: I agree. Updated

alf/bin/play.py
hnyu: updated docstring

le-horizon (dismissed):
LGTM, thanks Haonan.

hnyu:
@emailweixu Do you have any comment?

alf/bin/train.py
hnyu: Actually I decided to remove _alf_root global variable.

le-horizon (approved):
LG.

emailweixu (approved):
Great! This is very helpful.

PR #824, opened 2021-03-26
Perspective projection and visualization of world/ego points in Carla

Haichao-Zhang:
Added support to project and visualize points in world and ego coordinates in the camera image.
This is useful to get a better understanding of items such as waypoints, trajectories etc. through visualization.

This type of rendering is different from ```debug.draw_point()``` function from Carla in the following aspects:
1. ```debug.draw_point()``` will paint the point in the world, which can modify the sensor perception. This is not desirable and has side-effects in some cases, e.g. it will impact the input to a learned model.
2. ```debug.draw_point()``` has one time-step lag. The points only become visible in the next time step. This is not desirable when we want to debug items that are changing at each time step.

An example usage for visualizing the waypoint (with interpolation) is shown here:
ezgif com-gif-maker (10)

### DIFF
--- a/alf/environments/carla_sensors.py
+++ b/alf/environments/carla_sensors.py
@@ -679,7 +679,7 @@ class CameraSensor(SensorBase):
             image_height (int): the height of the image to be rendered on
         Output:
             np.ndarray: representing the image plane coordinates for the set
-                of points that are visible in the camera. Its shape is [N', 3],
+                of points that are visible in the camera. Its shape is [N', 2],
                 with N' <= N.
         """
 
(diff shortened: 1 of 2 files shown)


alf/environments/carla_sensors.py
```diff
@@ -616,10 +618,124 @@ def render(self, display):
...
+        point_in_camera_coords = point_in_camera_coords[:, valid_ind]
+
+        # use projection matrix K to do the mapping from 3D to 2D
+        points_2d = np.dot(K, point_in_camera_coords)
+
+        # normalize the x, y values by the third value
+        points_2d = np.array([
+            points_2d[0, :] / points_2d[2, :],
+            points_2d[1, :] / points_2d[2, :], points_2d[2, :]
+        ])
```
emailweixu: Do we need the third dimension in the return value of points_2d?
```python
points_2d = points_2d[:2, :] / points_2d[2, :]
```

alf/environments/carla_sensors.py
```diff
@@ -616,10 +618,124 @@ def render(self, display):
...
+        # transform the points from world space to camera space through the
+        # following several steps
+
+        # [N, 3] -> [3, N]
+        world_points = np.transpose(world_points, (1, 0))
+        # [3, N] -> [4, N]
+        world_points = np.concatenate(
+            (world_points, np.ones_like(world_points[0:1, ...])), axis=0)
+
+        sensor_points = np.dot(world_2_camera, world_points)
```
emailweixu: is np.dot same as np.matmul here?

alf/environments/carla_sensors.py
```diff
@@ -616,10 +618,124 @@ def render(self, display):
...
+
+        # change from UE4's left handed coordinate system to a typical
+        # right handed camera coordinate system, which is equivalent to
+        # axis swapping: (x, y, z) -> (y, -z, x)
+        point_in_camera_coords = np.array(
+            [sensor_points[1], sensor_points[2] * -1, sensor_points[0]])
```
emailweixu: why not `-sensor_points[2]`?

alf/environments/carla_sensors.py
```diff
@@ -616,10 +618,124 @@ def render(self, display):
...
+        points_2d = np.dot(K, point_in_camera_coords)
+
+        # normalize the x, y values by the third value
+        points_2d = np.array([
+            points_2d[0, :] / points_2d[2, :],
+            points_2d[1, :] / points_2d[2, :], points_2d[2, :]
+        ])
+
+        # [3, N] -> [N, 3]
+        points_2d = np.transpose(points_2d, (1, 0))
```
emailweixu: `points_2d = points_2d.transpose()`

alf/environments/carla_sensors.py
Haichao-Zhang: Returning the third dimension is unnecessary. Updated according to suggestion.

alf/environments/carla_sensors.py
Haichao-Zhang: Yes they are the same as the arguments are 2D arrays. 
Changed ```np.dot``` to ```np.matmul```, as the latter one is preferred according to np documents for 2D case.

alf/environments/carla_sensors.py
Haichao-Zhang: Changed

emailweixu (approved):

PR #825, opened 2021-03-26
Utility to verify algorithm loaded from checkpoint is same as the original algorithm

emailweixu:


alf/algorithms/config.py
```diff
@@ -181,48 +181,45 @@ def __init__(self,
...
-            clear_replay_buffer=clear_replay_buffer,
-            replay_buffer_length=replay_buffer_length,
-            priority_replay=priority_replay,
-            priority_replay_alpha=priority_replay_alpha,
-            priority_replay_beta=priority_replay_beta,
-            priority_replay_eps=priority_replay_eps,
-            num_envs=num_envs)
-        for k, v in parameters.items():
-            self.__setattr__(k, v)
+        self.root_dir = root_dir
```
emailweixu, line 184: The reason of making these changes is so that vscode can know the attributes of config for error reporting and hint.

alf/bin/verify_checkpoint.py
```diff
@@ -0,0 +1,211 @@
...
+flags.DEFINE_float('epsilon_greedy', 1., "probability of sampling action.")
+flags.DEFINE_integer('random_seed', None, "random seed")
+flags.DEFINE_integer('num_train_iterations', 2,
+                     "number of training iterations")
+flags.DEFINE_integer('num_test_steps', 10, "number of test steps")
+flags.DEFINE_string('gin_file', None, 'Path to the gin-config file.')
+flags.DEFINE_multi_string('gin_param', None, 'Gin binding parameters.')
+flags.DEFINE_string('conf', None, 'Path to the alf config file.')
+flags.DEFINE_multi_string('conf_param', None, 'Config binding parameters.')
+flags.DEFINE_float('tolerance', 0., "Allowed difference between to runs")
```
hnyu: two runs

alf/bin/verify_checkpoint.py
```diff
@@ -0,0 +1,211 @@
...
+
+
+def main(_):
+    with tempfile.TemporaryDirectory() as root_dir:
+        FLAGS.root_dir = root_dir
+        conf_file = common.get_conf_file()
+        step_num = FLAGS.num_train_iterations
+        ckpt_dir = os.path.join(root_dir, 'ckpt')
+        algorithm1, env1 = _create_algorithm_and_env(root_dir)
+        for i in range(FLAGS.num_train_iterations):
```
hnyu, line 163: Better comment that here training is used to make sure the parameters of the algorithm actually changes to a certain state different from the initial state before checkpoint.

alf/bin/verify_checkpoint.py
```diff
@@ -0,0 +1,211 @@
+# Copyright (c) 2021 Horizon Robotics and ALF Contributors. All Rights Reserved.
```
hnyu, line 1: Should we add this to CI for all checked-in algorithms?

hnyu (changes requested):

alf/bin/verify_checkpoint.py
emailweixu: fixed

alf/bin/verify_checkpoint.py
emailweixu: Added.

alf/bin/verify_checkpoint.py
emailweixu: I thought about this. Then decided to do it later until this is more mature.

alf/bin/verify_checkpoint.py
```diff
@@ -0,0 +1,215 @@
...
+        else:
+            time_step = time_steps[i]
+
+        policy_steps.append(policy_step)
+        policy_state = policy_step.state
+
+    return policy_steps, time_steps
+
+
+def _create_algorithm_and_env(root_dir):
```
le-horizon, line 124: should we set random seed here as well, so env randomness will also be the same?

alf/bin/verify_checkpoint.py
```diff
@@ -0,0 +1,215 @@
...
+from alf.algorithms.data_transformer import create_data_transformer
+from alf.environments.utils import create_environment
+from alf.trainers import policy_trainer
+from alf.utils import common, dist_utils
+import alf.utils.checkpoint_utils as ckpt_utils
+
+flags.DEFINE_string('root_dir', os.getenv('TEST_UNDECLARED_OUTPUTS_DIR'),
+                    'Root directory for writing logs/summaries/checkpoints.')
+flags.DEFINE_float('epsilon_greedy', 1., "probability of sampling action.")
+flags.DEFINE_integer('random_seed', None, "random seed")
```
le-horizon, line 60: perhaps random_seed default shouldn't be None?  So default behavior will be consistent?

le-horizon (commented):
Nice checker Wei!  Only two comments if they make sense.

Thanks,
Le

alf/bin/verify_checkpoint.py
emailweixu: env randomness doesn't matter, as I store all the env timesteps so the second run will use the exact same timesteps.

alf/bin/verify_checkpoint.py
emailweixu: My current thinking is that the test should always pass no matter what the seed is. If not, we can consider fixed seed in the future.###

le-horizon: make sense.  given env randomness is gone, this should work.

alf/bin/verify_checkpoint.py
le-horizon: I see, this makes a more reliable test!

alf/bin/verify_checkpoint.py
```diff
@@ -0,0 +1,215 @@
...
+        ckpt_mngr2.load(step_num)
+
+        seed = common.set_random_seed(FLAGS.random_seed)
+        policy_steps1, time_steps = _run_steps(algorithm1, env1,
+                                               FLAGS.num_test_steps)
+        # We calulate policy_steps1 again to make sure policy_steps1 and
+        # policy_steps2 go through exactly same computation sequence so that
+        # they can be compared with each other.
+        common.set_random_seed(seed)
+        policy_steps1, _ = _run_steps(algorithm1, env1, FLAGS.num_test_steps,
```
le-horizon: instead of env1 just use None to be clear?

alf/bin/verify_checkpoint.py
```diff
@@ -0,0 +1,215 @@
...
+                                      time_steps)
+        common.set_random_seed(seed)
+        policy_steps2, _ = _run_steps(algorithm2, env2, FLAGS.num_test_steps,
```
le-horizon: env2->None?

le-horizon (approved):
LG, only one nit.

hnyu (dismissed):

alf/bin/verify_checkpoint.py
hnyu: > instead of env1 just use None to be clear?

None will break the code because env.batch_size is used in the function

le-horizon: Oh, good point, then maybe change env2 below to env1, so it's clear the only difference comes from algorithm1 & 2?

le-horizon (dismissed):

hnyu (approved):

PR #826, opened 2021-03-26
More flexible Sequential

emailweixu:


alf/layers.py
```diff
@@ -2338,21 +2344,144 @@ def reset_parameters(self):
...
-class Sequential(nn.Sequential):
-    """A thin wrapper over nn.Sequential to allow stateless alf.networks.Network."""
+def _get_nested_field(nest, nest_fields):
+    return map_structure(lambda f: get_field(nest, f), nest_fields)
+
+
+class Sequential(nn.Module):
+    """A more flexible Sequential than torch.nn.Sequential.
+
+    ``alf.layers.Sequentail`` is similar to ``alf.nn.Sequential``, but does not
```
Haichao-Zhang: Sequentail -> Sequential

alf/networks/containers.py
```diff
@@ -51,34 +101,70 @@ def Sequential(*modules, input_tensor_spec=None, name="Sequential"):
...
-    return _Sequential(modules, input_tensor_spec=input_tensor_spec, name=name)
+    return _Sequential(
+        modules,
+        named_modules,
+        output=output,
+        input_tensor_spec=input_tensor_spec,
+        name=name)
+
+
+def _get_nested_field(nest, nest_fields):
```
Haichao-Zhang: This function also appears in layers.py and can potentially be used in other cases as well.
Maybe we can move it to nest.utils?

alf/layers_test.py
```diff
@@ -626,6 +626,23 @@ def test_causal_conv1d_output_causality(self, batch_size=5):
...
+        batch_size = 24
+        x = torch.randn((batch_size, 4))
+        y = net(x)
+
+        x1 = net[0](x)
+        x2 = net[1](x1)
+        x3 = net[2](x2)
+        x4 = net[3]((x2, x3))
+        self.assertEqual(x4, y)
+
```
le-horizon, line 645: let's add a few more tests to cover the code better:
1) using 'input': e.g. `c=(('a', 'b', 'input'), alf.layers.NestConcat()))`
2) throws error when out of sequence named_modules: e.g. move c=('a', module) ahead of a= and b=
3) using output argument: output=c

alf/layers.py
```diff
@@ -2338,21 +2344,144 @@ def reset_parameters(self):
...
+            named_modules.items())
+        modules = []
+        inputs = []
+        outputs = []
+        simple = True
+        is_nested_str = lambda s: all(
+            map(lambda x: type(x) == str, alf.nest.flatten(s)))
+        for i, (out, element) in enumerate(named_elements):
+            input = ''
+            if isinstance(element, tuple) and len(element) == 2:
```
le-horizon, line 2436: to be safer, let's use this?
```
if isinstance(element, tuple):
    assert len(element) == 2
    input, module = element
```

alf/layers.py
```diff
@@ -2338,21 +2344,144 @@ def reset_parameters(self):
...
+                input, module = element
+            else:
+                module = element
+            if not (isinstance(module, nn.Module) and is_nested_str(input)):
```
le-horizon: can module here be alf.Network?

should we assert module to be either Network or Module regardless of what input is?

alf/layers.py
```diff
@@ -2338,21 +2344,144 @@ def reset_parameters(self):
...
+                the dictionary of the current ``named_results``. ``named_results``
+                is updated once the result of a named module is calculated.
+            output (nested str): if not provided, the result from the last module
+                will be used as output. Otherwise, it will be used to retrieve
+                results from ``named_results`` after the results of all modules
+                have been calculated.
+        """
+        super().__init__()
+        named_elements = list(zip([''] * len(modules), modules)) + list(
+            named_modules.items())
```
le-horizon, line 2424: change setup.py to require python >= 3.6?  Otherwise named_modules may be out of order?  https://www.python.org/dev/peps/pep-0468/

alf/layers.py
le-horizon: Is this comment correct?  Code seems to allow use of stateless networks?

le-horizon (commented):
Nice improvements Wei.  This makes stacking networks more fun.

Since it's more flexible, it'll be good to add a few more tests.

The layers and networks Sequentials have lots of code overlap, it's Ok as is, but would be good to have a future plan to somehow simplify it.

Thanks,
Le

alf/layers.py

alf/layers.py
emailweixu: good point

alf/layers.py
emailweixu: It's ok here. Wrong cases can be handled by line 2437

alf/networks/containers.py
emailweixu: moved.

alf/layers.py
emailweixu: alf Network is always an nn.Module

alf/layers_test.py
emailweixu: added

Haichao-Zhang (approved):

le-horizon (approved):
LG, thanks for the quick improvements!

PR #828, opened 2021-03-27
make discounted return same with tf_agents

hnyu:
Make our discounted_return behave the same with tf_agents. The reason for this is that the returned values for StepType.LAST don't matter in our existing use cases because they will be masked out when computing critic loss. But in rare cases where is_lasts have customized meanings, StepType.LAST might need accumulated discounted return (since the next StepType.LAST) instead of its own value.


hnyu:
Closed due to inconsistency with other value operator behaviors

PR #829, opened 2021-03-29
fix use_entropy_reward for multi-step TD

hnyu:
currently if use_entropy_reward=True in SAC, the entropy reward is only added to target_critics. This is fine for one-step TD. But for multi-step TD, the intermediate steps don't have entropy rewards. So I follow the SARSA's way to add entropy into experience.reward. 

I'm doing some experiments (SAC+N-TDLoss) for verification. But it seems that this fix works better than before so far.

### DIFF
--- a/alf/algorithms/td_loss.py
+++ b/alf/algorithms/td_loss.py
@@ -99,7 +99,12 @@ class TDLoss(nn.Module):
 
     @property
     def gamma(self):
-        return self._gamma
+        """Return the :math:`\gamma` value for discounting future rewards.
+
+        Returns:
+            Tensor: a rank-0 or rank-1 (multi-dim reward) floating tensor.
+        """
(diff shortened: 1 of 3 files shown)


alf/algorithms/td_loss.py
```diff
@@ -97,6 +97,10 @@ def __init__(self,
         self._normalize_target = normalize_target
         self._target_normalizer = None
 
+    @property
+    def gamma(self):
```
emailweixu, line 101: comment

emailweixu (approved):

Haichao-Zhang (approved):

hnyu:
Update: actually with this fix SAC did better on the four Fetch environments but significantly worse on BipedalWalker{HardCore}. I feel that having entropy reward in SAC in general needs discussion?

PR #832, opened 2021-04-01
Make sure docs are generated for the scripts under alf/bin

emailweixu:
Previously, the docs of some of them are not generated. It seems to be caused
by the fact that some of those scripts try to define same flags when the module
is imported. So I moved the flags definitions into a function so that they are
defined only when the script is invoked.


hnyu (approved):

hnyu:
I've noticed this issue for a while

PR #834, opened 2021-04-02
rendering interface

hnyu:
Rerun 'pip install -e .' under ALF root to install a newer version of matplotlib.

Currently support three kinds of renderings in render.py: curve, heatmap, and bar. I've deleted mostly rendering code in video_recorder.py. @Haichao-Zhang Please modify your encode_future_info code using the new interface. 

Example rendering is below:

image

By adding the following code block in SAC's `predict_step()`:

```python
            action_img = render.render_action(name="action",
                                              action=action,
                                              action_spec=self._action_spec)
            action_heatmap = render.render_heatmap(name="action_heatmap",
                                                   data=action,
                                                   val_label="action")
            act_dist_curve = render.render_action_distribution(
                name="action_dist",
                act_dist=action_dist,
                action_spec=self._action_spec)

        return AlgStep(
            output=action,
            state=SacState(action=action_state),
            info=dict(
                action_img=action_img,
                action_heatmap=action_heatmap,
                action_dist_curve=act_dist_curve,
                info=SacInfo(action_distribution=action_dist)))
```

### DIFF
--- a/alf/bin/play.py
+++ b/alf/bin/play.py
@@ -74,9 +74,9 @@ def _define_flags():
     flags.DEFINE_bool(
         'render', True,
         "Whether render ('human'|'rgb_array') the frames or not")
-    # use '--render_prediction' to enable pred info rendering
-    flags.DEFINE_bool('render_prediction', False,
-                      "Whether render prediction info at every frame or not")
+    # use '--alg_render' to enable algorithm specific rendering
+    flags.DEFINE_bool('alg_render', False,
+                      "Whether enable algorithm specific rendering")
     flags.DEFINE_string('gin_file', None, 'Path to the gin-config file.')
     flags.DEFINE_multi_string('gin_param', None, 'Gin binding parameters.')
     flags.DEFINE_string('conf', None, 'Path to the alf config file.')
@@ -97,6 +97,8 @@ FLAGS = flags.FLAGS
 def play():
     if torch.cuda.is_available():
         alf.set_default_device("cuda")
+
+    alf.summary.render.enable_rendering(FLAGS.alg_render)
 
     seed = common.set_random_seed(FLAGS.random_seed)
     alf.config('create_environment', nonparallel=True)
@@ -141,7 +143,6 @@ def play():
             future_steps=FLAGS.future_steps,
             append_blank_frames=FLAGS.append_blank_frames,
             render=FLAGS.render,
-            render_prediction=FLAGS.render_prediction,
             ignored_parameter_prefixes=FLAGS.ignored_parameter_prefixes.split(
                 ",") if FLAGS.ignored_parameter_prefixes else [])
     finally:
--- a/alf/bin/play.py
+++ b/alf/bin/play.py
@@ -53,9 +53,6 @@ def _define_flags():
     flags.DEFINE_float('epsilon_greedy', 0., "probability of sampling action.")
     flags.DEFINE_integer('random_seed', None, "random seed")
     flags.DEFINE_integer('num_episodes', 10, "number of episodes to play")
-    flags.DEFINE_integer('max_episode_length', 0,
-                         "If >0,  each episode is limited "
-                         "to so many steps")
     flags.DEFINE_integer(
         'future_steps', 0, "If >0, display information from so many "
         "number of future steps in addition to the current step "
@@ -137,7 +134,6 @@ def play():
             checkpoint_step=FLAGS.checkpoint_step or "latest",
             epsilon_greedy=FLAGS.epsilon_greedy,
             num_episodes=FLAGS.num_episodes,
-            max_episode_length=FLAGS.max_episode_length,
             sleep_time_per_step=FLAGS.sleep_time_per_step,
             record_file=FLAGS.record_file,
             future_steps=FLAGS.future_steps,
--- a/alf/summary/render.py
+++ b/alf/summary/render.py
@@ -19,7 +19,10 @@ import matplotlib
 import matplotlib.pyplot as plt
 # Style gallery: https://tonysyu.github.io/raw_content/matplotlib-style-gallery/gallery.html
 plt.style.use('seaborn-dark')
-import rpack
+try:
+    import rpack
+except ImportError:
+    rpack = None
 
 import cv2
 
@@ -149,6 +152,8 @@ class Image(object):
         Returns:
             Image: the big mosaic image
         """
+        assert rpack is not None, "You need to install rectangle-packer first!"
+
         imgs = nest.flatten(imgs)
         if len(imgs) == 0:
             return
--- a/alf/summary/render.py
+++ b/alf/summary/render.py
@@ -147,7 +147,7 @@ class Image(object):
         that is used for building CSS sprites, for an approximate solution.
 
         Args:
-            imgs (nested Image): a nest of ``Image`` instances images
+            imgs (nested Image): a nest of ``Image`` instances
 
         Returns:
             Image: the big mosaic image
@@ -197,13 +197,6 @@ def _rendering_wrapper(rendering_func):
     """A wrapper function to gate the rendering function based on if rendering
     is enabled, and if yes generate a scoped rendering identifier before
     calling the rendering function. It re-uses the scope stack in ``alf.summary.summary_ops.py``.
-
-    Example:
-
-    .. code-block:: python
-
-        with alf.summary.scope("root"):
-            render_heatmap("val", ...)  # tag is "root/val"
     """
 
     @functools.wraps(rendering_func)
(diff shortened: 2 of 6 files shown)


hnyu:
I also added a rectangle packing solution so that several pred_info images can be packed into a larger image with an area as small as possible (approximate solution). An example is below (the left side is packed with a smallest possible area):

image

le-horizon (dismissed):
Thanks for the nice improvement @hnyu !

It's a fairly large PR, I only took a brief look, please wait for others' comments.

In the future, it would be best to separate the refactor changes and the new codes into separate commits, so reviewing would be easier.

Thanks,
Le

alf/summary/render.py
```diff
@@ -0,0 +1,658 @@
...
+
+import functools
+import io
+import numpy as np
+import matplotlib
+import matplotlib.pyplot as plt
+# Style gallery: https://tonysyu.github.io/raw_content/matplotlib-style-gallery/gallery.html
+plt.style.use('seaborn-dark')
+try:
+    import rpack
```
emailweixu, line 23: Since rpack is one of the requirement in setup.py, there is no need to do try except here.

alf/summary/render.py
```diff
@@ -0,0 +1,658 @@
...
+        assert rpack is not None, "You need to install rectangle-packer first!"
+
+        imgs = nest.flatten(imgs)
+        if len(imgs) == 0:
+            return
+
+        # first get all images' sizes (w,h)
+        sizes = [(i.shape[1], i.shape[0]) for i in imgs]
+        # call rpack for an approximate solution: [(x,y),...] positions
+        positions = rpack.pack(sizes)
```
emailweixu, line 164: Is rpack fast? If not, we can cache the solutions and reuse it because the sizes are going to be same.

alf/summary/render.py
```diff
@@ -0,0 +1,658 @@
...
+def _rendering_wrapper(rendering_func):
+    """A wrapper function to gate the rendering function based on if rendering
+    is enabled, and if yes generate a scoped rendering identifier before
+    calling the rendering function. It re-uses the scope stack in ``alf.summary.summary_ops.py``.
+
+    Example:
+
+    .. code-block:: python
+
+        with alf.summary.scope("root"):
```
emailweixu: this is not an example of using `_rendering_wrapper`

alf/summary/render.py
```diff
@@ -0,0 +1,658 @@
...
+
+def _convert_to_image(name, fig, dpi, height, width):
+    """First putting the rendering identifier on top of the figure and then
+    convert it to an instance of ``Image``. Also release the resources of
+    ``fig``.
+
+    Args:
+        name (str): a scoped identifier
+        fig (pyplot.figure): the figure holding the rendering
+        dpi (int): resolution of each rendered image
```
emailweixu, line 219: dpi seems to be redundant if height and width are given.

alf/summary/render.py
```diff
@@ -0,0 +1,658 @@
...
+    Args:
+        name (str): rendering identifier
+        data (Tensor|np.ndarray): a tensor/np.array of shape ``[H,W]``
+        val_label (str): The label for the rendered values.
+        row_labels (list[str]): A list of length ``H`` with the labels for
+            the rows.
+        col_labels (list[str]): A list of length ``W`` with the labels for
+            the columns.
+        cbar_kw (dict): A dictionary with arguments to ``matplotlib.Figure.colorbar``.
+        annotate_format (str): The format of the annotations on the heatmap.
```
emailweixu: not clear what annotation does.

alf/summary/render.py
```diff
@@ -0,0 +1,658 @@
...
+        y_range (tuple[float]): a tuple of ``(min_y, max_y)`` for showing on
+            the figure. If None, then it will be decided according to the
+            ``y`` values.
+        x_ticks (list[float]): x ticks shown along x axis
+        x_label (str): shown besides x-axis
+        y_label (str): shown besides y-axis
+        legends (list[str]): label for each curve. No legends are shown if
+            None.
+        legend_kwargs (dict): optional legend kwargs
+        annotate_format (str): The format of the annotations inside the heatmap.
```
emailweixu: not heatmap

alf/trainers/policy_trainer.py
```diff
@@ -736,14 +731,17 @@ def play(root_dir,
                                                max_episode_length > 0)
 
         if recorder:
+            # Note that because we capture the frame *after* calling ``env.step()``
+            # so the ``policy_step`` here contains prediction info of the frame
+            # *before* the currently rendered frame by the side. (i.e., pred info is
+            # always delayed by one frame).
```
emailweixu: Seems to me we should make a change so that there is no delay.

alf/summary/render.py
hnyu: This is just because the current CI will fail if I don't have this. I need to submit another PR first to add rpack in requirements.txt.

alf/summary/render.py
hnyu: It's very fast. No need to cache the results. Only takes about 3ms to pack 100 rectangles.

alf/summary/render.py
hnyu: dpi is the resolution. height and width are only for resizing. If dpi is low but height and width are big, then the resized image will look blurred.

alf/summary/render.py
hnyu: annotation is for overlaying a number on each heatmap slot to show the value of that slot.

alf/summary/render.py
```diff
@@ -0,0 +1,658 @@
...
+        fig.savefig(buf, format="png", dpi=dpi, bbox_inches="tight")
+        buf.seek(0)
+        img_arr = np.frombuffer(buf.getvalue(), dtype=np.uint8)
+        buf.close()
+        img = cv2.imdecode(img_arr, 1)
+        img = cv2.cvtColor(img, cv2.COLOR_BGR2RGB)
+        return cls(img)
+
+    @classmethod
+    def from_image_nest(cls, imgs):
```
Haichao-Zhang, line 140: ```from_image_nest```: can change the name of function to be more informative, e.g. ```pack_image_nest```?

alf/summary/render.py
```diff
@@ -0,0 +1,658 @@
...
+        """Given a nest of images, pack them into a larger image so that it has
+        an area as small as possible. This problem is generally known as
+        "rectangle packing" and its optimal solution is
+        `NP-complete <https://en.wikipedia.org/wiki/Rectangle_packing>`_.
+
+        Here we just rely on a third-party lib `rpack <https://pypi.org/project/rectangle-packer/>`_
+        that is used for building CSS sprites, for an approximate solution.
+
+        Args:
+            imgs (nested Image): a nest of ``Image`` instances images
```
Haichao-Zhang: ``Image`` instances images -> ``Image`` instances

alf/utils/video_recorder.py
```diff
@@ -173,20 +112,17 @@ class VideoRecorder(GymVideoRecorder):
...
-                 render_prediction=False,
                  **kwargs):
         """
         Args:
             env (Gym.env):
-            img_plot_width (int): the image width for displaying prediction info,
-                excluding the env frame width
-            value_range (float): if quantities plotted as heatmaps, the values
-                will be clipped according to ``[-value_range, value_range]``.
+            frame_width (int): the max width of a video frame. Scale if the original
```
Haichao-Zhang: frame_width -> frame_max_width

alf/summary/render.py
hnyu: This is a class method. I think it's quite clear Image.from_image_nest.

hnyu:
I've simplified the logic of play() in policy_trainer.py by removing the flag 'max_episode_length'. We can just use the `max_episode_steps` defined in a conf file.

alf/trainers/policy_trainer.py
hnyu: changed

alf/summary/render.py
emailweixu: Is it possible to set dpi according to width and height so the image is not blur?

hnyu: I think this depends on what kind of plot you have. Their relationship is different between curves and bars, for example. And it also has something to do with the figsize. https://stackoverflow.com/questions/47633546/relationship-between-dpi-and-figure-size

emailweixu (approved):

alf/bin/play.py
```diff
@@ -53,9 +53,6 @@ def _define_flags():
     flags.DEFINE_float('epsilon_greedy', 0., "probability of sampling action.")
     flags.DEFINE_integer('random_seed', None, "random seed")
     flags.DEFINE_integer('num_episodes', 10, "number of episodes to play")
-    flags.DEFINE_integer('max_episode_length', 0,
```
Haichao-Zhang, line 56: This flag is actually useful in some cases, e.g. when you only want to render a short video clips.

hnyu: You can just change the conf file, or passing a conf param? The point is that we don't want to handle the time out logic here in play.py because alf env will do that for us.

Haichao-Zhang (approved):

PR #837, opened 2021-04-06
Add support to speed limit penalty in Carla

Haichao-Zhang:
This PR is only on adding overspeed reward/penalty, without modifying observations.
Observation part can be done separately if required.

### DIFF
--- a/alf/environments/suite_carla.py
+++ b/alf/environments/suite_carla.py
@@ -343,8 +343,11 @@ class Player(object):
                 Note that this reward is only given once at the first step of
                 contiguous red light violation.
             overspeed_penalty_weight (float): if > 0, a penalty propotional to
-                the overspeed magnitude will be applied, multiplied by the weight
-                of ``overspeed_penalty_weight``. A negative value is the same as 0.
+                the overspeed magnitude will be applied, multiplied by the step
+                time (seconds each step of simulation represents) to make the
+                penalty invariant to it, and then multiplied by the weight
+                of ``overspeed_penalty_weight``.
+                A negative value is the same as 0.
             sparse_reward (bool): If False, the distance reward is given at every
                 step based on how much it moves along the navigation route. If
                 True, the distance reward is only given after moving ``sparse_reward_distance``.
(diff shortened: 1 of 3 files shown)


alf/environments/suite_carla.py
```diff
@@ -745,6 +792,14 @@ def get_current_time_step(self, current_frame):
                 Player.PENALTY_RATE_RED_LIGHT * max(0., self._episode_reward))
         self._prev_violated_red_light_id = red_light_id
 
+        overspeed = self.get_overspeed_amount()
+        if overspeed > 0:
+            logging.info("actor=%d frame=%d OVERSPEED" % (self._actor.id,
+                                                          current_frame))
+            reward_vector[Player.REWARD_OVERSPEED] = 1.
+            info['overspeed'] = np.float32(1.0)
+            reward -= self._overspeed_penalty_weight * overspeed
```
emailweixu: To make this invariant to delta_seconds, it needs to be multiplied with ``self._delta_seconds`` thus with same unit as distance.

alf/environments/suite_carla.py
```diff
@@ -615,6 +626,41 @@ def _get_goal(self):
         return _calculate_relative_position(self._actor.get_transform(),
                                             self._goal_location)
 
+    def update_speed_limit(self, dis_threshold=1.0):
+        """Update the speed limit of the actor according to the active speed
+        limit sign. The speed limit is updated when passing by a speed limit sign.
+
+        Args:
+            dis_threshold (float): the distance within which to consider the
```
emailweixu: It the unit meter? I feel 1 meter is too low. It may miss the speed limit sign.

emailweixu (changes requested):

alf/environments/suite_carla.py
Haichao-Zhang: Good point. Done.

alf/environments/suite_carla.py
Haichao-Zhang: The unit is meter. Doc str has been updated. The threshold has been increased to a larger value.

emailweixu (approved):

Haichao-Zhang:
For reference, here is a list of all speed limit signs and their locations in Town01. The unit of speed limit is km/h:
image

PR #844, opened 2021-04-16
Fix a int/int bug for scheduler

emailweixu:
Some of the progress value is int64 torch tensor, which can cause problem for division for LinearScheduler.

Also:
1. implemented `__repr__` for the schedulers
2. support scheduling with respect to global_step

### DIFF
--- a/alf/utils/schedulers.py
+++ b/alf/utils/schedulers.py
@@ -12,8 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """Schedulers."""
-from alf.trainers.policy_trainer import Trainer
 import gin
+
+import alf
+from alf.trainers.policy_trainer import Trainer
 
 
 class Scheduler(object):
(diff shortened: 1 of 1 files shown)


Haichao-Zhang (approved):

alf/utils/schedulers.py
```diff
@@ -29,6 +31,7 @@ class Scheduler(object):
     * "percent": percent of training completed.
     * "iterations": the number training iterations.
     * "env_steps": the number of environment steps
+    * "global_counter": the value from ``alf.summary.get_global_counter()``
```
hnyu, line 34: Is the potential difference between "iterations" and "global_counter" only due to the flag "update_counter_every_mini_batch"?

hnyu (approved):

alf/utils/schedulers.py
emailweixu: For supervised learning,  iteration means epoch. So global_counter is increased many time in train_iter.

PR #849, opened 2021-04-17
simplify video recorder

hnyu:
Simplify video_recorder by removing encoding future info related stuff. @Haichao-Zhang For the current rendering interface, the same thing can be done much more easily in a customized way. See the example code below:

```python
def predict_step(self, time_step: TimeStep, state: SacState, epsilon_greedy=1.0):
        action_dist, action, _, action_state = self._predict_action(
            time_step.observation,
            state=state.action,
            epsilon_greedy=epsilon_greedy,
            eps_greedy_sampling=True)

        new_actions = torch.cat((state.actions[:, 1:, ...],
                                 action.unsqueeze(1)), dim=1) # [B, 100, ...]

        info = dict(info=SacInfo(action_distribution=action_dist))
        import alf.summary.render as render
        if render.is_rendering_enabled():
                action_val = torch.transpose(new_actions.squeeze(0), 0, 1)
                img = render.render_curve(
                    name="actions",
                    data=action_val,
                    y_range=(np.min(self._action_spec.minimum),
                             np.max(self._action_spec.maximum)),
                    img_width=1024,
                    img_height=512)
                info['actions'] = img

        return AlgStep(
            output=action,
            state=SacState(action=action_state,
                           actions=new_actions),
            info=info)
```

Also always put the env frame on the top/left and the pred info on the bottom/right.


alf/summary/render.py
```diff
@@ -174,6 +174,41 @@ def from_image_nest(cls, imgs):
...
+                vertically.
+
+        Returns:
+            Image: the stacked big image
+        """
+        assert isinstance(imgs, (list, tuple))
+        if horizontal:
+            H = max([i.shape[0] for i in imgs])
+            W = sum([i.shape[1] for i in imgs])
+            stacked_img = np.ones((H, W, 3), dtype=np.uint8) * 255
```
Haichao-Zhang: can use np.fill

alf/summary/render.py
```diff
@@ -174,6 +174,41 @@ def from_image_nest(cls, imgs):
...
+            offset_w = 0
+            for i in imgs:
+                stacked_img[:i.shape[0], offset_w:offset_w +
+                            i.shape[1], :] = i.data
+                offset_w += i.shape[1]
+        else:
+            H = sum([i.shape[0] for i in imgs])
+            W = max([i.shape[1] for i in imgs])
+            stacked_img = np.ones((H, W, 3), dtype=np.uint8) * 255
```
Haichao-Zhang: np.fill

Haichao-Zhang (approved):

PR #853, opened 2021-04-22
fix the bug of reward weight setting if checkpoint loaded

hnyu:
Turns out that for playing from a checkpoint, the reward weight is not correctly set, because checkpoint loading happens after class init.

### DIFF
--- a/alf/algorithms/agent.py
+++ b/alf/algorithms/agent.py
@@ -175,9 +175,6 @@ class Agent(OnPolicyAlgorithm):
             reward_weight_algorithm = reward_weight_algorithm_cls(
                 reward_spec=reward_spec, debug_summaries=debug_summaries)
             agent_helper.register_algorithm(reward_weight_algorithm, "rw")
-            # Initialize the reward weights of the rl algorithm
-            rl_algorithm.set_reward_weights(
-                reward_weight_algorithm.reward_weights)
 
         super().__init__(
             observation_spec=observation_spec,
@@ -228,6 +225,11 @@ class Agent(OnPolicyAlgorithm):
             new_state = new_state._replace(goal_generator=goal_step.state)
             info = info._replace(goal_generator=goal_step.info)
             observation = [observation, goal_step.output]
+
+        if self._reward_weight_algorithm is not None:
+            # Initialize the reward weights of the rl algorithm
+            self._rl_algorithm.set_reward_weights(
+                self._reward_weight_algorithm.reward_weights)
 
         rl_step = self._rl_algorithm.predict_step(
             time_step._replace(observation=observation), state.rl,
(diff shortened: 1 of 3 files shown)


alf/algorithms/lagrangian_reward_weight_algorithm.py
```diff
@@ -103,7 +102,7 @@ def _inv_softplus(self, tensor):
     def reward_weights(self):
         """Return the detached reward weights. These weights are expected not to
         be changed by external code."""
-        return self._reward_weights.detach().clone()
+        return F.softplus(self._lambdas).detach().clone()
```
hnyu, line 105: This function can be called after class init so if self._lambdas is updated by checkpoint, we should always return the newest value. For simplicity, we don't want a variable to cache its latest value.

alf/algorithms/agent.py
```diff
@@ -393,6 +395,12 @@ def after_train_iter(self, experience, train_info: AgentInfo = None):
         algorithms = list(filter(lambda a: a is not None, algorithms))
         self._agent_helper.after_train_iter(algorithms, experience, train_info)
 
+        # Strictly speaking, if we load a checkpoint for training, then in the
+        # first iteration ``self._rl_algorithm`` might have the wrong reward
+        # weights because the function call below is called only after the first
+        # iteration. But for simplicity (there is no easy way to set the weights
+        # after loading a checkpoint while before the first iteration) here we
+        # ignore its effect.
```
emailweixu: All the changes to parameters will be overridden by checkpoint.  See Trainer._restore_checkpoint(). But there will be some experiences written to replay buffer (if replay buffer is not checkpointed).

hnyu: > All the changes to parameters will be overridden by checkpoint. See Trainer._restore_checkpoint(). But there will be some experiences written to replay buffer (if replay buffer is not checkpointed).

After loading a checkpoint, the first thing is that self._reward_weight_algorithms.reward_weights is updated because it's from a nn.Parameter. Then we need to call set_reward_weights on self._rl_algorithm (otherwise it's all ones by default and it is not nn.Parameter in RLAlgorithm).  Even though there are some experiences written to replay buffer in the first iteration (generated by wrong reward weights), I think it's generally OK for an off-policy algorithm.

alf/algorithms/agent.py
```diff
@@ -175,9 +175,6 @@ def __init__(self,
             reward_weight_algorithm = reward_weight_algorithm_cls(
                 reward_spec=reward_spec, debug_summaries=debug_summaries)
             agent_helper.register_algorithm(reward_weight_algorithm, "rw")
-            # Initialize the reward weights of the rl algorithm
-            rl_algorithm.set_reward_weights(
-                reward_weight_algorithm.reward_weights)
```
emailweixu: why do you remove these lines?

hnyu: > why do you remove these lines?

reverted

emailweixu (approved):

PR #857, opened 2021-05-04
add soft boundary transforms and functions

hnyu:
Added Softplus, Softlower, Softupper, and Softclip transforms. Also the corresponding functions with hinge_softness.

### DIFF
--- a/alf/utils/dist_utils.py
+++ b/alf/utils/dist_utils.py
@@ -61,6 +61,7 @@ SoftmaxTransform = get_invertable(td.SoftmaxTransform)
 @alf.configurable
 class Softplus(td.Transform):
     r"""Transform via the mapping :math:`\text{Softplus}(x) = \log(1 + \exp(x))`.
+
     Code adapted from `pyro <https://docs.pyro.ai/en/latest/_modules/pyro/distributions/transforms/softplus.html>`_
     and `tensorflow <https://github.com/tensorflow/probability/blob/v0.12.2/tensorflow_probability/python/bijectors/softplus.py#L61-L189>`_.
     """
@@ -81,11 +82,11 @@ class Softplus(td.Transform):
         assert self._hinge_softness > 0, "Must be a positive softness number!"
 
     def __eq__(self, other):
-        return isinstance(other, Softplus)
+        return (isinstance(other, Softplus)
+                and self._hinge_softness == other._hinge_softness)
 
     def _call(self, x):
-        return (nn.functional.softplus(x / self._hinge_softness) *
-                self._hinge_softness)
+        return nn.functional.softplus(x, beta=1. / self._hinge_softness)
 
     def _inverse(self, y):
         return (y / self._hinge_softness).expm1().log() * self._hinge_softness
@@ -131,7 +132,7 @@ def Softupper(high, hinge_softness=1.):
 
 
 @alf.configurable
-def Softclip(low, high, hinge_softness=1.):
+def SoftclipTF(low, high, hinge_softness=1.):
     """Create a Softclip transform by composing Softlower, Softupper, and Affine
     transforms, adapted from `tensorflow <https://www.tensorflow.org/probability/api_docs/python/tfp/bijectors/SoftClip>`_.
     Mathematically,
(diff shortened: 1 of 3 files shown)


alf/utils/dist_utils.py
```diff
@@ -58,6 +58,117 @@ class NewCls(cls):
...
+            hinge_softness (float): this positive parameter changes the transition
+                slope. A higher softness results in a smoother transition from
+                0 to identity.
+        """
+        super().__init__(cache_size=1)
+        self._hinge_softness = float(hinge_softness)
+        assert self._hinge_softness > 0, "Must be a positive softness number!"
+
+    def __eq__(self, other):
+        return isinstance(other, Softplus)
```
emailweixu: `and self.hinge_softness == other.highe_softness`

alf/utils/dist_utils.py
```diff
@@ -58,6 +58,117 @@ class NewCls(cls):
 SoftmaxTransform = get_invertable(td.SoftmaxTransform)
 
 
+@alf.configurable
+class Softplus(td.Transform):
+    r"""Transform via the mapping :math:`\text{Softplus}(x) = \log(1 + \exp(x))`.
+    Code adapted from `pyro <https://docs.pyro.ai/en/latest/_modules/pyro/distributions/transforms/softplus.html>`_
```
emailweixu, line 65: Blank line before this.

alf/utils/math_ops.py
```diff
@@ -82,6 +83,86 @@ def swish(x):
     return x * torch.sigmoid(x)
 
 
+@alf.configurable
+def softlower(x, low, hinge_softness=None):
```
emailweixu: why not let hinge_sortness default to 1.0 here instead of line 104? And same for other related functions.

alf/utils/dist_utils.py
```diff
@@ -58,6 +58,117 @@ class NewCls(cls):
...
+@alf.configurable
+def Softclip(low, high, hinge_softness=1.):
+    """Create a Softclip transform by composing Softlower, Softupper, and Affine
+    transforms, adapted from `tensorflow <https://www.tensorflow.org/probability/api_docs/python/tfp/bijectors/SoftClip>`_.
+    Mathematically,
+
+    .. code-block:: python
+
+        clipped = softupper(softlower(x, low), high)
+        softclip(x) = (clipped - high) / (high - softupper(low, high)) * (high - low) + high
```
emailweixu, line 143: It seems that upper and lower are not treated symmetrically, which feels not very nice.
The following is a symmetric one which has analytical inverse and bounded within (low, high)
```
softlow(x, low) + softhigh(x, high) - x
```

alf/utils/dist_utils.py
```diff
@@ -58,6 +58,117 @@ class NewCls(cls):
...
+
+    The second scaling step is beause we will have
```
Haichao-Zhang, line 145: beause -> because

alf/utils/dist_utils.py
hnyu: This one is indeed better. But it seems that I can't use ComposedTransform to configure this transform from existing ones (there is no built-in transform that combines two transforms). Do you think we should write a custom Softclip transform from scratch (like Softplus)?

hnyu: > It seems that upper and lower are not treated symmetrically, which feels not very nice.
> The following is a symmetric one which has analytical inverse and bounded within (low, high)
> 
> ```
> softlow(x, low) + softhigh(x, high) - x
> ```

This one is indeed better. But it seems that I can't use ComposedTransform to configure this transform from existing ones (there is no built-in transform that combines two transforms). Do you think we should write a custom Softclip transform from scratch (like Softplus)?

emailweixu: It's not that difficult to implement it directly. The actual formula is quite simple if you plugin softlow and softhigh.

hnyu: > It's not that difficult to implement it directly. The actual formula is quite simple if you plugin softlow and softhigh.

I think the main time-consuming part is to derive the inverse function and log_abs_jacobian and verify their correctness.

emailweixu: Inverse is not that difficult for this one, not much difficult to calculate compared to softhigh. Jacobian is just the absolute value of first order derivative.

hnyu: It seems that this solution has some issues when x is very large or small. In either case the clipped values will be slightly beyond the (low, high) range. While the TFP solution won't have this problem. 

```python
import torch
import alf

def softclip(x, low, high, hinge_softness=1.):
    """Softly bound ``x`` in between ``[low, high]``, namely,
    ``softlower(x, low) + softupper(x, high) - x``. Unlike ``softclip_tf``,
    this transform is symmetric regarding the lower and upper bound when
    squashing.

    Args:
        x (Tensor): input
        low (float|Tensor): the lower bound
        high (float|Tensor): the upper bound
        hinge_softness (float): this positive parameter changes the transition
            slope. A higher softness results in a smoother transition from
            ``low`` to ``high``. Default to 1.
    """
    return (alf.math.softlower(x, low, hinge_softness)
            + alf.math.softupper(x, high, hinge_softness) - x)

if __name__ == "__main__":
    torch.set_printoptions(precision=8)
    x = torch.arange(-5, 5, 0.01)
    b = 1
    hinge_softness = 1e-1
    clipped_x = softclip(x, -b, b, hinge_softness)
    print(clipped_x)
    print(clipped_x <= b)
```

Interestingly, if I look at the clipped values near the two ends, they seem not to increase monotonically.

emailweixu: Interesting. Mathematically, it is guaranteed to be within (l, h) and increasing. Not sure if there is a way to write it so that it has better numerical precision.

emailweixu: Sorry, indeed it is not monotonically increasing close to the end

emailweixu: Sorry again, it is actually monotonically increasing

emailweixu: Colab here: https://colab.research.google.com/drive/1677E5AePlK5fx2uSWS3p84BEjJ5P3Qel?usp=sharing

hnyu: added a new softclip function

alf/utils/math_ops.py
```diff
@@ -82,6 +83,114 @@ def swish(x):
...
+        \end{array}
+
+    Args:
+        x (Tensor): input
+        low (float|Tensor): the lower bound
+        high (float|Tensor): the upper bound
+        hinge_softness (float): this positive parameter changes the transition
+            slope. A higher softness results in a smoother transition from
+            ``low`` to ``high``. Default to 1.
+        threshold (float): For numerical stability, the function becomes (2) when
```
emailweixu: why is this parameter needed? It seems to me that threshold=0. is always a better choice because (1) and (2) are guaranteed to be exactly same at low and (1) and (3) are guaranteed to be exactly same at high even considering the finite precision calculation.

alf/utils/dist_utils.py
```diff
@@ -58,6 +58,176 @@ class NewCls(cls):
...
+    r"""Transform via the mapping defined in ``alf.math_ops.softclip()``.
+    Unlike ``SoftclipTF``, this transform is symmetric regarding the lower and
+    upper bound when squashing.
+    """
+    domain = constraints.real
+    codomain = constraints.real
+    bijective = True
+    sign = +1
+
+    def __init__(self, low, high, hinge_softness=1., threhsold=10.):
```
emailweixu: threhsold => threshold

alf/utils/math_ops.py
```diff
@@ -82,6 +83,114 @@ def swish(x):
...
+            ``x < low - hinge_softness * threshold``
+            and becomes (3) ``x > high + hinge_softness * threshold``. Default
+            value: 10.
+    """
+    l, h, s = low, high, hinge_softness
+    u = ((l - x) / s).exp()
+    v = ((x - h) / s).exp()
+    return torch.where(
+        x < l - s * threshold, l + s * ((1 + 1 / u) / (1 + v)).log(),
```
emailweixu: Using log1p() is better for numerical precision:
```
    u1 = u.log1p()
    v1 = v.log1p()
    return torch.where(
        x < l, l + s * ((1 / u).log1p() - v1),
        torch.where(
            x > h, h + s * (u1 - (1 / v).log1p()),
            x + s * (u1 - v1)))
```

alf/utils/math_ops.py
hnyu: Removed

emailweixu (approved):

PR #858, opened 2021-05-05
Pr algorithm container

emailweixu:
Sorry for a large PR as the change of interface affecting many algorithms.

Summary:
1. Adding a new class `AlgorithmInterface` which only define the interfaces without actual implementation. The reason for doing this is that Algorithm is too long and it's not easy to know that the actual interface is.
2. Change the signature of `train_step` from `train_step(inputs, state)` to `train_step(inputs, state, rollout_info)` so that train_step behave same for non-RL and RL algorithms. Previously, for RL algorithm, the interface assume that the type of inputs is Experience and rollout_info is part of inputs. While for non-RL does not have the behavior. 
3. change the signature of `calc_loss` from `calc_loss(experience, train_info)` to `calc_loss(info)`.  The reason is that experience is not suitable for non-RL algorithm.
4. change `preprocess_experience` from `preprocess_experience(experience)` to `preprocess_experience(root_inputs, rollout_info, batch_info` to clearly separate different type of data.
5. Change after_update from `after_update(experience, train_info)` to `after_update(root_inputs, info)` and after_train_iter from `after_train_iter(experience, info)` to `after_train_iter(root_inputs, rollout_info)`. The different name root_inputs is used to emphasize it is the input to the root algorithm.
6. add SequantialAlg to allow flexible composition of algorithms.


alf/algorithms/actor_critic_algorithm.py
```diff
@@ -103,37 +115,44 @@ def __init__(self,
         if loss is None:
             loss = loss_class(debug_summaries=debug_summaries)
         self._loss = loss
+        self._epsilon_greedy = alf.get_config_value(
```
hnyu: This line is duplicated with line 93? Also why not directly do "config.epsilon_greedy"?

alf/algorithms/actor_critic_loss.py
```diff
@@ -93,13 +93,19 @@ def __init__(self,
...
-                time-major.
+            info (namedtuple): information for calculating loss. All tensors are
+                time-major. It should contain the following fields:
+                - reward:
+                - step_type:
+                - discount:
+                - action:
+                - action_distribution:
+                - value:
             train_info (nest): information collected for training. It is batched
```
hnyu: remove the arg "train_info"

alf/algorithms/agent.py
```diff
@@ -96,7 +92,8 @@ def __init__(self,
                 the observation for downstream algorithms such as ``rl_algorithm``.
             intrinsic_reward_module (Algorithm): an algorithm whose outputs
                 is a scalar intrinsic reward.
-            goal_generator (Algorithm): an algorithm with output a goal vector
+            goal_generator (Algorithm): an algorithm with output a tuple of goal
```
hnyu: "with output" => "which outputs"

alf/algorithms/algorithm_interface.py
```diff
@@ -0,0 +1,295 @@
...
+
+            nest.get_field(experience.rollout_info, self.path)
+
+        Returns:
+            str: path from the root algorithm to this algorithm
+        """
+        raise NotImplementedError()
+
+    def set_path(self, path):
+        """Set the path from the root algorithm to this algorithm.
```
hnyu, line 101: Can be more specific about how the path will be used. It seems to me that this function adds another layer of complexity when writing a new custom algorithm that contains a sub-algorithm.

alf/algorithms/agent.py
```diff
@@ -250,75 +251,87 @@ def rollout_step(self, time_step: TimeStep, state: AgentState):
             info = info._replace(repr=repr_step.info)
             observation = repr_step.output
 
+        training_reward = time_step.reward
```
hnyu: Better not to call it "training_reward". For off-policy algorithm, sometimes we want to show rollout reward. So this name is confusing with the actual training reward.

alf/algorithms/agent.py
```diff
@@ -329,47 +342,24 @@ def train_step(self, exp: Experience, state):
...
-    def calc_loss(self, experience, train_info: AgentInfo):
+    def calc_loss(self, train_info: AgentInfo):
         """Calculate loss."""
-        if experience.rollout_info is ():
-            experience = experience._replace(
-                reward=self.calc_training_reward(experience.reward,
-                                                 train_info))
+
+        if train_info.rewards is not ():
+            for name, reward in train_info.rewards.items():
```
hnyu: When using reward normalizer, "rewards" contains normalized external reward during rollout. It's not the same with the normalized ones during training for off-policy algorithms.

alf/algorithms/agent.py
```diff
@@ -329,47 +342,24 @@ def train_step(self, exp: Experience, state):
...
+                self.summarize_reward("calc_training_reward/%s" % name, reward)
```
hnyu: calc_training_reward -> calc_loss

alf/algorithms/agent.py
```diff
@@ -250,75 +251,87 @@ def rollout_step(self, time_step: TimeStep, state: AgentState):
...
+            info = info._replace(rewards=rewards)
+
         return AlgStep(output=rl_step.output, state=new_state, info=info)
 
-    def train_step(self, exp: Experience, state):
+    def train_step(self, time_step: TimeStep, state, rollout_info):
         new_state = AgentState()
-        info = AgentInfo()
-        observation = exp.observation
+        info = AgentInfo(rewards=rollout_info.rewards)
```
hnyu, line 303: It's not good to assign normalized rollout reward here.

alf/algorithms/agent.py
```diff
@@ -391,36 +381,30 @@ def after_train_iter(self, experience, train_info: AgentInfo = None):
...
         """Add intrinsic rewards to extrinsic rewards if there is an intrinsic
         reward module. Also call ``preprocess_experience()`` of the rl
         algorithm.
         """
-        reward = self.calc_training_reward(exp.reward, exp.rollout_info)
-        exp = exp._replace(reward=reward)
+        exp = root_inputs
+        rewards = rollout_info.rewards
+        if rewards is not ():
+            exp = exp._replace(reward=rewards['overall'])
```
hnyu, line 408: This is not correct when using normalized reward, and if we want to generate intrinsic rewards using the up-to-date goal_generator or irm instead of the version during rollout for off-policy training.

alf/algorithms/agent.py
```diff
@@ -329,47 +342,24 @@ def train_step(self, exp: Experience, state):
                 "`action_distribution`, which is required by "
                 "`enforce_entropy_target`")
             et_step = self._entropy_target_algorithm.train_step(
-                rl_step.info.action_distribution, step_type=exp.step_type)
+                (rl_step.info.action_distribution, time_step.step_type))
             info = info._replace(entropy_target=et_step.info)
 
         return AlgStep(output=rl_step.output, state=new_state, info=info)
 
-    def calc_training_reward(self, external_reward, info: AgentInfo):
```
hnyu, line 337: We should be very careful with this calc_training_reward function, especially in the off-policy training case.

hnyu (changes requested):
Only finished reviews up to the Agent class. I assume that this PR will have many rounds of comments and revisions, so maybe doing it in an incremental style to avoid being overwhelmed by comments&revisions.

alf/algorithms/encoding_algorithm.py
```diff
@@ -41,42 +49,123 @@ def __init__(self,
             action_spec (nested BoundedTensorSpec): not used
             encoder_cls (type): The class or function to create the encoder. It
                 can be called using ``encoder_cls(input_tensor_spec)``.
+            time_step_as_input (bool): If True, use the whole TimeStep strucuture
+                as the input to the encoder instead of the observation.
+            output_fields (None | nested str): if None, all the output from the
+                encoder will be used as the output. Otherwise, ``output_fields``
+                will be used to retrieve the fields from the encoder output.
+            loss_fields (None | nested str): there is not lsss for this is None.
```
Haichao-Zhang: ```there is not lsss for this is None``` -> ```there is no loss for this if None```?

alf/algorithms/muzero_algorithm.py
```diff
@@ -303,8 +314,8 @@ def preprocess_experience(self, experience: Experience):
             if not self._train_game_over_function:
                 game_overs = ()
 
-            action = replay_buffer.get_field('action', env_ids,
-                                             positions[:, :-1])
+            action = replay_buffer.get_field('prev_action', env_ids,
```
Haichao-Zhang, line 317: What is the reason for this change?

alf/algorithms/predictive_representation_learner.py
```diff
@@ -515,24 +511,12 @@ def preprocess_experience(self, experience: Experience):
                                              positions)
 
             # [B, T, unroll_steps+1]
```
Haichao-Zhang: Need to update the comments related to ```unroll_steps+1```

alf/algorithms/sac_algorithm.py
```diff
@@ -681,38 +696,42 @@ def train_step(self, exp: Experience, state: SacState):
...
-            exp, state.critic, action, log_pi, action_distribution)
+            inputs, state.critic, rollout_info, action, action_distribution)
         alpha_loss = self._alpha_train_step(log_pi)
 
         state = SacState(
             action=action_state, actor=actor_state, critic=critic_state)
         info = SacInfo(
+            reward=inputs.reward,
+            step_type=inputs.step_type,
+            discount=inputs.discount,
```
Haichao-Zhang, line 715: Since the adding of these fields are the same for many instance algorithms, it would be good if they can be done automatically or with a helper function.

alf/algorithms/algorithm_interface.py
```diff
@@ -0,0 +1,295 @@
...
+       parameters after one rollout (multiple rollout steps) is performed; or it
+       can put these collected experiences into a replay buffer.
+    3. ``train_step()``: only used by algorithms that put experiences into
+       replay buffers. The training data are sampled from the replay buffer
+       filled by ``rollout_step()``.
+    4. ``train_from_unroll()``: perform a training iteration from the unrolled
+       result.
+    5. ``train_from_replay_buffer()``: perform a training iteration from a
+       replay buffer.
+    6. ``update_with_gradient()``: Do one gradient update based on the loss. It
```
Haichao-Zhang: Do -> do, to be consistent with the rest

alf/algorithms/containers.py
```diff
@@ -0,0 +1,502 @@
...
+        """Set the path for each sub-algorithm."""
+        super().set_path(path)
+        prefix = path
+        if path:
+            prefix = prefix + '.'
+        for name, alg in self._algs.items():
+            alg.set_path(path + name)
+
+    def calc_loss(self, info):
+        """Call calc_loss of each sub-algorithm and accumalte the loss."""
```
Haichao-Zhang: accumalte -> accumulate

alf/algorithms/algorithm.py
```diff
@@ -1200,9 +1063,10 @@ def train_from_unroll(self, experience, train_info):
         else:
             valid_masks = None
         experience = experience._replace(rollout_info_field='rollout_info')
-        loss_info = self.calc_loss(experience, train_info)
+        loss_info = self.calc_loss(train_info)
         loss_info, params = self.update_with_gradient(loss_info, valid_masks)
-        self.after_update(experience, train_info)
+        time_step = experience_to_time_step(experience)
+        self.after_update(time_step, train_info)
```
hnyu, line 1069: This conversion from experience back time_step is inconsistent with the docstring of after_update: "root_inputs (nest): experiences collected for the most recent ``update_with_gradient()``." Why bother converting experience back to time_step?

alf/algorithms/actor_critic_algorithm.py
emailweixu: removed.
The reason of using get_config_value is that config is not always provided to the constructor

alf/algorithms/algorithm.py
```diff
@@ -1280,13 +1144,29 @@ def _train_experience(self, experience, batch_info, num_updates,
...
-        experience = self._add_batch_info(experience, batch_info)
         if self._exp_replayer_type != "one_time":
             # The experience put in one_time replayer is already transformed
             # in unroll().
+            experience = self._add_batch_info(experience, batch_info)
             experience = self.transform_experience(experience)
-        experience = self.preprocess_experience(experience)
-        experience = self._clear_batch_info(experience)
+            experience = self._clear_batch_info(experience)
+        time_step = TimeStep(
```
hnyu: experience_to_time_step()

alf/algorithms/actor_critic_loss.py
emailweixu: removed

alf/algorithms/agent.py
emailweixu: fixed.

alf/algorithms/algorithm.py
```diff
@@ -1420,7 +1301,9 @@ def _collect_train_info_sequentially(self, experience):
                     "Policy state is non-empty but the experience doesn't "
                     "contain the 'step_type' field. No way to reinitialize "
                     "the state but will simply keep updating it.")
-            policy_step = self.train_step(exp, policy_state)
+            time_step = experience_to_time_step(exp)
+            policy_step = self.train_step(time_step, policy_state,
```
hnyu, line 1299: converting exp back to time_step in order to be compatible with the signature of the interface seems awkward in many different places

alf/algorithms/agent.py
emailweixu: changed to overall_reward

alf/algorithms/agent.py
emailweixu: removed external reward

alf/algorithms/agent.py
emailweixu: Now extrinsic reward is always from TimeStep

alf/algorithms/agent.py
emailweixu: changed to reward/

alf/algorithms/agent.py
emailweixu: Fixed.

alf/algorithms/algorithm_interface.py
emailweixu: fixed

alf/algorithms/algorithm_interface.py
emailweixu: Added more explanation. It seems hard to avoid this if we want to allow the algorithm to access its own rollout_info in replay buffer.

alf/algorithms/algorithm_interface.py
```diff
@@ -0,0 +1,295 @@
...
+        can have its own logic of a training iteration (e.g.,
+        ``train_from_unroll()`` and ``train_from_replay_buffer()``) which can be
+        triggered by a parent ``RLAlgorithm`` inside its ``after_train_iter()``.
+    """
+
+    @property
+    def path(self):
+        """Path from the root algorithm to this algorithm.
+
+        The input state to rollout_step() can be retrieved by
```
hnyu: "For example, ..."

alf/algorithms/containers.py
emailweixu: fixed

alf/algorithms/encoding_algorithm.py
emailweixu: fixed.

alf/algorithms/muzero_algorithm.py
emailweixu: This is because root_inputs now is changed to TimeStep, which does not have action field.

alf/algorithms/algorithm_interface.py
```diff
@@ -0,0 +1,295 @@
...
+
+        For on-policy training, ``train_step()`` will not be called. And ``info``
+        passed to ``calc_loss()`` is info collected from ``rollout_step()``.
+
+        For off-policy training, ``train_step()`` will be called with the experience
+        from replay buffer. And ``info`` passed to ``calc_loss()`` is info collected
+        from ``train_step``.
+
+        An algorithm can override this to indicate whether it is an on-policy or
+        off-policy algorithm. If an algorithm does not override this, it support
```
hnyu: "If If an algorithm does not override this, it support both ..." From this sentence it is not clear how this algorithm supports both: which functions to select between self._train_iter_on_policy and self._train_iter_off_policy?

alf/algorithms/predictive_representation_learner.py
emailweixu: updated

alf/algorithms/algorithm_interface.py
```diff
@@ -0,0 +1,295 @@
...
+            - output (nested Tensor): prediction result.
+            - state (nested Tensor): should match ``predict_state_spec``.
+        """
+        raise NotImplementedError()
+
+    def rollout_step(self, inputs, state):
+        """Rollout for one step of inputs.
+
+        It is called to calculate output for every environment step. For on-policy
+        training, it also needs to generate necessary information for calc_loss().
```
hnyu: \``calc_loss()\``   \``train_step()\``

alf/algorithms/sac_algorithm.py
emailweixu: I thought about that before. But decided to do this way to make the code clearer.

alf/algorithms/algorithm_interface.py
```diff
@@ -0,0 +1,295 @@
...
+              batched and passed as ``info`` for calc_loss(). If this is
+              ``LossInfo``, ``calc_loss()`` in ``Algorithm`` can be used.
+              Otherwise, the user needs to override ``calc_loss()`` to
+              calculate loss or override ``update_with_gradient()`` to do
+              customized training.
+        """
+        raise NotImplementedError()
+
+    def calc_loss(self, info):
+        """Calculate the loss at each step for each sample.
```
hnyu: each sample -> each mini-batch

alf/algorithms/algorithm_interface.py
```diff
@@ -0,0 +1,295 @@
...
+              and retrieved for ``train_step()`` as ``rollout_info``.
+        """
+        raise NotImplementedError()
+
+    def train_step(self, inputs, state, rollout_info):
+        """Perform one step of training computation.
+
+        It is called to calculate output for every time step for a batch of
+        experience from replay buffer. It also needs to generate necessary
+        information for calc_loss().
```
hnyu: \``calc_loss()\``

alf/algorithms/algorithm_interface.py
```diff
@@ -0,0 +1,295 @@
...
+                this algorithm during ``unroll()``.
+        """
+
+    def train_iter(self):
+        """Perform one iteration of training.
+
+        Users may choose to implement their own ``train_iter()``.
+
+        Returns:
+            int:
```
hnyu, line 287: int: number of samples ...

alf/algorithms/algorithm_interface.py
```diff
@@ -0,0 +1,295 @@
...
+            - number of samples being trained on (including duplicates).
+        """
+
+    def train_from_unroll(self, experience, train_info):
+        """Train given the info collected from ``unroll()``. This function can
+        be called by any child algorithm that doesn't have the unroll logic but
+        has a different training logic with its parent (e.g., off-policy).
```
hnyu: with its off-policy parent.

alf/algorithms/agent.py
```diff
@@ -391,36 +388,33 @@ def after_train_iter(self, experience, train_info: AgentInfo = None):
...
         reward module. Also call ``preprocess_experience()`` of the rl
         algorithm.
         """
-        reward = self.calc_training_reward(exp.reward, exp.rollout_info)
-        exp = exp._replace(reward=reward)
+        exp = root_inputs
+        rewards = rollout_info.rewards
+        if rewards is not ():
+            rewards = copy.copy(rewards)
+            rewards['overall'] = self._calc_overall_reward(
```
hnyu, line 406: What's the purpose of storing an overall reward in rewards? It will never be used later.

alf/algorithms/containers.py
```diff
@@ -0,0 +1,502 @@
...
+        assert isinstance(algs, dict)
+        if is_on_policy is not None:
+            for aname, alg in algs.items():
+                if alg.on_policy is not None:
+                    assert alg.on_policy == is_on_policy, (
+                        "is_on_policy=%s "
+                        "is different from algs[%s].on_policy=%s" %
+                        (is_on_policy, aname, alg.on_policy))
+        else:
+            is_on_policy = None
```
hnyu: This line is unnecessary.

alf/algorithms/containers.py
```diff
@@ -0,0 +1,502 @@
...
+            ]
+            if on_policy_algs and off_policy_algs:
+                raise ValueError(
+                    "%s is on-policy, but %s is off-policy." %
+                    (on_policy_algs[0].name, off_policy_algs[0].name))
+            if on_policy_algs or off_policy_algs:
+                is_on_policy = bool(on_policy_algs)
+        if is_on_policy is not None:
+            for alg in algs.values():
+                alg.set_on_policy(is_on_policy)
```
hnyu, line 79: Line 79-80 can directly be `self.set_on_policy(is_on_policy)`. What if is_on_policy is still None at this point?

alf/algorithms/containers.py
```diff
@@ -0,0 +1,502 @@
...
+                  name="SequentialAlg",
+                  debug_summaries=False,
+                  **named_modules):
+    """
+
+
+    The following examples shows an example of using SequentialAlg to build
+    actor-critic algorithm.
+
+    Example:
```
hnyu: It's still not straightforward for me to figure out from the example below how to use SequentialAlg, because almost all the underlying work is done by a private class _SequentialAlg.

alf/algorithms/containers.py
```diff
@@ -0,0 +1,502 @@
...
+            var_dict[output] = x
+        if self._output:
+            x = get_nested_field(var_dict, self._output)
+        return AlgStep(output=x, state=new_state, info=info)
+
+
+class EchoAlg(Algorithm):
+    """Echo Algorithm.
+
+    EchoAlg uses part of the output of ``alg`` of current step as part of
```
hnyu: "Echo algorithm uses part of ..."

alf/algorithms/containers.py
```diff
@@ -0,0 +1,502 @@
...
+    will be the 'echo' input of the next step. 'input' of ``alg``'s input is from
+    the input of ``EchoAlg`` and 'output' of ``alg``'s output is the output of
+    ``EchoAlg``.
+    """
+
+    def __init__(self, alg, echo_spec, debug_summaries=False, name='EchoAlg'):
+        """
+        Args:
+            alg (Algorithm): the module for performing the actual computation
+            echo_spec ():
```
hnyu: docstring incomplete

alf/algorithms/containers.py
```diff
@@ -0,0 +1,502 @@
...
+                 observation_spec,
+                 action_spec,
+                 algorithm,
+                 env=None,
+                 reward_spec=alf.TensorSpec(()),
+                 reward_weights=None,
+                 config: TrainerConfig = None,
+                 optimizer=None,
+                 debug_summaries=False,
+                 name="RLAlgWrapper"):
```
hnyu, line 558: docstring missing for args

alf/algorithms/containers.py
```diff
@@ -0,0 +1,502 @@
...
+    def after_update(self, root_inputs, info):
+        self._alg.after_update(root_inputs, info)
+
+    def after_train_iter(self, root_inputs, rollout_info):
+        self._alg.after_train_iter(root_inputs, rollout_info)
+
+
+@alf.configurable
+class RLAlgWrapper(RLAlgorithm):
+    """Wrap an Algorithm as RLAlgorithm so that it can be used for RLTrainer"""
```
hnyu: "Wrap an \``Algorithm\`` instance as an \``RLAlgorithm\`` instance so that it can be used by \``RLTrainer\``."

alf/algorithms/containers.py
```diff
@@ -0,0 +1,502 @@
...
+
+    def set_on_policy(self, is_on_policy):
+        super().set_on_policy(is_on_policy)
+        self._algorithm.set_on_policy(is_on_policy)
+
+    def set_path(self, path):
+        super().set_path(path)
+        self._alg.set_path(path)
+
+    def rollout_step(self, inputs: TimeStep, state):
```
hnyu, line 605: For the following functions, can we directly do a forward call? For example in OpenAI wrapper:

```python
def __getattr__(self, name):
        if name.startswith('_'):
            raise AttributeError("attempted to get missing private attribute '{}'".format(name))
        return getattr(self.env, name)
```

alf/algorithms/containers.py
```diff
@@ -0,0 +1,502 @@
...
+                                                     batch_info)
+
+    def after_update(self, root_inputs, info):
+        self._algorithm.after_update(root_inputs, info)
+
+    def after_train_iter(self, root_inputs, rollout_info):
+        self._algorithm.after_train_iter(root_inputs, rollout_info)
+
+
+class Loss(Algorithm):
```
hnyu: This class shouldn't be put under "containers.py"

alf/algorithms/dynamic_action_repeat_agent.py
```diff
@@ -290,20 +286,20 @@ def train_step(self, rl_exp: Experience, state: ActionRepeatState):
         repr_state = ()
         if self._repr_learner is not None:
             repr_step = self._repr_learner.train_step(rl_exp, state.repr)
```
hnyu: rl_exp -> inputs

alf/algorithms/algorithm_interface.py
```diff
@@ -0,0 +1,309 @@
...
+                for this algorithm.
+            batch_info (BatchInfo): information about this batch of data
+        Returns:
+            tuple:
+            - processed root_inputs
+            - processed rollout_info
+        """
+        return root_inputs, rollout_info
+
+    def after_update(self, root_inputs, info):
```
hnyu, line 247: Why it is always inputs from the root? I think for subalgorithms, the inputs should be children inputs

alf/algorithms/encoding_algorithm.py
```diff
@@ -31,7 +33,13 @@ class EncodingAlgorithm(Algorithm):
     def __init__(self,
```
hnyu, line 33: It is no longer just a wrapper for ``EncodingNetwork``

alf/algorithms/goal_generator.py
```diff
@@ -134,18 +134,20 @@ def train_step(self, exp: Experience, state):
         experience trajectory.
 
         Args:
-            exp (Experience): the experience data whose ``rollout_info`` has been
+            inputs (TimeStep): the experience data whose ``rollout_info`` has been
```
hnyu: inputs has no ``rollout_info`` anymore

alf/algorithms/predictive_representation_learner.py
```diff
@@ -473,13 +473,9 @@ def preprocess_experience(self, experience: Experience):
         used as the target for the corresponding decoder.
 
         """
-        if self._num_unroll_steps == 0:
```
hnyu, line 476: Why removing this case?

alf/algorithms/rl_algorithm.py
```diff
@@ -97,6 +99,7 @@ def __init__(self,
             predict_state_spec (nested TensorSpec): for the network state of
                 ``predict_step()``. If None, it's assumed to be the same as
                 ``rollout_state_spec``.
+            is_on_policy:
```
hnyu: incomplete docstring

alf/algorithms/tasac_algorithm.py
```diff
@@ -49,8 +47,10 @@
     default_value=())
 
 TasacInfo = namedtuple(
-    "TasacInfo",
-    ["action_distribution", "action", "actor", "critic", "alpha", "repeats"],
+    "TasacInfo", [
+        "reward", "step_type", "discount", "action", "action_distribution",
+        "rollout_beta", "beta", "actor", "critic", "alpha", "repeats"
```
hnyu: rollout_beta -> rollout_b; beta -> b (beta is the policy name and b is the output)

alf/algorithms/tasac_algorithm.py
```diff
@@ -147,44 +147,40 @@ def gamma(self):
...
             # [T, B, D] or [T, B, 1]
-            discounts = experience.discount.unsqueeze(-1) * self._gamma
+            discounts = info.discount.unsqueeze(-1) * self._gamma
         else:
             # [T, B]
-            discounts = experience.discount * self._gamma
+            discounts = info.discount * self._gamma
 
-        rollout_b = experience.rollout_info.action[0]
+        rollout_b = info.beta
```
hnyu: rollout_b = info.rollout_b

hnyu (changes requested):

alf/algorithms/algorithm.py
emailweixu: changed.

alf/algorithms/algorithm.py
emailweixu: Agree. What am thinking is to change Experience to a tuple of time_step, state, rollout_info and action. So there will be no need to convert. I can do that in a different PR to limit the size of this PR.

alf/algorithms/algorithm_interface.py
emailweixu: updated as: If an algorithm does not override this, it needs to support both on-policy and off-policy training, which means that ``rollout_step()`` and ``train_step()`` need to have the correct behavior for on-policy and off-policy training.

alf/algorithms/algorithm_interface.py
emailweixu: fixed.

alf/algorithms/algorithm_interface.py
emailweixu: I thought about using inputs for the sub-algorithm. Though it might seem more reasonable and convenient, I decided to use root_inputs for several reasons: 1. using the inputs to subalgorithm will consume more memory; 2. it will makes writing algorithm containing sub-algorithm more difficult as it needs to keep track of the inputs (may be computed by some other steps) provided to the subalgorithm. 3. if the sub-algorithm really need its inputs for after_update(), it can put them in info.

Note the root_inputs is the actual behavior of previous interface.

alf/algorithms/algorithm_interface.py
emailweixu: Its parent may not be off-policy?

alf/algorithms/containers.py
emailweixu: removed.

alf/algorithms/containers.py
emailweixu: set_on_policy() will call super().set_on_policy(). super() is not well-defined at this moment.
It is fine as long as the on-policyness of the root algorithm can be decided.

alf/algorithms/containers.py
emailweixu: Added more documentation.

alf/algorithms/containers.py
emailweixu: Updated.

alf/algorithms/containers.py
emailweixu: added.

alf/algorithms/containers.py
emailweixu: No, because the base class defined those functions.

alf/algorithms/dynamic_action_repeat_agent.py
emailweixu: fixed.

alf/algorithms/encoding_algorithm.py
emailweixu: updated.

alf/algorithms/goal_generator.py
emailweixu: updated.

alf/algorithms/algorithm_interface.py
emailweixu: updated.

alf/algorithms/algorithm_interface.py
emailweixu: updated.

alf/algorithms/containers.py
emailweixu: added

alf/algorithms/containers.py
emailweixu: updated.

alf/algorithms/containers.py
emailweixu: Couldn't think of good place. Moved to algorithm.py

alf/algorithms/predictive_representation_learner.py
emailweixu: In this case, the updated EncodingAlgorithm is better choice.

alf/algorithms/rl_algorithm.py
emailweixu: added.

alf/algorithms/tasac_algorithm.py
emailweixu: hmm. My bad. But strangely, it didn't affect its performance on bipedal walker

alf/algorithms/tasac_algorithm.py
emailweixu: updated.

alf/algorithms/agent.py
emailweixu: For summary in calc_loss()

alf/algorithms/algorithm.py
emailweixu: updated ``after_update()`` as ``after_update(root_inputs, info)``. The reason for root_inputs instead of Experience is that root_inputs is well-defined for generic Algorithm.

alf/algorithms/algorithm_interface.py
hnyu: I think the meaning of "train_from_unroll" is that when the parent doesn't train from unroll (off-policy) but the child wants to? Otherwise the child can just train together with its parent during unroll.

alf/algorithms/containers.py
hnyu: It seems that if line 75 "if on_policy_args or off_policy_args" is not True, then is_on_policy is still None at line 79? So the child algorithms won't be set correctly.

alf/algorithms/containers.py
```diff
@@ -0,0 +1,622 @@
...
+                  output='',
+                  is_on_policy=None,
+                  name="SequentialAlg",
+                  **named_modules):
+    """Compose Algorithms Networks sequentially as a new Algorithm.
+
+    All the modules provided through ``modules`` and ``named_modules`` are calculated
+    sequentially in the same order as they appear in the call to ``SequentialAlg``.
+    By default, each module takes the output of the previous module as its input
+    (or the input to the Sequential if it is the first module), and the output of
```
hnyu: or the input to ``SequentialAlg``

alf/algorithms/containers.py
```diff
@@ -0,0 +1,622 @@
...
+    The following contructs an algorithm which predicts the future of its input:
+
+    .. code-block:: python
+
+        predictor = EncodingNetwork(...)
+
+        alg = SequentialAlg(
+            predicted=predictor,
+            delayed=networks.Delay(),
+            error=(('delayed', 'input'), lambda xy: (xy[0] - xy[1]) ** 2,
```
hnyu: missing a ')' at the end

alf/algorithms/containers.py
```diff
@@ -0,0 +1,622 @@
...
+    module as the input to a module. To do this,
+    one can pass a tuple of (nested_str, module) instead of module as an argument
+    to ``SequentialAlg``. With this, the inputs to the module will be obtained using
+    ``get_nested_field(named_results, nested_str)``, where ``named_results``
+    is a dictionary containing the inputs to ``SequentialAlg`` and all the results calulcated
+    by previous modules. More specifically, ``named_results['input']`` is the
+    inputs to this algorithm. ``named_results['a']`` is the output of the module
+    named 'a'. ``named_results['info']['a']`` is the info output of the algorithm
+    named 'a'.  And ``named_results['state']['a']`` is state output of the
+    algorithm/network named 'a'.
```
hnyu, line 173: Seems still missing the explanation for loss computation. Is it always use the 'loss=' field when retrieving the loss?

alf/algorithms/containers.py
```diff
@@ -0,0 +1,622 @@
...
+                    action_distribution=action_dist,
+                    action=action,
+                    value=value))
+                )
+                return AlgStep(
+                    output=action,
+                    state=(value_state, actor_state),
+                    info=loss_step.info)
+
+            def calc_loss(self, info):
```
hnyu, line 277: In the two examples, both "calc_loss" is fairly simple: directly call self._loss.calc_loss(info). What if the user wants to customize the calc_loss function and still use this SequentialAlg?

alf/algorithms/containers.py
hnyu: Yeah I mean instead of calling self._algorithm.xxx() one by one, we can just use one function to forward the call? For example:

```python
def __getattr__(self, name):
    if name in ["rollout_step", "train_step", ...]:
          return getattr(self._algorithm, name)
```

alf/algorithms/goal_generator.py
```diff
@@ -114,17 +114,17 @@ def _step(self, time_step: TimeStep, state):
...
 
-    def predict_step(self, time_step: TimeStep, state, epsilon_greedy):
-        return self._step(time_step, state)
+    def predict_step(self, inputs: TimeStep, state):
+        return self._step(inputs, state)
 
-    def train_step(self, exp: Experience, state):
+    def train_step(self, inputs: TimeStep, state, rollout_info):
         """For off-policy training, the current output goal should be taken from
         the goal in ``exp.rollout_info`` (historical goals generated during rollout).
```
hnyu: This docstring is still outdated?

alf/algorithms/containers.py
emailweixu: line 79 executes only when is_on_policy is not None.

alf/algorithms/containers.py
emailweixu: fixed.

alf/algorithms/containers.py
emailweixu: No. Loss is an algorithm which implements calc_loss().

alf/algorithms/containers.py
emailweixu: fixed.

alf/algorithms/containers.py
emailweixu: If the loss is per step loss, using `Loss` class should be fine. If the loss needs to use the sequence, they need to implement an Algorithm for it, such as `ActorCriticLoss`

alf/algorithms/containers.py
emailweixu: I guess I didn't explain it clearly. AlgorithmInterface is the ancestor of RLAlgWrapper. All these functions have been defined in AlgorithmInterface. `__getattr__` won't be called if the attribute has been defined.

alf/algorithms/goal_generator.py
emailweixu: updated.

alf/algorithms/containers.py
hnyu: If is_on_policy is always None, and also child algortihms' is_on_policy is None, there will be a problem

emailweixu: If this algorithm is the subalgorithm of another algorithm, then the parent algorithm will decide whether it's on-policy or not. In the end, at the root RLAlgorithm, train_iter() will check whether on_policy is None.

hnyu: Oh I didn't see the check in train_iter.

hnyu (approved):

PR #862, opened 2021-05-14
zero_grad() -> zero_grad(set_to_none=True)

emailweixu:
This will allow the possiblity of optimizing different sets of model parameters for different update_with_gradient() calls.


Haichao-Zhang (approved):

emailweixu:
I made this change when I was using pytorch 1.8. It turns out that set_to_none option is not available in pytorch 1.4.

PR #866, opened 2021-05-18
Upgrade to pytorch 1.8 and python 3.7 and fix corresponding errors

### LINKED ISSUE #867: Upgrade to PyTorch 1.8 + Python 3.7
1. Upgraded pytorch and python used in ALF to pytorch1.8, python3.7
2. Fixed broken tests and errors due to upgrade
3. Updated docker files and images
4. Improved naming/versioning for better correspondence between ALF and docker images
5. Added pytorch notes on some upgrading issues

Haichao-Zhang:
Brief summaries (will write a detailed doc with examples in the future):

1. Fixed an error when converting np.float64 scalar to torch tensors (1.4 will generate torch tensor with dtype of torch.float32, but 1.8 will have dtype of torch.float64, which will cause issues in some cases)
2. Fixed optimizer test. ```optimizer``` now uses param indices instead of ids (affect some test cases but does not affect training)
3. Fixed norm distribution construction. ```torch.Distribution``` introduced ```validate_args```, this affected some test cases.
4. Fixed a number of IndexError, due to automatic type promotion in pytorch1.8 when computing indices
5. Fixed adamw optimizer. The adamw implementation in pytorch1.8 (both 1.8.0 and 1.8.1) has a bug, which impacts our optimizers where an empty ```params``` is specified by default.
6. Removed trunc_normal and used the one from pytorch.
7. Removed MixtureSameFamily and used the one from pytorch.
8. Fixed gym rendering issue when upgrading to python3.7 from 3.6. This is due to a higher version of Pillow used by python 3.7 by default.


.ci-cd/Dockerfile.cpu
```diff
@@ -82,14 +82,16 @@ RUN pip3 install git+https://github.com/HorizonRobotics/gin-config.git
...
-    && wget \
-      'https://static.roms.download/roms/nintendo/Super Mario Bros (E).zip' \
-    && unzip 'Super Mario Bros (E).zip' \
-    && mv 'Super Mario Bros (E).nes' \
-      /usr/local/lib/python3.6/dist-packages/retro/data/stable/SuperMarioBros-Nes/rom.nes \
-    && rm 'Super Mario Bros (E).zip'
+# The source is inaccessible anymore. We leave it here as a reference and
+# install it manually when building the image if needed.
+# # import public ROM for super mario (ignore shasum)
+# RUN cd /tmp \
```
hnyu, line 93: FYI, our internal Gitlab alf_k8s has a checked-in image for this. 

http://gitlab.hobot.cc/hobotrl/docker_build/alf_k8s/raw/pytorch/alf/package/retro/SuperMarioBros.nes?inline=false

alf/algorithms/entropy_target_algorithm.py
```diff
@@ -240,9 +242,9 @@ def adjust_alpha(self, entropy):
         target_entropy = self._target_entropy()
 
         if target_entropy > 0:
-            fast_stage_thresh = 0.5 * target_entropy
+            fast_stage_thresh = (0.5 * target_entropy).astype(np.float32)
```
hnyu: From my understanding, now we have to explicitly specify float32 whenever a floating number is involved?

alf/utils/dist_utils.py
```diff
@@ -370,7 +370,11 @@ def __init__(self, loc, scale):
             loc (Tensor): mean of the distribution
             scale (Tensor): standard deviation. Should have same shape as ``loc``.
         """
-        super().__init__(td.Normal(loc, scale), reinterpreted_batch_ndims=1)
+        # set validate_args to False here to enable the construction of Normal
+        # distribution with zero scale.
```
hnyu, line 374: Is zero scale needed in any case?

hnyu (changes requested):
We have too many docker images for CI, local usage, and cluster usage. I think probably a documentation page will help.

Also, the update for cluster dockers is missing.

alf/algorithms/entropy_target_algorithm.py
Haichao-Zhang: Whether it requires special attention to dtype mainly depends on the pytorch function under consideration.

This line is not needed here and has been removed. The actual conversion happens in L157-160, where we convert np.float64 to torch tensors. In pytorch1.4, it will return a tensor with type of torch.float32. But in 1.8, it will return torch.float64.  

In the concrete case of entropy target algorithm, when we apply torch.where(cond, a, b) laster, we will get an error since a and b are of different dtypes. 

In summary, in this case, because of the torch.where function, we have to pay attention to this conversion issue. For any other similar situations where such conversion is needed, we will get an error message from torch at least.

alf/utils/dist_utils.py
Haichao-Zhang: We have a test case for this:

https://github.com/HorizonRobotics/alf/blob/3ae0b511e63fb0bf9586712863b63267a43aaff4/alf/networks/projection_networks_test.py#L65-L86

.ci-cd/Dockerfile.cpu
Haichao-Zhang: > We have too many docker images for CI, local usage, and cluster usage. I think probably a documentation page will help.

Will do. But probably not put it here, maybe as an internal doc?
 
> Also, the update for cluster dockers is missing.

I'm aware of this and working on it. Since it is in different repos, it is not reflected here.

alf/algorithms/entropy_target_algorithm.py
```diff
@@ -240,9 +247,9 @@ def adjust_alpha(self, entropy):
         target_entropy = self._target_entropy()
 
         if target_entropy > 0:
-            fast_stage_thresh = 0.5 * target_entropy
+            fast_stage_thresh = (0.5 * target_entropy)
```
emailweixu: Why this change?

.ci-cd/Dockerfile.cpu
hnyu: > 
> Will do. But probably not put it here, maybe as an internal doc?

A Feishu doc would be good.

alf/algorithms/entropy_target_algorithm.py
Haichao-Zhang: The brackets are unnecessary. They are not removed when I revert some unnecessary conversions.  Will remove them.

.ci-cd/Dockerfile.cpu
Haichao-Zhang: Here is a doc:
https://oshcjymfmk.feishu.cn/docs/doccnr4Qg0kb55mIU2ueG7wYs9d

alf/algorithms/entropy_target_algorithm.py
Haichao-Zhang: Removed now.

emailweixu (dismissed):

README.md
```diff
@@ -25,7 +25,14 @@ Agent Learning Framework (ALF) is a reinforcement learning framework emphasizing
 
 ## Installation
 
-Python3.7 is currently supported by ALF. [Virtualenv](https://virtualenv.pypa.io/en/latest/) is recommended for the installation. After activating a virtual env, you can run the following commands to install ALF
+Python3.7 is currently supported by ALF. [Virtualenv](https://virtualenv.pypa.io/en/latest/) is recommended for the installation.
+
+Make sure python3.7-dev is installed before creating an virtual enivronment:
+```
+apt install -y python3.7-dev
```
hnyu: "sudo apt install" What's the difference between libpython3.7-dev and python3.7-dev?

Haichao-Zhang: Added sudo.

hnyu (approved):

PR #870, opened 2021-05-20
do not allow root_dir under alf_root; always use abs_path

hnyu:
Do not allow creating a root_dir under ALF_ROOT to avoid infinitely recursive copying by rsync. Also make sure train.py, play.py and grid_search.py always use absolute path of root_dir.

### DIFF
--- a/alf/utils/common.py
+++ b/alf/utils/common.py
@@ -1082,6 +1082,10 @@ def generate_alf_root_snapshot(alf_root, dest_path):
         dest_path (str): the path to generate a snapshot of ALF repo
     """
 
+    def _is_subdir(path, directory):
+        relative = os.path.relpath(path, directory)
+        return not relative.startswith(os.pardir)
+
     def rsync(src, target, includes):
         args = ['rsync', '-rI', '--include=*/']
         args += ['--include=%s' % i for i in includes]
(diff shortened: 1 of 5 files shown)


alf/utils/common.py
```diff
@@ -1082,6 +1082,10 @@ def generate_alf_root_snapshot(alf_root, dest_path):
         dest_path (str): the path to generate a snapshot of ALF repo
     """
 
+    def _is_subdir(path, directory):
+        relative = os.path.relpath(path, directory)
+        return not relative.startswith(os.pardir)
```
emailweixu, line 1087: This does not always work. For example:

```
>>> os.path.relpath("/abc/def", "/abc")
'def'
```

emailweixu (changes requested):

alf/utils/common.py
hnyu: > This does not always work. For example:
> 
> ```
> >>> os.path.relpath("/abc/def", "/abc")
> 'def'
> ```

'def' is the expected result in this case. This function decides if 'path' is a subdirectory of 'directory'. If not, 'relative' should always start with '..' (os.pardir)

emailweixu: You right. I was confused.

emailweixu (approved):

PR #872, opened 2021-05-20
BetaProjectionNetwork

emailweixu:
Beta distribution is naturally bounded. So it does not need to be wrapped with a squashing transformation, which often causes numerical issue.

### DIFF
--- a/alf/networks/projection_networks_test.py
+++ b/alf/networks/projection_networks_test.py
@@ -21,7 +21,7 @@ import alf
 from alf.networks import CategoricalProjectionNetwork
 from alf.networks import NormalProjectionNetwork
 from alf.networks import StableNormalProjectionNetwork
-from alf.networks import BetaProjectionNetwork, DirichletProjectionNetwork
+from alf.networks import BetaProjectionNetwork
 from alf.tensor_specs import TensorSpec, BoundedTensorSpec
 from alf.utils.dist_utils import DistributionSpec
 import alf.utils.math_ops as math_ops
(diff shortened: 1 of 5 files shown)


alf/utils/dist_utils.py
```diff
@@ -745,6 +797,8 @@ def get_mode(dist):
             mode = base_mode
             for transform in dist.transforms:
                 mode = transform(mode)
+    elif isinstance(dist, td.Beta):
+        mode = dist.mean
```
emailweixu: mode can be calculated analytically, which is (alpha -1)/(alpha+beta - 2) for alpha, beta >1. The mean is alpha/(alpha+beta). Do we need the actual mode?

alf/networks/projection_networks_test.py
```diff
@@ -174,6 +175,27 @@ def test_stable_normal_projection_net_minmax_std(self):
...
+                                        maximum=1.,
+                                        dtype=torch.float32)
+
+        net = BetaProjectionNetwork(
+            input_spec.shape[0], action_spec, projection_output_init_gain=.0)
+
+        dist, _ = net(embedding)
+        self.assertTrue(isinstance(net.output_spec, DistributionSpec))
+        samples = dist.sample()
+        print(samples)
```
Haichao-Zhang: ``print(samples)`` can be removed?

alf/utils/dist_utils.py
hnyu: > mode can be calculated analytically, which is (alpha -1)/(alpha+beta - 2) for alpha, beta >1. The mean is alpha/(alpha+beta). Do we need the actual mode?

Yeah I think we need mode instead of mean in most cases.

Haichao-Zhang: Maybe mode is better? In cases where the distribution is not uni-mode, the mean and mode could be very different?

emailweixu: Changed to mode.

alf/networks/projection_networks_test.py
emailweixu: removed

Haichao-Zhang (dismissed):

hnyu (dismissed):

Haichao-Zhang (approved):

PR #875, opened 2021-05-22
improve plot tb curves

hnyu:
Improve the TB curve plotting functions. 
1. Extend the `MeanCurve` class to include two members functions for computing the normalized AUC and final score.
2. Change the `max_n_scalars` argument of `MeanCurveReader` to `x_steps`. The reason is that several curves might have different x steps (due to TB logging) even though they are identical and independent runs. So misalignment might exist. Now use a common `x_steps` for interpolation.

### DIFF
--- a/alf/utils/plot_tb_curves.py
+++ b/alf/utils/plot_tb_curves.py
@@ -35,9 +35,8 @@ class MeanCurve(
         namedtuple(
             "MeanCurve", ['x', 'y', 'min_y', 'max_y', 'name'],
             default_value=())):
-    def normalized_auc(self):
+    def average_y(self):
         """Return the averaged ``y``, ``min_y``, and ``max_y`` over the x-axis.
-        It can also approximate the normalized area under the curve.
         """
         return self.final_y(N=len(self.y))
 
--- a/alf/utils/plot_tb_curves.py
+++ b/alf/utils/plot_tb_curves.py
@@ -29,9 +29,22 @@ from alf.data_structures import namedtuple
 
 HOME = os.getenv("HOME")
 
+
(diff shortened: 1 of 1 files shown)


alf/utils/plot_tb_curves.py
```diff
@@ -29,9 +29,22 @@
...
-    "MeanCurve", ['x', 'y', 'min_y', 'max_y', 'name'], default_value=())
+class MeanCurve(
+        namedtuple(
+            "MeanCurve", ['x', 'y', 'min_y', 'max_y', 'name'],
+            default_value=())):
+    def normalized_auc(self):
+        """Return the averaged ``y``, ``min_y``, and ``max_y`` over the x-axis.
+        It can also approximate the normalized area under the curve.
+        """
+        return self.final_y(N=len(self.y))
```
emailweixu, line 41: Why does this approximate normalized AUC?

hnyu: > Why does this approximate normalized AUC?

Assuming x ticks are evenly separated by x_delta. The area under curve can be approximated by \sum_t y[t] * x_delta. Then the normalized AUC is this value divided by x_delta * N, which is the average of y values.

emailweixu: So it is not the normalized auc in your paper? I feel average_y seems to be a more accurate name.

hnyu: > So it is not the normalized auc in your paper? I feel average_y seems to be a more accurate name.

Unless there is an official definition of normalized auc, I actually defined this same concept in the paper, meaning auc normalized by x range. If not normalized, the area could be any magnitude depending on how many x ticks there are.

emailweixu (approved):

PR #876, opened 2021-05-25
Function to summarize the intermediate gradient during backward.

emailweixu:


alf/utils/summary_utils.py
```diff
@@ -367,3 +367,69 @@ def __exit__(self, type, value, traceback):
...
+    Args:
+        name (str): name of the summary
+        tensor (nested Tensor): tensor of which the gradient is to be summarized.
+        batch_dims (int): first so many dimensions are treated as batch dimensions
+        clone (bool): If True, ``tensor`` will first be cloned. This is useful
+            if ``tensor`` is used in multiple places and you only want to summarize
+            the gradient from one place. If False, the gradient will be the sum
+            from all gradients backpropped to ``tensor``.
+    Returns:
+        ``tensor`` or cloned ``tensor``: the cloned ``tensor`` should be used for
```
Haichao-Zhang, line 384: What will happen if ``clone`` is True,  but for downstream calculation, the cloned tensor is not used, but the original tensor is used mistakenly? 

I'm trying to think of how to help avoid this type of mistakes. It might be helpful to change the name of the function to something like ``tensor_gradients_summarize_layer``, which implies the output should be used in downstream computation?

emailweixu: In that case, the user will not be able to find the related summary in tensorboard. So the mistake will not be silently ignored.

Haichao-Zhang (approved):

PR #879, opened 2021-05-27
Fix dtype when constructing tensor spec from gym space

Haichao-Zhang:
Fix dtype issue when constructing tensor spec from gym space to avoid error due to a mis-match between the dtype of parameters and input data of a network.

### DIFF
--- a/alf/utils/spec_utils.py
+++ b/alf/utils/spec_utils.py
@@ -92,10 +92,7 @@ def zeros_from_spec(nested_spec, batch_size):
         shape = [batch_size]
 
     def _zero_tensor(spec):
-        res = spec.zeros(shape)
-        if res.dtype == torch.float64:
-            res = res.float()
-        return res
+        return spec.zeros(shape)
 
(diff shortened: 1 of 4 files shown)


alf/utils/spec_utils.py
```diff
@@ -92,7 +92,10 @@ def zeros_from_spec(nested_spec, batch_size):
         shape = [batch_size]
 
     def _zero_tensor(spec):
-        return spec.zeros(shape)
+        res = spec.zeros(shape)
+        if res.dtype == torch.float64:
```
emailweixu: I think if spec.dtype is explicitly set as torch.float64, then we should honor the intention. This means that the right fix is to pass torch.float32 instead of torch.float64 when construct the spec.

Haichao-Zhang: Agree. Now do the dtype conversion at the place when constructing the spec.

emailweixu (approved):

PR #881, opened 2021-05-27
Fix Algorithm._setup_optimizers_()

emailweixu:
Previously, _setup_optimizers_() include all its children into handled.
This is not correct when the algorithm does not have a default optimizer.

### DIFF
--- a/alf/algorithms/algorithm.py
+++ b/alf/algorithms/algorithm.py
@@ -70,7 +70,8 @@ class Algorithm(AlgorithmInterface):
         """Each algorithm can have a default optimimzer. By default, the parameters
         and/or modules under an algorithm are optimized by the default
         optimizer. One can also specify an optimizer for a set of parameters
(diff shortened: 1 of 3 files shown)


alf/algorithms/algorithm.py
```diff
@@ -470,46 +471,53 @@ def _setup_optimizers_(self):
             tuple:
 
             - list of parameters not handled by any optimizers under this algorithm
```
hnyu, line 473: This sentence might be ambiguous. It should be "list of parameters not handled by optimizers of child modules but handled by this algorithm itself"

hnyu (changes requested):

alf/algorithms/algorithm.py
emailweixu: they are not handled by this algorithm either.

hnyu (approved):

PR #882, opened 2021-05-27
revert pytorch version in RTD requirements

hnyu:
There is still a problem of running a higher version of pytorch on the RTD server. I've reverted pytorch versions to 1.4.


Haichao-Zhang:
It might be better to figure out why building fails.
Otherwise, the mismatch of the package version might be an issue in the future?

hnyu:
> It might be better to figure out why building fails.
> Otherwise, the mismatch of the package version might be an issue in the future?

It's hard to figure out. 

image

I think it's more about RTD's own server environment. This issue doesn't appear in a  local doc build.

Regarding the version issue, I think as long as the new version doesn't have new interfaces to be called globally, then an old version in RTD is OK. It's mainly used to generate a parsing of a code file.

Haichao-Zhang (approved):

PR #886, opened 2021-05-28
do not create wrapped_env when using alf conf

hnyu:


alf/trainers/policy_trainer.py
```diff
@@ -319,11 +319,11 @@ def __init__(self, config: TrainerConfig):
 
         # Create an unwrapped env to expose subprocess gin confs which otherwise
         # will be marked as "inoperative". This env should be created last.
-        # DO NOT register this env in self._envs because AsyncOffPolicyTrainer
-        # will use all self._envs to init AsyncOffPolicyDriver!
-        if self._evaluate or isinstance(
+        # This unwrapped env will only be created when training with gin files.
```
emailweixu: This can also be a problem for alf config.

hnyu: > This can also be a problem for alf config.

For now use a config flag to indicate

emailweixu (approved):

PR #892, opened 2021-06-07
fix mini_batch_length

hnyu:
Currently for off-policy training, if mini_batch_length=None, self._exp_replayer.replay will throw an error

### DIFF
--- a/alf/algorithms/algorithm.py
+++ b/alf/algorithms/algorithm.py
@@ -1110,6 +1110,8 @@ class Algorithm(AlgorithmInterface):
           for every update in ``num_updates_per_train_iter``, the data will
           be shuffled and divided into
           ``buffer_size//(mini_batch_size * mini_batch_length)`` "mini-updates".
+          If ``mini_batch_length`` is None, then ``unroll_length`` will be used
+          for this calculation.
 
         Args:
             update_global_counter (bool): controls whether this function changes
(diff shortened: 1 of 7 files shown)


alf/environments/process_environment.py
```diff
@@ -246,7 +246,7 @@ def _worker(self, conn, env_constructor, env_id=None, flatten=False):
         """
         try:
             alf.set_default_device("cpu")
-            env = env_constructor(env_id)
+            env = env_constructor(env_id=env_id)
```
hnyu, line 249: This is for a more informative error message when a user writes a load function but forgets to have "env_id" as its argument. Otherwise the value of env_id will be treated as for the next argument.

alf/algorithms/algorithm.py
```diff
@@ -1110,6 +1110,8 @@ def train_from_replay_buffer(self, update_global_counter=False):
           for every update in ``num_updates_per_train_iter``, the data will
           be shuffled and divided into
           ``buffer_size//(mini_batch_size * mini_batch_length)`` "mini-updates".
+          If ``mini_batch_length`` is None, then ``unroll_length`` will be used
```
emailweixu, line 1113: For off policy training, I think mini_batch_length needs to be explicitly set by the user. Using unroll_length is often a bad choice.

hnyu: > For off policy training, I think mini_batch_length needs to be explicitly set by the user. Using unroll_length is often a bad choice.

This is for replaying the whole buffer, and I think it's reasonable. I've disabled this for off-policy training

emailweixu (dismissed):

emailweixu (approved):

PR #896, opened 2021-06-08
Fix sampling issue when using a large batch size

Haichao-Zhang:
cuda version of ``torch.randperm(n)`` seems to have a bug when n is a large number, sometimes generating negative or very large values that cause out of bound kernel error, when running together with other codes.

- however, when running cuda version of ``torch.randperm(n)`` alone, no issue observed. 
- when running cpu version of ``torch.randperm(n)`` either alone or embedded into the same piece of code, no issue observed.

Some other users have reported similar issues (https://github.com/pytorch/vision/issues/3816).

### DIFF
--- a/alf/algorithms/algorithm.py
+++ b/alf/algorithms/algorithm.py
@@ -1264,7 +1264,7 @@ class Algorithm(AlgorithmInterface):
                 # have a bug when n is a large number, generating negative or
                 # very large values that cause out of bound kernel error
                 # https://github.com/pytorch/vision/issues/3816
-                indices = np.random.permutation(batch_size)
+                indices = torch.as_tensor(np.random.permutation(batch_size))
                 experience = alf.nest.map_structure(lambda x: x[indices],
                                                     experience)
                 if batch_info is not None:
--- a/alf/algorithms/algorithm.py
+++ b/alf/algorithms/algorithm.py
@@ -18,6 +18,7 @@ import copy
 from collections import OrderedDict
 import itertools
 import json
+import numpy as np
 import os
 import psutil
 import torch
--- a/alf/algorithms/algorithm.py
+++ b/alf/algorithms/algorithm.py
@@ -1258,7 +1258,12 @@ class Algorithm(AlgorithmInterface):
 
         for u in range(num_updates):
             if mini_batch_size < batch_size:
-                indices = torch.randperm(batch_size)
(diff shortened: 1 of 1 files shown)


alf/algorithms/algorithm.py
```diff
@@ -1258,7 +1258,12 @@ def _make_time_major(nest):
 
         for u in range(num_updates):
             if mini_batch_size < batch_size:
-                indices = torch.randperm(batch_size)
+                # here we use numpy random.permutation to generate the permuted
+                # indices, as the cuda version of torch.randperm(n) seems to
+                # have a bug when n is a large number, generating negative or
+                # very large values that cause out of bound kernel error
+                # https://github.com/pytorch/vision/issues/3816
+                indices = np.random.permutation(batch_size)
```
emailweixu: indices will be used several times at line 1267 and 1271. It's better to convert to tensor first `indices = torch.as_tensor(indices)`

emailweixu:
pytorch people haven't confirm it's a bug yet. Did you try to assert the min and max are in the right range immediately after randperm? Just in case it's not caused by something else.

Haichao-Zhang:
> pytorch people haven't confirm it's a bug yet. Did you try to assert the min and max are in the right range immediately after randperm? Just in case it's not caused by something else.


``torch.randperm(200000)`` could generate results as follows:

```
tensor([-5330359564414012330, -5330628352066552185, -5333646077689376993,
         ..., -5426296998407581841,  3819835045095460268,
         3781822724807834884])
```

The issue mentioned above is what I found through a search after I fixed the issue.
It is not confirmed by pytorch people as it seems to be a bit hard to reproduce with several lines of standalone codes.
But the behavior mentioned in that issue is very similar to what I observed.

emailweixu:
> `torch.randperm(200000)` could generate results as follows:
> 
> ```
> tensor([-5330359564414012330, -5330628352066552185, -5333646077689376993,
>          ..., -5426296998407581841,  3819835045095460268,
>          3781822724807834884])
> ```
Can a single randperm get the error? Or it has to be mixed with other code?

Haichao-Zhang:
> Can a single randperm get the error? Or it has to be mixed with other code?

Single randperm does not have the error. It seems to be impacted by something else and has to be mixed with other code. That's what I mean by hard to reproduce with several lines of standalone codes.

emailweixu (approved):

PR #899, opened 2021-06-09
Optimistic actor critic

runjerry:
- Optimistic Actor-Critic algorithm
- Test on HalfCheetah and Humanoid


alf/algorithms/oac_algorithm.py
```diff
@@ -0,0 +1,242 @@
...
+from alf.utils import dist_utils
+
+
+class OacNormalProjectionNetwork(NormalProjectionNetwork):
+    """NormalProjectionNetwork for OacAlgorithm. 
+
+    Differences between NormalProjectionNetwork:
+
+    1. If squashed, normal_dist outputs both the squashed_dist and the normal_dist
+    before squash.
```
hnyu: You can directly use `dist_utils.get_base_dist(squashed_dist)` to access the unsquashed normal dist.

alf/algorithms/oac_algorithm.py
```diff
@@ -0,0 +1,242 @@
...
+                 use_parallel_network=False,
+                 num_critic_replicas=2,
+                 env=None,
+                 config: TrainerConfig = None,
+                 critic_loss_ctor=None,
+                 target_entropy=None,
+                 prior_actor_ctor=None,
+                 target_kld_per_dim=3.,
+                 initial_log_alpha=0.0,
+                 explore=True,
```
emailweixu, line 60: Is there any reason to use "explore=False" for this algorithm? If the answer is no, we can simply remove it.

alf/algorithms/oac_algorithm.py
```diff
@@ -0,0 +1,242 @@
...
+
+        2. Add a switch for explore mode where OAC explore policy is constructed 
+        from the target policy (actor_network) and used for action prediction.
+        """
+        new_state = SacActionState()
+        action_dist, actor_network_state = self._actor_network(
+            observation, state=state.actor_network)
+        assert isinstance(action_dist, tuple), (
+            "both squashed dist and original normal dist are expected.")
+        action_dist, normal_dist = action_dist
```
emailweixu: I think it's better not to implement a special projection network that is only used by this algorithm. Otherwise, you have to subclass every projection network (e.g. StableNormalProjectionNetwork) for this algorithm.

You can get normal_dist by:
```
assert isinstance(action_dist, TransformedDistribution):
assert isinstance(action_dist.base_dist, DiagMultivariateNormal)
normal_dist = action_dist.base_dist
```

alf/algorithms/oac_algorithm.py
```diff
@@ -0,0 +1,242 @@
...
+                (observation, critic_action), state=state.critic)
+            critics = critics.view(-1)
+            q_mean = critics.mean()
+            q_std = torch.abs(critics[0] - critics[1]) / 2.0
+            q_ub = q_mean + self._beta_ub * q_std
+            dqda = nest_utils.grad(critic_action, q_ub.sum())
+            shifted_mean = nest.map_structure(mean_shift_fn, unsquashed_mean,
+                                              dqda, unsquashed_var)
+            action_dist, _ = self._actor_network.projection_net.normal_dist(
+                shifted_mean, unsquashed_std)
```
emailweixu: You can build the distribution directly so that you don't have to add new method `projection_net`:
```
normal_dist = dist_utils.DiagMultivariateNormal(loc=shifted_mean, scale=unsquashed_std)
action_dist = td.TransformedDistribution(
                base_distribution=normal_dist, transforms=action_dist.transforms)
```

alf/examples/oac_half_cheetah_conf.py
```diff
@@ -0,0 +1,82 @@
+# Copyright (c) 2021 Horizon Robotics and ALF Contributors. All Rights Reserved.
```
emailweixu: please add this to bin/train_play_test.py and upload the training curve.

alf/algorithms/oac_algorithm.py
```diff
@@ -0,0 +1,242 @@
...
+    def rollout_step(self, time_step: TimeStep, state: SacState):
+        """Same as SacAlgorithm.rollout_step except that `explore` is set to be
+        `self._explore` when calling `_predict_action`.
+        """
+        action_dist, action, _, action_state = self._predict_action(
+            time_step.observation,
+            state=state.action,
+            epsilon_greedy=1.0,
+            eps_greedy_sampling=True,
+            explore=self._explore)
```
hnyu, line 196: I think for this algorithm, ``self._explore`` has always to be True. Otherwise it reduces to SAC and is pointless?

alf/algorithms/oac_algorithm.py
```diff
@@ -0,0 +1,242 @@
...
+                                   self._dqda_clipping)
+            norm = torch.sqrt(torch.sum(torch.mul(dqda * dqda, sigma))) + 1e-6
+            shift = self._explore_delta * torch.mul(sigma, dqda) / norm
+            return mu + shift
+
+        if explore:
+            critic_action = normal_dist.mean.detach().clone()
+            critic_action.requires_grad = True
+            critics, critic_state = self._critic_networks(
+                (observation, critic_action), state=state.critic)
```
hnyu: ``new_state`` should be updated after getting ``critic_state``

alf/algorithms/oac_algorithm.py
```diff
@@ -0,0 +1,242 @@
...
+            critics = critics.view(-1)
+            q_mean = critics.mean()
+            q_std = torch.abs(critics[0] - critics[1]) / 2.0
+            q_ub = q_mean + self._beta_ub * q_std
+            dqda = nest_utils.grad(critic_action, q_ub.sum())
+            shifted_mean = nest.map_structure(mean_shift_fn, unsquashed_mean,
+                                              dqda, unsquashed_var)
+            action_dist, _ = self._actor_network.projection_net.normal_dist(
```
hnyu: Probably you could directly replace the mean and std of ``action_dist`` computed above at line 168, without adding the two new properties ``projection_net`` and ``normal_dist``.

hnyu (changes requested):

alf/algorithms/oac_algorithm.py
```diff
@@ -0,0 +1,242 @@
...
+                dqda = torch.clamp(dqda, -self._dqda_clipping,
+                                   self._dqda_clipping)
+            norm = torch.sqrt(torch.sum(torch.mul(dqda * dqda, sigma))) + 1e-6
+            shift = self._explore_delta * torch.mul(sigma, dqda) / norm
+            return mu + shift
+
+        if explore:
+            critic_action = normal_dist.mean.detach().clone()
+            critic_action.requires_grad = True
+            critics, critic_state = self._critic_networks(
```
emailweixu: critic_action should be transformed according to action_dist.transforms before feeding to critic_network

emailweixu (changes requested):
Also please update README.md to include this algorithm.

alf/algorithms/oac_algorithm.py
runjerry: Good point, didn't know this.

alf/algorithms/oac_algorithm.py
runjerry: No, the only reason was to make comparisons with SAC baseline easier, will remove it.

alf/algorithms/oac_algorithm.py
runjerry: Yes, good point, will update.

alf/algorithms/oac_algorithm.py
runjerry: Yep, updated.

alf/algorithms/oac_algorithm.py
runjerry: Will update.

alf/algorithms/oac_algorithm.py
runjerry: Yes, just for the sake of running comparison experiments, will remove it.

alf/examples/oac_half_cheetah_conf.py
runjerry: Will do

alf/algorithms/oac_algorithm.py
runjerry: Yes, will do

alf/algorithms/oac_algorithm.py
runjerry: Will do.

alf/algorithms/oac_algorithm.py
```diff
@@ -109,14 +73,15 @@ def __init__(self,
         Refer to SacAlgorithm for Args besides the following.
 
         Args:
-            explore (bool): whether to use the explore policy in rollout_step. Only
-                continuous action space is supported when `explore` is True.
+            explore (bool): default is True for OAC algorithm, where 
+                'unroll_with_grad' has to be True in the 'TrainerConfig'
```
emailweixu: This requirement is not necessary. You can use the following to enable gradient:
```python
with torch.enable_grad():
   ...
```

alf/bin/train_play_test.py
```diff
@@ -612,6 +617,20 @@ def test_trac_sac_pendulum(self):
             conf_file='trac_sac_pendulum.gin',
             extra_train_params=OFF_POLICY_TRAIN_PARAMS)
 
+    @unittest.skip(SKIP_TODO_MESSAGE)
```
emailweixu: don't skip

alf/bin/train_play_test.py
```diff
@@ -612,6 +617,20 @@ def test_trac_sac_pendulum(self):
...
+    def test_oac_halfcheetah(self):
+        self._test(
+            conf_file='oac_halfcheetah_conf.py',
+            skip_checker=self._skip_if_mujoco_unavailable,
+            extra_play_params=OAC_TRAIN_PARAMS)
+
+    @unittest.skip(SKIP_TODO_MESSAGE)
```
emailweixu: don't skip

alf/algorithms/oac_algorithm.py
runjerry: updated.

alf/bin/train_play_test.py
runjerry: updated

alf/algorithms/oac_algorithm.py
```diff
@@ -0,0 +1,220 @@
...
+        """
+        Differences between SacAlgorithm._predict_action:
+
+        1. Only continuous actions are supported.
+
+        2. Add a switch for explore mode where OAC explore policy is constructed 
+        from the target policy (actor_network) and used for action prediction.
+        """
+        new_state = SacActionState()
+        with torch.enable_grad():
```
emailweixu: you only need to use enable_grad for line 160-171

runjerry: That have tried that in the first place, but it did not work. It seems that I have to include forwarding of self._actor_network into the "with torch.enable_grad()" to make things work.

runjerry: I have tried that in the first place and it did not work. I just confirmed that it seems that "with torch.enable_grad()" has to start before the "if explore" to make things work. Will update it.

runjerry: updated.

alf/bin/train_play_test.py
runjerry: updated

alf/bin/train_play_test.py
```diff
@@ -165,6 +165,11 @@ def _to_conf_params(parameters):
 ]
 MBRL_TRAIN_PARAMS = _to_conf_params(MBRL_TRAIN_CONF)
 
+OAC_TRAIN_CONF = OFF_POLICY_TRAIN_CONF + [
+    'TrainerConfig.unroll_with_grad=True',
```
hnyu: Why do you need unroll_with_grad? I think OAC only changes the exploration but not the training. This option will consume lots of memory during rollout.

alf/algorithms/oac_algorithm_test.py
```diff
@@ -0,0 +1,141 @@
...
+
+class OACAlgorithmTest(parameterized.TestCase, alf.test.TestCase):
+    @parameterized.parameters(True, False)
+    def test_oac_algorithm(self, explore=True):
+        reward_dim = 3
+        num_env = 1
+        config = TrainerConfig(
+            root_dir="dummy",
+            unroll_length=1,
+            unroll_with_grad=True,
```
hnyu: Why do you need unroll_with_grad? I think OAC only changes the exploration but not the training.

hnyu (changes requested):

alf/bin/train_play_test.py
runjerry: This was previously used for enabling grad computation during unroll. But already updated following Wei's reviews by using "with torch.enable_grad()" instead. Forgot to updated here.

alf/algorithms/oac_algorithm_test.py
runjerry: Same as above, just updated

hnyu:
I don't have other comments.

alf/examples/oac_humanoid_conf.py
```diff
@@ -0,0 +1,84 @@
...
+alf.config('OneStepTDLoss', td_error_loss_fn=element_wise_squared_loss)
+
+# training config
+alf.config('Agent', rl_algorithm_cls=OacAlgorithm)
+
+alf.config(
+    'TrainerConfig',
+    initial_collect_steps=10000,
+    mini_batch_length=2,
+    unroll_length=1000,
```
emailweixu, line 72: I still think we should use the best setting we know instead of the exact setting of the paper so that if somebody later want to use this, he/she will be get better result. We want to know how better it is compared to our best SAC runs.

emailweixu (commented):
And it's better to include the title in the figures of raining curves when you crop the images.

alf/examples/oac_humanoid_conf.py
runjerry: Let me re-run oac using unroll_length=1, previously it performs similarly as unroll_length=1000, but there has been update since then.

runjerry: I have updated the oac_halfcheetah_conf to "1 rollout/1 update" for better performance. On humanoid, the "1000 rollouts/1000 updates" works slightly better. Both images are updated with titles.

emailweixu (approved):

hnyu (approved):

PR #902, opened 2021-06-12
Fix Timelimit wrapper

Haichao-Zhang:
When ``TimeLimitWrapper`` is used to wrap an env with tensor typed time_step, the dtype of the ``time_step.step_type`` will be reset to np.array instead of tensor for the last step. 

This could crash our current code due to mis-match with the expected dtype.

### DIFF
--- a/alf/environments/alf_wrappers.py
+++ b/alf/environments/alf_wrappers.py
@@ -130,6 +130,9 @@ class TimeLimit(AlfEnvironmentBaseWrapper):
         super(TimeLimit, self).__init__(env)
         self._duration = duration
         self._num_steps = None
+        assert self.batch_size is None or self.batch_size == 1, (
+            "does not support batched environment with batch size larger than one"
+        )
 
     def _reset(self):
         self._num_steps = 0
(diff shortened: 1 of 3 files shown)


hnyu:
What's the use case of TimeLimit for tensor? The current wrap_env always assumes that TimeLimit is applied before numpy arrays are converted to tensors. 

In fact, many other wrappers inside alf_wrappers.py have this assumption.

Haichao-Zhang:
> What's the use case of TimeLimit for tensor? The current wrap_env always assumes that TimeLimit is applied before numpy arrays are converted to tensors.

For example, when I want to render a short video for Carla env, I need to add a time limit wrapper, since the previous interface for setting number of rendering steps has been removed.

There could be other cases as well.

alf/environments/alf_wrappers.py
```diff
@@ -130,6 +130,9 @@ def __init__(self, env, duration):
         super(TimeLimit, self).__init__(env)
         self._duration = duration
         self._num_steps = None
+        assert self.batch_size is None or self.batch_size == 1, (
```
hnyu, line 133: Would be best to also support self.batch_size>1 using conditional_update or torch.where to reset the step_type.

Haichao-Zhang: I have thought about this, but decided not to do it right now since the reset behavior for batch_size > 1 is coupled with the actual implementation of the reset function of the environment. For example, to be able to do a conditional reset, the reset function from env should be able to take this condition information as input, which might not be the case ( e.g. CarlaEnv).

hnyu (approved):

PR #904, opened 2021-06-14
replace all gin.configurable with alf.configurable

hnyu:
Replace all gin.configurable with alf.configurable one time in case someone doesn't know about the transition from gin to alf.


emailweixu (commented):
Should remove 'import gin'

emailweixu (approved):

PR #907, opened 2021-06-16
Compositional FC

Haichao-Zhang:
An FC layer which maintains a set of ``n`` FC parameters for learning. It also accepts as input a compositional weight for combining the parameters.

### DIFF
--- a/alf/layers.py
+++ b/alf/layers.py
@@ -734,6 +734,159 @@ class ParallelFC(nn.Module):
 
 
 @alf.configurable
+class CompositionalFC(nn.Module):
+    """Compositional FC layer."""
+
+    def __init__(self,
+                 input_size,
+                 output_size,
+                 n,
+                 activation=identity,
+                 use_bias=True,
+                 use_bn=False,
+                 use_ln=False,
+                 kernel_initializer=None,
+                 kernel_init_gain=1.0,
+                 bias_init_value=0.0):
+        """
+        It maintains a set of ``n`` FC parameters for learning. During forward
+        computation, it composes the set of parameters using weighted average
+        with the compositional weight provided as input and then performs the
+        FC computation, which is equivalent to combine the pre-activation output
+        from each of the ``n`` FC layers using the compositional weight, and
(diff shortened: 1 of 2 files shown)


alf/layers.py
```diff
@@ -733,6 +733,159 @@ def bias(self):
...
+            self._bn = nn.BatchNorm1d(output_size)
+        else:
+            self._bn = None
+        if use_ln:
+            self._ln = nn.LayerNorm(output_size)
+        else:
+            self._ln = None
+        self.reset_parameters()
+
+    def forward(self, inputs, comp_weight=None):
```
emailweixu: It might be worth to it easier to composite multiple CompositeFC using nn.Sequential. For example, FCBatchEnsemble takes a tuple as input, and also re-output emsemble_id so it can be used by the next layer in a Sequential

alf/layers.py
```diff
@@ -733,6 +733,159 @@ def bias(self):
...
+                self.weight.transpose(1, 2))  # [n, B, k]
+        else:
+            y = torch.bmm(inputs, self._weight.transpose(1, 2))  # [n, B, k]
+        y = y.transpose(0, 1)  # [B, n, k]
+
+        if comp_weight is not None:
+            assert comp_weight.ndim == 2, (
+                "Wrong comp_weight.ndim=%d" % comp_weight.ndim)
+
+            y = y * comp_weight.unsqueeze(-1)
```
emailweixu: you can save the sum at line 835 by using matmul
```python
y = torch.matmul(comp_weight.unsqueeze(1), y).squeeze(1)
```

Haichao-Zhang: Changed to use ``torch.bmm`` to avoid the additional ``sum``. Used ``bmm`` since ``comp_weight`` is a batched tensor.

alf/layers.py
Haichao-Zhang: Good point. Changed to support better chaining and added test.

emailweixu (approved):

PR #908, opened 2021-06-16
lean_function

emailweixu:
Wrap a function to save memory for backward. The returned function performs same
computation as ``func``, but save memory by discarding intermediate results.
It calculates the gradient by recomputing ``func`` using the same input during backward.

### DIFF
--- a/alf/utils/lean_function.py
+++ b/alf/utils/lean_function.py
@@ -1,0 +1,99 @@
+# Copyright (c) 2021 Horizon Robotics and ALF Contributors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import torch
+
+
+class _LeanFunction(torch.autograd.Function):
+    @staticmethod
+    def forward(ctx, func, keywords, *args):
+        # The last len(keywords) of args are keyword arguments for func.
+        ctx.func = func
+        ctx.keywords = keywords
+        tensors = tuple(arg for arg in args if isinstance(arg, torch.Tensor))
+        ctx.args = tuple((isinstance(arg, torch.Tensor),
+                          None if isinstance(arg, torch.Tensor) else arg)
+                         for arg in args)
+        ctx.save_for_backward(*tensors)
+        if keywords:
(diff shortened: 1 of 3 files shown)


alf/utils/lean_function.py
```diff
@@ -0,0 +1,99 @@
...
+    ignored.
+
+    Example:
+
+    .. code-block:: python
+
+        def myfunc(x, w, b, scale=1.0):
+            return torch.sigmoid(scale * (x @ w) + b)
+
+        lean_myfunc = lean_function(myfunc)
```
le-horizon, line 83: would it be too much trouble to implement as an annotator? e.g.
```
@lean_function
def myfunc(..)
```

an example is here:
```
def atomic(func):
    """Make class member function atomic by checking ``class._lock``.

    Can only be applied on class methods, whose containing class
    must have ``_lock`` set to ``None`` or a ``multiprocessing.Lock`` object.

    Args:
        func (callable): the function to be wrapped.

    Returns:
        the wrapped function
    """

    def atomic_deco(func):
        @functools.wraps(func)
        def atomic_wrapper(self, *args, **kwargs):
            lock = getattr(self, '_lock')
            if lock:
                with lock:
                    return func(self, *args, **kwargs)
            else:
                return func(self, *args, **kwargs)

        return atomic_wrapper

    return atomic_deco(func)
```

alf/utils/lean_function.py
```diff
@@ -0,0 +1,99 @@
...
+
+def lean_function(func):
+    """Wrap ``func`` to save memory for backward.
+
+    The returned function performs same computation as ``func``, but save memory
+    by discarding intermediate results. It calculates the gradient by recomputing
+    ``func`` using the same input during backward.
+
+    Note: All the Tensor inputs to ``func`` must be explicitly listed as arguments
+    of ``func``. For example, a tuple of Tensors as argument is not allowed.
```
Haichao-Zhang, line 68: Is a module with learnable parameters considered a valid ``func``? For such a module, during backward, since the parameters are unmodified yet, it can still get the same output given the same input. But an issue might be the gradient of the learnable parameters are lost.

alf/utils/lean_function.py
le-horizon: Does it make sense to do this?  Are there cases where both the before wrapping and after wrapping functions are needed in the code? (except for in tests).

le-horizon (commented):
Look almost good, just one suggestion.

Thanks,
Le

alf/utils/lean_function.py
emailweixu: No. All tensors used by `func` must be from its argument.

alf/utils/lean_function.py
emailweixu: It can be used as a decorator in the current form.

Haichao-Zhang (approved):

alf/utils/lean_function.py
le-horizon: Oh, you are right.

le-horizon (approved):
LG

PR #911, opened 2021-06-18
Move parallel_network flag from algorithm to network

Haichao-Zhang:
The ``use_parallel_network`` flag is an option more on the network side and can be decoupled with the algorithm.

By moving the flag out of algorithm, we do not need to repeatedly maintain a similar logic on deciding which types of parallel network to be used within each kind of algorithms (ddpg/sac/sarsa etc).

By providing a flag in CriticNetwork, the user can still be able to explicitly specify which one to use on the network side.

### DIFF
--- a/alf/algorithms/ddpg_algorithm.py
+++ b/alf/algorithms/ddpg_algorithm.py
@@ -63,7 +63,6 @@ class DdpgAlgorithm(OffPolicyAlgorithm):
                  reward_spec=TensorSpec(()),
                  actor_network_ctor=ActorNetwork,
                  critic_network_ctor=CriticNetwork,
-                 use_parallel_network=False,
                  reward_weights=None,
                  epsilon_greedy=None,
                  calculate_priority=False,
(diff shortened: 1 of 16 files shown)


emailweixu:
I remember that the reason of having this option is that parallelnetwork can be slower.

Haichao-Zhang:
> I remember that the reason of having this option is that parallelnetwork can be slower.

Now  moved to the flag to the network side so that the user can still specify which one to use.
The benefit of doing this is that we do not need to repeatedly maintain a similar logic on deciding which types of parallel network to be used within each kind of algorithms (ddpg/sac/sarsa etc).

emailweixu:
> > I remember that the reason of having this option is that parallelnetwork can be slower.
> 
> Is that the case when the network is very small? In that case, the impact of the speed difference might not be very large? Since we mostly use `use_parallel_network=True` in practice and the other option does not seem to be used very often.

According to the comment in sarsa_algorithm.py, parallelnetwork "can be useful when ``mini_batch_size * mini_batch_length`` (when ``temporally_independent_train_step``  is True)  or ``mini_batch_size`` is not very large". What I remember was that when mini_batch_size is very big, parallelnetwork can be slower. But it's possible that the pytorch improved since then and parallel version is never slower.

I guess a reasonable choice for now is to default it to True. Or you can do an extensive evaluation on the speed of parallelnetwork in different scenarios to make sure it's always faster now.

emailweixu (approved):

PR #915, opened 2021-06-29
Add the directory of the entry point gin file to search path (#914)

### LINKED ISSUE #914: Allow executing train and play outside alf/examples/
There is currently a restriction that we have to run the program from within `alf/examples` (or, the directory that directly contains the entry point `.gin` file). This is bad especially for development phase when the recommended way is to run `python -m alf.bin.train` from the repo's root.

## Root Cause

The directory of the entry point `.gin` file should be added to the search path by default IMHO, but it is not. This means that e.g. `include ddpg.gin` will fail if the current working directory is not `alf/examples/`.

## Solution

Whenever calling the parser of gin, we can add the directory of the entry point `.gin` file to the search path with `gin.add_config_file_search_path()`. Therefore `include ddpg.gin` will always succeed when you specify `--gin_file alf/examples/ddpg_pendulum.gin` for example.

## How to Verify the Fix

From the root of the `alf` repo run

``` bash
python -m alf.bin.train --root_dir ~/tmp/pendulum --gin_file alf/examples/ddpg_pendulum.gin
```

Previously this would fail because `ddpg.gin` is not found. It succeeds after the fix.

breakds:
This allows running alf train, play and other programs from anywhere (no longer restricted to `alf/examples/`). Aiming to fix #914.

Tested locally as stated in #914.

### DIFF
--- a/alf/utils/common.py
+++ b/alf/utils/common.py
@@ -484,6 +484,16 @@ def parse_conf_file(conf_file):
     """
     if conf_file.endswith(".gin"):
         gin_params = getattr(flags.FLAGS, 'gin_param', None)
+
+        # Add the directory of the entry gin file to the search path.
+        # This is to bypass the issue
+        # https://github.com/google/gin-config/issues/104
+        #
+        # This is not ideal because add_config_file_search_path has
+        # side effects, but this should be fine for alf.
+        conf_file_direcotry = pathlib.Path(conf_file).parent
+        gin.add_config_file_search_path(conf_file_direcotry)
+
         gin.parse_config_files_and_bindings([conf_file], gin_params)
         ml_type = alf.get_config_value('TrainerConfig.ml_type')
         if ml_type == 'rl':


alf/utils/common.py
```diff
@@ -484,6 +484,16 @@ def parse_conf_file(conf_file):
...
         gin_params = getattr(flags.FLAGS, 'gin_param', None)
+
+        # Add the directory of the entry gin file to the search path.
+        # This is to bypass the issue
+        # https://github.com/google/gin-config/issues/104
+        #
+        # This is not ideal because add_config_file_search_path has
+        # side effects, but this should be fine for alf.
+        conf_file_direcotry = pathlib.Path(conf_file).parent
+        gin.add_config_file_search_path(conf_file_direcotry)
```
hnyu, line 495: This might have some issue when using an ALF snapshot (old repo stored in a training dir), in which case the conf search path should be carefully set.

breakds: Sorry about my lacking context of ALF snapshot. We can chat offline about it when you have time :)

breakds:
Discussed with @hnyu and realized that `.gin` is being deprecated. Will close this PR.

hnyu:
> Discussed with @hnyu and realized that `.gin` is being deprecated. Will close this PR.

Thanks for bringing this issue up. If you find any alf conf file that uses relative importing (e.g., `from xxx import yyy` or `import yyy`), it would be good to submit PRs to correct them by explicitly using `from alf.xxx import yyy` or `import alf.xxx.yyy`.

hnyu:
> > Discussed with @hnyu and realized that `.gin` is being deprecated. Will close this PR.
> 
> Thanks for bringing this issue up. If you find any alf conf file that uses relative importing (e.g., `import yyy`), it would be good to submit PRs to correct them by explicitly using `from alf.xxx import yyy` or `import alf.xxx.yyy`.

breakds:
> > Discussed with @hnyu and realized that `.gin` is being deprecated. Will close this PR.
> 
> Thanks for bringing this issue up. If you find any alf conf file that uses relative importing (e.g., `from xxx import yyy` or `import yyy`), it would be good to submit PRs to correct them by explicitly using `from alf.xxx import yyy` or `import alf.xxx.yyy`.

Sure I will. Thanks!

PR #916, opened 2021-06-30
Add additional time option to Carla

Haichao-Zhang:
Add additional time option to Carla environment. This is useful especially for the episodes with short route_lengths, as it takes some time for the car to be able to move (because of initial spawning phase with z > 0 and acceleration phase).


breakds (approved):
LGTM, in case this is not too late and this PR is still active.

Issue #917, opened 2021-07-07
Create Nix-based Hermetic Development Environment For Alf


breakds:
# Motivation

One of Nix's advantages is to create deterministic/hermetic development environment. This is especially useful when the project has many dependencies, and some of the dependencies can be tricky to build (i.e. Cuda, pytorch, ...)

Since we are not aiming at deploy alf any time soon, the most important environment we are facing at is the development environment. 

**Ideally, the development environment should not make any assumption about what has been installed and what hasn't been installed in your system**.

# Goal

- [ ] Create nix packages for the dependencies that is not in the [official repo](https://github.com/NixOS/nixpkgs)
- [ ] Being able to switch to Alf development (called `alf dev shell`) with a single command, or none at all, on all Linux machines which has Nvidia cards
- [ ] Document how to activate such dev environment

breakds:
This is done and can be closed.

PR #920, opened 2021-07-07
Part III: Create nix package for (pybind) python library cnest (#917)

breakds:
# Description

This is part of a series of PRs to establish a consistent development environment for alf based on Nix. This series of PRs includes

1. Some of the dependencies are not packaged yet, therefore I would package them individually. There are 5 such packages.
2. Create a `devShell`, which is also a package but is used to establish the environment
3. Create `flake.nix` which enables single line `nix develop` to activate the `devShell`

This PR is part of No.1. Unlike the previous PRs, this PR packages `cnest`, which is a nested project inside alf. In the future we can probably move it out to its own repository.

The current nixpkgs release provides `pybind11` 2.6. If there is no strict reason that we need to have `2.5.0`, I'll just override the version in the requirement like this.

# Test

This alone does not have an impact on alf yet. The build of this package is tested with

```bash
cnest $ nix-build -E 'with import <nixpkgs> {}; with pkgs; python3Packages.callPackage ./default.nix {}'
```
And it succeeded.

### DIFF
--- a/alf/nest/cnest/setup.py
+++ b/alf/nest/cnest/setup.py
@@ -123,7 +123,7 @@ class BuildExt(build_ext):
 
 setup(
     name="cnest",
-    setup_requires=['pybind11>=2.5.0'],
+    setup_requires=['pybind11>=2.6.2'],
     ext_modules=ext_modules,
     cmdclass={'build_ext': BuildExt},
 )
--- a/alf/nest/cnest/setup.py
+++ b/alf/nest/cnest/setup.py
@@ -123,7 +123,7 @@ class BuildExt(build_ext):
 
 setup(
     name="cnest",
-    setup_requires=['pybind11==2.5.0'],
+    setup_requires=['pybind11>=2.5.0'],
     ext_modules=ext_modules,
     cmdclass={'build_ext': BuildExt},
 )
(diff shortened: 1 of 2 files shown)


alf/nest/cnest/setup.py
```diff
@@ -123,7 +123,7 @@ def build_extensions(self):
 
 setup(
     name="cnest",
-    setup_requires=['pybind11==2.5.0'],
+    setup_requires=['pybind11>=2.5.0'],
```
hnyu: For some historical reason in this PR https://github.com/HorizonRobotics/alf/pull/711 , pybind11 was fixed to 2.5.0.

breakds: Not an expert in Python, but I think maybe a certain version of `2.5.x` is harmful. The current latest version is `2.6.2` and we just need to avoid the bad version.

In my development environment the version is fixed to `2.6.2` and it works well.

hnyu: OK then maybe the requirement here should be >=2.6.2 instead?

breakds:
Deprecating this PR as the major part of the nix implementation of the environment has been moved to [alf-nix-devenv](https://github.com/HorizonRobotics/alf-nix-devenv)

PR #921, opened 2021-07-07
Part IV: Create nix package for python library atari-py, with atari ROMS automatically installed

breakds:
# Description

This is part of a series of PRs to establish a consistent development environment for alf based on Nix. This series of PRs includes

1. Some of the dependencies are not packaged yet, therefore I would package them individually. There are 5 such packages.
2. Create a `devShell`, which is also a package but is used to establish the environment
3. Create `flake.nix` which enables single line `nix develop` to activate the `devShell`

This PR is part of No.1. When installed, this package provides the latest `atari-py`, which is used for our `gym`. Also, unlike the `pip` version, this package is battery-included, i.e. it will install all the Atari ROMs automatically as well.

# Test

This alone does not have an impact on alf yet. The build of this package is tested with

```bash
atari-py-with-rom $ nix-build -E 'with import <nixpkgs> {}; with pkgs; python3Packages.callPackage ./default.nix {}'
```
And it succeeded.

# Integration Test

I also tested it with the example `ac_breakout` and it works seamlessly.

### DIFF
--- a/nix/pkgs/atari-py-with-rom/default.nix
+++ b/nix/pkgs/atari-py-with-rom/default.nix
@@ -1,0 +1,131 @@
+# This packages the atari-py together with the Atari 2600 roms.
+
+{ lib
+, buildPythonPackage
+, autoPatchelfHook
+, numpy
+, six
+, isPy37
+, isPy38
+, isPy39
+, stdenv
+, unrar
+, unzip
+, writeShellScriptBin
+}:
+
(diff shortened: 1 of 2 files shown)


breakds:
Deprecating this PR as the major part of the nix implementation of the environment has been moved to [alf-nix-devenv](https://github.com/HorizonRobotics/alf-nix-devenv)

PR #922, opened 2021-07-07
Add Nix-based Development Environment for Alf (#917)

breakds:
# Description

This PR added a `flake.nix` file which can be used to activate the development environment of alf. The development environment is actually imported from [alf-nix-devenv](https://github.com/HorizonRobotics/alf-nix-devenv).

# Integration Test

1. Run `nix develop` to activate the development environment.
2. Train and play `ac_cart_pole_conf.py` to make sure that `pytorch` works correctly.
3. Train and play `taac_bipedal_walker` to make sure that more complicated model works correctly.
4. Train and play `ac_breakout` to make sure that atari based gym works correctly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,14 @@ cd alf
 pip install -e .
 ```
 
+**For Nix Users**: There is a built-in Nix-based development environment defined in [flake.nix](./flake.nix). To activate it, run
+
+```bash
+$ nix develop
+```
+
+in the root of your local repository.
+
 ## Documentation
 
 You can read the [ALF documentation here](https://alf.readthedocs.io/).
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,80 @@
+{
+  "nodes": {
+    "alf-devenv": {
+      "inputs": {
+        "nixpkgs": [
+          "nixpkgs"
+        ],
+        "utils": "utils"
+      },
+      "locked": {
+        "lastModified": 1626997680,
+        "narHash": "sha256-osKD7YhkFy2VfYOqcFQzrsT0MmATUdp7VGIfr/0B76s=",
+        "owner": "HorizonRobotics",
+        "repo": "alf-nix-devenv",
+        "rev": "97f7f1847345a3df094513f74e6e5eb53bfd46e8",
+        "type": "github"
+      },
+      "original": {
+        "owner": "HorizonRobotics",
+        "repo": "alf-nix-devenv",
+        "type": "github"
+      }
+    },
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1622516815,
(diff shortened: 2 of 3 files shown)


le-horizon:
Hey Break,

Thanks for preparing this large change and nicely separating into separate PRs.

Do you hope to merge these PRs soon?  Or could it wait for your presentation of nix?  I really don't have the knowledge to review the PRs, and hoping to attend your lecture on nix before I review.

Really looking forward to use nix to one command setup alf!

Thanks,
Le

breakds:
As discussed yesterday, I have updated the PR, it is now as simple as adding a `flake.nix` that imports the development environment defined elsewhere. It also has a `flake.lock` to lock the versions of the dependencies. Please take another look. Thanks!

le-horizon (commented):
Hey Break, thanks for making this so much easier to review.

Two bigger points:
1) I see that alf-nix-devenv has a separate copy of cnest and fixes the version of gin-config.  If we want to update either, should we also update the copy in this repo?
We update our repo fairly often, and the old version of cnest may not be compatible with the new code from time to time.

2) if we update this repo, will there be any automatic, e.g. CI, tests that would check that the env still works?  Again, the alf repo changes often, so it's perhaps a good idea to have some basic test to protect the devenv?

What do you think?

Thanks,
Le

breakds:
> Hey Break, thanks for making this so much easier to review.
> 
> Two bigger points:
> 
> 1. I see that alf-nix-devenv has a separate copy of cnest and fixes the version of gin-config.  If we want to update either, should we also update the copy in this repo?
>    We update our repo fairly often, and the old version of cnest may not be compatible with the new code from time to time.
> 2. if we update this repo, will there be any automatic, e.g. CI, tests that would check that the env still works?  Again, the alf repo changes often, so it's perhaps a good idea to have some basic test to protect the devenv?
> 
> What do you think?
> 
> Thanks,
> Le

Hi Le,

Thanks for the prompt review! My plan was to create a separate repo for cnest and list it as a dependency of alf instead of part of alf. (That is why I left a TODO in alf-nix-devenv). 

Have discussed with @hnyu about this and I think it is more proper for Haonan to create that repo as he is the author, so that people can attribute the credit correctly :)

Before that happens, I will maintain the separate copy. I do not see `cnest` being updated for the past 6 months, and based on my understanding it is not something that will change rapidly.

Hope that helps!

le-horizon (approved):
Sounds good.

Would be better to have some unittest protecting the devenv.  But the PR looks good to me as is.

Thanks for the contribution Break!

Le

breakds:
> Sounds good.
> 
> Would be better to have some unittest protecting the devenv. But the PR looks good to me as is.
> 
> Thanks for the contribution Break!
> 
> Le

That's a really good suggestion, Le! Sure I will be working on a unit test for the dev env as side project.

PR #924, opened 2021-07-07
Convert ac_breakout to using python configurable

breakds:
# Description

This converts the original `.gin` file to the configuration file. The original motivation is to test the atari based gym.

# Integration Test

Trained on both the `.gin` and `_conf.py` and the comparison looks like below:

Left `_confg.py` vs Right `.gin`

comparison_ac_breakout


hnyu:
Probably run each conf file longer to confirm the performance is good enough? The "AverageReturn" metric should be above 250 after a while.

breakds:
> Probably run each conf file longer to confirm the performance is good enough? The "AverageReturn" metric should be above 250 after a while.

Acknowledged. Will run training for longer time and monitor.

breakds:
After some more extended training, evaluating and digging I think I am close to what have happened. The `AverageReturn` I saw from yesterday was misleading because in the tensorboard graph the range of the `x`-axis (denoting the number of episode steps is only 1k), and that is why it reaches only around `2.5`.

I have no idea how I can tweak the visualization to show more than `1k` episode steps during training. However, I found that if I call `alf.bin.play` for an extensive time I can observe:

1. First, the trained agent is actually pretty good, and it reaches score = 400 easily on every game (and then get stuck in the dead loop).
2. After the `play`, if I look at the tensorboard again, it correctly shows `AverageReturn` with 50k steps or so. And the Average Return looks like below:

average_return

I think this also concludes the integration test of this PR.

hnyu:
> 2. After the `play`, if I look at the tensorboard again, it correctly shows `AverageReturn` with 50k steps or so. And the Average Return looks like below:

The play shouldn't affect how tensorboard shows. Can you share me with your tensorboard link?

breakds:
> > 1. After the `play`, if I look at the tensorboard again, it correctly shows `AverageReturn` with 50k steps or so. And the Average Return looks like below:
> 
> The play shouldn't affect how tensorboard shows. Can you share me with your tensorboard link?

Yep sure, are you coming to the office tomorrow? I can demonstrate that and consult you on this topic.

hnyu:
> > > 1. After the `play`, if I look at the tensorboard again, it correctly shows `AverageReturn` with 50k steps or so. And the Average Return looks like below:
> > 
> > 
> > The play shouldn't affect how tensorboard shows. Can you share me with your tensorboard link?
> 
> Yep sure, are you coming to the office tomorrow? I can demonstrate that and consult you on this topic.

Sounds good

alf/examples/ac_breakout_conf.py
```diff
@@ -0,0 +1,70 @@
...
+# use 'NoFrameSkip' Atari environments from OpenAI gym
+alf.config(
+    'create_environment',
+    env_name='BreakoutNoFrameskip-v4',
+    num_parallel_environments=64)
+
+# Neural Network Configuration
+CONV_LAYER_PARAMS = ((32, 8, 4), (64, 4, 2), (64, 3, 1))
+alf.config(
+    'ActorDistributionNetwork',
```
hnyu: A better way of configuring networks would be using partial, to avoid configuring the network class in a global scope. 

```python
actor_network_cls = partial(ActorDistributionNetwork, fc_layer_params=(512,), conv_layer_params=CONV_LAYER_PARAMS)

alf.config(
    'ActorCriticAlgorithm',
    actor_network_ctor=actor_network_cls,
    ...)
``` 

Same for the value network.

alf/examples/atari_conf.py
```diff
@@ -0,0 +1,34 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import alf
+from alf.environments import suite_gym, gym_wrappers
+from alf.algorithms.data_transformer import FrameStacker, ImageScaleTransformer, RewardClipping
+
+alf.config('create_environment', env_load_fn=suite_gym.load)
```
hnyu, line 18: Would be better specifying a max episode length:

```python
alf.config('suite_gym.load', max_episode_steps=4500) # DQN paper setting, 18000 frames assuming frameskip=4
```

breakds:
Updated the PR addressing @hnyu 's comments:

1. Removing the original `.gin` file
2. Using `functools.partial` for both the actor network and value network
3. Default max episode steps to 4500 (originally implicitly set as 400,000)
4. Also, updated `README.md` to link to the python conf version

Will run training again and update the thread.

breakds:
Ran the experiment with the updated commit and the result is similar. Was able to finish training 5M steps in 22 minutes.

alf/bin/train_play_test.py
```diff
@@ -319,6 +320,7 @@ def _test_train(self, conf_file, extra_params, root_dir):
         if conf_file.endswith('.gin'):
             cmd.append('--gin_file=%s' % conf_file)
         else:
+            print(f'conf_file={conf_file}')
```
hnyu: remove this print

hnyu (changes requested):
Also when you make additional changes to a PR, please do not rebase and force pushing. Just push on top of the existing commits so that the delta change can be seen on github.

alf/bin/train_play_test.py
breakds: Thought I have removed this. Sorry!

hnyu (approved):

breakds:
@hnyu Appreciate the thorough review!

PR #925, opened 2021-07-09
write pre configs to conf file when grid searching

hnyu:
currently using grid search the searched parameters won't be written into the conf file for individual jobs, because only command flag "conf_param" will be recorded. This PR fixes this issue by looking at _HANDLED_PRE_CONFIGS and write them to the file.

Note that gin-config didn't have this issue because we wrote parsed configurations to `configured.gin`.

### DIFF
--- a/alf/utils/common.py
+++ b/alf/utils/common.py
@@ -528,20 +528,17 @@ def write_config(root_dir):
     root_dir = os.path.expanduser(root_dir)
     alf_config_file = os.path.join(root_dir, ALF_CONFIG_FILE)
     os.makedirs(root_dir, exist_ok=True)
-    conf_params = getattr(flags.FLAGS, 'conf_param', None)
+    pre_configs = alf.config_util.get_handled_pre_configs()
     config = ''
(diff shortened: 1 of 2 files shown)


alf/utils/common.py
```diff
@@ -528,20 +528,17 @@ def write_config(root_dir):
     root_dir = os.path.expanduser(root_dir)
     alf_config_file = os.path.join(root_dir, ALF_CONFIG_FILE)
     os.makedirs(root_dir, exist_ok=True)
-    conf_params = getattr(flags.FLAGS, 'conf_param', None)
+    pre_configs = alf.config_util.get_handled_pre_configs()
     config = ''
-    if conf_params:
+    if pre_configs:
         config += "########### config from commandline ###########\n\n"
```
Haichao-Zhang: Is the message in L534 still valid after the change?

hnyu: > Is the message in L534 still valid after the change?

corrected

Haichao-Zhang (approved):

PR #933, opened 2021-07-17
Load BSuite Environment

pd-perry:
Added suite_bsuite to load the DeepMind BSuite environment on ALF.


alf/examples/oac_halfcheetah_conf.py
```diff
@@ -23,7 +23,7 @@
 from alf.utils.math_ops import clipped_exp
 from alf.utils.losses import element_wise_squared_loss
 
-from alf.examples import sac_conf
+import sac_conf
```
hnyu: Please revert all import changes to "from alf.examples import ...". It seems that your PR is not based on the most update-to-date alf. Make sure to pull first.

alf/environments/suite_bsuite.py
```diff
@@ -0,0 +1,106 @@
...
+
+    Returns:
+        An AlfEnvironment instance.
+    """
+
+    env = bsuite.load_from_id(environment)
+    gym_env = BSuiteWrapper(env)
+
+    if max_episode_steps is None:
+        max_episode_steps = 0
```
hnyu, line 67: For those envs that do have max_episode_steps, do you need to obtain it from the env so that ALF will correctly set the StepType?

alf/environments/suite_bsuite.py
```diff
@@ -0,0 +1,106 @@
...
+    """Loads the selected environment and wraps it with the specified wrappers.
+
+    Note that by default a TimeLimit wrapper is used in wrap_env to limit episode 
+    lengths to the default benchmarks defined by the registered environments.
+
+    Args:
+        environment_name (str): Name for the environment to load.
+        env_id (int): (optional) ID of the environment.
+        discount (float): Discount to use for the environment.
+        max_episode_steps (int): If None the max_episode_steps will be set to zero as not
```
hnyu, line 48: The description of this argument has an incorrect format.

alf/environments/suite_bsuite.py
```diff
@@ -0,0 +1,106 @@
+import alf
```
hnyu, line 15: Please add the copyright header at the top. See other files for an example.

alf/environments/suite_bsuite.py
```diff
@@ -0,0 +1,106 @@
...
+        reward = timestep.reward or 0.
+        if timestep.last():
+            self.game_over = True
+        return np.reshape(timestep.observation, (timestep.observation.shape[1], )), reward, timestep.last(), {}
+
+    def reset(self) -> np.ndarray:
+        self.game_over = False
+        timestep = self._env.reset()
+        self._last_observation = timestep.observation
+        return np.reshape(timestep.observation, (timestep.observation.shape[1], ))
```
hnyu: A simple test file would be needed. See ``suite_safety_gym_test.py`` or ``suite_babyai_test.py`` for examples.

hnyu (changes requested):
Thanks for your PR! I have several comments to make it better.

alf/environments/suite_bsuite.py
```diff
@@ -0,0 +1,106 @@
+import alf
+from alf.environments import gym_wrappers, alf_wrappers, alf_gym_wrapper
+from alf.environments.suite_gym import wrap_env
+
+import bsuite
```
breakds, line 19: Since this is a new dependency, it's probably a good idea to add it to `setup.py` or `requirements.txt` as well.

alf/environments/suite_bsuite.py
pd-perry: Added copyright header

alf/environments/suite_bsuite.py
pd-perry: Added bsuite to requirements.txt

alf/environments/suite_bsuite.py
pd-perry: Changed environment parameter to environment_name and reformatted the max_episode_steps description.

alf/environments/suite_bsuite.py
pd-perry: It does not seem like the author of bsuite intended others to directly obtain the max_time or max_steps variable, since it is marked with an underscore.

alf/environments/suite_bsuite.py
pd-perry: Added suite_bsuite_test.py for some simple tests.

alf/examples/oac_halfcheetah_conf.py
pd-perry: Reverted all import changes to "from alf.examples import ..."

alf/environments/suite_bsuite.py
hnyu: > It does not seem like the author of bsuite intended others to directly obtain the max_time or max_steps variable, since it is marked with an underscore.

The reason why we want ALF to differentiate between a normal episode end (success/failure) and a timeout is detailed on this page: https://alf.readthedocs.io/en/latest/notes/knowledge_base.html#about-timestep-discount
(starting from "Each TimeStep is associated with a discount value....")

In a word, for timeout step, ALF has the step type `LAST` with discount=1, while for a normal termination step, ALF has the step type `LAST` with discount=0. 

Unfortunately, most envs don't know this subtle difference, including bsuite. For example, in its MountainCar env:

https://github.com/deepmind/bsuite/blob/master/bsuite/environments/mountain_car.py#L89

It treats a success end and timeout the same, which is not correct. 

By obtaining `max_epsiodes` info from the env and pass `max_steps-1` to `suite_gym.wrap_env`, ALF will make sure the correct step type and discount is used. The reason why we should have `max_steps-1` is to prevent bsuite assigning incorrect step type and discount at the last step (which cannot be corrected by ALF later). 

And I guess it doesn't matter if the variable has an underscore or not, as long as the right thing is done.

alf/environments/suite_bsuite.py
```diff
@@ -1,3 +1,17 @@
+# Copyright (c) 2020 Horizon Robotics and ALF Contributors. All Rights Reserved.
```
hnyu: 2020 -> 2021?

alf/environments/suite_bsuite_test.py
```diff
@@ -0,0 +1,69 @@
...
+
+        constructor = functools.partial(ctor)
+
+        self._env = parallel_environment.ParallelAlfEnvironment(
+            [constructor] * env_num)
+        self.assertTrue(self._env.batched)
+        self.assertEqual(self._env.batch_size, env_num)
+        self.assertEqual(torch.float32, self._env.observation_spec().dtype)
+        # self.assertEqual((suite_bsuite.VectorReward.REWARD_DIMENSION, ),
+        #                  self._env.reward_spec().shape)
```
hnyu: Please remove this comment if not used.

hnyu (changes requested):
Thank you for the change! Just some additional minor comments.

alf/environments/suite_bsuite.py
pd-perry: Thanks so much for the clarifications! I have added a check for if the environment has instance variable _max_steps, and if it does, set the max_episode_steps to env._max_steps - 1.

alf/environments/suite_bsuite.py
pd-perry: Changed to 2021 for both suite_bsuite and suite_bsuite_test.

alf/environments/suite_bsuite_test.py
pd-perry: This comment has been removed.

hnyu (approved):

PR #937, opened 2021-07-19
add config num_summaries and num_evals

hnyu:
If num_iterations or num_env_steps are fixed, it might be more convenient to directly specify num_summaries or num_evals in order to have a better control of the tensorboard event file size or evaluation time.


runjerry:
I had the same feeling when running oac experiments and agree that this might be more convenient.

breakds (dismissed):

breakds (approved):

Issue #913, opened 2021-06-28
Support Distributed Training in alf


breakds:
As discussed with @emailweixu, it would be nice to have alf support multi-GPU training. The goals are

- [x] As a starting project, help me gain enough knowledge about the repository itself
- [x] Being able to to train models on machines with multiple GPU (single machine)

hnyu:
@breakds FYI, two reference papers I came across a while ago (RL scenario): 

https://openreview.net/pdf?id=H1gX8C4YPr
https://ai.googleblog.com/2020/03/massively-scaling-reinforcement.html

Although they are proposed for multi-machine training, our multi-gpu single-machine case is a special and simpler case.

Or refer to Pytorch official multi-gpu support (general DL scenario):
https://pytorch.org/tutorials/beginner/former_torchies/parallelism_tutorial.html

breakds:
Thanks @hnyu for the references!

breakds:
DD-PPO, the first paper's main idea is about early stopping the slow simulation during rollout with a batched environment (potentially distributed over different machines in a cluster), and try to use the full experience from some of the environment and partial experience from the early-stopped ones during the training in each iteration. I think we can borrow the ideas in the near future.

As the first step, I will look into how pytorch's `DataParallel` is implemented and use it (or similar techniques) to enable multi-GPU single-machine's

1. Network(s) forward evaluation during rollout
2. Network(s) forward/backward evaluation during training

in each of the training iteration.

breakds:
Currently I am hitting two problems with `DataParallel`:

1. When tried on a simple `backward()` operation, `DataParallel` version (2 GPU) is taking 1.8 seconds while the sinlge GPU version only takes 0.5 seconds, and I haven't figured out why. I think this at least suggest the overhead of `DataParallel` is pretty significant.
2. Directly applying `DataParallel` on our network and it will crash. This is another thing that I am working on.

hnyu:
> 1. When tried on a simple `backward()` operation, `DataParallel` version (2 GPU) is taking 1.8 seconds while the sinlge GPU version only takes 0.5 seconds, and I haven't figured out why. I think this at least suggest the overhead of `DataParallel` is pretty significant.

I think multi-gpu only makes sense for a large mini-batch with intensive computation. What is your setup?

breakds:
[...]

PR #938, opened 2021-07-19
[Refcator] Extract message types as MessageType Enum in ProcessEnvironment

breakds:
This is a refactor, which means that no logic is being changed.

# Motivation

This is part of the effort to address #913. A sub-task requires extract the worker logic to be out of the class (for some reason it will prevent `multiprocessing` to work correctly). Without such change the `multiprocessing.Process` will just be stuck on `start()`.

### DIFF
--- a/alf/environments/process_environment.py
+++ b/alf/environments/process_environment.py
@@ -19,6 +19,7 @@ Adapted from TF-Agents Environment API as seen in:
 
 from absl import logging
 import atexit
+from enum import Enum
 import multiprocessing
 import numpy as np
 import sys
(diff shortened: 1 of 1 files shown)


runjerry (approved):

emailweixu:
> This is a refactor, which means that no logic is being changed.
> 
> # Motivation
> This is part of the effort to address #913. A sub-task requires extract the worker logic to be out of the class (for some reason it will prevent `multiprocessing` to work correctly). Without such change the `multiprocessing.Process` will just be stuck on `start()`.

How does this change solve the problem of stucking on `start()`?

breakds:
> > This is a refactor, which means that no logic is being changed.
> > # Motivation
> > This is part of the effort to address #913. A sub-task requires extract the worker logic to be out of the class (for some reason it will prevent `multiprocessing` to work correctly). Without such change the `multiprocessing.Process` will just be stuck on `start()`.
> 
> How does this change solve the problem of stucking on `start()`?

I do not know why as well, this is just an observation. My guess is that without passing `self` the multiprocessing do not have to pickle the `ProcessEnvironment` itself, which might be triggering something that contribute to the "getting stuck" problem.

PR #939, opened 2021-07-19
[Refactor] Extract worker logic out of ProcessEnvironment #913

breakds:
This is a refactor PR that does not incur any logic change

# Motivation

This is part of the effort to unblock #913. Two reasons for this change

1. `worker` definitely does not rely on `ProcessEnvironment` at all, and therefore it is cleaner to make it independent of `ProcessEnvironment`.
2. If it stays as a member method of `ProcessEnvironment`, `multiprocess.Process` will get stuck on `start()` if the parent process is also a `multiprocess.Process`, for unknown reason though (tried investigation but haven't figured out).

### DIFF
--- a/alf/environments/process_environment.py
+++ b/alf/environments/process_environment.py
@@ -64,6 +64,64 @@ class _MessageType(Enum):
     CLOSE = 6
 
 
+def _worker(conn, env_constructor, env_id=None, flatten=False):
+    """The process waits for actions and sends back environment results.
+
+    Args:
+        conn (multiprocessing.connection): Connection for communication to the main process.
+        env_constructor (Callable): callable environment creator.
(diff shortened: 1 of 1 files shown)


alf/environments/process_environment.py
```diff
@@ -64,6 +64,64 @@ class _MessageType(Enum):
     CLOSE = 6
 
 
+def _worker(conn, env_constructor, env_id=None, flatten=False):
```
breakds, line 67: Note that this is a full copy of the original method, **without any** modification.

runjerry (dismissed):

breakds:
Updated with just a rebase.

emailweixu (approved):

PR #940, opened 2021-07-19
Replace the official gin-config with customized gin-config in Nix dev…

breakds:
# Motivation

This is to pin the `gin-config` to the exact version we have in https://github.com/HorizonRobotics/gin-config in the Nix based development environment


breakds:
Absorbed to the nix dev environment PR #922

PR #943, opened 2021-07-22
update oac algorithm with latest alf updates

runjerry:
oac algorithm fell behind recent alf updates due to my forgetting to include its tests in build.sh.

### DIFF
--- a/alf/algorithms/oac_algorithm_test.py
+++ b/alf/algorithms/oac_algorithm_test.py
@@ -32,9 +32,8 @@ from alf.utils import common
 from alf.utils.math_ops import clipped_exp
 
 
-class OACAlgorithmTest(parameterized.TestCase, alf.test.TestCase):
-    @parameterized.parameters(True, False)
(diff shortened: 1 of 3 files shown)


emailweixu (dismissed):

alf/algorithms/oac_algorithm_test.py
```diff
@@ -90,8 +87,8 @@ def test_oac_algorithm(self, explore=True):
             use_entropy_reward=reward_dim == 1,
             env=env,
             config=config,
-            explore=explore,
-            explore_delta=2.,
+            explore=True,
```
emailweixu, line 90: explore=explore

runjerry: I just removed the explore=False from the test to save some time, since it is the same as sac. Should I add it back?

emailweixu: I see. Didn't notice that.

emailweixu (approved):

PR #944, opened 2021-07-23
[REFACTOR] train.py to consolidate common logic for both single GPU and multi GPU training (#913)

breakds:
# Motivation

The effort of #913 requires updating the entry point script `alf/bin/train.py`. Such updates will rearrange some of the code logic in order to **minimize code duplication**. I think it is a good practice to split the work and submit the refactor earlier so that the work is done in a more incremental fashion, with the benefits of

1. Since this is mainly a refactor PR, much easier to review (and to revert if needed)
2. The DDP project can take a few days to finish, and if someone has to update the original `train.py`, the risk of resolving conflicts increases. Submitting this PR first will minimize such risk.

# How is it refactored?

1. Abstract functions that does the setup such as logging and snashop.
2. Added a new flag `--distributed` which takes enum value. The `multi-gpu` is currently disabled explicitly.
2. Convert `train_eval` into `training_worker` so that it does the training/evaluation job as a single proce
3. Later, the `multi-gpu` mode is just about running several processes via `multiprocessing`, while each process just runs `train_eval_worker`.

# How is it tested?

1. **Disable multi-gpu** - Verified that when `multi-gpu` is specified in `--distributed`, it prompts to tell the user that DDP is currently unavailable
2. **No-regression** - Verified that when running with single gpu mode by not specifying `--distributed` (i.e. the same as before), it runs `ac_breakout` successfully and the score hits 50 in 4k training steps.

### DIFF
--- a/alf/bin/train.py
+++ b/alf/bin/train.py
@@ -53,7 +53,6 @@ import pathlib
 import torch
 import torch.distributed as dist
 import torch.multiprocessing as mp
-import traceback
 from typing import Optional
 
 from alf.utils import common
@@ -71,8 +70,9 @@ def _define_flags():
     flags.DEFINE_multi_string('conf_param', None, 'Config binding parameters.')
     flags.DEFINE_bool('store_snapshot', True,
                       'Whether store an ALF snapshot before training')
-    flags.DEFINE_enum('distributed', 'none', ['none', 'multi-gpu'],
-                      'Whether store an ALF snapshot before training')
+    flags.DEFINE_enum(
+        'distributed', 'none', ['none', 'multi-gpu'],
+        'Set whether and how to run trainning in distributed mode.')
     flags.mark_flag_as_required('root_dir')
 
 
@@ -83,8 +83,8 @@ def _setup_logging(rank: int = 0, log_dir: Optional[str] = None):
     """Setup logging for each process
 
     Args:
-        rank: The ID of the process among all of the DDP processes
-        log_dir: path to the direcotry where log files are written to
+        rank (int): The ID of the process among all of the DDP processes
+        log_dir (Optional[str]): path to the direcotry where log files are written to
     """
     FLAGS.alsologtostderr = True
     logging.set_verbosity(logging.INFO)
@@ -94,36 +94,30 @@ def _setup_logging(rank: int = 0, log_dir: Optional[str] = None):
 def _setup_device(rank: int = 0):
     """Setup the GPU device for each process
 
-    Each process will use the GPU with the same rank (device id) by default.
+    All tensors of the calling process will use the GPU with the
+    specified rank by default.
 
     Args:
-        rank: The ID of the process among all of the DDP processes
+        rank (int): The ID of the process among all of the DDP processes
+
     """
     if torch.cuda.is_available():
         alf.set_default_device('cuda')
         torch.cuda.set_device(rank)
 
 
-def _snapshot_alf(root_dir: str):
-    """Snapshot the alf repository to the specified path
-
-    Args:
-        root_dir: the direcotry that stores various states of this training session
-    """
-    # ../<ALF_REPO>/alf/bin/train.py
-    file_path = os.path.abspath(__file__)
-    alf_root = str(pathlib.Path(file_path).parent.parent.parent.absolute())
-    # generate a snapshot of ALF repo as ``<root_dir>/alf``
-    common.generate_alf_root_snapshot(alf_root, root_dir)
-
-
 def training_worker(rank: int, world_size: int, conf_file: str, root_dir: str):
     """An executable instance that trains and evaluate the algorithm
+
+    Args:
+        rank (int): The ID of the process among all of the DDP processes.
+        world_size (int): The number of processes in total. If set to 1, it is interpreted as "non distributed mode".
+        conf_file (str): Path to the training configuration.
+        root_dir (str): Path to the directory for writing logs/summaries/checkpoints.
     """
     try:
         _setup_logging(log_dir=root_dir, rank=rank)
         _setup_device(rank)
-        print(world_size)
         if world_size > 1:
             # Specialization for distributed mode
             dist.init_process_group('nccl', rank=rank, world_size=world_size)
@@ -154,9 +148,7 @@ def training_worker(rank: int, world_size: int, conf_file: str, root_dir: str):
         # If the training worker is running as a process in multiprocessing
         # environment, this will make sure that the exception raised in this
         # particular process is captured and shown.
-        tb = traceback.format_exc()
-        print(e)
-        print(tb)
+        logging.exception(e)
     finally:
         # Note that each training worker will have its own child processes
         # running the environments. In the case when training worker process
@@ -170,7 +162,7 @@ def main(_):
     os.makedirs(root_dir, exist_ok=True)
 
     if FLAGS.store_snapshot:
-        _snapshot_alf(root_dir)
+        common.generate_alf_root_snapshot(common.alf_root(), root_dir)
 
     conf_file = common.get_conf_file()
 
--- a/alf/bin/train.py
+++ b/alf/bin/train.py
@@ -88,11 +88,7 @@ def _setup_logging(rank: int = 0, log_dir: Optional[str] = None):
     """
     FLAGS.alsologtostderr = True
     logging.set_verbosity(logging.INFO)
-
-    # When there are multiple processes, only the process with rank = 0 writes
-    # to the log directory to avoid race condition.
-    if rank == 0 and log_dir is not None:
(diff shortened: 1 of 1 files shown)


alf/bin/train.py
```diff
@@ -67,54 +71,149 @@ def _define_flags():
...
-        trainer = policy_trainer.SLTrainer(trainer_conf)
-    else:
-        raise ValueError("Unsupported ml_type: %s" % trainer_conf.ml_type)
+    FLAGS.alsologtostderr = True
+    logging.set_verbosity(logging.INFO)
+
+    # When there are multiple processes, only the process with rank = 0 writes
+    # to the log directory to avoid race condition.
+    if rank == 0 and log_dir is not None:
+        logging.get_absl_handler().use_absl_log_file(log_dir=log_dir)
```
emailweixu: I think process id is the suffix of the name of absl log file. So there should be no race condition.

alf/bin/train.py
```diff
@@ -67,54 +71,149 @@ def _define_flags():
...
+                raise RuntimeError(
+                    "Multi-GPU DDP training does not support supervised learning"
+                )
+            trainer = policy_trainer.SLTrainer(trainer_conf)
+        else:
+            raise ValueError("Unsupported ml_type: %s" % trainer_conf.ml_type)
+
+        trainer.train()
+
+        # Load the configuration file, note that this will also bring up the
```
emailweixu: unfinished sentence.

alf/bin/train.py
```diff
@@ -67,54 +71,149 @@ def _define_flags():
...
+        # Force setting multiprocessing's start method to spawn. This is
+        # required for DDP to run correctly.
+        try:
+            mp.set_start_method('spawn', force=True)
+        except RuntimeError:
+            pass
+
+        processes = []
+        for i in range(world_size):
+            processes.append(
```
emailweixu: why not use mp.spawn?

alf/bin/train.py
```diff
@@ -67,54 +71,149 @@ def _define_flags():
...
+    common.generate_alf_root_snapshot(alf_root, root_dir)
+
+
+def training_worker(rank: int, world_size: int, conf_file: str, root_dir: str):
+    """An executable instance that trains and evaluate the algorithm
+    """
+    try:
+        _setup_logging(log_dir=root_dir, rank=rank)
+        _setup_device(rank)
+        print(world_size)
```
hnyu: Should remove this 'print'?

alf/bin/train.py
```diff
@@ -67,54 +71,149 @@ def _define_flags():
     flags.DEFINE_multi_string('conf_param', None, 'Config binding parameters.')
     flags.DEFINE_bool('store_snapshot', True,
                       'Whether store an ALF snapshot before training')
+    flags.DEFINE_enum('distributed', 'none', ['none', 'multi-gpu'],
+                      'Whether store an ALF snapshot before training')
```
hnyu: The comment message of this flag is not correct

alf/bin/train.py
```diff
@@ -67,54 +71,149 @@ def _define_flags():
...
+        trainer.train()
+
+        # Load the configuration file, note that this will also bring up the
+    except Exception as e:
+        # If the training worker is running as a process in multiprocessing
+        # environment, this will make sure that the exception raised in this
+        # particular process is captured and shown.
+        tb = traceback.format_exc()
+        print(e)
+        print(tb)
```
hnyu: better use logging.error or logging.info

alf/bin/train.py
```diff
@@ -67,54 +71,149 @@ def _define_flags():
...
+    # to the log directory to avoid race condition.
+    if rank == 0 and log_dir is not None:
+        logging.get_absl_handler().use_absl_log_file(log_dir=log_dir)
 
-    trainer.train()
+
+def _setup_device(rank: int = 0):
+    """Setup the GPU device for each process
+
+    Each process will use the GPU with the same rank (device id) by default.
```
hnyu: You mean "all tensors of the process will use the GPU with the same rank"?

alf/bin/train.py
```diff
@@ -67,54 +71,149 @@ def _define_flags():
...
 
-@alf.configurable
-def train_eval(root_dir):
-    """Train and evaluate algorithm
+def _setup_logging(rank: int = 0, log_dir: Optional[str] = None):
+    """Setup logging for each process
 
     Args:
-        root_dir (str): directory for saving summary and checkpoints
+        rank: The ID of the process among all of the DDP processes
```
hnyu: For all the args, please add types after the arg names.

alf/bin/train.py
```diff
@@ -67,54 +71,149 @@ def _define_flags():
...
+    # ../<ALF_REPO>/alf/bin/train.py
+    file_path = os.path.abspath(__file__)
+    alf_root = str(pathlib.Path(file_path).parent.parent.parent.absolute())
+    # generate a snapshot of ALF repo as ``<root_dir>/alf``
+    common.generate_alf_root_snapshot(alf_root, root_dir)
+
+
+def training_worker(rank: int, world_size: int, conf_file: str, root_dir: str):
+    """An executable instance that trains and evaluate the algorithm
+    """
```
hnyu, line 116: Add comments on arguments unless it's private usage.

alf/bin/train.py
```diff
@@ -67,54 +71,149 @@ def _define_flags():
...
+
+def _snapshot_alf(root_dir: str):
+    """Snapshot the alf repository to the specified path
+
+    Args:
+        root_dir: the direcotry that stores various states of this training session
+    """
+    # ../<ALF_REPO>/alf/bin/train.py
+    file_path = os.path.abspath(__file__)
+    alf_root = str(pathlib.Path(file_path).parent.parent.parent.absolute())
```
hnyu: Now you can directly call common.alf_root(). So probably we don't need a function for this. 

https://github.com/HorizonRobotics/alf/blob/pytorch/alf/utils/common.py#L1145

alf/bin/train.py
breakds: Verified that there are suffix of the process together with the timestamp, so we should be fine. Thanks!

I removed this piece of code.

alf/bin/train.py
breakds: Sorry about the mistake. I removed this sentence and put the complete comments where it should be.

alf/bin/train.py
breakds: Because I did not know about `mp.spawn` ... Thanks for mentioning that. Rewrite this with `mp.spawn` and it looks much cleaner and nicer.

hnyu (changes requested):

alf/bin/train.py
breakds: Absolutely. Removed.

alf/bin/train.py
breakds: A sloppy copy-paste. This is fixed now.

alf/bin/train.py
breakds: Replaced this with `logging.exception`. Thanks for pointing this out.

alf/bin/train.py
breakds: Pretty much. The original documentation is indeed unclear. Rephrased now.

alf/bin/train.py
breakds: Done.

alf/bin/train.py
breakds: Done.

alf/bin/train.py
breakds: Sure. This function is removed. I think in this case we can probably update `generate_alfjroot_snapshot` to take just one argument `root_dir`.

breakds:
All comments resolved. PTAL.

alf/bin/train.py
hnyu: > Sure. This function is removed. I think in this case we can probably update `generate_alfjroot_snapshot` to take just one argument `root_dir`.

Potentially we could have multiple ALF repos in the python path. The current common.alf_root() only returns the activated one. If using alf snapshot, common.alf_root() will return the snapshot root path.

alf/bin/train.py
```diff
@@ -180,6 +180,11 @@ def main(_):
                 rank=0, world_size=1, conf_file=conf_file, root_dir=root_dir)
             return
 
+        # The other process will communicate with the authoritative
+        # process via network protocol on localhost:12355.
+        os.environ['MASTER_ADDR'] = 'localhost'
```
hnyu, line 184: Is it better to add prefix to these env variables, e.g., "ALF_"?

breakds: I do agree. Sadly this is hard-coded in pytorch. This is a pytroch distributed protocol.

emailweixu (approved):

hnyu (approved):

PR #949, opened 2021-07-29
Disable optimizer's mnemonic naming in algorithm.state_dict()

breakds:
# Motivation

Previously the name for `optimizer` is mnemonic. This means that if an optimizer can be traced from both `A` and `B` in their attribute tree, the name of the `optimizer` in their `state_dict` will be the same. Such behavior is not functional - whether the name is based on path from `A` or `B` depends on whose `state_dict()` is called first.

This might be an old design choice (or a bug, because the mnemonic container is being used) and is not desired at this moment. For example, during writing checkpoint and reading checkpoint, if the name of a specific optimizer is decided differently, the read and write might of checkpoint can be incompatible and results in exception.

# Solution

Remove the mnemonic naming handling and the mnemonic cache container.

# Testing

Added a test case for this. Before the fix the test fails, and after the fix the test succeeds.

### DIFF
--- a/alf/algorithms/algorithm_test.py
+++ b/alf/algorithms/algorithm_test.py
@@ -39,10 +39,6 @@ class MyAlg(Algorithm):
         return ['ignored_param']
 
 
-# class HigherLevelAlgorithm(Algorithm):
-#     def __init__(self
-
-
 class AlgorithmTest(alf.test.TestCase):
     def test_flatten_module(self):
         a = nn.Module()
--- a/alf/algorithms/algorithm_test.py
+++ b/alf/algorithms/algorithm_test.py
@@ -39,6 +39,10 @@ class MyAlg(Algorithm):
         return ['ignored_param']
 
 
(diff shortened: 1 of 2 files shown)


alf/algorithms/algorithm_test.py
```diff
@@ -39,6 +39,10 @@ def _trainable_attributes_to_ignore(self):
         return ['ignored_param']
 
 
+# class HigherLevelAlgorithm(Algorithm):
```
Haichao-Zhang: These commented lines can be removed

alf/algorithms/algorithm_test.py
```diff
@@ -39,6 +39,10 @@ def _trainable_attributes_to_ignore(self):
...
+#     def __init__(self
+
+
```
emailweixu: remove

emailweixu (changes requested):

emailweixu (approved):

Haichao-Zhang (approved):

PR #952, opened 2021-07-30
support multi-dim reward for AC and PPO

hnyu:
For both, the multi-dim advantages are weighted averaged before pg_loss is computed. The td_loss is still multi-dim. 

Some other minor changes:
1. Added a Lagrangian method using prediction rewards.
2. Added an option of projecting the positive weights to a simplex before returning them.
3. Added a common ppo_conf.py file

### DIFF
--- a/alf/algorithms/actor_critic_loss.py
+++ b/alf/algorithms/actor_critic_loss.py
@@ -15,10 +15,12 @@
 from collections import namedtuple
 
 import torch
+import numpy as np
 
 import alf
 from alf.data_structures import LossInfo
 from alf.utils.losses import element_wise_squared_loss
+from alf.utils.summary_utils import safe_mean_hist_summary
 from alf.utils import tensor_utils, dist_utils, value_ops
 from .algorithm import Loss
 
@@ -27,13 +29,18 @@ ActorCriticLossInfo = namedtuple("ActorCriticLossInfo",
 
 
 def _normalize_advantages(advantages, variance_epsilon=1e-8):
-    # advantages is of shape [rollout_steps, num_envs]
+    # advantages is of shape [T, B] or [T, B, N], where N is reward dim
     # this function normalizes over all elements in the input advantages
-    adv_mean = advantages.mean()
-    adv_var = torch.var(advantages, unbiased=False)
+    shape = advantages.shape
+    # shape: [TB, 1] or [TB, N]
+    advantages = advantages.reshape(np.prod(advantages.shape[:2]), -1)
+
+    adv_mean = advantages.mean(0)
+    adv_var = torch.var(advantages, dim=0, unbiased=False)
+
     normalized_advantages = (
         (advantages - adv_mean) / (torch.sqrt(adv_var) + variance_epsilon))
-    return normalized_advantages
+    return normalized_advantages.reshape(*shape)
 
 
 @alf.configurable
@@ -59,6 +66,9 @@ class ActorCriticLoss(Loss):
             - entropy_regularization * entropy)
 
         Args:
+            gamma (float|list[float]): A discount factor for future rewards. For
+                multi-dim reward, this can also be a list of discounts, each
+                discount applies to a reward dim.
             td_errors_loss_fn (Callable): A function for computing the TD errors
                 loss. This function takes as input the target and the estimated
                 Q values and returns the loss for each element of the batch.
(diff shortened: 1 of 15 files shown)


alf/algorithms/actor_critic_loss.py
```diff
@@ -130,10 +153,15 @@ def forward(self, info):
             advantages = torch.clamp(advantages, -self._advantage_clip,
                                      self._advantage_clip)
 
+        if info.reward_weights is not ():
```
emailweixu: It seems that now pytorch can correctly handle `tensor != ()`. So we don't need to use `is not`. `is not` will generate warning with python 3.9.

alf/algorithms/ppo_algorithm.py
```diff
@@ -52,23 +52,30 @@ def train_step(self, inputs: TimeStep, state, rollout_info):
...
+            discounts=discounts,
             td_lambda=self._loss._lambda,
             time_major=False)
-        advantages = torch.cat([
-            advantages,
-            torch.zeros(*advantages.shape[:-1], 1, dtype=advantages.dtype)
-        ],
-                               dim=-1)
+        advantages = tensor_utils.tensor_extend_zero(
+            advantages.transpose(0, 1)).transpose(0, 1)
```
emailweixu: If you want to change to use tensor_extend_zero, I feel it's better to enchance tensor_extend_zero to add a dim argument.

alf/algorithms/lagrangian_reward_weight_algorithm.py
```diff
@@ -59,6 +69,9 @@ def __init__(self,
                 its init value is always used.
             optimizer (optimizer): optimizer for learning the reward weights.
             init_weights (float|list[float]): the initial reward weights.
+            max_weight (float): the reward weights will be clipped up to this value
+            reward_weight_projection (bool): whether project the weights to
```
emailweixu: reward_weight_normalization might be a more descriptive name than reward_weight_projection

alf/algorithms/lagrangian_reward_weight_algorithm.py
```diff
@@ -47,6 +54,9 @@ def __init__(self,
                  reward_thresholds,
                  optimizer,
                  init_weights=1.,
+                 max_weight=None,
+                 reward_weight_projection=True,
+                 lambda_transform=F.softplus,
```
emailweixu, line 61: comment for lambda_transform. Only F.softplus and torch.exp are supported.

alf/algorithms/lagrangian_reward_weight_algorithm.py
```diff
@@ -59,6 +69,9 @@ def __init__(self,
                 its init value is always used.
```
emailweixu: This not exactly true when reward_weight_projection is used.

emailweixu (changes requested):

alf/utils/tensor_utils.py
```diff
@@ -75,16 +75,22 @@ def tensor_extend(x, y):
...
     Returns:
-        Tensor: the extended tensor. Its shape is ``(x.shape[0]+1, x.shape[1:])``
+        Tensor: the extended tensor. Its shape is
+            ``(*x.shape[:dim], x.shape[dim]+1, *x.shape[dim+1:])``
     """
-    return torch.cat((x, torch.zeros(1, *x.shape[1:], dtype=x.dtype)))
+    if dim < 0:
+        return tensor_extend_zero(x, dim + x.ndim)
+
+    zero_shape = x.shape[:dim] + (1, ) + x.shape[dim + 1:]
```
emailweixu: Code can be simplified a little bit (line 88-89 can be removed):
```python
zero_shape = list(x.shape)
zero_shape[dim] = 1
```

alf/algorithms/lagrangian_reward_weight_algorithm.py
```diff
@@ -79,52 +95,167 @@ def __init__(self,
...
+        """
+        Args:
+            reward_spec (TensorSpec): a rank-1 tensor spec representing multi-dim
+                rewards.
+            reward_thresholds (list[float|None]): a list of floating numbers,
+                each representing a desired minimum reward threshold in expectation.
+                If any entry is None, then that reward weight won't be tuned and
+                its init value (after normalization if ``reward_weight_normalization=True``)
+                is always used. If a single floating value, it's shared by all
+                weights.
```
emailweixu: The comment for base class also needs to be updated.

emailweixu (approved):

PR #953, opened 2021-07-30
Adjust number of environments in DDP mode based on number of processes

breakds:
# Motivation

To make DDP as transparent to the user as possible, we should not expect the user to have to manually update a configuration to achieve parity between DDP mode and non DDP mode. More specifically, if single process has 64 environments in parallel, we would expect in dual process the per-process number of parallel environments to be reduced accordingly (to 32).

# Solution

This PR adds a function that adjust the configuration (by overriding). Currently it only adjust `num_parallel_environments` but there could potentially be more.

# Testing

Test run on dual process settings with `num_parallel_environments = 64`. Confirmed that each process actually starts 32.

### DIFF
--- a/alf/config_helpers.py
+++ b/alf/config_helpers.py
@@ -20,6 +20,7 @@ based TrainerConfig in this module.
 """
 
 import runpy
+import math
 
 from alf.algorithms.config import TrainerConfig
 from alf.algorithms.data_transformer import create_data_transformer
@@ -29,7 +30,8 @@ from alf.utils.common import set_random_seed
 
 __all__ = [
     'close_env', 'get_raw_observation_spec', 'get_observation_spec',
-    'get_action_spec', 'get_env', 'parse_config'
+    'get_action_spec', 'get_env', 'parse_config',
+    'adjust_config_by_multi_process_divider'
 ]
 
 _env = None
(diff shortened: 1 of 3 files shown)


alf/config_helpers.py
```diff
@@ -134,6 +168,10 @@ def parse_config(conf_file, conf_params):
     finally:
         _is_parsing = False
 
+    # TODO(breakds): Decouple this with configuration parsing after get_env is
+    # decoupled from parse_config.
+    adjust_config_by_multi_process_divider(multi_process_divider)
```
emailweixu, line 173: It might be too late to do this here. get_env() might already be called by the config file (e.g. though get_observation_spec()).

emailweixu (changes requested):

alf/config_helpers.py
breakds: Thanks. Yes `get_observationjspec()` can be called from `*_conf.py`. However I feel reluctant about moving it further to `get_env` because unlike `parse_conf`, `get_env()` is not a "once per process" function and we probably want to keep its argument clean.

How does the following idea appeal to you? We could add a global variable context that stores the rank and number of processes (read only), so that all functions can get access to it without having to pass them down along the stack. I hate globals too but these read-only global variables might help reduce the damage to readability.

emailweixu: That sounds a reasonable solution to me.

alf/utils/common.py
```diff
@@ -481,17 +481,21 @@ def parse_conf_file(conf_file):
...
+            We need it to adjust specific config to achieve number of process
+            parity, if the value is greater than 1.
     """
     if conf_file.endswith(".gin"):
         gin_params = getattr(flags.FLAGS, 'gin_param', None)
         gin.parse_config_files_and_bindings([conf_file], gin_params)
         ml_type = alf.get_config_value('TrainerConfig.ml_type')
         if ml_type == 'rl':
             # Create the global environment and initialize random seed
+            alf.adjust_config_by_multi_process_divider(multi_process_divider)
```
le-horizon, line 494: I don't think adjust_config_by_multi_process_divider function can handle the case of overriding gin config?
If that's correct, then we should raise error when gin config is used with DDP?

le-horizon (commented):
Thanks for the nice and clean PR, Break.

I think the PR is good as is, but it could be improved by creating separate config params for per process num_envs (or separate params for total_across_process_num_envs).  Overriding a config for two meanings (total vs per process) can be confusing in the long run, and make code less readable.

What do you think?

alf/utils/common.py
breakds: In #955, this has been moved to `get_env()` so both gin and py conf can enjoy this :)

breakds:
> Thanks for the nice and clean PR, Break.
> 
> I think the PR is good as is, but it could be improved by creating separate config params for per process num_envs (or separate params for total_across_process_num_envs). Overriding a config for two meanings (total vs per process) can be confusing in the long run, and make code less readable.
> 
> What do you think?

Thanks for taking time to review this, Le! In #955, I still decide to make this transparent because this seems nicer to the user.

breakds:
Closing this in favor of #955.

PR #954, opened 2021-07-30
[Small] Disable summary for non-master processes in DDP training mode

breakds:
# Description

Similar to only having process 0 (master process) writing checkpoint, we want to have only process 0 writing the summary. This is done by not enabling  summary unless the process rank is 0 (or -1) when setting up `PolicyTrainer`.

# Testing

Tested training in dual GPU settings with this PR. There are still 2 events file generated but only one is being written to.

### DIFF
--- a/alf/trainers/policy_trainer.py
+++ b/alf/trainers/policy_trainer.py
@@ -138,7 +138,10 @@ class Trainer(object):
     def train(self):
         """Perform training."""
         self._restore_checkpoint()
-        alf.summary.enable_summary()
+        if self._rank <= 0:
+            # Only enable summary for single process (rank is -1) training or
+            # the master process (rank is 0) in multi-process training.
+            alf.summary.enable_summary()
 
         self._checkpoint_requested = False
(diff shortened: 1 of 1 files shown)


alf/trainers/policy_trainer.py
```diff
@@ -138,7 +138,10 @@ def __init__(self, config: TrainerConfig, ddp_rank: int = -1):
     def train(self):
         """Perform training."""
         self._restore_checkpoint()
-        alf.summary.enable_summary()
+        if self._rank <= 0:
```
emailweixu, line 141: In Algorithm._train_experience(), enable_summary() is also called, so this will not be effective.

emailweixu (changes requested):

alf/trainers/policy_trainer.py
emailweixu: You may consider to change run_under_record_context so that _cond() always returns False.

breakds:
Based on the feedback, I will replace this PR with a new one.

PR #957, opened 2021-08-03
prevent Beta NaN grad

hnyu:
This change solved the gradient NaN issue in my experiments.

### DIFF
--- a/alf/utils/dist_utils.py
+++ b/alf/utils/dist_utils.py
@@ -413,8 +413,11 @@ class Beta(td.Beta):
         """
         self._concentration1 = concentration1
         self._concentration0 = concentration0
-        self._eps = eps
         super().__init__(concentration1, concentration0, validate_args)
+        if eps is None:
+            self._eps = torch.finfo(self._dirichlet.concentration.dtype).eps
+        else:
+            self._eps = float(eps)
 
     @property
     def concentration0(self):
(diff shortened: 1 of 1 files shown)


alf/utils/dist_utils.py
```diff
@@ -412,6 +424,19 @@ def concentration0(self):
...
+        """We override the original ``rsample()`` in order to clamp the output
+        to avoid `NaN` and `Inf` values in the gradients. See Pyro's
+        ``rsample()`` implementation in
+        `<https://docs.pyro.ai/en/dev/_modules/pyro/distributions/affine_beta.html#AffineBeta>`_.
+        """
+        x = super(Beta, self).rsample(sample_shape)
+        if self._eps is None:
+            eps = torch.finfo(x.dtype).eps
+        else:
+            eps = self._eps
```
emailweixu: move 434-437 to constructor

emailweixu (changes requested):

alf/utils/dist_utils.py
hnyu: > move 434-437 to constructor

I thought of doing this, but there is no way to get x.dtype in the constructor, unless the init accepts a dtype in advance.

emailweixu: It's self._dirichlet.concentration.dtype

emailweixu (approved):

PR #958, opened 2021-08-03
Convert ppo_cart_pole to python configuration

breakds:
# Motivation

Need a ppo example to play with to learn about the details on

1. Off-policy training in alf
2. Policy Gradient algorithm implementation

`ppo_cart_pole` is picked due to its simplicity. The original `.gin` is not working.

# Solution

Converted the `.gin` into `_conf.py`.

# Testing

ppo_cart_pole

Converges to `AverageReward = 200.0` almost immediately after training start.


breakds:
Need another pair of eyes to check whether the above makes sense. Note that the on-policy counter part `ac_breakout_conf.py` with its current configurations reaches 200 on `AverageReturn` almost immediately. I think the difference does make sense. Can someone confirm?

hnyu:
> Need another pair of eyes to check whether the above makes sense. Note that the on-policy counter part `ac_breakout_conf.py` with its current configurations reaches 200 on `AverageReturn` almost immediately. I think the difference does make sense. Can someone confirm?

It seems that the training curve hasn't converged yet while the eval curve is much better. You can try setting entropy_regularization=0. Also please remove the old .gin file.

alf/examples/ppo_cart_pole_conf.py
```diff
@@ -0,0 +1,60 @@
...
+
+# algorithm config
+alf.config('ActorDistributionNetwork', fc_layer_params=(100, ))
+alf.config('ValueNetwork', fc_layer_params=(100, ))
+
+alf.config(
+    'ActorCriticAlgorithm',
+    actor_network_ctor=ActorDistributionNetwork,
+    value_network_ctor=ValueNetwork,
+    optimizer=alf.optimizers.Adam(lr=1e-3))
```
hnyu: This should be AdamTF instead because the original code was TF.

alf/examples/ppo_cart_pole_conf.py
```diff
@@ -0,0 +1,60 @@
...
+
+alf.config(
+    'PPOLoss',
```
hnyu, line 40: Now PPOLoss is not used by PPOAlgorithm by default. 

https://github.com/HorizonRobotics/alf/blob/pytorch/alf/algorithms/ppo_algorithm.py#L34

hnyu: Probably should fix PPOAlgorithm so that by default it uses PPOLoss.

emailweixu:
> Need another pair of eyes to check whether the above makes sense. Note that the on-policy counter part `ac_breakout_conf.py` with its current configurations reaches 200 on `AverageReturn` almost immediately. I think the difference does make sense. Can someone confirm?

Why don't you run ppo_cart_pole.gin to get its curve first? In general, ppo should be more sample efficient than ac.

alf/examples/ppo_cart_pole_conf.py
breakds: Sure, updated.

emailweixu (commented):
Can you also upload the training curve?

breakds (commented):
> > Need another pair of eyes to check whether the above makes sense. Note that the on-policy counter part `ac_breakout_conf.py` with its current configurations reaches 200 on `AverageReturn` almost immediately. I think the difference does make sense. Can someone confirm?
> 
> Why don't you run ppo_cart_pole.gin to get its curve first? In general, ppo should be more sample efficient than ac.

The original `.gin` file is a bit out of date. However with Haonan's hint, the converted `_conf.py` now converges really quickly.

breakds:
> Can you also upload the training curve?

Yep, already updated in the description.

hnyu:
Can you also remove ppo.gin?

breakds:
> Can you also remove ppo.gin?

I can't do that yet because there are other `.gin` files importing it at this moment. Once they are cleaned up and converted, I will.

alf/examples/ppo_cart_pole_conf.py
```diff
@@ -0,0 +1,64 @@
...
+    mini_batch_size=128,
+    num_updates_per_train_iter=4,
+    num_iterations=200,
+    num_checkpoints=5,
+    evaluate=True,
+    eval_interval=50,
+    debug_summaries=False,
+    summary_interval=5)
+
+alf.config('ReplayBuffer', max_length=2048)
```
emailweixu: This one can be removed. It's not used.

emailweixu (commented):
> > Can you also upload the training curve?
> 
> Yep, already updated in the description.

I mean uploading it to repo, like examples/ac_cart_pole.png.

hnyu:
> > Can you also upload the training curve?
> 
> Yep, already updated in the description.

Maybe check in the curve figure under examples?

emailweixu:
> > > Can you also upload the training curve?
> > 
> > 
> > Yep, already updated in the description.
> 
> I mean uploading it to repo, like examples/ac_cart_pole.png.

And use figure from Metrics_vs_EnvironmentSteps, as we typically plot the metrics against environment steps.

alf/examples/ppo_cart_pole_conf.py
breakds: Fixed. Thank you!

breakds:
> > > Can you also upload the training curve?
> > 
> > 
> > Yep, already updated in the description.
> 
> Maybe check in the curve figure under examples?



> > > > Can you also upload the training curve?
> > > 
> > > 
> > > Yep, already updated in the description.
> > 
> > 
> > I mean uploading it to repo, like examples/ac_cart_pole.png.
> 
> And use figure from Metrics_vs_EnvironmentSteps, as we typically plot the metrics against environment steps.

Checked in the graph accoordingly.

hnyu (approved):

emailweixu (approved):

PR #960, opened 2021-08-04
revised chap2 and added chap3

hnyu:
Reorganized chap2 according to feedbacks. 
https://alf-fork.readthedocs.io/en/latest/tutorial/understanding_ALF_via_the_minimal_working_example.html

Added chap3 on algorithm interfaces
https://alf-fork.readthedocs.io/en/latest/tutorial/algorithm_interfaces.html


breakds:
Chapter 3 rst file is not included - did I miss anything?

hnyu:
> Chapter 3 rst file is not included - did I miss anything?

It's  docs/tutorial/algorithm_interfaces.rst . Might have been folded by Github.

docs/tutorial/algorithm_interfaces.rst
```diff
@@ -1,2 +1,464 @@
...
+algorithms could consist of an *algorithm hierarchy* so that the calling of
+their interfaces of the same name is performed in a nested way. We leave this
+advantaged usage to a later chapter :doc:`./customize_algorithms`.
+
+* ``inputs`` denotes the observations/states based on which
+  an algorithm's interface does computation at every step.
+* ``output`` is specially used to represent a step function's main output, e.g.,
+  action of an RL algorithm, latent embedding of an encoder algorithm, or goal
+  representation of a goal proposal algorithm.
+* ``state`` denotes episodic memory (e.g., RNN state) that is automatically reset
```
emailweixu: episodic memory usually refers to long term memory (e.g. replay buffer). `state` for algorithm is more like working memory.

breakds:
> > Chapter 3 rst file is not included - did I miss anything?
> 
> It's docs/tutorial/algorithm_interfaces.rst . Might have been folded by Github.

Found it. My bad. Thanks!

docs/tutorial/algorithm_interfaces.rst
```diff
@@ -1,2 +1,464 @@
...
+Now we will take a look at how different algorithm interfaces interact with each
+other in the on-policy/off-policy pipelines. We will try modifying some
+algorithm interfaces of the previous example :doc:`./a_minimal_working_example`,
+for an illustration purpose. Then we introduce a simple off-policy training
+example and again explore its algorithm interfaces.
+
+Interface Overview
+------------------
+
+.. image:: images/algorithm_interfaces.png
```
breakds, line 14: In big picture, the procedure is mainly divided into `unroll` stage and `update` stage. It would be good for the reader to find a corresponding between the steps here and the two stages.

To be more specific, I think there is a gap for the reader and first-time user between `unroll`/`update` and the function name `train_from_unroll`, `train_from_replay_buffer`.

breakds: Here in the graph, `evaluation` is isolated. Also there are no text talking about it, which might be a bit confusing, especially when it is said "`predict_step` is called at every time step"

docs/tutorial/algorithm_interfaces.rst
```diff
@@ -1,2 +1,464 @@
...
+:meth:`~.AlgorithmInterface.calc_loss`                  :math:`[T,B,\ldots]`                         ``info``                                 ``loss_info``                   On/Off-policy
+:meth:`~.AlgorithmInterface.after_update`               :math:`[T,B,\ldots]`                         ``root_inputs``, ``info``                ``None``                        On/Off-policy
+:meth:`~.AlgorithmInterface.after_train_iter`           :math:`[T,B,\ldots]`                         ``root_inputs``, ``rollout_info``        ``None``                        On/Off-policy
+======================================================  ===========================================  =======================================  =============================== ===============
+
+:math:`T` is :attr:`.TrainerConfig.unroll_length` for on-policy algorithms and
+:attr:`.TrainerConfig.mini_batch_length` for off-policy algorithms. :math:`B` is
+:attr:`.create_environment.num_parallel_environments` for on-policy algorithms and
+:attr:`.TrainerConfig.mini_batch_size` for off-policy algorithms. We can see that
+the first three interfaces are called at every time step, while the last three
```
breakds: Nit: as a reader I found it slightly easier to read if we explicitly list what are the "first three interfaces" here

docs/tutorial/algorithm_interfaces.rst
```diff
@@ -1,2 +1,464 @@
...
+:meth:`~.AlgorithmInterface.calc_loss`                  :math:`[T,B,\ldots]`                         ``info``                                 ``loss_info``                   On/Off-policy
+:meth:`~.AlgorithmInterface.after_update`               :math:`[T,B,\ldots]`                         ``root_inputs``, ``info``                ``None``                        On/Off-policy
+:meth:`~.AlgorithmInterface.after_train_iter`           :math:`[T,B,\ldots]`                         ``root_inputs``, ``rollout_info``        ``None``                        On/Off-policy
+======================================================  ===========================================  =======================================  =============================== ===============
+
+:math:`T` is :attr:`.TrainerConfig.unroll_length` for on-policy algorithms and
+:attr:`.TrainerConfig.mini_batch_length` for off-policy algorithms. :math:`B` is
+:attr:`.create_environment.num_parallel_environments` for on-policy algorithms and
+:attr:`.TrainerConfig.mini_batch_size` for off-policy algorithms. We can see that
```
emailweixu: For rollout_step, B is always create_environment.num_parallel_environments

docs/tutorial/algorithm_interfaces.rst
```diff
@@ -1,2 +1,464 @@
...
+the first three interfaces are called at every time step, while the last three
+are called once for multiple time steps. The off-policy pipeline has one
+additional :meth:`~.AlgorithmInterface.train_step` interface compared to on-policy.
```
emailweixu: Need to explain what these functions do.

docs/tutorial/algorithm_interfaces.rst
```diff
@@ -1,2 +1,464 @@
...
+    :align: center
+    :alt: interaction between algorithm interfaces
+
+
+There are several key interfaces (highlighted in the figure above) that play
+important roles in assembling a training pipeline of an ALF algorithm, and their
+signatures are summarized below.
+
+======================================================  ===========================================  =======================================  =============================== ===============
+Interface                                               Tensor shape                                 Input                                    Output                          Which pipeline?
```
breakds: Not sure whether the terminologies are fixed, but those 6 look to me more like `hook`s, or to be more specifically lifecycle callbacks or lifecycle hooks. Lifecycle hooks are very commonly used in algorithmic loops (like our training iteration loop), for example in [k8s](https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/).

My mental model of Alf training pipeline is that 

1. The training pipeline loop: a bunch of sequentially training iterations
2. The lifecycle of each training iteration is divided into two major stages, `unroll` and `update`
3. Both the `unroll` and `update` stages have sub stages
4. At the boundaries of stages and sub-stages, lifecycle stages are called

So there is a hierarchy of "pipeline -> iteration -> iteration stage -> sub (iteration) stage".

And the (general) users use alf by injecting their logic at lifecycle hooks to implement their (training) algorithms.

docs/tutorial/algorithm_interfaces.rst
```diff
@@ -1,2 +1,464 @@
...
+======================================================  ===========================================  =======================================  =============================== ===============
+:meth:`~.AlgorithmInterface.predict_step`               :math:`[1,\ldots]`                           ``inputs``, ``state``                    ``output``, ``state``, ``info`` On/Off-policy
```
breakds: Not related to this tutorial, but I think one potential approach to lower the learning cost of the first time user and also the cost of composing tutorial is to unify the names of the lifecycle hooks, i.e. `xxx_step` or `xxx_hook` or `on_xxx_hook`.

docs/tutorial/algorithm_interfaces.rst
breakds: Might be clearer if we can explain whether the `Tensor Shape` here in the table stands for the input, the output or something else.

docs/tutorial/algorithm_interfaces.rst
```diff
@@ -1,2 +1,464 @@
...
+.. note::
+
+    If :attr:`.TrainerConfig.temporally_independent_train_step=True`, we will call
+    :meth:`~.AlgorithmInterface.train_step` on steps along a temporal extent in
+    parallel. So the tensor shape is :math:`[TB,\ldots]`. This speeds up the
+    training forward step and the result will be reshaped to
+    :math:`[T,B,\ldots]` when fed to :meth:`~.AlgorithmInterface.calc_loss`.
+
+In most cases, we only need to override these interfaces while following their
+input/output signatures. Potentially we are able to accomplish many things through
```
breakds: Rephrase suggestion - Just one sentence like below would work:

> Most of the time, all we have to do is overriding these interfaces while adhering to their input/output signatures.

docs/tutorial/algorithm_interfaces.rst
```diff
@@ -1,2 +1,464 @@
...
+them.
+
+Argument conventions
+--------------------
+
+Here we assume that there is only a single algorithm in the pipeline. In fact, multiple
+algorithms could consist of an *algorithm hierarchy* so that the calling of
+their interfaces of the same name is performed in a nested way. We leave this
+advantaged usage to a later chapter :doc:`./customize_algorithms`.
```
breakds: I found it debatable whether to talk about hierarchical algorithms here, as it might increase the mental burden of the inexperienced audience. Can we move this to a later place?

docs/tutorial/algorithm_interfaces.rst
```diff
@@ -1,2 +1,464 @@
...
+  on the context, it might have two alias names: ``train_info`` and ``rollout_info``.
+  ``train_info`` usually appears in the context where ``info`` is used for computing losses
+  and gradients. ``train_info`` is output by either an on-policy algorithm's
+  :meth:`~.AlgorithmInterface.rollout_step` or an off-policy algorithm's
+  :meth:`~.AlgorithmInterface.train_step`. ``rollout_info`` exists in the
+  context of off-policy training, where the ``info`` of :meth:`~.AlgorithmInterface.rollout_step`
+  is used for reference (e.g., it contains rollout action distributions for computing the
+  probability ratio in PPO).
+* ``loss_info`` should be an instance of :class:`~.data_structures.LossInfo`
+  that can be correctly processed by :meth:`~.Algorithm.update_with_gradient`.
```
breakds: `update_with_gradient` hasn't been in the context yet. Might be good to rephrase so that we can explain it without relying mostly on knowledge of `update_with_gradient`.

docs/tutorial/algorithm_interfaces.rst
```diff
@@ -1,2 +1,464 @@
...
+with the field ``ac`` storing the :class:`.ActorCriticAlgorithm` info and ``zeros``
+storing a zero tensor that we're going to create during rollout. Then we derive
+from :class:`.ActorCriticAlgorithm` for our new algorithm class and override its
+:meth:`~.ActorCriticAlgorithm.rollout_step`:
+
+.. code-block:: python
+
+  class MyACAlgorithm(ActorCriticAlgorithm):
+    def rollout_step(self, inputs, state):
+        alg_step = super(MyACAlgorithm, self).rollout_step(inputs, state)
```
breakds: AFAIK, in python 3 this can be simplified as `super().rollout_step()`, which is shorter and much more readable (https://docs.python.org/3/library/functions.html#super)

docs/tutorial/algorithm_interfaces.rst
```diff
@@ -1,2 +1,464 @@
...
+On-policy interfaces
+--------------------
+
+We modify the minimal working example by introducing a new actor-critic algorithm
+that wraps :class:`ActorCriticAlgorithm` and does an additional trivial job of
+inspecting tensor shapes. First we define a new data structure
+
+.. code-block:: python
+
+    MyACInfo = namedtuple("MyACInfo", ["ac", "zeros"])
```
breakds, line 202: IMHO good implementation approaches are usually bottom-up and good tutorials goes top-down. I would suggest stating the goals first instead of the `MyACInfo` definition which is more like implementation details. This is because the user will most likely ask "why do I need `MyACInfo`" after reading this line.

The order that seems a bit more natural is

1. Explain what does "inspecting the shape" means
2. In order to do that, we need to inject code in `rollout_step`, `calc_loss` hooks
3. Explain why we need new data structure to pass information around

docs/tutorial/algorithm_interfaces.rst
breakds: Potential questions that might be asked here because of the audience lacking the context: 

1. What is the type of `alg_step`? If such type is not explained before, can we explain it here?
2. What does the parent `rollout_step` do?

Not saying I do not know the answer to them - but trying to stand from audience's perspective.

breakds (commented):
As a new user to Alf I am learning a lot from the tutorial. What you have done is extremely beneficial to us. Thank you @hnyu for the great work!

In the first part of the review on Chapter 3, I summarized what I have thought while reading through it. Still reading.

docs/tutorial/algorithm_interfaces.rst
```diff
@@ -1,2 +1,464 @@
...
+
+Off-policy interfaces
+---------------------
+
+As another example, we override :class:`~.sac_algorithm.SacAlgorithm` for the same
+experimentation. First we set up some basic training configurations:
+
+.. code-block:: python
+
+    alf.config('create_environment', num_parallel_environments=10)
```
breakds, line 297: Though it is implicitly set to default, but might still be good to explicitly write the target task/environment here? Otherwise the audience may ask.

docs/tutorial/algorithm_interfaces.rst
```diff
@@ -1,2 +1,464 @@
...
+       num_iterations=1)
+
+And define a new class ``MySacAlgorithm`` with a new ``train_step()``
+
+.. code-block:: python
+
+    MySacInfo = namedtuple("MySacInfo", ["sac", "zeros"])
+
+    class MySacAlgorithm(SacAlgorithm):
+        def train_step(self, inputs, state, rollout_info: MySacInfo):
```
breakds, line 313: Looking around and did not find explanations on what does `train_step` hook do. The audience might be able to find that in the documentation, but it's probably better to explicitly explain it here as the target audiences are first-timer users, and we would like to lower the barrier for them as much as possible. 

Such explanation might need to include "why does `train_step` requires `rollout_info` as an input?"

docs/tutorial/algorithm_interfaces.rst
```diff
@@ -1,2 +1,464 @@
...
+            alg_step = super(MySacAlgorithm, self).train_step(
+                inputs, state, rollout_info.sac)
+            print("train_step rollout zeros:  ", rollout_info.zeros.shape)
+            train_zeros = torch.zeros_like(alg_step.output, dtype=torch.uint8)
+            print("train_step train ones:" , train_zeros.shape)
```
breakds, line 318: Nit: s/ones/zeros/ ? I know ones probably also work here semantically - but "zeros" seems less ambiguous.

docs/tutorial/algorithm_interfaces.rst
```diff
@@ -1,2 +1,464 @@
...
+
+Recall that for this example, we have ``TrainerConfig.unroll_length=8`` and
+``create_environment.num_parallel_environments=30``. Thus the tensor shapes match our
+expectation.
+
+
+Off-policy interfaces
+---------------------
+
+As another example, we override :class:`~.sac_algorithm.SacAlgorithm` for the same
```
breakds, line 292: One question that the audience might ask about this experiment is that

> Why do we want to pass around a full-zero tensor, if we only care about its shape? Is this for demonstration purpose only or we actually need this tensor?

docs/tutorial/algorithm_interfaces.rst
hnyu: Good point. I will explain their correspondences.

docs/tutorial/algorithm_interfaces.rst
hnyu: I'll move this paragraph to the beginning right under the table so that it's clearer.

docs/tutorial/algorithm_interfaces.rst
hnyu: Your understanding is correct and I do think "hook" is a more appropriate term regarding software. But I'm not sure if we want to expose too many details of injecting their logic into the pipeline, which might be even more confusing to them. 

As for the function name, I'm OK to change from ``AlgorithmInterface`` to something like ``AlgorithmHook``. I think we can discuss with Wei.

docs/tutorial/algorithm_interfaces.rst
hnyu: I think the reason why we have "step" in some hooks but not in others is that "predict_step", "rollout_step" and "train_step" are indeed called on a per-step basis, i.e., t, t+1, t+2, ... But for "calc_loss", "after_udpate", "after_train_iter" or other hooks they are not called on a per-step basis. So "step" here actually means the time step concept in RL.

docs/tutorial/algorithm_interfaces.rst
hnyu: It's for all I/O tensors, with some rare exception that you might return a different shape (but it's dangerous to do so). I've made the title clearer.

docs/tutorial/understanding_ALF_via_the_minimal_working_example.rst
```diff
@@ -401,23 +428,6 @@ In summary, different components on a ALF pipeline are connected by using
...
-    alf/networks/
-    alf/layers.py
-    alf/environments/
-
-A user can easily experiment them via the conf file. In the example conf,
-:class:`.ActorCriticAlgorithm`, :class:`.ActorDistributionNetwork`, and
-:class:`.NormalProjectionNetwork` are representatives.
-
 Summary
 -------
```
runjerry, line 433: Minor typo in the first sentence of the last paragraph: It might still be unclear to a user

docs/tutorial/algorithm_interfaces.rst
hnyu: I've simplified this paragraph.

docs/tutorial/algorithm_interfaces.rst
hnyu: I've added a sentence for a brief explanation about update_with_gradient.

docs/tutorial/algorithm_interfaces.rst
hnyu: Updated

docs/tutorial/algorithm_interfaces.rst
breakds: Got it. That totally make sense - worth explaining in the tutorial as well?

docs/tutorial/algorithm_interfaces.rst
hnyu: Good point. I've added more explanations and re-arranged the presentation order.

docs/tutorial/algorithm_interfaces.rst
hnyu: I will explain AlgStep earlier in the chapter. So for what the parent `rollout_step` does, I don't think it's relevant to this example. But I can just mention something like "we still let the parent does what it should do".

docs/tutorial/algorithm_interfaces.rst
hnyu: I don't think this example has anything to do with which specific task? No matter what task is picked, the demonstration result will be the same.

docs/tutorial/algorithm_interfaces.rst
hnyu: I've added explanations in "Argument conventions" section.

docs/tutorial/understanding_ALF_via_the_minimal_working_example.rst
hnyu: updated

docs/tutorial/algorithm_interfaces.rst
hnyu: Sure, I've explained their difference in temporal logic. But I won't talk about their naming logic in the tutorial.

hnyu:
@breakds Thanks for your detailed comments! They help a lot. I've updated this PR.

breakds (approved):

docs/tutorial/algorithm_interfaces.rst
hnyu: changed to working memory

docs/tutorial/algorithm_interfaces.rst
hnyu: updated

docs/tutorial/algorithm_interfaces.rst
hnyu: Added a section for this

PR #963, opened 2021-08-05
Fix the default value of state for NestedEntropyTargetAlgorithm

emailweixu:
state is not used by NestedEntropyTargetAlgorithm and it is called without providing state in Agent. So it should have a default value for state.

### DIFF
--- a/alf/algorithms/algorithm.py
+++ b/alf/algorithms/algorithm.py
@@ -1020,7 +1020,12 @@ class Algorithm(AlgorithmInterface):
 
         if isinstance(loss_info.loss, torch.Tensor):
             loss = weight * loss_info.loss
-            loss.backward()
+            # In theory, there might be unfinished kernel before backward is called
+            # and and after backward() returns and torch.cuda.synchronize() should
+            # be called to make sure we only measure the computation time for backward.
+            # In practice, we found that the measured time is almost unchanged.
+            with record_time("time/backward"):
+                loss.backward()
 
         all_params = []
         for optimizer in optimizers:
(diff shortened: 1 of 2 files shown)


alf/algorithms/algorithm.py
```diff
@@ -1020,7 +1020,12 @@ def update_with_gradient(self,
 
         if isinstance(loss_info.loss, torch.Tensor):
             loss = weight * loss_info.loss
-            loss.backward()
+            # In theory, there might be unfinished kernel before backward is called
+            # and and after backward() returns and torch.cuda.synchronize() should
+            # be called to make sure we only measure the computation time for backward.
+            # In practice, we found that the measured time is almost unchanged.
```
hnyu: Not sure what this paragraph means?

hnyu:
EntropyTargetAlgorithm has no issue because its predict_step and rollout_step already have state=None. It is NestedEntropyTargetAlgorithm that this PR fixes.

alf/algorithms/algorithm.py
emailweixu: In general, there is no guarantee that its computation is finished when a torch function returns. See https://pytorch.org/docs/stable/notes/cuda.html#asynchronous-execution

emailweixu:
> EntropyTargetAlgorithm has no issue because its predict_step and rollout_step already have state=None. It is NestedEntropyTargetAlgorithm that this PR fixes.

Changed PR title and description.

alf/algorithms/algorithm.py
hnyu: OK, I think it's a little difficult to parse the first sentence. Maybe 

"In theory, there might be unfinished kernel before backward being called, and after backward returns, torch.cuda.synchronize() should be called to make sure..."

hnyu (approved):

PR #969, opened 2021-08-11
reproduce TasacValue performance on carla

hnyu:
Found out that for TaacL, directly generating the action diff will be much better and basically reproduces our previous `TasacValueAlgorithm`'s result. I've updated taac_algorithm.py to include an option `inverse_mode` for both `TaacLAlgorithm` and `TaacQAlgorithm`. 

The current curve is
image

with weather turned on. I'll update curve once finished.

### DIFF
--- a/alf/algorithms/taac_algorithm.py
+++ b/alf/algorithms/taac_algorithm.py
@@ -58,8 +58,9 @@ TaacActorInfo = namedtuple(
 
 TaacInfo = namedtuple(
     "TaacInfo", [
-        "reward", "step_type", "tau", "discount", "action_distribution",
-        "rollout_b", "b", "actor", "critic", "alpha", "repeats"
+        "reward", "step_type", "tau", "prev_tau", "discount",
+        "action_distribution", "rollout_b", "b", "actor", "critic", "alpha",
+        "repeats"
     ],
     default_value=())
 
@@ -323,7 +324,7 @@ class TaacAlgorithmBase(OffPolicyAlgorithm):
                 'TrainerConfig.epsilon_greedy')
         self._epsilon_greedy = epsilon_greedy
 
-        tau_spec, critic_networks, actor_network = self._make_networks(
+        self._tau_spec, critic_networks, actor_network = self._make_networks(
             observation_spec, action_spec, reward_spec, actor_network_cls,
             critic_network_cls)
 
@@ -331,7 +332,8 @@ class TaacAlgorithmBase(OffPolicyAlgorithm):
                      nn.Parameter(torch.zeros(())))
 
         train_state_spec = TaacState(
-            tau=tau_spec, repeats=TensorSpec(shape=(), dtype=torch.int64))
+            tau=self._tau_spec,
+            repeats=TensorSpec(shape=(), dtype=torch.int64))
         super().__init__(
             observation_spec,
             action_spec,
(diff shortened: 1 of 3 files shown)


alf/examples/sac_carla_conf.py
```diff
@@ -60,17 +63,50 @@
...
     # 1000 second day length means 4.5 days in replay buffer of 90000 length
     day_length=1000,
     max_weather_length=500,
 )
 
-alf.define_config('taac', False)
+
+class NormalizedActionWrapper(AlfEnvironmentBaseWrapper):
+    """An AlfEnvironment wrapper that converts the last action dim ('reverse') from
+    [0,1] to [-1,1].
```
hnyu: This is because TaacL and TaacQ both require action to be in [-1,1]

alf/algorithms/taac_algorithm.py
```diff
@@ -411,10 +411,7 @@ def _make_parallel(net):
 
         tau_spec = nest.map_structure(lambda m: action_spec if m else (),
                                       tau_mask)
-        tau_embedding = nest.map_structure(
-            lambda _: torch.nn.Sequential(
-                alf.layers.FC(action_spec.numel, observation_spec.numel)),
-            tau_spec)
+        tau_embedding = nest.map_structure(lambda _: None, tau_spec)
```
emailweixu: This change makes the algorithm behave differently. Does it change the experiment results?
And if all None, there is no need to set input_preprocessors below.

alf/examples/sac_carla_conf.py
```diff
@@ -60,17 +63,50 @@
...
+            dtype=env.action_spec().dtype,
+            minimum=-1.,
+            maximum=1.)
+        self._time_step_spec = env.time_step_spec()._replace(
+            prev_action=self._action_spec)
+
+    def _step(self, action):
+        # in: [-1,1] -> [0,1]
+        action[..., -1] = (action[..., -1] + 1.) / 2.
+        return self._env.step(action)
```
emailweixu: prev_action of the returned TimeStep is not correctly handled

alf/examples/sac_carla_conf.py
emailweixu: This requirement needs to be checked in TaacAlgorithm

hnyu: > This requirement needs to be checked in TaacAlgorithm

TaacAlgorithm is fine with any action spec. Only TaacL and TaacQ needs this requirement and they already check this.

alf/examples/sac_carla_conf.py
hnyu: > prev_action of the returned TimeStep is not correctly handled

Correct. I think this will affect the training of TaacL and TaacQ. (Taac doesn't require action_spec to be [-1,1] and my paper experiments should not be affected). I'll rerun the current reproducing job.

alf/algorithms/taac_algorithm.py
hnyu: Just in case, I've reverted this change to be consistent with the experiments in the paper.

alf/examples/sac_carla_conf.py
hnyu: > > This requirement needs to be checked in TaacAlgorithm
> 
> TaacAlgorithm is fine with any action spec. Only TaacL and TaacQ needs this requirement and they already assert this.

hnyu:
It turns out that the secret of TasacValue's success is because of generating 'reverse' signal in [-1, 1] even though the simulator checks >0.5. This puts a prior of inactivating 'reverse' during training. 

https://github.com/hnyu/action_repetition/blob/master/alf/algorithms/tasac_algorithm.py#L267

I also applied this setting to TaacL, and get the figure below so far:

image

which is much better than before.

alf/algorithms/taac_algorithm.py
```diff
@@ -261,6 +262,7 @@ def __init__(self,
                  critic_optimizer=None,
                  alpha_optimizer=None,
                  debug_summaries=False,
+                 randomize_first_state_tau=False,
```
hnyu, line 265: As I checked, the paper's version didn't randomize. However, it was used by TasacValueAlgorithm running on Carla.

emailweixu (approved):

PR #975, opened 2021-08-18
Update the lock file to use the up-to-date nix devenv

breakds:
# Motivation

This is to update the [nix development environment](https://github.com/HorizonRobotics/alf-nix-devenv) to its latest version. This will bring in the change of 

1. `cnest` from Pypi
2. `pytorchviz` (see [here](https://github.com/szagoruyko/pytorchviz)) for visualizing the structure of network

Note that the latter is only for development time usage, similar to tools like `clang-format`.

# Solution

Update the `flake.lock` file so that when activate, the nix based development environment will now fetch the one with the specified hash. It will not have any effect of non nix-based development.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -2,22 +2,46 @@
   "nodes": {
     "alf-devenv": {
       "inputs": {
+        "ml-pkgs": "ml-pkgs",
         "nixpkgs": [
           "nixpkgs"
         ],
         "utils": "utils"
       },
       "locked": {
-        "lastModified": 1626997680,
-        "narHash": "sha256-osKD7YhkFy2VfYOqcFQzrsT0MmATUdp7VGIfr/0B76s=",
+        "lastModified": 1629241250,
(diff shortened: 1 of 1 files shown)


runjerry:
Should we consider also bringing pytorchviz to alf main branch?

breakds:
> Should we consider also bringing pytorchviz to alf main branch?

Yes, but I am not sure whether `setup.py` or `requirements.txt` is the right place for development dependencies (i.e. the dependencies that **ARE NOT** needed for runtime), because they do not distinguish between those based on my limited knowledge. `pyproject.toml` on the other hand support this.

Not having `pytorchviz` will not damage other's development environment or so because it is only used in development time and such code will not be checked in.

Even if we want to add that, I think that should be a separate PR :)

runjerry:
Thanks for the explanations! It makes sense to me.

runjerry (approved):

Issue #965, opened 2021-08-06
Implement phasic policy gradient (PPG) in alf


breakds:
# Motivation

PPG is claimed to be significantly better in terms of sample efficiency compared to PPO, and we heard that it has become the default algorithm in OpenAI. Therefore, it would be great if we can have an implementation of PPG in alf so that it can be used as baseline, or combined with other techniques/algorithms to create new solutions for future experiments.

# Tasks

- [x] Study and understand in detail the architecture of PPG
- [x] Implement PPG based on its official (open source) [implementation](https://github.com/openai/phasic-policy-gradient).
- [x] Tune it on cart pole
- [ ] Tune it on procgen, which is the benchmark released by OpenAI, where the original implementation is tested against

breakds:
A few design choices.

If the main network and the auxiliary network are completely separate, the graphic card memory consumption will almost double. Per the paper and OpenAI's official implementation, a compromise (at the cost slightly worse performance) is proposed to **share the encoding network, but detach the gradient update for the auxiliary head**. We can put an option in `PPGAlgorithm` so that the user can choose.

breakds:
![train_iter_off_policy](https://user-images.githubusercontent.com/1111035/128581804-b4496f29-b944-4302-84fb-fef678d6de93.png)

PPG requires running **two different kinds of behavior in the `_update()` loop** above. As discussed with @emailweixu  and @hnyu, it is probably better to override the `train_step` and `calc_loss` of `PPGAlgorithm` to inject the "loop ID" in `train_info`, so that we can decide what kind of behavior we want based on the loop ID.

Also it seems that `preprocess_experience()` looks like the right place to inject such loop ID and set it to `0`.

breakds:
Went through [OpenAI's Implementation](https://github.com/openai/phasic-policy-gradient) in more detail, I created the following diagram to show the network architecture and loss function construction. 

There are always **3 heads** in the network architecture and 2 different loss functions. One of the loss functions is for normal training iteration where roll-outs are performed, and the other one is for aux phase updates (I call it extra iteration so that the "aux" does not confuse with the aux network).

![ppg_dual_network](https://user-images.githubusercontent.com/1111035/129265833-19c85a99-8fed-498d-9d3f-591074ff0e7e.png)

breakds:
[...]

PR #977, opened 2021-08-18
Phasic Policy Gradient [Part 1]: Disjoint Policy Value Network for PPG

breakds:
# Motivation

This is part of the effort to implement #965.

More specifically, PPG requires a disjoint network architecture to operates its two separate phases. Such disjoint network has two architecture configuration, "shared" vs "dual". Both network are graphically illustrated in #965, but the "dual" architecture is also depicted below for convenience.

dual

There is no simple way to re-use the existing networks without hurting too much on readability. Therefore, we need a new implementation.
 
# Solution

Implemented the disjoint network with the help of the network containers. There are 3 branches for the policy head, aux value head and true value head respectively.

# Testing

1. Added unit tests as a simple verification
2. Tested together with the PPG algorithm on cart pole and it converges to 200.0 almost immediately.

Plan to do more tests after adding more PPG-based examples.

# Note

Plan to have this reviewed and checked-in while working on cleaning up the other parts, adding documentation and more tests. If something needs to be updated after the submission of this PR, we can still do that incrementally.

### DIFF
--- a/alf/algorithms/ppg/disjoint_policy_value_network.py
+++ b/alf/algorithms/ppg/disjoint_policy_value_network.py
@@ -18,9 +18,7 @@ import torch
 import alf
 from alf.tensor_specs import TensorSpec
 from alf.data_structures import namedtuple
-from .network import Network
-from .projection_networks import NormalProjectionNetwork, CategoricalProjectionNetwork
-from .encoding_networks import EncodingNetwork
+from alf.networks import Network, NormalProjectionNetwork, CategoricalProjectionNetwork, EncodingNetwork
 
 
 def _create_projection_net_based_on_action_spec(
@@ -126,7 +124,7 @@ class DisjointPolicyValueNetwork(Network):
                 network is shared between the value component and the policy
                 component, 1 or 2 encoding network will be created using this
                 constructor.
-
+        
             is_sharing_encoder (bool): When set to true, the encoding network is
                 shared between the value and the policy component, resulting in
                 a "shared" architecture disjoint network. When set to false, the
--- a/alf/algorithms/ppg/disjoint_policy_value_network.py
+++ b/alf/algorithms/ppg/disjoint_policy_value_network.py
@@ -58,7 +58,6 @@ def _create_projection_net_based_on_action_spec(
     return alf.nest.map_structure(_create_individually, action_spec)
 
 
-# TODO(breakds): Add its recurrent network counterpart
 @alf.configurable
 class DisjointPolicyValueNetwork(Network):
     """A composite network with a policy component and a value component.
@@ -124,7 +123,7 @@ class DisjointPolicyValueNetwork(Network):
                 network is shared between the value component and the policy
                 component, 1 or 2 encoding network will be created using this
                 constructor.
-        
+
             is_sharing_encoder (bool): When set to true, the encoding network is
                 shared between the value and the policy component, resulting in
                 a "shared" architecture disjoint network. When set to false, the
@@ -168,11 +167,6 @@ class DisjointPolicyValueNetwork(Network):
             continuous_projection_net_ctor=continuous_projection_net_ctor,
             input_size=self._actor_encoder.output_spec.shape[0],
             action_spec=action_spec)
-        # In the case when the projection net is multi-fold (because actor
-        # spec is a nest), force picking up the parameters inside those
-        # project networks into this DisjointPolicyValueNetwork instance.
-        self._projection_net_module_list = torch.nn.ModuleList(
-            alf.nest.flatten(self._policy_head))
 
         # +------------------------------------------+
         # | Step 3: Value head of the aux branch     |
@@ -190,43 +184,43 @@ class DisjointPolicyValueNetwork(Network):
         # | Step 4: Assemble network + value head    |
         # +------------------------------------------+
 
-        if is_sharing_encoder:
-            # Use the same encoder, but the encoder is DETACHED.
-            self._value_head = alf.nn.Sequential(
-                alf.layers.Detach(),
-                alf.layers.FC(
-                    input_size=self._actor_encoder.output_spec.shape[0],
-                    output_size=1),
-                alf.layers.Reshape(shape=()),
-                input_tensor_spec=self._actor_encoder.output_spec)
+        encoder_output_size = self._actor_encoder.output_spec.shape[0]
 
-            all_branches = ((self._policy_head, self._aux_head),
-                            self._value_head)
+        if is_sharing_encoder:
             self._composition = alf.nn.Sequential(
                 self._actor_encoder,
-                alf.nn.Branch(all_branches, name='AllBranches'))
+                alf.nn.Branch(
+                    self._policy_head,
+                    alf.nn.Sequential(
+                        # Use the same encoder, but the encoder is DETACHED.
+                        alf.layers.Detach(),
+                        alf.layers.FC(
+                            input_size=encoder_output_size, output_size=1),
+                        alf.layers.Reshape(shape=()),
+                        input_tensor_spec=self._actor_encoder.output_spec),
+                    self._aux_head))
         else:
-            policy_component = alf.nn.Sequential(
-                self._actor_encoder,
-                alf.nn.Branch((self._policy_head, self._aux_head),
-                              name='PolicyComponent'))
-
             # When not sharing encoder, create a separate encoder for the value
             # component.
             self._value_encoder = encoding_network_ctor(
                 input_tensor_spec=observation_spec,
                 kernel_initializer=kernel_initializer)
 
-            self._value_head = alf.nn.Sequential(
-                alf.layers.FC(
-                    input_size=self._actor_encoder.output_spec.shape[0],
-                    output_size=1), alf.layers.Reshape(shape=()))
-
-            value_component = alf.nn.Sequential(self._value_encoder,
-                                                self._value_head)
-
-            self._composition = alf.nn.Branch(
-                (policy_component, value_component), name='AllBranches')
+            self._composition = alf.nn.Sequential(
+                alf.nn.Branch(
+                    alf.nn.Sequential(
+                        self._actor_encoder,
+                        alf.nn.Branch(
+                            self._policy_head,
+                            self._aux_head,
+                            name='PolicyComponent')),
+                    alf.nn.Sequential(
+                        self._value_encoder,
+                        alf.layers.FC(
+                            input_size=encoder_output_size, output_size=1),
+                        alf.layers.Reshape(shape=()))),
+                # Order: policy, value, aux value
+                lambda heads: (heads[0][0], heads[1], heads[0][1]))
 
     def forward(self, observation, state):
         """Computes the action distribution, aux value and value estimation
@@ -236,15 +230,15 @@ class DisjointPolicyValueNetwork(Network):
             observation (nested torch.Tensor): a tensor that is consistent with
                 the encoding network
 
-            state: the state for RNN based network for interface compatibility,
-                which is always set to empty in this case because this network
-                does not have an RNN-based encoder
+            state: the state(s) for RNN based network
 
         Returns:
-            act_dist (torch.distributions): action distribution
-            state: empty
+
+            output (Triplet): network output in the order of policy (action
+                distribution), value function estimation, auxiliary value
+                function estimation
+
+            state (Triplet): RNN states in the order of policy, value, aux value
 
         """
-        ((action_distribution, aux), value), _ = self._composition(
-            observation, state=())
-        return (action_distribution, value, aux), ()
+        return self._composition(observation, state=state)
--- a/alf/algorithms/ppg/disjoint_policy_value_network.py
+++ b/alf/algorithms/ppg/disjoint_policy_value_network.py
@@ -16,15 +16,16 @@ from typing import Callable, Tuple
 
 import torch
 import alf
-from alf.tensor_specs import TensorSpec
+from alf.tensor_specs import TensorSpec, BoundedTensorSpec
 from alf.data_structures import namedtuple
 from alf.networks import Network, NormalProjectionNetwork, CategoricalProjectionNetwork, EncodingNetwork
 
 
 def _create_projection_net_based_on_action_spec(
-        discrete_projection_net_ctor: Callable[..., Network],
-        continuous_projection_net_ctor: Callable[..., Network],
-        input_size: int, action_spec):
+        discrete_projection_net_ctor: Callable[[int, BoundedTensorSpec],
+                                               Network],
+        continuous_projection_net_ctor: Callable[
+            [int, BoundedTensorSpec], Network], input_size: int, action_spec):
     """Create project network(s) for the potentially nested action spec.
 
     This function basically creates a projection network for each of the leaf
@@ -33,17 +34,14 @@ def _create_projection_net_based_on_action_spec(
 
     Args:
 
-        discrete_projection_net_ctor (Callable[..., Network]): constructor that
-            generates a discrete projection network that outputs discrete
-            actions.
-
-        continuous_projection_net_ctor (Callable[..., Network): constructor that
-            generates a continuous projection network that outputs continuous
-            actions.
-
+        discrete_projection_net_ctor (Callable[[int, BoundedTensorSpec],
+            Network]): constructor that generates a discrete projection network
+            that outputs discrete actions.
+        continuous_projection_net_ctor (Callable[[int, BoundedTensorSpec],
+            Network): constructor that generates a continuous projection network
+            that outputs continuous actions.
         input_size (int): the input_size for the projection network, which usually
             comes from the output of an encoding network.
-
         action_spec (nest of TensorSpec): speficifies the shape and type of the
             output action. The type of each invidual projection network in the
             output is derived from this.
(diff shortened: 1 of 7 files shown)


alf/networks/disjoint_policy_value_network.py
```diff
@@ -0,0 +1,263 @@
...
+            continuous_projection_net_ctor=continuous_projection_net_ctor,
+            input_size=self._actor_encoder.output_spec.shape[0],
+            action_spec=action_spec)
+
+        if alf.nest.is_nested(projection_net):
+            # In the case when the projection net is multi-fold (because actor
+            # spec is a nest), force picking up the parameters inside those
+            # project networks into this DisjointPolicyValueNetwork instance.
+            self._projection_net_module_list = torch.nn.ModuleList(
+                alf.nest.flatten(projection_net))
```
emailweixu: No need to have 178-179. `nn.Branch` should be able to handle it correctly.

alf/networks/disjoint_policy_value_network.py
```diff
@@ -0,0 +1,263 @@
...
+
+        action_distribution, actor_state = self._actor_branch(
+            observation, state=state.actor)
+
+        value, value_state = self._value_branch(observation, state=state.value)
+
+        aux, aux_state = self._value_branch(observation, state=state.aux)
+
+        return action_distribution, value, aux, DisjointPolicyValueNetworkState(
+            actor=actor_state, value=value_state, aux=aux_state)
```
emailweixu: Network should return a tuple of two elements, the first element is the output, the second element is the state.

alf/networks/disjoint_policy_value_network.py
```diff
@@ -0,0 +1,263 @@
...
+            state: empty
+
+        """
+
+        action_distribution, actor_state = self._actor_branch(
+            observation, state=state.actor)
+
+        value, value_state = self._value_branch(observation, state=state.value)
+
+        aux, aux_state = self._value_branch(observation, state=state.aux)
```
emailweixu: when sharing encoder, actor_encoder is called twice with the same input. It's better to call it once.

alf/networks/disjoint_policy_value_network.py
```diff
@@ -0,0 +1,263 @@
...
+            # component.
+            value_encoder = encoding_network_ctor(
+                input_tensor_spec=observation_spec,
+                kernel_initializer=kernel_initializer)
+
+            self._value_branch = alf.nn.Sequential(
+                value_encoder,
+                alf.layers.FC(
+                    input_size=self._actor_encoder.output_spec.shape[0],
+                    output_size=1), alf.layers.Reshape(shape=()))
```
emailweixu: Have you considered putting every thing in a single container? That might simplify the code.

breakds: Yep I was thinking about using one `nn.Branch`, but hesitated because in certain cases we might want to run just one head. But given that the head is usually very lite, maybe putting them under one `nn.Branch` is better. Also good for readability.

alf/networks/disjoint_policy_value_network.py
```diff
@@ -0,0 +1,263 @@
+# Copyright (c) 2021 Horizon Robotics and ALF Contributors. All Rights Reserved.
```
hnyu: I wonder do we need such a big class to hold several networks? It seems to me that these can be included in the algorithm.

breakds: I think there are clear benefits about having it in a separate `.py` file:

1. It is cleaner this way as all network related logic are grouped together (so does all the documentation, the network class looks big but it is half documentation half implementation).
2. On the other side, if we put so many lines into the algorithm, it will make the algorithm much bigger and therefore much less readable. 
3. We can put unit test in a file that is specifically tied to this class.
4. We can hide implementations in this file so that people only cares about the algorithm don't have to read them.
5. We can hide the private/non-public methods/functions in this file so that they do not appear in the algorithm implementation, which can harm readability too.

I might have missed but did not see significant cost of having a separate `.py` file (or, a separate library).

alf/networks/disjoint_policy_value_network.py
breakds: I read the code on `Sequential` and `Branch` and did not find how those in `Branch` can get into the closure. Also did an experiment in the unit test and found that when `action_spec` is a nest, those projection networks' parameters do not propagate back to the `DisjointPolicyValueNetwork`'s state closure.

If this is not intentional, we probably need to fix `Sequential`/`Branch` to add this feature.

alf/networks/disjoint_policy_value_network.py
breakds: Good catch! Sorry I missed this. Will update.

alf/networks/disjoint_policy_value_network.py
breakds: Thanks for the reminder - did not know this convention. Will fix.

alf/networks/disjoint_policy_value_network.py
hnyu: > I think there are clear benefits about having it in a separate `.py` file:
> 
> 1. It is cleaner this way as all network related logic are grouped together (so does all the documentation, the network class looks big but it is half documentation half implementation).
> 2. On the other side, if we put so many lines into the algorithm, it will make the algorithm much bigger and therefore much less readable.
> 3. We can put unit test in a file that is specifically tied to this class.
> 4. We can hide implementations in this file so that people only cares about the algorithm don't have to read them.
> 5. We can hide the private/non-public methods/functions in this file so that they do not appear in the algorithm implementation, which can harm readability too.
> 
> I might have missed but did not see significant cost of having a separate `.py` file (or, a separate library).

I think my point is that this network is a highly specialized one that probably won't be used by any other algorithm in the future. Other networks in alf.networks are very general. 

The details in this class might be critical for understanding PPG. If someone wants to understand PPG, he/she also needs to look at this file eventually.

alf/networks/disjoint_policy_value_network.py
emailweixu: It's supposed to be handled at line 413 and 160 in networks/containers.py

alf/networks/disjoint_policy_value_network.py
emailweixu: Agree that networks in this folder should be more generic.

breakds: Right this network is high specialized, thought I still think it should be in its own `.py` file, I do agree it might be too specialized for `network`.

I have two proposals here:

1. put it under `networks/ppg/`
2. put it under `algorithms/ppg`

I understand that currently `algorithms/` is a flat directory, but having a secondary directory structure there can help grouping related files there. In this case we can have the network, `ppg_aux_phase_loss.py` and `ppg_algorithm.py`.

What do you think?

alf/networks/disjoint_policy_value_network.py
breakds: Thanks. I can confirm that after the refactor all the parameters are picked up in `_composition`. Line 413 and 160 in `continers.py` should work.

> Also did an experiment in the unit test and found that when action_spec is a nest, those projection networks' parameters do not propagate back to the DisjointPolicyValueNetwork's state closure.

Could not get back to where the experiment fails, it is likely that I missed the projection parameters when checked with my naked eye.

alf/networks/disjoint_policy_value_network.py
breakds: Now combined into one container.

breakds:
Thanks for the prompt review, @emailweixu  and @hnyu ! Have updated the PR based on the comments and suggestions, mainly

1. Combined all branches in to one container
2. Moved the network to `algorithms/ppg`
3. Add the unit test to CI/CD

This is tested again with the updated unit tests (in this PR) and the updated algorithm (next PR).

PTAL!

alf/algorithms/ppg/disjoint_policy_value_network.py
```diff
@@ -0,0 +1,250 @@
...
+            self._value_head = alf.nn.Sequential(
+                alf.layers.FC(
+                    input_size=self._actor_encoder.output_spec.shape[0],
+                    output_size=1), alf.layers.Reshape(shape=()))
+
+            value_component = alf.nn.Sequential(self._value_encoder,
+                                                self._value_head)
+
+            self._composition = alf.nn.Branch(
+                (policy_component, value_component), name='AllBranches')
```
emailweixu: It's possible to make the network return the right structure as the following. And we don't even need to implement forward. Simply return the constructed network is enough. And the network states will be correctly handled by the container, which means that RNN encoder is automatically supported.
```python
    self._composition = alf.nn.Sequential(
        alf.nn.Branch(policy_component, value_component),
        lambda heads: (heads[0][0], heads[1], heads[0][1]))
```

alf/algorithms/ppg/disjoint_policy_value_network.py
```diff
@@ -0,0 +1,250 @@
...
+        self._policy_head = _create_projection_net_based_on_action_spec(
+            discrete_projection_net_ctor=discrete_projection_net_ctor,
+            continuous_projection_net_ctor=continuous_projection_net_ctor,
+            input_size=self._actor_encoder.output_spec.shape[0],
+            action_spec=action_spec)
+        # In the case when the projection net is multi-fold (because actor
+        # spec is a nest), force picking up the parameters inside those
+        # project networks into this DisjointPolicyValueNetwork instance.
+        self._projection_net_module_list = torch.nn.ModuleList(
+            alf.nest.flatten(self._policy_head))
```
emailweixu: Branch() is intended to handle this. If not, should fix Branch instead.

alf/algorithms/ppg/disjoint_policy_value_network.py
```diff
@@ -0,0 +1,250 @@
...
+                input_tensor_spec=observation_spec,
+                kernel_initializer=kernel_initializer)
+
+            self._value_head = alf.nn.Sequential(
+                alf.layers.FC(
+                    input_size=self._actor_encoder.output_spec.shape[0],
+                    output_size=1), alf.layers.Reshape(shape=()))
+
+            value_component = alf.nn.Sequential(self._value_encoder,
+                                                self._value_head)
```
emailweixu: Can be merged with line 220
```python
    value_component = alf.nn.Sequential(
        self._value_encoder,
        alf.layers.FC(input_size=encoder_out_size, output_size=1),
        alf.layers.Reshape(shape=()))
```

alf/algorithms/ppg/disjoint_policy_value_network.py
emailweixu: Do you think a single giant statement like the following is easier to understand?
```python
    composition = alf.nn.Sequential(
        alf.nn.Branch(
            alf.nn.Sequential(
                self._actor_encoder,
                alf.nn.Branch(self._policy_head, self._aux_head)),
            alf.nn.Sequential(
                self._value_encoder,
                alf.layers.FC(input_size=encoder_out_size, output_size=1),
                alf.layers.Reshape(shape=()))),
        lambda heads: (heads[0][0], heads[1], heads[0][1]))
```

alf/algorithms/ppg/disjoint_policy_value_network.py
breakds: Sorry I thought I have removed this. It is not needed.

breakds:
@emailweixu The PR is updated accordingly. I have not yet removed `forward()`, though it is one-liner. Should I merge that into `__init__` instead? I am indifferent on that.

alf/algorithms/ppg/disjoint_policy_value_network.py
```diff
@@ -0,0 +1,244 @@
...
+        # +------------------------------------------+
+        # | Step 3: Value head of the aux branch     |
+        # +------------------------------------------+
+
+        # Note that the aux branch value head belongs to the policy component.
+
+        # Like the value head Aux head is outputing value estimation
+        self._aux_head = alf.nn.Sequential(
+            alf.layers.FC(
+                input_size=self._actor_encoder.output_spec.shape[0],
```
emailweixu: encoder_output_size

alf/algorithms/ppg/disjoint_policy_value_network.py
```diff
@@ -0,0 +1,244 @@
...
+            kernel_initializer=kernel_initializer)
+
+        # +------------------------------------------+
+        # | Step 2: Projection for the policy branch |
+        # +------------------------------------------+
+
+        self._policy_head = _create_projection_net_based_on_action_spec(
+            discrete_projection_net_ctor=discrete_projection_net_ctor,
+            continuous_projection_net_ctor=continuous_projection_net_ctor,
+            input_size=self._actor_encoder.output_spec.shape[0],
```
emailweixu: encoder_output_size

alf/algorithms/ppg/disjoint_policy_value_network.py
```diff
@@ -0,0 +1,244 @@
...
+    tensor spec in the action spec. Those networks are packed into the same
+    nested structure as the input action spec and returned as a whole.
+
+    Args:
+
+        discrete_projection_net_ctor (Callable[..., Network]): constructor that
+            generates a discrete projection network that outputs discrete
+            actions.
+
+        continuous_projection_net_ctor (Callable[..., Network): constructor that
```
emailweixu: Callable[[int, BoundedTensorSpec], Network]

alf/algorithms/ppg/disjoint_policy_value_network.py
```diff
@@ -0,0 +1,244 @@
...
+        input_size: int, action_spec):
+    """Create project network(s) for the potentially nested action spec.
+
+    This function basically creates a projection network for each of the leaf
+    tensor spec in the action spec. Those networks are packed into the same
+    nested structure as the input action spec and returned as a whole.
+
+    Args:
+
+        discrete_projection_net_ctor (Callable[..., Network]): constructor that
```
emailweixu: Callable[[int, BoundedTensorSpec], Network]

alf/algorithms/ppg/disjoint_policy_value_network.py
```diff
@@ -0,0 +1,244 @@
...
+                shared between the value and the policy component, resulting in
+                a "shared" architecture disjoint network. When set to false, the
+                encoding network is not shared, resulting in a "dual"
+                architecture disjoint network.
+
+            kernel_initializer (Callable): initializer for all the layers
+                excluding the projection net. If none is provided a default
+                xavier_uniform will be used.
+
+            discrete_projection_net_ctor (Callable[..., Network]): constructor
```
emailweixu: Callable[[int, BoundedTensorSpec], Network]

alf/algorithms/ppg/disjoint_policy_value_network.py
```diff
@@ -0,0 +1,244 @@
...
+                that generates a discrete projection network that outputs
+                discrete actions.
+
+            continuous_projection_net_ctor (Callable[..., Network): constructor
```
emailweixu: Callable[[int, BoundedTensorSpec], Network]

alf/algorithms/ppg/disjoint_policy_value_network.py
```diff
@@ -0,0 +1,244 @@
...
+                component, 1 or 2 encoding network will be created using this
+                constructor.
+
+            is_sharing_encoder (bool): When set to true, the encoding network is
+                shared between the value and the policy component, resulting in
+                a "shared" architecture disjoint network. When set to false, the
+                encoding network is not shared, resulting in a "dual"
+                architecture disjoint network.
+
+            kernel_initializer (Callable): initializer for all the layers
```
emailweixu: Callable[[Tensor], None]

alf/algorithms/ppg/disjoint_policy_value_network.py
```diff
@@ -0,0 +1,244 @@
...
+        because in the case when the action spec is a nest of different types
+        where some of them are discrete and some of them are continuous,
+        corresponding projection networks can be created for the two parties
+        individually and respectively.
+
+        Args:
+
+            observation_spec (nest of TesnorSpec): specifies the shape and type
+                of the input observation.
+
```
hnyu: I think the convention is to remove the empty line between two argument descriptions.

alf/algorithms/ppg/disjoint_policy_value_network.py
```diff
@@ -0,0 +1,244 @@
...
+        Returns:
+
+            output (Triplet): network output in the order of policy (action
+                distribution), value function estimation, auxiliary value
+                function estimation
+
+            state (Triplet): RNN states in the order of policy, value, aux value
+
+        """
+        return self._composition(observation, state=state)
```
hnyu, line 237: If you always output the three values, will there be redundant computation? For example, sometimes you might only need one or two of them.

breakds: Valid point. Yep, as mentioned above this is one of the concern. However since currently the layer is rather thin, it should not account for noticeable overhead. 

On the other hand making up for this will make the logic a bit more complicated. Plan to add a `TODO` here - if this becomes a problem in the future, we know how to fix it.

alf/algorithms/ppg/disjoint_policy_value_network.py
breakds: Sure, will do. I understand this is for consistency, but personally I found the empty lines make it easier to read.

breakds:
Updated the type hints and added `TODO` based on comments.

emailweixu:
> @emailweixu The PR is updated accordingly. I have not yet removed `forward()`, though it is one-liner. Should I merge that into `__init__` instead? I am indifferent on that.

What I meant is that given the current implementation, this doesn't have to a class. Simply returning the constructed composition net can work (and hence forward() is not needed).

breakds:
> > @emailweixu The PR is updated accordingly. I have not yet removed `forward()`, though it is one-liner. Should I merge that into `__init__` instead? I am indifferent on that.
> 
> What I meant is that given the current implementation, this doesn't have to a class. Simply returning the constructed composition net can work (and hence forward() is not needed).

Got it. After thinking about it I'd like to keep the `class` here as it can be handy for explicitly specifying the type hints in the algorithm. Do you think that is valid?

emailweixu (approved):

PR #978, opened 2021-08-19
fix rtd pip install

hnyu:
RTD build failed because of trying to pip install local cnest dir

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -18,5 +18,3 @@ python:
   version: 3.7
   install:
     - requirements: docs/requirements.txt
-    - method: pip
-      path: ./alf/nest/cnest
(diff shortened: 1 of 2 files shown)


breakds (dismissed):

breakds:
Thanks Haonan!

Just did another search and found this: https://github.com/HorizonRobotics/alf/blob/pytorch/alf/utils/common.py#L1136

Should we just remove it from the environment variable snapshot?

hnyu:
> Thanks Haonan!
> 
> Just did another search and found this: https://github.com/HorizonRobotics/alf/blob/pytorch/alf/utils/common.py#L1136
> 
> Should we just remove it from the environment variable snapshot?

You're right. I've removed it from snapshot paths.

breakds (approved):

PR #979, opened 2021-08-20
(WIP) Phasic Policy Gradient [Part 2]: Implement the PPG Algorithm

breakds:
# Motivation

Implement the PPG algorithm and hopefully we can use it as 

1. A good baseline
2. An example of algorithm implementation with an irregular training pipeline

# Solution

**NOTE**: I understand this PR might need a few rounds of reviews and updates, and I think it would help both parties to send it out as a whole for preview as early as possible. I have tried my best to add documentation but feel free to let me know if it can be improved.

1. A custom training pipeline by overriding `train_iter()`
2. Switching between different phase with `PhaseContext` objects, which aims to promote readability

# Testing

Tested on cart pole which converges to 200.0 almost immediately under current settings of parameters. 

Plan to do more tests after adding more PPG-based examples.

### DIFF
--- a/alf/algorithms/ppg_algorithm.py
+++ b/alf/algorithms/ppg_algorithm.py
@@ -1,0 +1,377 @@
+# Copyright (c) 2021 Horizon Robotics and ALF Contributors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Phasic Policy Gradient Algorithm."""
+
+import torch
+
+from typing import Optional, Tuple
+from contextlib import contextmanager
+
+import alf
+from alf.algorithms.ppg import DisjointPolicyValueNetwork, PPGAuxPhaseLoss, PPGAuxPhaseLossInfo, PPGRolloutInfo, PPGTrainInfo
+from alf.algorithms.off_policy_algorithm import OffPolicyAlgorithm
+from alf.algorithms.config import TrainerConfig
+from alf.algorithms.ppo_loss import PPOLoss
+from alf.algorithms.algorithm import Loss
+from alf.networks.encoding_networks import EncodingNetwork
+from alf.experience_replayers.experience_replay import OnetimeExperienceReplayer
+from alf.data_structures import TimeStep, AlgStep, LossInfo
+from alf.tensor_specs import TensorSpec
+
+from alf.utils import common, dist_utils, value_ops, tensor_utils
+from alf.utils.summary_utils import record_time
+
+
+class PPGPhaseContext(object):
+    """The context for each of the two phases of PPG
+
+    The PPG algorithm has two phases. Each phase will
+
+    1. use a DIFFERENT "loss function"
+    2. on a DIFFERENT "replay buffer"
+    3. and use a DIFFERENT "optimizer"
+    4. on a DIFFRENT copy of the "network"
+
+    Therefore, this context class is created to group such (stateful)
+    information for each phase so as to make switching between phases
+    as easy as switching the context to use. See the implementation of
+    the ``PPGAlgorithm`` for the details.
+
+    Note that in the above list No.4 is not needed in theory because
+    the two phases technically operates on the same network. However,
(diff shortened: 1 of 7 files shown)


alf/algorithms/ppg_algorithm.py
```diff
@@ -0,0 +1,376 @@
...
+
+    @property
+    def exp_replayer(self):
+        return self._exp_replayer
+
+    @property
+    def network(self):
+        return self._network
+
+    def on_context_switch_from(self, previous_context: 'PPGPhaseContext'):
```
emailweixu, line 111: PPGPhaseContext should not be quoted.

emailweixu:
The current implementation makes PPG an isolated algorithm, meaning that it cannot be embedded in other algorithms (e.g. Agent).

A possible way to make it more like other existing RL algorithms is to maintain a replay buffer by the algorithm itself. It will be very much like the current PPO with the following changes:
1. storing experiences to replay buffer at rollout_step()
2. implementing after_train_iter() for aux phase.

breakds:
> The current implementation makes PPG an isolated algorithm, meaning that it cannot be embedded in other algorithms (e.g. Agent).
> 
> A possible way to make it more like other existing RL algorithms is to maintain a replay buffer by the algorithm itself. It will be very much like the current PPO with the following changes:
> 
> 1. storing experiences to replay buffer at rollout_step()
> 2. implementing after_train_iter() for aux phase.

Thanks for the prompt comments, Wei! I will look into this.

alf/algorithms/ppg_algorithm.py
```diff
@@ -0,0 +1,391 @@
...
+            exp_replayer=None)
+
+        # The auxiliary phase uses an extra experience replayer with the loss
+        # specific to the auxiliary update phase.
+        self._aux_phase = PPGPhaseContext(
+            main_algorithm=self,
+            network=dual_actor_value_network.copy(),
+            optimizer=aux_optimizer,
+            loss=PPGAuxPhaseLoss(),
+            exp_replayer=OnetimeExperienceReplayer())
```
emailweixu, line 282: This will not work if PPGAlgorithm is used as a sub-algorithm. Currently, only the root algorithm automatically stores experience to replay buffer. You have to store experience to the replay buffer for aux phase by yourself at rollout_step().

I suggest you put PPG in Agent to test if the whole thing works.

breakds: > This will not work if PPGAlgorithm is used as a sub-algorithm. Currently, only the root algorithm automatically stores experience to replay buffer. You have to store experience to the replay buffer for aux phase by yourself at rollout_step().

Yep I picked and remembered the part from last round of review already - trusted your suggestion. Haven't done this because I haven't fully understood why it does not work (yet). Will need to try this with `Agent` to see how it works.

> I suggest you put PPG in Agent to test if the whole thing works.

Good point. I will try this.

breakds:
Sorry I should have commented here: this is still WIP - will let the reviewers know when it is ready for review.

alf/algorithms/ppg_algorithm.py
breakds: Will do.

breakds:
After some trial and error I was able to run PPG in `Agent` (PR not updated yet). One of the remaining problem is that when running under `Agent`, there is a significant drop in performance (throughput 12000 vs 8000) in normal iterations (i.e. non-aux-iteration).

I am working on hunting down the source of performance degeneration.

breakds:
> After some trial and error I was able to run PPG in `Agent` (PR not updated yet). One of the remaining problem is that when running under `Agent`, there is a significant drop in performance (throughput 12000 vs 8000) in normal iterations (i.e. non-aux-iteration).
> 
> I am working on hunting down the source of performance degeneration.

This comes from how I am adding the experience replayer. I was trying to add it more elegantly but such approach has bad side effects. Will come up with another solution.

alf/algorithms/ppg_algorithm.py
breakds: Added `annotation` to allow forward reference like this.

alf/algorithms/ppg_algorithm.py
breakds: Updated the design and now it works both standalone and under `Agent`. Tried to make this as elegant as possible.

breakds:
I have updated the PR so that it works with `Agent` now and tested. I'll wait for the CI/CD to pass before removing the `(WIP)` label.

breakds:
Discussed with @emailweixu the other day that I will create another PR to make `train_from_replay_buffer` more flexible before resuming work on this PR.

breakds:
Just came back to this PR and I am suspecting there are (GPU) memory leaks or something like that in this implementation. Even without auxiliary step (which rendered the algorithm to be just "PPO") the memory keeps growing as iterations go. Will investigate.

breakds:
Not removing the branch yet as some of the information in this PR might still be useful.
The PPG implementation effort will be in a new PR.

PR #980, opened 2021-08-20
Format error, some red font is in the wrong range

www2171668:
Format error, some red font is in the wrong range:
1、ProcessEnvironment``s and obtain a batched ``ALfEnvironmnet ->  ProcessEnvironment`` s and obtain a batched ``ALfEnvironmnet
2、TimeStep``s from ``ProcessEnvironment -> TimeStep`` s from ``ProcessEnvironment


hnyu (approved):
Thanks for the PR

PR #981, opened 2021-08-22
fix grid search after ddp

hnyu:
After the DDP PR, there is no longer 'train_eval' function to be called by grid_search.py, so I moved a code block out from `training_worker()`. Also fix the grid search naming issue when a value contains special chars.


breakds:
LGTM. Sorry about not realizing grid search depend on `train.py`. Thank you Haonan!

breakds (approved):

PR #985, opened 2021-08-25
make_parallel for network containers

### LINKED ISSUE #762: Implement make_parallel for network containers
So we can have parallel version for any network built using Sequential, Parallel and Branch.

emailweixu:
This change supports creating parallel network for network containers.  This is achieved by implementing make_parallel for various layers and transforming all the modules in the container.

With this change, it will be trivial to implement make_parallel for all kinds of networks if they are implemented using network containers.

### DIFF
--- a/alf/layers.py
+++ b/alf/layers.py
@@ -13,14 +13,14 @@
 # limitations under the License.
 """Some basic layers."""
 
+from absl import logging
 import copy
-
 from functools import partial
 import numpy as np
 import torch
 import torch.nn as nn
 import torch.nn.functional as F
-from typing import Callable
+from typing import Optional, Union, Callable
 
 import alf
 from alf.initializers import variance_scaling_init
@@ -84,6 +84,9 @@ class Cast(nn.Module):
     def forward(self, x):
         return x.to(self._dtype)
 
+    def make_parallel(self, n: int):
+        return Cast(self._dtype)
+
 
 class Transpose(nn.Module):
     """A layer that perform the transpose of channels.
@@ -109,6 +112,9 @@ class Transpose(nn.Module):
     def forward(self, x):
         return x.transpose(self._dim0, self._dim1)
 
+    def make_parallel(self, n: int):
+        return Transpose(self._dim0, self._dim1)
+
 
 class Permute(nn.Module):
     """A layer that perform the permutation of channels."""
@@ -126,6 +132,9 @@ class Permute(nn.Module):
 
     def forward(self, x):
         return x.permute(*self._dims)
+
+    def make_parallel(self, n: int):
+        return Permute(*self._dims)
 
 
 class BatchSquash(object):
@@ -182,6 +191,9 @@ class OneHot(nn.Module):
     def forward(self, input):
         return nn.functional.one_hot(
             input, num_classes=self._num_classes).to(torch.float32)
+
+    def make_parallel(self, n: int):
+        return OneHot(self._num_classes)
 
 
 @alf.configurable
@@ -432,7 +444,7 @@ class FC(nn.Module):
     def bias(self):
         return self._bias
 
-    def make_parallel(self, n):
+    def make_parallel(self, n: int):
         """Create a ``ParallelFC`` using ``n`` replicas of ``self``.
         The initialized layer parameters will be different.
         """
@@ -639,6 +651,8 @@ class ParallelFC(nn.Module):
             bias_init_value (float): a constant
         """
         super().__init__()
+        self._input_size = input_size
+        self._output_size = output_size
         self._activation = activation
         self._weight = nn.Parameter(torch.Tensor(n, output_size, input_size))
         if use_bias:
@@ -1085,6 +1099,11 @@ class Conv2D(nn.Module):
                 ``kernel_initializer`` is not None.
             bias_init_value (float): a constant
         """
+        # get the argument list with vals
+        self._kwargs = copy.deepcopy(locals())
+        self._kwargs.pop('self')
+        self._kwargs.pop('__class__')
+
         super(Conv2D, self).__init__()
         if use_bias is None:
             use_bias = not use_bn
@@ -1135,6 +1154,9 @@ class Conv2D(nn.Module):
     @property
     def bias(self):
         return self._conv2d.bias
+
+    def make_parallel(self, n: int):
+        return ParallelConv2D(n=n, **self._kwargs)
 
 
 @alf.configurable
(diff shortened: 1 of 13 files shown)


alf/layers.py
```diff
@@ -2854,3 +2912,146 @@ def reset_parameters(self):
...
+    Examples:
+
+    Applying parallel net on same input:
+
+    .. code-block:: python
+
+        pnet = make_parallel_net(net, n)
+        # replicate input.
+        # pinput will have shape [batch_size, n, ...], if input has shape [batch_size, ...]
+        pinput = make_parallel_input(input)
```
le-horizon: `make_parallel_input(input, n)`

alf/layers.py
```diff
@@ -13,14 +13,14 @@
...
+from absl import logging
 import copy
-
 from functools import partial
 import numpy as np
 import torch
 import torch.nn as nn
 import torch.nn.functional as F
-from typing import Callable
+from typing import Optional, Union, Callable
```
Haichao-Zhang: `Optional` is not used

alf/layers.py
```diff
@@ -2854,3 +2912,146 @@ def reset_parameters(self):
...
+        return NaiveParallelLayer(module, n)
+
+
+class NaiveParallelLayer(nn.Module):
+    def __init__(self, module: Union[nn.Module, Callable], n: int):
+        """
+        A parallel network has ``n`` copies of network with the same structure but
+        different indepently initialized parameters.
+
+        ``NaiveParallelLayer`` creats ``n`` independent networks with the same
```
Haichao-Zhang: creats -> creates

alf/layers_test.py
```diff
@@ -932,6 +963,82 @@ def test_sequential4(self):
...
+        self._test_make_parallel(layer, input_spec)
+        layer = alf.layers.Transpose(-2, -1)
+        self._test_make_parallel(layer, input_spec)
+
+    def test_permute(self):
+        input_spec = alf.BoundedTensorSpec((8, 4, 10))
+        layer = alf.layers.Permute(2, 1, 0)
+        self._test_make_parallel(layer, input_spec)
+
+    def test_onehost(self):
```
Haichao-Zhang: test_onehost -> test_onehot

alf/networks/networks.py
```diff
@@ -171,7 +178,8 @@ def __init__(self,
                  stack_size,
                  pooling_size=1,
                  dtype=torch.float32,
-                 mode='skip'):
+                 mode='skip',
+                 name='TemporalPool'):
```
le-horizon, line 182: Looking at the TemporalPool example above, it seems if pool size == 2, we always ignore every other (half) of the timesteps.
Have we considered pooling every timestep, instead of pooling every pool_size timesteps?

le-horizon (commented):
Thanks for the nice improvement Wei.

Looks good as is, just small comments.

Thanks,
Le

alf/layers.py
```diff
@@ -2624,6 +2664,9 @@ def __init__(self, dim):
     def forward(self, input):
         return input.sum(dim=self._dim)
 
+    def make_parallel(self, n: int):
+        return Sum(self._dim)
```
Haichao-Zhang, line 2647: Might be useful to add comments on the assumed shape of parallel batch `[B, n, ...]`, and both the batch dimension and replica dimension are not counted for dim.

alf/layers.py
```diff
@@ -109,6 +112,9 @@ def __init__(self, dim0=0, dim1=1):
     def forward(self, x):
         return x.transpose(self._dim0, self._dim1)
 
+    def make_parallel(self, n: int):
```
Haichao-Zhang, line 116: Would be useful to add comments on the assumed shape of parallel batch `[B, n, ...]`, and both the batch dimension and replica dimension are not considered for transpose.

alf/layers.py
hnyu: > Might be useful to add comments on the assumed shape of parallel batch `[B, n, ...]`, and both the batch dimension and replica dimension are not counted for dim.

If the replica dim is not counted, then the value of `self._dim` is not correct here.

alf/layers.py
```diff
@@ -2854,3 +2912,146 @@ def reset_parameters(self):
...
+        return Sequential(
+            *new_networks, output=self._output, **new_named_networks)
+
+
+def make_parallel_net(module, n: int):
+    """Make a parallelized version of ``module``.
+
+    A parallel network has ``n`` copies of network with the same structure but
+    different independently initialized parameters. The parallel network can
+    process a batch of the data with shape [batch_size, n, ...] using ``n``
```
hnyu, line 2921: Should we automatically convert data with shape [batch_size, ...] to [batch_size, n, ...]?

alf/nest/utils.py
```diff
@@ -104,6 +105,10 @@ def _combine_flat(self, tensors):
         else:
             return torch.cat(tensors, dim=self._dim)
 
+    def make_parallel(self, n):
+        dim = self._dim if self._dim < 0 else self._dim + 1
```
hnyu: Should add comments saying that self._dim excludes batch dim and parallel dim. The current implementation seems not to ignore the batch dim.

hnyu (changes requested):

alf/layers.py
Haichao-Zhang: Base on my understanding, the replica dim is excluded. In this case the ``self._dim`` is correct as it is increased by `1` during the construction of the new``Sum`` instance. The same is for `Transpose` and several other related operators. Some more documentation on this would be helpful.

alf/layers.py
emailweixu: As network can have sub-networks, doing this check for all of them can be wasteful. So it's intended to use make_parallel_input to do the conversion by the user.

alf/nest/utils.py
emailweixu: Changed the behavior of NestConcat to not including batch dim.
Comments added.

alf/networks/networks.py
emailweixu: Is mode "avg" what you want?

alf/layers.py
```diff
@@ -2665,6 +2634,16 @@ def forward(self, input):
         return input.sum(dim=self._dim)
 
     def make_parallel(self, n: int):
+        """Create a Sum layer to handle parallel batch.
+
+        It is assumed that a parallel batch has shape [B, n, ...] and both the
+        batch dimension and replica dimension are counted for ``dim``
```
Haichao-Zhang: `are counted` --> `are not counted`

Haichao-Zhang (commented):
The comments are very helpful. Just some more minor points.

alf/utils/spec_utils.py
```diff
@@ -17,13 +17,57 @@
...
 from typing import Iterable
 
-from alf.layers import BatchSquash
 import alf.nest as nest
 from alf.nest.utils import get_outer_rank
 from alf.tensor_specs import TensorSpec, BoundedTensorSpec
 from . import dist_utils
 
 
+class BatchSquash(object):
```
hnyu: It's better to put this in tensor_utils?

Haichao-Zhang (approved):

hnyu (approved):

PR #986, opened 2021-08-25
tutorial chap 5

hnyu:
environments and wrappers
preview: https://alf-fork.readthedocs.io/en/latest/tutorial/environments_and_wrappers.html


docs/tutorial/environments_and_wrappers.rst
```diff
@@ -0,0 +1,424 @@
...
+
+    class SparseReward(gym.Wrapper):
+        """Binarize the original :math:`-/+` rewards to :math:`0/1`.
+        """
+
+        def __init__(self, env):
+            gym.Wrapper.__init__(self, env)
+
+        def step(self, action):
+            ob, reward, done, info = self.env.step(action)
```
breakds, line 61: I know this is probably common knowledge, but it does not harm to add a comment here suggesting that `self.env` is actually the wrapped environment that gets passed into `gym.Wrapper.__init__(self, env)` above.

docs/tutorial/environments_and_wrappers.rst
```diff
@@ -0,0 +1,424 @@
...
+
+    class TimeStep(
+        namedtuple(
+            'TimeStep', [
+                'step_type', 'reward', 'discount', 'observation',
+                'prev_action', 'env_id', 'untransformed', "env_info"
+            ],
+            default_value=())):
+
+Fields of ``TimeStep``
```
breakds, line 148: Do you think we can add a hyperlink from Chapter 3 to this? Some of the hooks takes `TimeStep` as inputs.

docs/tutorial/environments_and_wrappers.rst
```diff
@@ -0,0 +1,424 @@
...
+     the step generated from ``env.reset()``.
+  2. :attr:`.StepType.LAST` is the last step of an episode.
+  3. :attr:`.StepType.MID` is for all the other steps in an episode.
+
+* ``reward``: reward from the previous action. In some RL literature, the
+  reward for an action :math:`a_t` at time :math:`t` is usually written as
+  :math:`r_t`. However, in ALF, :math:`r_t` will always represent the reward for
+  the previous action at time :math:`t-1`.
+
+* ``discount``: When calculating the cumulative discounted return, ``discount``
```
breakds, line 166: A question that might be raised here is: why the environment needs the discount information, and why is environment responsible for computing the cumulative return (instead of being algorithm dependent)?

breakds: Ah I found it right below: https://alf-fork.readthedocs.io/en/latest/tutorial/environments_and_wrappers.html?highlight=discount#step-type-and-discount  maybe add the link to above?

breakds (dismissed):

hnyu:
Updated the PR to address some questions.

breakds:
Thanks Haonan!

breakds (approved):

Issue #984, opened 2021-08-24
Incorporate `procgen` environments into alf


breakds:
# Motivation

In PPG and maybe other baseline algorithms the authors use [procgen](https://github.com/openai/procgen) as the benchmarking environment. To be able to test our implementations and algorithms on such environment as well, we should add procgen into the suite of environments as well.

Also, I would like to take this opportunity to learn a bit more into details about how environment works with algorithms in alf.

# Goals

1. [x] Package procgen and be able to run procgen environments
2. [x] Be able to train algorithms with procgen
3. [x] Be able to play trained model
4. [x] Be able to record the play of trained model
5. [ ] Create a set of procgen examples in alf

hnyu:
FYI, I plan to write the environment chapter tomorrow.

breakds:
Though procgen itself uses `gym3` interface, fortunately they provide a `gym` wrapper as well. This means that we can use `gym_suite` to load procgen environments e.g. 

```python
alf.config(
    'create_environment',
    env_name='procgen:procgen-chaser-v0',
    num_parallel_environments=8)
```

Based on their official README, the `gym3` version is vectorized and is said to be faster. Without looking into their implementation I doubt using `gym` will be too much slower as I thought the vectorization is mainly used for simulating the environments internally.

breakds:
> FYI, I plan to write the environment chapter tomorrow.

Thanks Haonan! I peeked into the outline of the tutorial and know that there will be on chapter about this. Looking forward to it!

breakds:
Currently the algorithm can pick up the observations in 64 x 64 resolution, which looks like

![bigfish_000](https://user-images.githubusercontent.com/1111035/130538319-4c0dd6d1-a66b-442d-aceb-1cb59ccba40a.png)

![chaser](https://user-images.githubusercontent.com/1111035/130538329-215c0f71-686f-4596-b85f-6c7da0fcd93d.png)

This is clearly downsampled from the original procgen frames. However, I checked the openai's implementation of PPG and confirmed that they are using 64 x 64 resolution as well, so that I will keep it like this.

breakds:
Now I was able to train on procgen environments and play with the trained model, working around the gym3 issue with a special wrapper (thanks to @hnyu 's suggestion).

Video recording in `alf.bin.play` is still not working as expected and I am working on solving it right now.

breakds:
[...]

PR #987, opened 2021-08-25
Procgen Integration (#984)

breakds:
# Motivation

In PPG and maybe other baseline algorithms the authors use [procgen](https://github.com/openai/procgen) as the benchmarking environment. To be able to test our implementations and algorithms on such environment as well, we should add procgen into the suite of environments as well.

However, `procgen` itself is a [gym3](https://github.com/openai/gym3/) based environment which is slightly different from `gym`, even though it provides a compatible layer to `gym`. We need to bridge the gap.

# Solution

`procgen` works out of the box in alf **for training** thanks to its `gym` compatible layer interface. However, there is a gap on rendering - for `gym3` based environments `render()` does nothing and if we want to render it we will need to do it at the construction time.

@hnyu  and I had a discussion and proposed a work around with an environment wrapper. In detail:

1. The wrapper `AlfGym3Wrapper` hijacks the `render()` function. When `render()` is called the first time, it will reconstruct the underlying environment with correct arguments that enables rendering.
2. Ask the `render()` to return the underlying `rgb_array` frame for recording purposes.
3. For `suite_gym`, it will automatically wrap the environment with the above `AlfGym3Wrapper` if it detects the underlying environment is a `procgen` env.

So that the whole process becomes transparent to the user. See the `procgen_conf.py` for an example of how to use it (without seeing the dirty work behind it).

# Testing

Tested with `ActorCriticAlgorithm` on 2 procgen environments `coinrun` and `bigfish` with training, testing and recording.

1. The configuration `ac_procgen_bigfish.py` and `ac_procgen_coinrun.py` aren't in the PR because the trained models are not good. However they were able to demonstrate that `procgen` environment works which is the purpose of this PR.
2. Recorded play movies are attached. Note that for bigfish it learned the trivial policy (lol)

https://user-images.githubusercontent.com/1111035/130845728-44ae7787-5ffd-41cd-bd60-e2e9eb5ca503.mp4

https://user-images.githubusercontent.com/1111035/130845533-05fa1343-e37b-4780-8ed6-42a11909b8cc.mp4


breakds:
I think I am still going to need to add `procgen` to the dependencies in `setup.py` and multiple `requirements.txt`. Other than that the PR is ready for review.

alf/environments/alf_gym3_wrapper.py
```diff
@@ -0,0 +1,69 @@
...
+"""Wrapper for gym-based Gym3 environments
+"""
+
+import gym
+import gym3
+
+from alf.environments.alf_wrappers import AlfEnvironmentBaseWrapper
+
+
+class AlfGym3Wrapper(AlfEnvironmentBaseWrapper):
```
hnyu: This class should inherit gym.Wrapper https://github.com/openai/gym/blob/3abd5d55afa9c43fe5f7da2b1a42f064c14127b0/gym/core.py#L208 because it doesn't involve any ALF data structures. 

And its name doesn't have to include ALF. Maybe just Gym3RenderWrapper? 

(see https://alf-fork.readthedocs.io/en/latest/tutorial/environments_and_wrappers.html and https://alf-fork.readthedocs.io/en/latest/tutorial/environments_and_wrappers.html#alfenvironmentbasewrapper for the difference between wrappers)

alf/environments/alf_wrappers.py
```diff
@@ -113,6 +113,21 @@ def seed(self, seed):
     def wrapped_env(self):
         return self._env
 
+    def replace_wrapped_env(self, new_env):
```
hnyu: As noted above, seems not appropriate to have this in this class.

hnyu (changes requested):

alf/environments/alf_gym3_wrapper.py
breakds: Let me read the tutorial and update the PR.

alf/environments/alf_wrappers.py
breakds: I am so glad that I don't have to add this function. Thank you!

breakds:
PR is updated and tested again. PTAL.

alf/environments/gym_wrappers.py
```diff
@@ -689,3 +689,55 @@ def __init__(self, env):
...
+
+        # Once enabled (set to True), it will be True forever.
+        self._render_enabled = False
+
+    def render(self, mode: str):
+        """Enables rendering by reconstructing the environment
+        """
+        if not self._render_enabled:
+            self.__init__(
+                self.env.spec.make(render=(mode == 'human'), render_mode=mode))
```
hnyu, line 735: Why pass in render=False when mode!="human"?

alf/environments/gym_wrappers.py
```diff
@@ -689,3 +689,55 @@ def __init__(self, env):
...
+            self._render_enabled = True
+
+        # For rgb_array rendering, return the image frame
+        wrapped_env = self.env.env
```
hnyu, line 739: Should be OK to just use `self.env`. In fact there might be a sequence of wrappers, so Gym already makes sure to forward the call recursively. See

https://github.com/openai/gym/blob/bb81e141ea7ae67ce339109095841592e8231185/gym/core.py#L228

alf/environments/suite_gym.py
```diff
@@ -123,8 +123,7 @@ def wrap_env(gym_env,
         # gym3 under the hood with a special rendering scheme. Apply the
         # AlfGym3Wrapper to them before all other wrappers to transparently work
         # that around.
-        from alf.environments.alf_gym3_wrapper import AlfGym3Wrapper
-        gym_env = AlfGym3Wrapper(gym_env)
+        gym_env = gym_wrappers.Gym3RenderWrapper(gym_env)
```
hnyu, line 126: Could potentially make Gym3RenderWrapper do nothing if gym_env is not gym3. So we don't have to check its id (not flexible).

hnyu (changes requested):

alf/environments/gym_wrappers.py
breakds: Because `render=True` means rendering on the screen in `gym3`. It should only be turned on when `render(mode='human')`.

alf/environments/gym_wrappers.py
breakds: Normally yes. But in this case `self.env` is `gym3.interop.ToGymEnv`, and `self.env.env` is `gym3.viewer.ViewerWrapper`. This mean that they are `gym3` wrappers that does not follow the convention. Therefore this is like specialized code logic for `gym3`.

I tested `self.env` does not work before adding this.

alf/environments/suite_gym.py
breakds: Acknowledged but,

1. I think "not wrapping it" is better than "having a wrapper that does nothing" for none-gym3 environments
2. Do not get what does "flexible" mean here - are you suggesting "in the case when there are non gym3 environments whose name starts with `procgen`, this will be broken"? If so I do agree. I can switch it to checking the type of it.

hnyu: > Acknowledged but,
> 
> 1. I think "not wrapping it" is better than "having a wrapper that does nothing" for none-gym3 environments
> 2. Do not get what does "flexible" mean here - are you suggesting "in the case when there are non gym3 environments whose name starts with `procgen`, this will be broken"? If so I do agree. I can switch it to checking the type of it.

If in the future there are other environments written by gym3 then we need to remember to add checks for those environment names, which is trouble,

breakds: Btw, I remembered now that I was trying to avoid `import gym3` if it is not necessary - that is why the previous implementation was on checking the prefix in the environment id.

hnyu (approved):

breakds:
With the updated docker image, CI passed. Will merge now. Thanks for the help, @hnyu !

PR #989, opened 2021-08-26
Implementing encoding networks using Sequential

### LINKED ISSUE #995: Implementing various networks using containers
Code can be significantly reduced.

emailweixu:
Since Sequential is very flexible, most networks can be implemented using it. And it automatically comes with support for make_parallel.

This PR only convert encoding networks to use containers. Future PRs will change other networks.

### DIFF
--- a/alf/algorithms/generator_test.py
+++ b/alf/algorithms/generator_test.py
@@ -142,7 +142,7 @@ class GeneratorTest(parameterized.TestCase, alf.test.TestCase):
                 inputs=None, loss_func=_neglogprob, batch_size=batch_size)
             generator.update_with_gradient(alg_step.info)
 
-        for i in range(2000):
+        for i in range(2100):
             _train()
             if functional_gradient:
                 weight = net._fc_layers[0].weight
--- a/alf/networks/actor_distribution_networks_test.py
+++ b/alf/networks/actor_distribution_networks_test.py
@@ -52,12 +52,13 @@ class TestActorDistributionNetworks(parameterized.TestCase, alf.test.TestCase):
                 actor_fc_layer_params=(64, 32))
             if isinstance(lstm_hidden_size, int):
                 lstm_hidden_size = [lstm_hidden_size]
-            state = []
+            state = [()]
             for size in lstm_hidden_size:
                 state.append((torch.randn((
                     1,
                     size,
                 ), dtype=torch.float32), ) * 2)
+            state.append(())
         else:
             network_ctor = ActorDistributionNetwork
             state = ()
@@ -112,6 +113,7 @@ class TestActorDistributionNetworks(parameterized.TestCase, alf.test.TestCase):
             conv_layer_params=self._conv_layer_params,
             continuous_projection_net_ctor=functools.partial(
                 NormalProjectionNetwork, scale_distribution=True))
+        logging.info("---- %s" % str(actor_dist_net.state_spec))
         act_dist, _ = actor_dist_net(self._image, state)
         actions = act_dist.sample((100, ))
 
@@ -157,7 +159,8 @@ class TestActorDistributionNetworks(parameterized.TestCase, alf.test.TestCase):
         if lstm_hidden_size is None:
             self.assertEqual(state, ())
         else:
-            self.assertEqual(len(state), len(lstm_hidden_size))
+            self.assertEqual(
+                len(alf.nest.flatten(state)), 2 * len(lstm_hidden_size))
 
     def test_make_parallel(self):
         obs_spec = TensorSpec((3, 20, 20), torch.float32)
--- a/alf/networks/actor_distribution_networks_test.py
+++ b/alf/networks/actor_distribution_networks_test.py
@@ -113,7 +113,6 @@ class TestActorDistributionNetworks(parameterized.TestCase, alf.test.TestCase):
             conv_layer_params=self._conv_layer_params,
             continuous_projection_net_ctor=functools.partial(
                 NormalProjectionNetwork, scale_distribution=True))
-        logging.info("---- %s" % str(actor_dist_net.state_spec))
         act_dist, _ = actor_dist_net(self._image, state)
         actions = act_dist.sample((100, ))
 
--- a/alf/networks/encoding_networks.py
+++ b/alf/networks/encoding_networks.py
@@ -400,9 +400,6 @@ class EncodingNetwork(_Sequential):
         """Make a parallelized version of ``module``.
 
         A parallel network has ``n`` copies of network with the same structure but
-        different independently initialized parameters.
-
-        A parallel network has ``n`` copies of network with the same structure but
         different independently initialized parameters. The parallel network can
         process a batch of the data with shape [batch_size, n, ...] using ``n``
         networks with same structure.
@@ -460,8 +457,8 @@ def ParallelEncodingNetwork(input_tensor_spec,
                             last_kernel_initializer=None,
                             last_use_fc_bn=False,
                             name="ParallelEncodingNetwork"):
-    """Parallel feed-forward network with FC layers which allows the last layer
-    to have different settings from the other layers.
+    """Parallel encoding network which effectively runs ``n`` individual encoding
+    network simultaneuosl.
 
     Args:
         input_tensor_spec (nested TensorSpec): the (nested) tensor spec of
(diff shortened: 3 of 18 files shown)


alf/networks/encoding_networks.py
```diff
@@ -566,49 +306,60 @@ def __init__(self,
...
                 variance_scaling_init,
                 mode='fan_in',
                 distribution='truncated_normal',
                 nonlinearity=activation)
 
-        self._img_encoding_net = None
+        spec = input_tensor_spec
+        nets = []
+
+        if input_preprocessors:
```
emailweixu, line 319: There is a slight change for the handling of input_preprocessors. Previously, in PreprocessorNetwork, each input preprocessor will be copied first. The new version will not copy. @Haichao-Zhang do you remember why copy is needed?

alf/algorithms/generator_test.py
```diff
@@ -142,7 +142,7 @@ def _train():
                 inputs=None, loss_func=_neglogprob, batch_size=batch_size)
             generator.update_with_gradient(alg_step.info)
 
-        for i in range(2000):
+        for i in range(2100):
```
emailweixu, line 145: This is due to input preprocessor is not copied and causes different parameter initialization. See comment for encoding network.

alf/networks/encoding_networks.py
Haichao-Zhang: > There is a slight change for the handling of input_preprocessors. Previously, in PreprocessorNetwork, each input preprocessor will be copied first. The new version will not copy. @Haichao-Zhang do you remember why copy is needed?

@emailweixu Yes, the reason for this is that it will make sure when we create parallel networks with pre-processors (could have learnable parameters), the pre-processors will also be copied, unless explicitly specified not to copy. 
This makes the expected behavior more natural, and fixed some previously encountered errors due to the mis-match between the thought behavior and actual behavior.

Some old related issues and PRs:
issue: https://github.com/HorizonRobotics/alf/issues/552
PR: https://github.com/HorizonRobotics/alf/pull/560

emailweixu: The make_parallel of the container will copy all the components. So there should be any problem with this change if the original motivation is to ensure make_parallel work correctly.

alf/networks/actor_distribution_networks_test.py
```diff
@@ -112,6 +113,7 @@ def test_continuous_actor_distribution(self, lstm_hidden_size):
             conv_layer_params=self._conv_layer_params,
             continuous_projection_net_ctor=functools.partial(
                 NormalProjectionNetwork, scale_distribution=True))
+        logging.info("---- %s" % str(actor_dist_net.state_spec))
```
Haichao-Zhang: can remove this line

alf/networks/encoding_networks.py
```diff
@@ -641,257 +392,148 @@ def __init__(self,
...
+                            fc_layer_params=None,
+                            activation=torch.relu_,
+                            kernel_initializer=None,
+                            use_fc_bn=False,
+                            last_layer_size=None,
+                            last_activation=None,
+                            last_kernel_initializer=None,
+                            last_use_fc_bn=False,
+                            name="ParallelEncodingNetwork"):
+    """Parallel feed-forward network with FC layers which allows the last layer
```
Haichao-Zhang: `feed-forward network with FC layers` -> `encoding network`

alf/networks/value_networks_test.py
```diff
@@ -66,6 +67,7 @@ def test_value_distribution(self, lstm_hidden_size):
                     conv_layer_params=conv_layer_params), None
             ],
             preprocessing_combiner=NestConcat())
+        logging.info("----%s" % str(value_net.state_spec))
```
Haichao-Zhang: can remove this line

alf/networks/encoding_networks.py
```diff
@@ -641,257 +392,148 @@ def __init__(self,
...
-            (alf.layers.Identity, alf.layers.NestSum, alf.layers.NestConcat))):
-            parallel_enc_net_args = dict(**self.saved_args)
-            parallel_enc_net_args.update(n=n, name="parallel_" + self.name)
-            return ParallelEncodingNetwork(**parallel_enc_net_args)
-        else:
-            common.warning_once(
-                " ``NaiveParallelNetwork`` is used by ``make_parallel()`` !")
-            return super().make_parallel(n)
+        A parallel network has ``n`` copies of network with the same structure but
+        different independently initialized parameters. The parallel network can
```
Haichao-Zhang: Line 405-406 repeats Line 402-403 and one of them can be removed.

alf/networks/encoding_networks_test.py
```diff
@@ -187,7 +183,7 @@ def test_encoding_network_input_preprocessor(self):
         self.assertEqual(output.size()[1], 1)
 
     @parameterized.parameters((True, ), (False, ))
-    def test_encoding_network_nested_input(self, lstm):
+    def test_encoding_network_nested_input(self, lstm=False):
```
Haichao-Zhang: why do we  need to set `lstm=False` here?

alf/networks/encoding_networks.py
```diff
@@ -641,257 +392,148 @@ def __init__(self,
...
+        process a batch of the data with shape [batch_size, n, ...] using ``n``
+        networks with same structure.
 
+        TODO: remove ``allow_non_parallel_input``. This means to make parallel network
```
Haichao-Zhang, line 407: The option `allow_non_parallel_input` seems to be useful and we can keep it. Does it make sense to set default value of `allow_non_parallel_input=True`, so that the input will always be handled correctly by default?

alf/networks/encoding_networks.py
emailweixu: fixed.

alf/networks/encoding_networks.py
emailweixu: A network can be a component of big network (container). It does not need the ability to handle non-parallel input in those situations.

alf/networks/encoding_networks.py
emailweixu: fixed

alf/networks/value_networks_test.py
emailweixu: removed.

alf/networks/encoding_networks_test.py
emailweixu: It was for debugging. Restored.

alf/networks/actor_distribution_networks_test.py
Haichao-Zhang: > can remove this line

This line is not removed yet.

alf/utils/checkpoint_utils_test.py
```diff
@@ -558,17 +558,17 @@ def _check_parallel_param(p_net_source):
...
         p_net_w_preprocessor = network_w_preprocessor.make_parallel(replicas)
 
-        # the number of parameters of parallel network with input_preprocessor
-        # should be equal to that of the naive parallel network without
-        # input processor + the number of parameters of input processor * replicas
-        self.assertEqual(
-            len(p_net_w_preprocessor.state_dict()),
-            len(p_net_wo_preprocessor.state_dict()) +
-            replicas * len(input_preprocessors.state_dict()))
+        if lstm:
```
Haichao-Zhang: Why do we need this `if lstm` condition here and also in L587?

emailweixu: This test relies on the fact the p_net_w_preprocessor is naively parallelized, which is no longer correct now.
Changed the test to not replying on this.

alf/utils/checkpoint_utils_test.py
```diff
@@ -584,14 +572,14 @@ def _check_parallel_param(p_net_source):
...
-            # the number of parameters of parallel network with a shared
-            # input_preprocessor should be equal to that of the naive parallel
-            # network without input processor + the number of parameters of input processor
-            self.assertEqual(
-                len(p_net_w_shared_preprocessor.state_dict()),
-                len(p_net_wo_preprocessor.state_dict()) + len(
-                    input_preprocessors.state_dict()))
+        # the number of parameters of a parallel network with a shared
+        # input_preprocessor should be equal to that of the parallel network
+        # with non-shared input processor - replicas * the number of parameters
```
Haichao-Zhang: In the comment, it should be `(replicas-1) *  ...`, same as the code.

Haichao-Zhang (commented):
Look great! Just one final comment.

Haichao-Zhang (approved):

PR #991, opened 2021-08-28
Implement the Impala CNN encoding network for procgen environments

breakds:
# Motivation

To create an encoder for the procgen environments because 

1. The encoder that works for the atari game "breakout" does not work well for procgen games
2. In the original OpenAI's implementation of PPG/PPO, such Impala CNN encoder (named after the Impala algorithm) is used

# Solution

Replicate the implementation of the Impala CNN encoder with network containers.

The network implementation is under `examples` because

1. It does not seem to be generally useful at this moment
2. In the future if we find it more generally useful, it is still flexible enough to move it to `alf/networks`

Therefore this is a "defer-the-decision" decision.

# Testing

1. Unit tests that verifies the expected output specs
2. Tested with a (not yet checked-in) [configuration](https://github.com/HorizonRobotics/alf/blob/PR_breakds_ppo_procgen/alf/examples/ppo_procgen_goldrun_conf.py) for ppo + procgen

Originally I tried training with a configuration adapted from `ac_breakout` and the average return remains around zero even after thousands of iterations. With this encoder and PPO reaches close to 10.0 (in `goldrun`, 10.0 means reaching the gold coin) after only 300 iterations. I think this confirms the effectiveness of this encoder. **However**, each iteration is much slower.

ppo_procgen

**NOTE**: nothing is specifically done about the parameters initialization (yet).


breakds:
> # Motivation
> To create an encoder for the procgen environments because
> 
> 1. The encoder that works for the atari game "breakout" does not work well for procgen games
> 2. In the original OpenAI's implementation of PPG/PPO, such Impala CNN encoder (named after the Impala algorithm) is used
> 
> # Solution
> Replicate the implementation of the Impala CNN encoder with network containers.
> 
> The network implementation is under `examples` because
> 
> 1. It does not seem to be generally useful at this moment
> 2. In the future if we find it more generally useful, it is still flexible enough to move it to `alf/networks`
> 
> Therefore this is a "defer-the-decision" decision.
> 
> # Testing
> 1. Unit tests that verifies the expected output specs
> 2. Tested with a (not yet checked-in) [configuration](https://github.com/HorizonRobotics/alf/blob/PR_breakds_ppo_procgen/alf/examples/ppo_procgen_goldrun_conf.py) for ppo + procgen
> 
> Originally I tried training with a configuration adapted from `ac_breakout` and the average return remains around zero even after thousands of iterations. With this encoder and PPO reaches close to 10.0 (in `goldrun`, 10.0 means reaching the gold coin) after only 300 iterations. I think this confirms the effectiveness of this encoder.

Note that nothing is done specifically about initialization of the parameters.
> 
> ppo_procgen

alf/examples/networks/impala_cnn_encoder.py
```diff
@@ -0,0 +1,186 @@
...
+            in each of the CNN downsampling stack. This parameter is uniform to
+            all the CNN downsampling stacks.
+        output_size (int): the size of the output vector for each of the input
+            image in the mini batch.
+        kernel_initializer (Callable[[Tensor], None]): initializer for
+            the Conv2D and FC layers in the network.
+    
+    Returns:
+        
+        A module representing the Impala CNN encoding network.
```
Haichao-Zhang, line 168: It would be useful to highlight the difference of `Impala CNN` compared to a standard resnet CNN.

breakds: Gotcha, will do. Let me read about the characteristics of a standard resnet CNN.

breakds: Added a `README.md` to explain that. To preview the README file, please visit: https://github.com/HorizonRobotics/alf/tree/PR_breakds_impala_cnn/alf/examples/networks

alf/examples/networks/impala_cnn_encoder.py
```diff
@@ -0,0 +1,186 @@
...
+            tensor omitting batch size should be (C, W, H), where C is the input
+            number of channels.
+        cnn_channel_list (List[int]): a list that specifies the number of output
+            channels of each CNN downsampling stack in the application order.
+        num_blocks_per_stack (int): specifies how many residual blocks to apply
+            in each of the CNN downsampling stack. This parameter is uniform to
+            all the CNN downsampling stacks.
+        output_size (int): the size of the output vector for each of the input
+            image in the mini batch.
+        kernel_initializer (Callable[[Tensor], None]): initializer for
```
Haichao-Zhang: It seems that this format on type (`Callable[[Tensor], None]`) is a bit different from existing ones. Maybe can consider using something like ``Callable[Tensor] | None``?

Also for Line 42.

Haichao-Zhang (commented):
LG. Just one minor comment.

alf/examples/networks/impala_cnn_encoder.py
breakds: Actually `Callable[[Tensor], None]` means it is a function like `f(x: Tensor) -> None`. This is python's type hints standard, while `Callable | None` means `Optional[Callable]`, meaning that the argument could be `None`.

Here the type hint is used to indicate the former.

breakds:
> LG. Just one minor comment.

Thanks. See reply.

alf/examples/networks/impala_cnn_encoder.py
Haichao-Zhang: > Actually `Callable[[Tensor], None]` means it is a function like `f(x: Tensor) -> None`. This is python's type hints standard, while `Callable | None` means `Optional[Callable]`, meaning that the argument could be `None`.
> 
> Here the type hint is used to indicate the former.

Understand. What I meant was indeed the latter case. Actually, if we pass a `None` value to `kernel_initializer`, a variance scaling based initializer will be used.

Haichao-Zhang (approved):

alf/examples/networks/impala_cnn_encoder.py
breakds: Ah I see! Sorry that I misunderstood your comments. I agree with you that the consistent interface would be better. In this case at least the public interface `create` is better to be consistent. Will update that in another PR. Appreciate the explanation!

PR #992, opened 2021-08-28
added igibson environment files

junyaoshi:


alf/environments/igibson/configs/turtlebot_clip.yaml
```diff
@@ -0,0 +1,76 @@
+# scene
```
hnyu, line 1: The name of this yaml file is confusing; the bot conf has nothing to do with "clip" actually.

alf/environments/igibson/igibson_env.py
```diff
@@ -0,0 +1,104 @@
...
+            elif self.config['task'] == 'interactive_nav_random':
+                self.task = InteractiveNavRandomTask(self)
+            elif self.config['task'] == 'dynamic_nav_random':
+                self.task = DynamicNavRandomTask(self)
+            elif self.config['task'] == 'reaching_random':
+                self.task = ReachingRandomTask(self)
+            elif self.config['task'] == 'room_rearrangement':
+                self.task = RoomRearrangementTask(self)
+            elif self.config['task'] == 'visual_object_nav':
+                self.task = VisualObjectNavTask(self)
```
hnyu, line 102: If the only difference with the base class is "VisualObjectNavTask", then you can simply overwrite this function as:

``` python
def load_task_setup(self):
    super().load_task_setup()
    if self.task is None and self.config['task'] == 'visual_object_nav':
        self.task = VisualObjectNavTask(self)
```

alf/environments/igibson/igibson_env.py
```diff
@@ -0,0 +1,104 @@
...
+from igibson.tasks.room_rearrangement_task import RoomRearrangementTask
+from igibson.tasks.point_nav_fixed_task import PointNavFixedTask
+from igibson.tasks.point_nav_random_task import PointNavRandomTask
+from igibson.tasks.interactive_nav_random_task import InteractiveNavRandomTask
+from igibson.tasks.dynamic_nav_random_task import DynamicNavRandomTask
+from igibson.tasks.reaching_random_task import ReachingRandomTask
+from alf.environments.igibson.igibson_tasks import VisualObjectNavTask
+
+
+class iGibsonCustomEnv(iGibsonEnv):
```
hnyu, line 25: Probably we should decorate this class with `@alf.configurable` so that it's possible for us to configure the env in an alf config file.

alf/environments/igibson/igibson_object.py
```diff
@@ -0,0 +1,48 @@
...
+                igibson/data/ig_dataset/objects/'name'
+            scale (float): scale of object
+        """
+        dirname = os.path.join(igibson.ig_dataset_path, 'objects', name)
+        object_dir = [f.path for f in os.scandir(dirname) if f.is_dir()][0]
+        object_name = os.path.basename(object_dir)
+        filename = os.path.join(object_dir, f'{object_name}.urdf')
+        super(iGibsonObject, self).__init__(filename, scale)
+
+    def load(self):
```
hnyu, line 36: Why do you want to have this function? It's exactly the same with `load()` of the base class `Object`:
https://github.com/StanfordVL/iGibson/blob/afe24c6aa95c8d4e74ce6d6b5eae2f479f2807ae/igibson/objects/object_base.py#L26

alf/environments/igibson/igibson_object.py
```diff
@@ -0,0 +1,48 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+import igibson
+from igibson.objects.articulated_object import ArticulatedObject
+
+
+class iGibsonObject(ArticulatedObject):
```
hnyu, line 20: Seems to me that this class is unnecessary. It only has some preprocessing of converting a name to a path. You can just do this in your local script.

alf/environments/igibson/suite_igibson.py
```diff
@@ -0,0 +1,384 @@
...
+        return self._flatten(observation, flatten)
+
+    def _need_flatten(self, shape):
+        if len(shape) == 2:
+            return True
+        return False
+
+    def _flatten(self, np_array, flatten=False):
+        """
+        Note that the extra copy() after np.transpose is crucial to pickle dump speed
```
hnyu, line 69: This docstring is incorrect.

alf/environments/igibson/suite_igibson.py
```diff
@@ -0,0 +1,384 @@
...
+import alf
+
+from igibson.objects.visual_marker import VisualMarker
+from alf.environments.igibson.igibson_tasks import VisualPointNavFixedTask, \
+    VisualPointNavRandomTask, VisualObjectNavTask
+from alf.environments.igibson.igibson_env import iGibsonCustomEnv
+
+
+@alf.configurable()
+class FlattenWrapper(BaseObservationWrapper):
```
hnyu, line 31: Probably we can directly use built-in gym.wrappers to do this (just provide all keys we are interested):

``` python
        try:  # for modern Gym (>=0.15.4)
            from gym.wrappers import FilterObservation, FlattenObservation
            env = FlattenObservation(FilterObservation(env, keys))
        except ImportError:  # for older gym (<=0.15.3)
            from gym.wrappers import FlattenDictWrapper  # pytype:disable=import-error
            env = FlattenDictWrapper(env, keys)
```

so no need to define our own wrapper.

alf/environments/igibson/suite_igibson.py
```diff
@@ -0,0 +1,384 @@
...
+        https://stackoverflow.com/questions/26998223/
+        """
+
+        if flatten:
+            np_array = np_array.flatten()
+        return np_array.copy()
+
+
+@alf.configurable()
+class StackRGBDWrapper(gym.ObservationWrapper):
```
hnyu, line 82: Probably we don't need this wrapper. We can instead define a network preprocessor ``NestConcat(dim=1)`` to concat two input tensors along the channels dim (assuming it's dim 1).

alf/environments/igibson/suite_igibson.py
```diff
@@ -0,0 +1,384 @@
...
+            observation_space.low, fill_value=0, dtype=np.uint8)
+        obs_high = np.full_like(
+            observation_space.high, fill_value=255, dtype=np.uint8)
+        assert obs_low.shape == obs_high.shape
+        observation_space = gym.spaces.Box(
+            low=obs_low, high=obs_high, dtype=np.uint8)
+        return observation_space
+
+    def transform_observation(self, observation):
+        return (observation * 255).astype(np.uint8)
```
hnyu, line 145: This is not "one-hot". One-hot vector is a zero vector of length N except that only one dim is 1. And you can only convert an integer 0<=n<N (not a floating number) to a one-hot vector of length N.

If the input task obs is an integer, then we don't need a wrapper to do this. Just use `EmbeddingPreprocessor` to convert it to an embedding when configuring the network.

alf/environments/igibson/suite_igibson.py
```diff
@@ -0,0 +1,384 @@
...
+
+
+@alf.configurable()
+class ImageIntWrapper(BaseObservationWrapper):
```
hnyu, line 149: Probably we should set the image type to uint8 in the first place when it's generated by the simulator, instead of using a wrapper to do this (which is inefficient).

alf/environments/igibson/suite_igibson.py
```diff
@@ -0,0 +1,384 @@
...
+        super().__init__(env)
+        self.env = env
+        self.metadata['render.modes'] = ['rgb_array', 'human']
+
+    def render(self, mode='human', **kwargs):
+        if 'vision' in self.env.sensors:
+            vision_obs = self.sensors['vision'].get_obs(self)
+            if 'rgb' in vision_obs:
+                rgb_obs = vision_obs['rgb']
+                rgb_obs = (rgb_obs * 255).astype(np.uint8)
```
hnyu, line 183: If both 'rgb' and 'depth' are generated as uint8 in the first place by the simulator, then the conversions here can be skipped.

alf/environments/igibson/suite_igibson.py
```diff
@@ -0,0 +1,384 @@
...
+                                              axis=1)
+                    return rgbd_obs
+            else:
+                raise AttributeError("robot doesn't have modality: rgb")
+        else:
+            raise AttributeError("robot doesn't have sensor: vision")
+
+
+@alf.configurable()
+class GFTTupleInputWrapper(gym.ObservationWrapper):
```
hnyu, line 200: Seems not quite necessary? Before calling GFT, if the `time_step.observation` is a nest (any structure including dict), you can first do `alf.nest.flatten(time_step.observation)` to convert it to a list/tuple.

hnyu (changes requested):
Good job! Some preliminary comments. I haven't looked at igibson_tasks.py and done the test yet. Will do them later.

alf/environments/igibson/igibson_tasks.py
```diff
@@ -0,0 +1,816 @@
...
+    def __init__(self, env, target_pos=[5, 5, 0], target_pos_vis_obj=None):
+        """
+        Args:
+            env (igibson.igibson_env.iGibsonCustomEnv): environment instance
+            target_pos (list[float]): x, y, z coordinates of the target position
+            target_pos_vis_obj (igibson.objects.object_base.Object): the visual object
+                at the target position
+        """
+        super(VisualPointNavFixedTask, self).__init__(env)
+        self.reward_type = self.config.get('reward_type', 'l2')
```
hnyu, line 53: According to ALF's name convention, please prefix all class members with '_'.

alf/environments/igibson/igibson_tasks.py
```diff
@@ -0,0 +1,816 @@
...
+import numpy as np
+import os
+import time
+import logging
+import pybullet as p
+
+from alf.environments.igibson.igibson_object import iGibsonObject
+
+
+class VisualPointNavFixedTask(BaseTask):
```
hnyu, line 38: It's better to derive this class from PointNavFixedTask because most functions have already been implemented there, so it's unnecessary to redo the work here.

alf/environments/igibson/igibson_tasks.py
```diff
@@ -0,0 +1,816 @@
...
+
+        x, y, z = target_pos
+        self.target_pos = np.array([x, y, z + self.z_offset])
+        self.initial_pos = initial_pos
+        self.initial_orn = initial_orn
+
+        super(VisualPointNavRandomTask, self).reset_agent(env)
+
+
+class VisualObjectNavTask(BaseTask):
```
hnyu, line 402: This class can be merged with VisualObjectNavFixedTask and VisualPointNavRandomTask. All three tasks can be formulated as randomizing (or fixing) a set of locations (number>=1) when resetting and asking the robot to reach the target location. In the case where there is no object concept, the object id can be an arbitrary integer (e.g., 0). 

It's best to reduce redundant code as much as possible, so it's less error-prone and easier to maintain in the future.

hnyu (changes requested):

PR #997, opened 2021-09-01
Implement the AlfEnvironment derived procgen envirnonment that is natively parallel

breakds:
# Motivation

Previously we depend on Gym3's official gym wrapper to bridge `procgen` environments and Alf algorithms. It is rather inefficient as creating one process for each instance of Gym3 environment will consume a lot of memory and runs slowly.

Since Gym3 supports batched environment natively, the new implementation should take advantage of that.

# Solution

Implemented a new wrapper `AlfGym3Wrapper` and with this it bypasses the needs to have `ProcessEnvironment` and `ParallelAlfEnvironment` to support batched environments.

The new loader `suite_procgen.load` is provided to be directly used in the configuration. Also

1. Removed the previous `Gym3RenderWrapper`.
2. Updated the base configuration for `procgen`.

# Testing

1. Tested training works and per iteration time goes from 10.8s on average to 7.8s on average.
2. Tested rendering and recording via `play.py`. Video attached.

https://user-images.githubusercontent.com/1111035/131761537-03ac6177-bd61-4f98-92d6-7fb2e4d9e62e.mp4


alf/environments/alf_gym3_wrapper.py
```diff
@@ -0,0 +1,407 @@
...
+                procgen allows sending action -1 to reset the environments. Set
+                this to True to enable such behavior.
+            render_activator (Optional[Callable[[], gym3.Env]]): when set to
+                None, it indicates that this environment does not support
+                rendering. Otherwise it will be a function that re-creates a
+                Gym3 environment with render enabled. See render() for details.
+            frame_extractor (Optional[Callable[[gym3.Env], Any]]): when set to
+                None, it indicates that this environment does not support
+                recording. Otherwise it will be a function that extracts the
+                rendered frame for recording from the environment.
```
emailweixu: It seems redundant to document the types of the arguments if type hints are used. We should update the coding standard at https://github.com/HorizonRobotics/alf/blob/pytorch/docs/contributing.rst#coding-standard and tools/codestyle.

alf/environments/alf_gym3_wrapper.py
```diff
@@ -0,0 +1,407 @@
...
+        self._gym3_env = gym3_env
+
+        self._support_force_reset = support_force_reset
+
+        # +--------------------------+
+        # | Render/Recording Related |
+        # +--------------------------+
+
+        # When initially constructed, render is not enabled until the first call
+        # to render() is invoked. USe self._render_enabled to make sure render
```
emailweixu: USe -> use

alf/environments/alf_gym3_wrapper.py
```diff
@@ -0,0 +1,407 @@
...
+        # | Stateful Contexts        |
+        # +--------------------------+
+
+        # A list representing whether the corresponding single environment
+        # finishes the current episode
+        self._done = [False] * self.batch_size
+
+    @property
+    def batched(self):
+        return self._gym3_env.num > 1
```
emailweixu: This should be True if it has a batch dimension.

alf/environments/alf_gym3_wrapper.py
```diff
@@ -0,0 +1,407 @@
...
+        np_action = nest.map_structure(lambda x: x.cpu().numpy(), action)
+        self._gym3_env.act(np_action)
+
+        reward, observation, first = self._gym3_env.observe()
+
+        # Override the obervation with the previous observation if that
+        # particular environment has ``first=True``.
+        for i in range(self.batch_size):
+            if first[i]:
+                observation[i] = prev_observation[i]
```
emailweixu: I think the type of observation  can be  `Dict[str, np.ndarray]`. The unittest should test both Tensor observation space and dict observation space.

alf/environments/alf_gym3_wrapper.py
```diff
@@ -0,0 +1,407 @@
...
+            step_type=step_type,
+            action=action)
+
+        self._done = first
+
+        return time_step
+
+    def render(self, mode: str):
+        """Enables rendering by re-activating the environment
+        """
```
emailweixu, line 427: document args

alf/environments/suite_procgen.py
```diff
@@ -0,0 +1,67 @@
...
+
+def _load_procgen(env_name: str, batch_size: int = 1, render: bool = False):
+    env = ProcgenGym3Env(
+        num=batch_size,
+        env_name=env_name,
+        render_mode='rgb_array' if render else None)
+    # This extracts the 'rgb' field in the observation dictionary
+    # every time when step() is called, so that the observation is
+    # an image that our algorithm can consume.
+    env = gym3.ExtractDictObWrapper(env, 'rgb')
```
emailweixu, line 29: Does the original observation have other field other than 'rgb'?

alf/environments/suite_procgen.py
```diff
@@ -0,0 +1,67 @@
...
+        batch_size (int): the number of parallel environments to run
+            simultaneously.
+
+    """
+    env = ProcgenGym3Env(
+        num=batch_size, env_name=env_name, render_mode='rgb_array')
+    # This extracts the 'rgb' field in the observation dictionary
+    # every time when step() is called, so that the observation is
+    # an image that our algorithm can consume.
+    env = gym3.ExtractDictObWrapper(env, 'rgb')
```
emailweixu: not used

alf/environments/alf_gym3_wrapper.py
```diff
@@ -0,0 +1,407 @@
...
+
+        With this wrapper we CHANGED that definition, so that the frame with
+        ``first=True`` actually becomes the LAST FRAME of the previous episode.
+        Because the observation of the new episode can be DRAMATICALLY DIFFERENT
+        from the actual last frame of the previous episode, the previous
+        observation is returned in this case.
+
+        Therefore to summarize, effectively we will
+
+        1. Repeat the end-of-episode observation twice for each episode.
```
emailweixu, line 367: Since openai ppg is directly trained on gym3 interface, which doesn't have the repeated frame at the and the skipped frame at the beginning, our training may behave slightly differently.

emailweixu (changes requested):

alf/environments/alf_gym3_wrapper.py
breakds: Agree with that. Added a PR for that https://github.com/HorizonRobotics/alf/pull/998

alf/environments/alf_gym3_wrapper.py
breakds: Thanks. Will be fixed.

alf/environments/alf_gym3_wrapper.py
breakds: Will fix.

alf/environments/suite_procgen.py
breakds: Will remove. Good catch!

alf/environments/alf_gym3_wrapper.py
breakds: Acknowledged. Right as we discussed about it, this seems the best short-term handling.

The problem with the `seed` and internal `done` approach is that after reading their code and experiment again, both of them are also "late signals", i.e. come at the next frame - the first frame of a new episode.

breakds:
Still working on the improvements. Will notify the reviewer when the updates are done.

alf/environments/alf_gym3_wrapper.py
breakds: Fixed.

alf/environments/alf_gym3_wrapper.py
breakds: Removed redundant type hints in the docstring.

alf/environments/alf_gym3_wrapper.py
emailweixu: How about this one?

alf/environments/suite_procgen.py
breakds: Yes for `procgen` there is only one field `rgb`.

alf/environments/alf_gym3_wrapper.py
breakds: Yes the observation could be of type dictionary. But instead of the normal `Dict`, it was `gym3`'s `DictType`, which is not recognized by `nest` as nested structure.

We can easily convert `DictType` to `Dict` if it only appears at the top level of observation structure, but I haven't found that assumption being mentioned yet, except for:

> The observations and actions can be either arrays, or "trees" of arrays, where a tree is a (potentially nested) dictionary with string keys.

emailweixu: If DictType cannot be handled,  map_structure at line 285 and 362 will not do the right job.

breakds: Yep I realized that. I am thinking about the best way to handle it. The quote above seem to suggest that only the first level in observation/action could be `DictType`. Can we note this assumption and handle accordingly?

emailweixu: Perhaps can do something similar to the _gym_space_to_nested_space() in gym_wrappers.py

alf/environments/alf_gym3_wrapper.py
```diff
@@ -0,0 +1,415 @@
...
+            if first[i]:
+                observation[i] = prev_observation[i]
+
+        # This does the trick of repeating end-of-episode frames and throwing
+        # away first-of-episode frames.
+        step_type = [
+            ds.StepType.FIRST if d else
+            (ds.StepType.LAST if f else ds.StepType.MID)
+            for d, f in zip(self._done, first)
+        ]
```
hnyu, line 405: Will this create one-frame misbatch between the rendered frame and policy action when playing a model?

alf/environments/alf_gym3_wrapper.py
```diff
@@ -0,0 +1,415 @@
...
+
+        time_step = self._create_time_step(
+            reward=reward,
+            observation=observation,
+            step_type=step_type,
+            action=action)
+
+        self._done = first
```
hnyu: Maybe self._done is not a good name anymore. Can be called `self._prev_first`.

breakds: It was actually called `_prev_first` before renaming to `_done`. In our interpretation, the frame whose `first` is `true` is actually "the last frame of the previous episode", which means `done`. But I am open to the suggestion.

alf/environments/alf_gym3_wrapper.py
hnyu: Actually I think there might be some problem for value learning if the end-of-episode frame is repeated twice, depending on the reward of a "first" frame and the discount gamma. In theory the value network will return the same value given the two identical frames, then the bellman equation won't be satisfied between the last two frames? 
Depending on how the episode ends, 
1. normal end like success or failure, then because the value of StepType.LAST is forced to be 0, so value network is only called once and there will be no issue.
2. timeout. The same value network will be called twice on the same frame to get V, for example. Then we have a TD learning loss between the last two identical frames: `(V + r - \gamma V)^2` which will never be zero. 

In practice, this might just have a slight impact in value learning.

alf/environments/alf_gym3_wrapper.py
```diff
@@ -0,0 +1,415 @@
...
+        self.metadata = {'render.modes': []}
+        if self._render_activator is not None and self._frame_extrator is not None:
+            self.metadata['render.modes'].append('rgb_array')
+
+        # +--------------------------+
+        # | Cache the Tensor Specs   |
+        # +--------------------------+
+
+        # NOTE(breakds): when needed, expose this and allow an user to set it.
+        self._discount = torch.full((self.batch_size, ), 1.0)
```
hnyu: Will procgen games have normal episode ends like success or failure or game-over? If so, then in those cases, self._discount should be 0.

breakds: Yep that is pretty much what `_done` is doing. Thanks a lot for pointing this out - I originally thought `step type` will have full information of this, but that is not used in computing advantages/accumulated returns.

alf/environments/alf_gym3_wrapper.py
breakds: Right, I thought about this when I see `gym3`'s official handling on this and it does put two identical frames at the end of each episode. My understanding is that when the length of the episode is long enough, the number of end frame is actually small compared to other frames, and hopefully this does not alter the training too much.

> timeout. The same value network will be called twice on the same frame to get V, for example. Then we have a TD learning loss between the last two identical frames: (V + r - \gamma V)^2 which will never be zero.

Totally agree with the analysis here.

alf/environments/alf_gym3_wrapper.py
breakds: This is now handled assuming the first layer of would be a `DictType` in `gym3`. Also updated the unit test to remove the `ExtractDictObWrapper`. This makes sure that both cases are tested because:

1. `observation_spec` in this case is a `DictType[Discrete Tensor]`, while
2. `action_spec` is a `Discrete Tensor`.

alf/environments/alf_gym3_wrapper.py
breakds: I added a `TODO` to 

1. acknowledge the issue here
2. as a reminder to more properly fix it in the future.

alf/environments/alf_gym3_wrapper.py
breakds: This is fixed.

alf/environments/alf_gym3_wrapper.py
breakds: Renamed back to `_prev_first`.

breakds:
The PR is updated based last round of review from @emailweixu  and @hnyu . Really appreciate the suggestions and comments! PTAL.

hnyu (dismissed):

alf/environments/alf_gym3_wrapper.py
emailweixu: I am wondering if observation is a dict, what is meaning of observation[i]?

breakds: Getting sloppy here because the unit test isn't exactly covering this case. In my original unit test it did not properly test the intermediate `FIRST` frame case. Let me find a way to actually test this - since this is `procgen`, I could achieve this by sending `-1` to one of the batch environments.

breakds: This is now fixed and added unit test to explicitly test such special handling. PTAL. Thanks for catching this!

alf/environments/alf_gym3_wrapper.py
```diff
@@ -381,9 +381,15 @@ def _step(self, action) -> ds.TimeStep:
...
-                observation[i] = prev_observation[i]
+
+        def __override_with_prev_observation(ob_array: np.ndarray,
+                                             prev_ob_array: np.ndarray):
+            new_ob_array = ob_array
+            new_ob_array[first] = prev_ob_array[first]
+            return new_ob_array
+
+        observation = nest.map_structure(__override_with_prev_observation,
+                                         observation, prev_observation)
```
hnyu: Could just simplify this as:

``` python 
observation = nest.map_structure(lambda ob, prev_ob: ob[first] = prev_ob[first], observation, prev_observation)
```

breakds: IMHO `map_structure` is functional, so side effect shouldn't be allowed here. Also the suggested simplification is not clear about what is returned in `nest.map_structure` call, right?

Tried in action and confirmed the idea:

1. `map_structure` does not support side effect
2. `lambda` does not support assignment (new finding, but reasonable lol)
3. Replace the `lambda` with a function that contains assignment, `map_structure` returns `None` (also reasonable).

hnyu: > IMHO `map_structure` is functional, so side effect shouldn't be allowed here. Also the suggested simplification is not clear about what is returned in `nest.map_structure` call, right?
> 
> Tried in action and confirmed the idea:
> 
> 1. `map_structure` does not support side effect
> 2. `lambda` does not support assignment (new finding, but reasonable lol)
> 3. Replace the `lambda` with a function that contains assignment, `map_structure` returns `None` (also reasonable).

You're right. Need to use a function instead of lambda. But map_structure does allow side effect, so it's OK to do:

```python
def __override(obs, prev_obs):
    obs[first] = prev_obs[first]

nest.map_structure(__override, observation, prev_observation)
```

breakds: Sure thing ~ Updated with the side-effect version now.

alf/environments/alf_gym3_wrapper.py
```diff
@@ -0,0 +1,441 @@
...
+
+        Gym3's official Gym wrapper DID THE SAME.
+
+        """
+        _, prev_observation, _ = self._gym3_env.observe()
+
+        np_action = nest.map_structure(lambda x: x.cpu().numpy(), action)
+        self._gym3_env.act(np_action)
+
+        reward, observation, first = self._gym3_env.observe()
```
hnyu, line 380: Not sure how high the cost is to call observe(). If takes much time, then probably we can avoid x2 times of calls by first calling observation() in reset() and then each time just call once in _step(). (If we do in this way, then observation should avoid side effect because it's passed down to the next _step() as prev_observation).

alf/environments/alf_gym3_wrapper.py
```diff
@@ -0,0 +1,441 @@
...
+        Otherwise the reset will be ignored.
+
+        """
+        if self._support_force_reset:
+            self._gym3_env.act(np.array([-1] * self.batch_size))
+            self._prev_first = [False] * self.batch_size
+        else:
+            logging.warning('reset() ignored by AlfGym3Wrapper')
+
+        reward, observation, _ = self._gym3_env.observe()
```
hnyu, line 336: If observation is of DictType, should we first convert it to dict here? I noticed that only observation_spec has been converted, but the observation itself it not.

hnyu (commented):
Sorry, two more questions about the observation.

alf/environments/alf_gym3_wrapper.py
breakds: The `observation` is stored in the buffer so accessing it in a read-only way should not incur extra cost. I was optimizing more towards "less stateful wrapper" here.

alf/environments/alf_gym3_wrapper.py
breakds: This is because I checked that `observation_spec` is `DictType`, but `observation` is actually a normal `dict`.

hnyu (approved):

emailweixu (approved):

PR #1000, opened 2021-09-08
Support Module and Network for lean_function

emailweixu:
Previously, only pure function is supported. It makes it very hard to use.
This change allows to wrap a Module or Network.

### DIFF
--- a/alf/utils/lean_function.py
+++ b/alf/utils/lean_function.py
@@ -28,10 +28,12 @@ class _LeanFunction(torch.autograd.Function):
     def forward(ctx, func, num_parameters, keywords, *args):
         """
         Args:
-            ctx (_LeanFunction):
+            ctx (_LeanFunction): context of the computation. It is the same object
+                passed for the corresponding backward().
             func (Callable): func/module to be wrapped
             num_parameters (int): the number of nn.Parameters of func if it is an
-                nn.Module. 0 otherwise
+                nn.Module. 0 otherwise. If ``func`` is a module, the first
+                ``num_parameters`` of arguments in args are the parameters of ``func``.
             keywords (tuple of str): the name of the keys of the keyword arguments
                 for ``func``
             args (Any): all the arguments (positional and keyword) for ``func``.
@@ -108,11 +110,15 @@ def lean_function(func: Callable) -> Callable:
 
         1. All the Tensor inputs to ``func`` must be explicitly listed as arguments
           of ``func``. For example, a tuple of Tensors as argument is not allowed.
-          Using Tensors outside of ``func`` is not allowed either unless ``func``
-          is a ``nn.Module``.
+          Using Tensors outside of ``func`` (e.g., tensors from class member variables)
+          is not allowed either unless ``func`` is a ``nn.Module``. On the other
+          hand, if ``func`` is a module, its parameters should not be put as arguments
+          as they are automatically taken care of.
+
         2. If ``func`` is not a ``Network``, its return value must be a Tensor
           or a tuple of Tensors. If it is a ``Network``, its return value (output
           and state) must be a nest of Tensors.
+
         3. ``func```` must be deterministic so that repeated evaluation with the
           same input will get same output.
 
--- a/alf/utils/lean_function.py
+++ b/alf/utils/lean_function.py
@@ -121,6 +121,12 @@ def lean_function(func: Callable) -> Callable:
          error if ``func`` does not satisfies these requirements and error will
          be silently ignored.
 
+    Note: pytorch also has a function with similar functionality. See https://pytorch.org/docs/stable/checkpoint.html
+    for detail. ``lean_function`` has several advantage over pytorch's implementation:
+
+        1. Keyword arguments are supported.
+        2. Both ``torch.autograd.grad`` and ``torch.autograd.backward`` are supported.
+
     Examples:
 
     1. Apply to simple function:
--- a/alf/utils/lean_function.py
+++ b/alf/utils/lean_function.py
@@ -13,37 +13,73 @@
 # limitations under the License.
 
 import torch
+import torch.nn as nn
+import types
+from typing import Callable
+
+from alf.nest import flatten, pack_sequence_as
+from alf.networks import Network
 
 
 class _LeanFunction(torch.autograd.Function):
+    # Reference: Pytorch: Defining new Autograd Functions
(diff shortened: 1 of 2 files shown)


alf/utils/lean_function.py
```diff
@@ -13,67 +13,123 @@
...
-    repeated evaluation with the same input will get same output. It is the
-    responsibility of the user of this function to make sure that ``func``
-    satifisies these requirements. ``lean_function`` will not report error if
-    ``func`` does not satisfies these requirements and error will be silently
-    ignored.
+    Note: There are several requirements for ``func``:
+
+        1. All the Tensor inputs to ``func`` must be explicitly listed as arguments
+          of ``func``. For example, a tuple of Tensors as argument is not allowed.
+          Using Tensors outside of ``func`` is not allowed either unless ``func``
```
hnyu: Not sure what does "using tensors outside of ``func``" mean?

alf/utils/lean_function.py
```diff
@@ -13,67 +13,123 @@
...
+          is a ``nn.Module``.
+        2. If ``func`` is not a ``Network``, its return value must be a Tensor
+          or a tuple of Tensors. If it is a ``Network``, its return value (output
+          and state) must be a nest of Tensors.
```
hnyu, line 120: A nest can also be a single tensor. So here "must" is better rephrased as "can".

alf/utils/lean_function.py
```diff
@@ -13,67 +13,123 @@
...
+    # Reference: Pytorch: Defining new Autograd Functions
+    # https://pytorch.org/tutorials/beginner/examples_autograd/two_layer_net_custom_function.html
     @staticmethod
-    def forward(ctx, func, keywords, *args):
+    def forward(ctx, func, num_parameters, keywords, *args):
+        """
+        Args:
+            ctx (_LeanFunction):
+            func (Callable): func/module to be wrapped
+            num_parameters (int): the number of nn.Parameters of func if it is an
```
breakds, line 34: I know this is not public facing documentation, but it would help me if we explain the pattern for `forward`: for modules, the first `num_parameters` arguments in `args` are the parameters of that module.

alf/utils/lean_function.py
breakds: Maybe explicitly explain that, for `nn.Module` we do not have to put the parameters in the argument list because it will be automatically taken care of.

breakds: I think it means "accessing tensors in the body of `func` which are not specified in the argument list of `func`".

alf/utils/lean_function.py
```diff
@@ -13,67 +13,123 @@
...
 
 class _LeanFunction(torch.autograd.Function):
+    # Reference: Pytorch: Defining new Autograd Functions
+    # https://pytorch.org/tutorials/beginner/examples_autograd/two_layer_net_custom_function.html
     @staticmethod
-    def forward(ctx, func, keywords, *args):
+    def forward(ctx, func, num_parameters, keywords, *args):
+        """
+        Args:
+            ctx (_LeanFunction):
```
breakds: Is is left blank intentionally? I think it is still worth it to explain this context variable even if it might be trivial for `autograd.Function`.

alf/utils/lean_function.py
```diff
@@ -84,16 +140,58 @@ def myfunc(x, w, b, scale=1.0):
...
+                                  *parameters, *args, *tuple(kwargs.values()))
+        if isinstance(func, Network):
+            ret = pack_sequence_as(specs, ret)
+        return ret
+
+    if isinstance(func, nn.Module):
+        func._lean_function = _wrapped
+        func._original_forward_for_lean_function = func.forward
+        func._inside_lean_function = False
+        func.forward = types.MethodType(_forward, func)
```
breakds, line 200: Forgive my ignorance on `Python` - what is the difference here between this and `func.forward = _forward` directly? Is this about adding `_forward` to the method list of `func` as well? 

Is `_forward` also part of pytorch's (hidden) standard API? This could help me understand why we need `_inside_lean_function` too.

hnyu: > Forgive my ignorance on `Python` - what is the difference here between this and `func.forward = _forward` directly? Is this about adding `_forward` to the method list of `func` as well?
> 
> Is `_forward` also part of pytorch's (hidden) standard API? This could help me understand why we need `_inside_lean_function` too.

You may find this helpful:
https://stackoverflow.com/questions/37455426/advantages-of-using-methodtype-in-python

breakds: Thanks! Coming from C++ I'd thought the method's first argument when called together with a class instance is automatically supplied but apparently it is not so in python. This helps a lot.

alf/utils/lean_function.py
emailweixu: added.

alf/utils/lean_function.py
emailweixu: added.

alf/utils/lean_function.py
emailweixu: What I meant are tensors from class member variables or global variables

emailweixu: updated the comment

alf/utils/lean_function.py
emailweixu: Since single Tensor is a special case of nest, so the comment is correct.

hnyu: > Since single Tensor is a special case of nest, so the comment is correct.

"Must" means forcing the user to do something right. Because either single Tensor or nested Tensor are supported in this case, then the word "can" will express the option of nested Tensor compared to "if ``func`` is not a ``Network``". Currently "must" seems strange because it reads like excluding single Tensor.

emailweixu: But single tensor is still nest of tensor.

hnyu: > But single tensor is still nest of tensor.

Yeah the statement itself is correct. I just feel that right now the second sentence has zero info because anything is a nest essentially. I thought your original meaning is to say "If it is a ``Network``, its return value (output and state) can be any structure of Tensors (not just a single or a tuple)."

emailweixu: It has to be a nest of "Tensor". For example, a nest of str is not allowed.

hnyu (approved):

breakds (approved):

PR #1004, opened 2021-09-12
A "callable" warning

www2171668:
Adjusting 198-200 lines and 338-342 lines. Especially in 198-200. If it writes as "[None] * num", original 400 line will show "Expression is not callable" warning. So I adjust them to match the methods as in "sac_algorithm.py". It will be better.


hnyu:
I think the current implementation is fine because ``None`` will always be replaced by the actual loss.

Issue #1005, opened 2021-09-13
Error in environment process: Traceback (most recent call last):


www2171668:
I meet a problem about sac algorithm. When I run "python -m alf.bin.train --gin_file=sac_fetchreach.gin --root_dir="sac_fetchreach" , it will show error as "ValueError: Configurable 'SacAlgorithm' doesn't have a parameter named 'actor_network_cls'". 

![image](https://user-images.githubusercontent.com/49861957/133020395-d4bbcc34-1cd7-4cb1-9831-84af46c8a38b.png)

I find a recent submit in Sac_fetchreach.gin, SacAlgorithm.actor_network=@actor/ActorDistributionNetwork() has been changed to SacAlgorithm.actor_network_cls=@actor/ActorDistributionNetwork(). In SacAlgorith, actor_network_cls is still been used in __init__. 
In my mind, I think the formmer is correct, but it shows error. Why SacAlgorithm.actor_network_cls shows error ,but SacAlgorithm.actor_network is OK?

hnyu:
Hi, please check your local version of the code. It seems that you're using a quite old version of SAC but a newer version of the gin file. The current SacAlgorithm takes a constructor instead of a constructed object for the actor network creation. You can find its init signature at:

https://alf.readthedocs.io/en/latest/api/alf.algorithms.html#module-alf.algorithms.sac_algorithm

![image](https://user-images.githubusercontent.com/51248379/133032723-c72831dd-0945-4e46-8217-a87aa3dc1805.png)

www2171668:
Mybe I use a old virtualenv and a chaotic version. When I change the newest version and create a new virtualenv, I meet a new problem never seen. 
![image](https://user-images.githubusercontent.com/49861957/133082270-67c5fd45-f8eb-48bf-8a0b-5d55d4fc05db.png)

![image](https://user-images.githubusercontent.com/49861957/133082421-df81e961-7624-4e80-8917-f13be15d3901.png)

...
![image](https://user-images.githubusercontent.com/49861957/133080862-4778edad-38d6-494b-b945-ffe24d1bd619.png)

www2171668:
Mybe I use a old virtualenv and a chaotic version. When I change the newest version and create a new virtualenv, I meet a new problem never seen.

![image](https://user-images.githubusercontent.com/49861957/133082718-5968e779-f758-45bb-a9b0-5c1fd2e24f0e.png)

[...]

PR #1006, opened 2021-09-13
fix gym's FlattenObservation wrapper

hnyu:
Related to issue #1005 

After upgrading gym >=0.15.3, suite_robotics will use ``env = FlattenObservation(FilterObservation(env, keys))`` to flatten the observation. However, the original implementation of ``FlattenObservation`` has a bug:

``` python 

class FlattenObservation(ObservationWrapper):
    r"""Observation wrapper that flattens the observation."""
    def __init__(self, env):
        super(FlattenObservation, self).__init__(env)

        flatdim = spaces.flatdim(env.observation_space)
        self.observation_space = spaces.Box(low=-float('inf'), high=float('inf'), shape=(flatdim,), dtype=np.float32)

        return spaces.flatten(self.observation_space, observation)
```

Because ``self.observation_space`` has already been changed in ``init``, when calling ``observation()``, its value is no longer correct. We should use the original observation space instead. 

**How to test:**
Can test with either ``sac_fetchreach.gin`` or ``ddpg_fetchreach.gin`` to reproduce and test.

### DIFF
--- a/alf/environments/suite_robotics.py
+++ b/alf/environments/suite_robotics.py
@@ -31,6 +31,7 @@ except ImportError:
 import functools
 import numpy as np
 import gym
+from gym import ObservationWrapper
 
 import alf
 from alf.environments import suite_gym, alf_wrappers, process_environment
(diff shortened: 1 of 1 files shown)


hnyu:
never mind, I found that later version of Gym has already fixed this. Maybe we should upgrade gym first.

PR #1007, opened 2021-09-13
upgrade gym to 0.15.4

hnyu:
This PR is a follow-up to #1006 . We can fix that issue by just upgrading gym from 0.15.3. to 0.15.4.

As I checked, 0.15.4 has an updated implementation:

``` python
class FlattenObservation(ObservationWrapper):
    r"""Observation wrapper that flattens the observation."""
    def __init__(self, env):
        super(FlattenObservation, self).__init__(env)

        flatdim = spaces.flatdim(env.observation_space)
        self.observation_space = spaces.Box(low=-float('inf'), high=float('inf'), shape=(flatdim,), dtype=np.float32)

        return spaces.flatten(self.env.observation_space, observation)
```

### DIFF
--- a/.ci-cd/requirements.txt
+++ b/.ci-cd/requirements.txt
@@ -1,7 +1,7 @@
 absl-py==0.7.0
 atari_py==0.1.7
 fasteners
-gym==0.15.3
+gym==0.15.4
 pyglet==1.3.2
 Box2D-kengz==2.3.3
 gym-retro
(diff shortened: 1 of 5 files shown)


le-horizon (dismissed):
LG.

breakds (commented):
Maybe we should also update `docker/requirements_carla.txt` and `docs/requirements.txt`?

breakds (approved):

PR #1009, opened 2021-09-15
update to relu_mlp

runjerry:
Update the following for implementing a more efficient helper network of GPVI

* add autograd-free jvp in relu_mlp

* enable vjp and jvp w.r.t. partial output-input Jacobian in relu_mlp

### DIFF
--- a/alf/networks/relu_mlp_test.py
+++ b/alf/networks/relu_mlp_test.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 from absl.testing import parameterized
+import numpy as np
 import torch
 
 import alf
@@ -59,8 +60,6 @@ class ReluMLPTest(parameterized.TestCase, alf.test.TestCase):
         # compute jac using direct approach
         x = torch.randn(batch_size, input_size, requires_grad=True)
         x1 = x.detach().clone()
-        x1.requires_grad = True
-        # z, state = mlp(x1, requires_jac=True)
         jac = mlp.compute_jac(x1)
 
         # compute jac using autograd
@@ -93,7 +92,6 @@ class ReluMLPTest(parameterized.TestCase, alf.test.TestCase):
         # compute jac diag using direct approach
         x = torch.randn(batch_size, input_size, requires_grad=True)
         x1 = x.detach().clone()
-        x1.requires_grad = True
         jac_diag = mlp.compute_jac_diag(x1)
 
         # compute jac using autograd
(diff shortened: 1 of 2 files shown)


alf/networks/relu_mlp_test.py
```diff
@@ -118,22 +116,77 @@ def test_compute_vjp(self, hidden_layers=(2, ), batch_size=2,
...
+        x2.requires_grad = True
+        y2, _ = mlp(x2)
+        mask_idx = np.setdiff1d(range(vec.shape[1]), partial_idx)
+        vec[:, mask_idx] = 0.
+        vjp2_partial = torch.autograd.grad(y2, x2, grad_outputs=vec)[0]
+
         self.assertArrayEqual(vjp, vjp2, 1e-6)
+        self.assertArrayEqual(vjp_partial, vjp2_partial, 1e-6)
+
+    @parameterized.parameters(
```
neale, line 162: since hidden_layers=() is explicitly handled in the _compute_jvp function, there should probably be a test case for it.

alf/networks/relu_mlp_test.py
```diff
@@ -118,22 +116,77 @@ def test_compute_vjp(self, hidden_layers=(2, ), batch_size=2,
...
+        dict(hidden_layers=(2, 3, 4)),
+    )
+    def test_compute_jvp(self, hidden_layers=(2, ), batch_size=3,
+                         input_size=5):
+        """
+        Check that the Jacobian-vec product computed by the direct(autograd-free)
+        approach is consistent with the one computed by calling autograd.
+        """
+        output_size = 4
+        partial_idx = [0, 2]
```
neale: Can we test this? Maybe add another test for partial_idx=[1, -1]. 
It's possible that a few years down the road we may end up with this only working for partial_idx=[0, 2]

neale (dismissed):
Overall looks good to me, just adding a couple of more test cases would be nice. 
I pulled this branch and tried to make the tests fail and couldn't. So that's good.

alf/networks/relu_mlp_test.py
runjerry: good point, just added for both vjp and jvp tests.

alf/networks/relu_mlp_test.py
runjerry: Just added partial_idx=[1, -1] into the tests.

neale (approved):
Thanks for the additions. Looks good to me

PR #1010, opened 2021-09-16
fix a spec bug in RewardNormalizer

hnyu:
Turns out I was using the same mean and std for different reward dims.

### DIFF
--- a/alf/algorithms/data_transformer.py
+++ b/alf/algorithms/data_transformer.py
@@ -599,6 +599,10 @@ class RewardNormalizer(SimpleDataTransformer):
         super().__init__(observation_spec)
         if normalizer is None:
             normalizer = AdaptiveNormalizer(
+                # ``get_reward_spec()`` is only a tmp solution. In some rare cases,
+                # reward spec might have been changed by data transformers before
+                # this one.
+                # TODO: we should pass a ``time_step`` spec to the constructor.
                 tensor_spec=alf.get_reward_spec(),
                 auto_update=False,
                 debug_summaries=True)
--- a/alf/algorithms/data_transformer.py
+++ b/alf/algorithms/data_transformer.py
@@ -597,7 +597,10 @@ class RewardNormalizer(SimpleDataTransformer):
         """
         super().__init__(observation_spec)
         if normalizer is None:
-            normalizer = ScalarAdaptiveNormalizer(auto_update=False)
+            normalizer = AdaptiveNormalizer(
+                tensor_spec=alf.get_reward_spec(),
+                auto_update=False,
+                debug_summaries=True)
         self._normalizer = normalizer
         self._clip_value = clip_value
         self._update_mode = update_mode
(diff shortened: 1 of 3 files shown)


le-horizon (commented):
Thanks for the fix, Haonan!  LG.

le-horizon (dismissed):

alf/utils/normalizers.py
```diff
@@ -122,8 +122,7 @@ def _summary(name, val):
                 with alf.summary.scope(self._name):
                     if val.ndim == 0:
                         alf.summary.scalar(name + "." + suffix, val)
-                    elif (val.shape[0] < self.MAX_DIMS_TO_OUTPUT
-                          and alf.summary.should_summarize_output()):
```
hnyu, line 126: It's not easy to relate ``should_summarize_output()`` with this function here without looking at the source code, so I just removed this check.

le-horizon: Just be aware that it could potentially dump a lot of tb plots, multiple for each dimension of observation.  It will slow down training and clog hard disk if ran many times.

hnyu: > Just be aware that it could potentially dump a lot of tb plots, multiple for each dimension of observation. It will slow down training and clog hard disk if ran many times.

It's kind of confusing when setting ``debug_summaries=True`` alone won't actually summarize, without turning on ``should_summarize_output``. Should we just decrease the value self.MAX_DIMS_TO_OUTPUT?

le-horizon: If we make MAX_DIMS_TO_OUTPUT configurable, then we can decrease the default value.

hnyu: > If we make MAX_DIMS_TO_OUTPUT configurable, then we can decrease the default value.

I've changed this parameter to an argument.

le-horizon (dismissed):
LG

alf/algorithms/data_transformer.py
```diff
@@ -597,7 +598,10 @@ def __init__(self,
         """
         super().__init__(observation_spec)
         if normalizer is None:
-            normalizer = ScalarAdaptiveNormalizer(auto_update=False)
+            normalizer = AdaptiveNormalizer(
+                tensor_spec=alf.get_reward_spec(),
```
emailweixu, line 606: It doesn't feel right to couple data transformer with config util. A hypothetical scenario where this may fail is that another data transformer make change to reward spec.

Perhaps we should make the data transformer take time_step_spec as input to the constructor so that it does not need to call alf.get_reward_spec().  This can be a TODO item.

hnyu: > It doesn't feel right to couple data transformer with config util. A hypothetical scenario where this may fail is that another data transformer make change to reward spec.
> 
> Perhaps we should make the data transformer take time_step_spec as input to the constructor so that it does not need to call alf.get_reward_spec(). This can be a TODO item.

added TODO

emailweixu (approved):

PR #1011, opened 2021-09-16
GPVI plus

runjerry:
Add an efficient block form for inverse_mvp computation of GPVI

### DIFF
--- a/alf/algorithms/generator.py
+++ b/alf/algorithms/generator.py
@@ -13,9 +13,10 @@
 # limitations under the License.
 """A generic generator."""
 
+import functools
 import numpy as np
 import torch
-import functools
+from torch.autograd.functional import jacobian
 
 import alf
 from alf.algorithms.algorithm import Algorithm
@@ -198,7 +199,7 @@ class InverseMVPAlgorithm(Algorithm):
         """
         z_inputs, vec = inputs
         assert z_inputs.ndim == 2 and z_inputs.shape[-1] >= self._z_dim
-        assert vec.shape[-1] == self._vec_dim
+        assert vec.shape[-1] >= self._vec_dim
         assert z_inputs.shape[0] == vec.shape[0]
 
         if z_inputs.shape[-1] > self._z_dim:
@@ -215,6 +216,9 @@ class InverseMVPAlgorithm(Algorithm):
             raise ValueError(
                 "vec must be dimension 2 or 3, got dimension {}".format(
                     vec.ndim))
+
+        if vec_inputs.shape[-1] > self._vec_dim:
+            vec_inputs = vec_inputs[:, :self._vec_dim]
 
         outputs = (self._net((z_inputs, vec_inputs))[0], z_inputs)
 
@@ -297,7 +301,10 @@ class Generator(Algorithm):
                  par_vi=None,
                  use_kernel_averager=False,
                  functional_gradient=False,
-                 fullrank_diag_weight=1.0,
+                 init_lambda=1.,
+                 lambda_trainable=False,
+                 block_inverse_mvp=False,
+                 direct_jac_inverse=False,
                  inverse_mvp_solve_iters=1,
                  inverse_mvp_hidden_size=100,
                  inverse_mvp_hidden_layers=1,
@@ -311,6 +318,7 @@ class Generator(Algorithm):
                  critic_optimizer=None,
                  inverse_mvp_optimizer=None,
                  optimizer=None,
+                 lambda_optimizer=None,
                  name="Generator"):
         r"""Create a Generator.
 
@@ -361,8 +369,14 @@ class Generator(Algorithm):
                 forwarding the first ``noise_dim`` components. We then add the 
                 full noise vector to the output, multiplied by the 
                 ``fullrank_diag_weight``.  
-            fullrank_diag_weight (float): weight on "extra" dimensions when 
-                forcing full rank Jacobian
+            init_lambda (float): weight on direct input-output link added to
+                the generator output. Only used for GPVI when forcing full rank 
+                Jacobian.
+            block_inverse_mvp(bool): whether to use the more efficient block form
+                for inverse_mvp when ``functional_gradient`` is True. This
+                option is recommended only when ``noise_dim`` < ``output_dim``.
+                as it is equivalent to the default form when ``noise_dim`` is 
+                equal to ``output_dim``.
             inverse_mvp_solve_iters (int): number of iterations of inverse_mvp
                 network training per single iteration of generator training.
             inverse_mvp_hidden_size (int): width of hidden layers in inverse_mvp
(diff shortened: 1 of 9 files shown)


alf/algorithms/generator.py
```diff
@@ -363,6 +368,9 @@ def __init__(self,
                 ``fullrank_diag_weight``.  
             fullrank_diag_weight (float): weight on "extra" dimensions when 
                 forcing full rank Jacobian
+            block_inverse_mvp(bool): whether to use the more efficient block form
+                for inverse_mvp when ``functional_gradient`` is True. This
+                option only makes sense when ``noise_dim`` < ``output_dim``.
```
neale: this could be more clear: "this option is recommended only when ``noise_dim`` < ``output_dim``, as it is equivalent to the default behavior when ``noise_dim`` == ``output_dim``

neale:
I can't find any errors, if we're not adding the autograd version of GPVI+ right now then it's pretty straightforward. 
Will the relu_mlp changes overwrite master? Because this relu_mlp now is stale with respect to the previous update to relu_mlp. 
Does Wei have any input on this PR?

alf/algorithms/generator.py
runjerry: updated

hnyu:
@runjerry I thought the formulation you shared last Friday was meant to be a part of publication in the future, so Feishu doc might be a good private place to go. But it seems that the corresponding code already has a PR here? If there is no plan for a publication for this, then probably a .rst file would be better for documenting the math instead.

runjerry:
> @runjerry I thought the formulation you shared last Friday was meant to be a part of publication in the future, so Feishu doc might be a good private place to go. But it seems that the corresponding code already has a PR here? If there is no plan for a publication for this, then probably a .rst file would be better for documenting the math instead.

I guess my collaborators at OSU might plan to include it in a future publication, depending on how well their further experiments based on this PR go. So maybe I can make up the .rst file later?

hnyu:
> > @runjerry I thought the formulation you shared last Friday was meant to be a part of publication in the future, so Feishu doc might be a good private place to go. But it seems that the corresponding code already has a PR here? If there is no plan for a publication for this, then probably a .rst file would be better for documenting the math instead.
> 
> I guess my collaborators at OSU might plan to include it in a future publication, depending on how well their further experiments based on this PR go. So maybe I can make up the .rst file later?

If so, should this code be private instead (to prevent public seeing it before being published)?

runjerry:
> > > @runjerry I thought the formulation you shared last Friday was meant to be a part of publication in the future, so Feishu doc might be a good private place to go. But it seems that the corresponding code already has a PR here? If there is no plan for a publication for this, then probably a .rst file would be better for documenting the math instead.
> > 
> > 
> > I guess my collaborators at OSU might plan to include it in a future publication, depending on how well their further experiments based on this PR go. So maybe I can make up the .rst file later?
> 
> If so, should this code be private instead (to prevent public seeing it before being published)?

Maybe it is fine given that the codes are not quite understandable without those formulas...

alf/algorithms/generator.py
```diff
@@ -311,6 +318,7 @@ def __init__(self,
                  critic_optimizer=None,
                  inverse_mvp_optimizer=None,
                  optimizer=None,
+                 lambda_optimizer=None,
```
neale, line 321: I think lambda_optimizer is not documented below

alf/algorithms/generator.py
```diff
@@ -973,37 +1017,84 @@ def _minmax_grad(self,
 
         return loss, loss_propagated
 
+    def _get_vec_for_jac_inv_vec_prod(self, z, vec):
+        r"""
+        Construct a vecor as input to the helper network for 
+        Jacobian-inverse vector product estimation, used for GPVI and GPVI_Plus.
```
neale: Is this really used for GPVI? I don't think we have a = vjp / \lambda term in GPVI.

alf/algorithms/generator.py
```diff
@@ -1067,6 +1177,64 @@ def _rkhs_func_grad(self,
         loss_propagated = torch.sum(grad.detach() * outputs, dim=1)
         return (loss, inverse_mvp_loss), loss_propagated
 
+    def _direct_jac_inverse_vec_prod(self, z, vec):
```
neale, line 1184: I tested direct_jac for GPVI in the generator test ( believe in this case GPVI=GPVI+), and it performs well. It is a just slightly better approximation than GPVI.

alf/algorithms/hypernetwork_algorithm_test.py
```diff
@@ -197,6 +194,189 @@ def _test(i, sampled_predictive=False):
         self.assertLess(mean_err, 0.5)
         self.assertLess(cov_err, 0.5)
 
+    @parameterized.parameters((True), (False, False, False, 48), (False, True),
+                              (True, False, True))
```
neale, line 198: I performed one more test that I think is crucial.  `(direct_jac=False, block_inverse_mvp=True, lambda_trainable=True)`
This performed nearly on-par with the final test you have: `(direct_jac=True, block_inverse_mvp=False, lambda_trainable=True)`, that is to say, quite well.

alf/algorithms/generator.py
runjerry: Nice catch, it is only used in GPVI_plus.

alf/algorithms/generator.py
runjerry: updated.

alf/algorithms/hypernetwork_algorithm_test.py
runjerry: I agree this is a crucial test case and have tested it. Did not include it just for CI speed concerns.

alf/algorithms/generator.py
runjerry: Yes, I found the direct_jac performs almost the same as GPVI+ while being slightly better than GPVI. It was very helpful for my debugging though.

neale (approved):
Because this code is so dense I did a lot of testing on the experimental cases that touch this PR, since I found no obvious errors in the implementation.
Besides correctness, I'm not sure any external user could debug this as it is. The changes to `generator.py` and `relu_mlp.py` are a little arcane with all the indexing and einsums. Good for now I suppose as publication work is ongoing.  
Though, when I implemented GPVI+ in the past my implementation was similarly difficult to read. I don't see a solution other than documentation.

runjerry:
> Because this code is so dense I did a lot of testing on the experimental cases that touch this PR, since I found no obvious errors in the implementation. Besides correctness, I'm not sure any external user could debug this as it is. The changes to `generator.py` and `relu_mlp.py` are a little arcane with all the indexing and einsums. Good for now I suppose as publication work is ongoing. Though, when I implemented GPVI+ in the past my implementation was similarly difficult to read. I don't see a solution other than documentation.

Many thanks Neale for your efforts in reviewing and testing this PR. Given how dense the computational details are, as you suggested, documentation is probably the only way to make these codes readable. I will definitely do it after we release a publication.

PR #1012, opened 2021-09-20
[Refactor] Retire OnetimeExperienceReplayer

breakds:
# Motivation

Currently there is a concept distinguishing 2 types of experience replayer, `one_time` and `uniform`, where the former does not count as "using replay buffer" and the latter counts. We would like to simplify this by having only "the experience replayer with replay buffer" because:

1. One less concept to learn for the new users, especially when such behavior is implicitly configured before
2. Less confusing as "uniform" does not directly imply "it can be used similar to one time experience replayer as well", i.e. always "replay_all"
3. When updating logic of replayers, we now have only one thing to worry about

# Solution

Remove `OnetimeExperienceReplayer` and experience replayer type related logic entirely. Several thoughts about this decision:

1. This will increase the time consumption of the training, but that increase is negligible. Such increase mainly comes from saving the experience in cpu during observation and copying them to GPU during updating. In procgen environment training (whose image observation is huge), this adds about 3ms to each training iteration.
2. To ensure `observe_for_replay` is called after `set_exp_replayer` is called, it will raise exception upon such incorrect calls.

# Testing

Tested `ppo_cart_pole` and `ppo_procgen_bossfight` where both used to use `one-time` experience replayer. Also it passes the CI.


emailweixu:
I feel that the concept of ExperienceReplayer seems not useful after removing OnetimeExperienceReplayer. Perhaps it can also be removed and merged with replay buffer.

@hynu what do you think?

hnyu:
> I feel that the concept of ExperienceReplayer seems not useful after removing OnetimeExperienceReplayer. Perhaps it can also be removed and merged with replay buffer.
> 
> @hynu what do you think?

Yeah if we don't need async off-policy drivers as in TF, then SyncExperienceReplayer is a very thin wrapper that can be removed.

breakds:
Sounds good. I think we can probably do this in two step - one to remove `OnetimeExperienceReplayer` and another one to remove experience replayer itself.

breakds:
> > I feel that the concept of ExperienceReplayer seems not useful after removing OnetimeExperienceReplayer. Perhaps it can also be removed and merged with replay buffer.
> > @hynu what do you think?
> 
> Yeah if we don't need async off-policy drivers as in TF, then SyncExperienceReplayer is a very thin wrapper that can be removed.

Here async meaning there is perhaps several learner drawing samples while the replay buffer is being populated?

hnyu:
> > > I feel that the concept of ExperienceReplayer seems not useful after removing OnetimeExperienceReplayer. Perhaps it can also be removed and merged with replay buffer.
> > > @hynu what do you think?
> > 
> > 
> > Yeah if we don't need async off-policy drivers as in TF, then SyncExperienceReplayer is a very thin wrapper that can be removed.
> 
> Here async meaning there is perhaps several learner drawing samples while the replay buffer is being populated?

Right. Such a training schedule only existed in the previous TF version. The current version hasn't implemented it.

breakds:
> > > > I feel that the concept of ExperienceReplayer seems not useful after removing OnetimeExperienceReplayer. Perhaps it can also be removed and merged with replay buffer.
> > > > @hynu what do you think?
> > > 
> > > 
> > > Yeah if we don't need async off-policy drivers as in TF, then SyncExperienceReplayer is a very thin wrapper that can be removed.
> > 
> > 
> > Here async meaning there is perhaps several learner drawing samples while the replay buffer is being populated?
> 
> Right. Such a training schedule only existed in the previous TF version. The current version hasn't implemented it.

I might be wrong but I think in the future it might be useful. Of course we could implement this "back" when we actually have that need.

hnyu:
> > > > > I feel that the concept of ExperienceReplayer seems not useful after removing OnetimeExperienceReplayer. Perhaps it can also be removed and merged with replay buffer.
> > > > > @hynu what do you think?
> > > > 
> > > > 
> > > > Yeah if we don't need async off-policy drivers as in TF, then SyncExperienceReplayer is a very thin wrapper that can be removed.
> > > 
> > > 
> > > Here async meaning there is perhaps several learner drawing samples while the replay buffer is being populated?
> > 
> > 
> > Right. Such a training schedule only existed in the previous TF version. The current version hasn't implemented it.
> 
> I might be wrong but I think in the future it might be useful. Of course we could implement this "back" when we actually have that need.

Yeah we can add it back when needed

alf/algorithms/algorithm.py
```diff
@@ -245,24 +242,25 @@ def _set_exp_replayer(self, sample_exp):
         Args:
             sample_exp (nested Tensor):
         """
+        if (self._exp_replayer_num_envs is None
+                or self._exp_replayer_length is None
+                or self._prioritized_sampling is None):
+            raise RuntimeError(
+                'Experience replayer cannot be initialized before set_exp_replayer() is called'
```
hnyu: "Experience replayer must be initialized first by calling ``set_exp_replayer()`` before ``observe_for_replay()`` is called!"

hnyu:
Just realized that one difference between OnetimeExperienceReplayer with SyncExperienceReplayer is that when clearing the buffer, Onetime will keep the last experience in the buffer as the first one for the next batch:

image

This makes sure the value of the last experience gets trained in the next batch instead of being directly thrown away.
If the env is deterministic and episode_length can be divided by the unroll_length , experience at every unroll_length will always be ignored using SyncExperienceReplayer. @emailweixu Do you think this matters?

breakds:
Additionally I found a problem in this PR while working on removing experience replayer for the good. Will update once I found the problem.

alf/algorithms/algorithm.py
breakds: Updated the message.

breakds:
> Additionally I found a problem in this PR while working on removing experience replayer for the good. Will update once I found the problem.

The problem is fixed in another PR #1013 and this PR has been rebased.

breakds:
> Just realized that one difference between OnetimeExperienceReplayer with SyncExperienceReplayer is that when clearing the buffer, Onetime will keep the last experience in the buffer as the first one for the next batch:
> 
> image
> 
> This makes sure the value of the last experience gets trained in the next batch instead of being directly thrown away.
> If the env is deterministic and episode_length can be divided by the unroll_length , experience at every unroll_length will always be ignored using SyncExperienceReplayer. @emailweixu Do you think this matters?

This does sound like a serious problem to me. Maybe we can implement this behavior when both `use_whole_buffer_training` and `clear_after_xxx` is set to `True`?

hnyu (approved):

PR #1014, opened 2021-09-23
[Refactor] Add series of ppo procgen configurations, grouped in a subdirectory

breakds:
# Motivation

We currently have `ppo_procgen_bossfight_conf.py`. There are 16 procgen games in the procgen bencmark and adding ppo configuration for another one has almost identical code except for the `env_name`.

# Solution

Create a base `ppo_procgen` configurations and have the other per-game configuration importing it. To make it nicer:

1. Put them under the same sub directory called `ppo_procgen`
2. Picked 3 games and explicitly put their configuration in the directory as examples
3. Added `README.md` for that directory to give some instructions/documentation

*EXTRA*: `mini_batch_length` updated from `1` to `256` to match the OpenAI implementation's hyper paramter.

# Testing

Tested that the training can run successfully with all the 3 configurations.


alf/examples/ppo_procgen/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+This directory provides configuration to train PPO algorithm on [procgen](https://openai.com/blog/procgen-benchmark/) benchmark environments.
+
+Procgen benchmark provide 16 different games (environments). Alf support all 16 of them and here we have configurations for 3 of them as examples:
+
+1. bossfight: [bossfight_conf.py](./bossfight_conf.py)
+2. bigfish: [bigfish_conf.py](./bigfish_conf.py)
+3. chaser: [chaser_conf.py](./chaser_conf.py)
+
+The main configuration parameters are in [base_conf.py](./base_conf.py). To create a configuration for a different game in Procgen, you just need to have 3 lines and fill `env_name` with the game of choice. For example, to create a configuration for `coinrun`, the following configuration will suffice.
```
emailweixu: A different way of training for a different environment is to use commandline flag `--conf_param`. For example:

```python
python -m alf.bin.train \
--conf ppo_procgen/base_conf.py \
--conf_param create_environment.env_name='coinrun'` \
--root_dir mydir
```

alf/examples/ppo_procgen/README.md
```diff
@@ -0,0 +1,19 @@
+# Procgen Environment with PPO
+
+This directory provides configuration to train PPO algorithm on [procgen](https://openai.com/blog/procgen-benchmark/) benchmark environments.
+
+Procgen benchmark provide 16 different games (environments). Alf support all 16 of them and here we have configurations for 3 of them as examples:
```
emailweixu: support => supports

hnyu:
Could also check in training curves for archiving purpose.

alf/examples/ppo_procgen/README.md
breakds: Actually this does look like a better approach. I can probably just leave `bossfight` here and use this `README.md` to instruct on training with other games. What do you think? @emailweixu @hnyu

emailweixu: Seems reasonable to me.

breakds: Updated accordingly.

alf/examples/ppo_procgen/README.md
breakds: Fixed.

breakds:
> Could also check in training curves for archiving purpose.

I have updated the PR to include only the `base_conf.py` and `bossfight_conf.py`. The other games are supposed to run with `base_conf.py` and the instructions in `README.md` are also updated accordingly. We probably do not need the training curves for this PR now. However when I have better curves (the current curves are based on original parameters that does not converge to very high reward) for them, I can put them in this folder in the future.

Please take another look.

alf/examples/ppo_procgen/base_conf.py
```diff
@@ -83,8 +82,8 @@ def value_network_ctor(input_tensor_spec):
 alf.config(
     'TrainerConfig',
     unroll_length=256,
-    mini_batch_length=1,
-    mini_batch_size=4096,
+    mini_batch_length=256,
```
emailweixu: batch length from the replay buffer is actually 257, so this will ignore the last step. It can be set to `None` so that min_batch_length is equal to the length from replay buffer

breakds: Yes this holds true if `keep_last_exp` is turned on. However, in current `alf` it is not turned on. But the `None` trick should be the most adaptive option. Thank you!

https://github.com/HorizonRobotics/alf/blob/breakds_exp/alf/algorithms/algorithm.py#L1198 Confirmed that, and updated here with `None` with documentation explaining that.

breakds:
PR is updated. Please take another look.

emailweixu (approved):

PR #1015, opened 2021-09-24
Support keep_last_exp in replay buffer

breakds:
# Motivation

After retiring the experience replayer (more specifically the `OnetimeExperienceReplayer`), we will use the replay buffer directly for whole buffer training + clear replay buffer situations. Previously there is a feature in `OnetimeExperienceReplayer` that handles a rare case and we do not want to lose that.

The reason we might need this is that in rare cases when the episodic MDP is deterministic and the episode length is a multiple of unroll length, we may find the last step of the episode is always ignored and never partipate in training. Keeping it will make it the first batch of the next iteration which guarantees its participation in training.

# Solution

Such logic is implemented based on `Ringbuffer` in `ReplayBuffer`. This PR is *only responsible for the implementation*, and the actual usage of it (to be more specific, the usage of the argument in `clear()`) will be in a future PR. Therefore, this PR does not change any logic yet.

# Testing

Added unit test to make sure this feature works.

### DIFF
--- a/alf/experience_replayers/replay_buffer.py
+++ b/alf/experience_replayers/replay_buffer.py
@@ -557,7 +557,7 @@ class ReplayBuffer(RingBuffer):
         return first_step_pos
 
     @atomic
-    def gather_all(self):
+    def gather_all(self, debug=False):
         """Returns all the items in the buffer.
 
         Returns:
@@ -568,6 +568,9 @@ class ReplayBuffer(RingBuffer):
         """
         size = self._current_size.min()
         max_size = self._current_size.max()
+        if debug:
+            print(f'size={size}')
+            print(f'current_pos={self._current_pos}')
         assert size == max_size, (
             "Not all environments have the same size. min_size: %s "
             "max_size: %s" % (size, max_size))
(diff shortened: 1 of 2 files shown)


alf/experience_replayers/replay_buffer.py
```diff
@@ -568,6 +568,9 @@ def gather_all(self):
         """
         size = self._current_size.min()
         max_size = self._current_size.max()
+        if debug:
+            print(f'size={size}')
```
hnyu: We can use logging.info to replace print for better message.

alf/experience_replayers/replay_buffer.py
```diff
@@ -622,6 +625,46 @@ def get_field(self, field_name, env_ids, positions):
         result = alf.nest.map_structure(lambda x: x[indices], field)
         return convert_device(result)
 
+    def clear(self, keep_last_exp: bool = False) -> None:
+        """Clear the replay buffer and remove all the batches.
```
hnyu: Should better not use the term "batch" because batch means a set of data sampled for a gradient update in ML context. I think we can just call data of size (B,1) a "step/sample/experience", assuming the buffer always stores multiple envs data.

alf/experience_replayers/replay_buffer.py
breakds: I removed those two lines. Did not intend to keep them here. Sorry!

alf/experience_replayers/replay_buffer.py
breakds: Sure thing. I got confused because of the method `add_batch`.

Updated batch to "experience".

alf/experience_replayers/replay_buffer.py
```diff
@@ -622,6 +622,47 @@ def get_field(self, field_name, env_ids, positions):
...
 
+    def clear(self, keep_last_exp: bool = False) -> None:
+        """Clear the replay buffer and remove all the experiences.
+
+        One of the exception is that when keep_last_exp is set to True and there
+        are at least 1 experience in the replay buffer, the latest experience
+        will be kept in the buffer after the clear.
+
+        The reason we might need this is that in rare cases when the episodic
+        MDP is deterministic and the episode length is a multiple of unroll
```
le-horizon: "when the episodic MDP is deterministic" => "when the episodic MDP has a fixed episode length"

emailweixu (dismissed):

le-horizon (dismissed):
LG.

emailweixu (approved):

PR #1016, opened 2021-09-28
[Bug] Do not call _set_exp_replayer if required parameters are not set

breakds:
# Motivation

The previous PR broke the build: https://github.com/HorizonRobotics/alf/commit/4068b3d5d66436bcc0f88509b4336378d3b1672a

The reason behind the breakage is that `_set_exp_replayer` is called without calling `set_exp_replayer` first. Normally this should not happen but in unit test this is frequently done. (One of) The correct logic should be just giving up calling `_set_exp_replayer` when the required parameters are not set.

I did not find this bug out before merging the previous PR because CI was green (not sure why yet).

# Fix

Calling `_set_exp_replayer` is now protected.

# Test

Manually tested some of the failed tests locally - they fail before the fix and succeed after the fix.

### DIFF
--- a/alf/algorithms/algorithm.py
+++ b/alf/algorithms/algorithm.py
@@ -280,7 +280,12 @@ class Algorithm(AlgorithmInterface):
                     exp.state, self.train_state_spec, value_to_match=()))
 
         if self._exp_replayer is None:
-            self._set_exp_replayer(exp)
+            # Do not even create the experience replayer if the
+            # required parameters are not set by set_exp_replayer
+            if (self._exp_replayer_num_envs is not None
+                    and self._exp_replayer_length is not None
+                    and self._prioritized_sampling is not None):
+                self._set_exp_replayer(exp)
 
         exp = dist_utils.distributions_to_params(exp)
         for observer in self._observers:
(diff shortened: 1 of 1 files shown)


alf/algorithms/algorithm.py
```diff
@@ -280,7 +280,12 @@ def observe_for_replay(self, exp):
                     exp.state, self.train_state_spec, value_to_match=()))
 
         if self._exp_replayer is None:
-            self._set_exp_replayer(exp)
+            # Do not even create the experience replayer if the
```
hnyu: Suggest directly changing line 245-250 to return instead of raising an error.

breakds: Sounds good. Updated.

hnyu (approved):

hnyu:
It seems that if I run the train_play_test locally, some cases will hang forever. This causes the CI to fail.

breakds:
> It seems that if I run the train_play_test locally, some cases will hang forever. This causes the CI to fail.

Yep I think I reproduced what you said. Working on investigation.

breakds:
This will be incorporated in a new PR. Dropping this PR.

PR #1020, opened 2021-10-05
Convert sac cart pole: gin -> python config

breakds:
# Motivation

In my recent quest on retiring one time experience replayer, I need an example of true off policy training (i.e. actually sample from replay buffer). I picked SAC cart pole and converted it to python conf as we are retiring `gin` as well.

# Solution

1. Converted gin -> py conf
2. Updated the unit test

# Testing
CI passed. Also tested the training locally and get

sac_cart_pole

Which resembles the current graph under examples.


alf/examples/sac_cart_pole_conf.py
```diff
@@ -0,0 +1,57 @@
...
+alf.config(
+    'TrainerConfig',
+    algorithm_ctor=SacAlgorithm,
+    initial_collect_steps=1000,
+    mini_batch_length=2,
+    mini_batch_size=64,
+    unroll_length=1,
+    num_updates_per_train_iter=1,
+    whole_replay_buffer_training=False,
+    clear_replay_buffer=False,
```
hnyu: Since you've imported sac_conf.py, you don't have to re-configure some options in this conf file.

hnyu (changes requested):

alf/examples/sac_cart_pole_conf.py
breakds: Ah yep. Thanks for pointing out. Fixed.

breakds:
Updated. PTAL.

alf/examples/sac_cart_pole_conf.py
```diff
@@ -0,0 +1,54 @@
...
+# environment config
+alf.config(
+    'create_environment', env_name="CartPole-v0", num_parallel_environments=8)
+
+# algorithm config
+alf.config('ActorDistributionNetwork', fc_layer_params=(100, ))
+alf.config('QNetwork', fc_layer_params=(100, ))
+alf.config(
+    'SacAlgorithm',
+    actor_network_cls=ActorDistributionNetwork,
```
hnyu: Could skip actor_network_cls for discrete actions

breakds: Thanks for pointing out. Updated.

hnyu (approved):

PR #1021, opened 2021-10-05
Convert ppo rnd mrevenge: gin -> python config

breakds:
# Motivation

See #1020. This is also the offending example for the previous build breakage. Converted it to py conf for easier testing.

# Testing

Updated unit tests for CI.
Also train locally and get

ppo_rnd_mrevenge


hnyu:
The gin file was written for TF version. Some differences might need to be paid attention to. For example, in the .gin file ``Adam`` actually corresponds to ``AdamTF`` in our pytorch version.

Anyway, it seems that the performance posted here is worse than the figure previously checked in.

breakds:
> The gin file was written for TF version. Some differences might need to be paid attention to. For example, in the .gin file `Adam` actually corresponds to `AdamTF` in our pytorch version.
> 
> Anyway, it seems that the performance posted here is worse than the figure previously checked in.

I did not realize that `rnd_mrevenge.png` is actually the graph for this one. Let me check.

breakds:
So I tried the original `.gin` and it produces almost the same graph:

ppo_rnd_mrevenge-2

Agree that we will want to tune the algorithm to reproduce what we have before - in this case, do you think we can do the conversion first and add a TODO in the configuration to tune it later?

hnyu:
> So I tried the original `.gin` and it produces almost the same graph:
> 
> ppo_rnd_mrevenge-2
> 
> Agree that we will want to tune the algorithm to reproduce what we have before - in this case, do you think we can do the conversion first and add a TODO in the configuration to tune it later?

Let me quickly run this example with AdamTF and see how it goes.

breakds:
> > So I tried the original `.gin` and it produces almost the same graph:
> > ppo_rnd_mrevenge-2
> > Agree that we will want to tune the algorithm to reproduce what we have before - in this case, do you think we can do the conversion first and add a TODO in the configuration to tune it later?
> 
> Let me quickly run this example with AdamTF and see how it goes.

Thanks! Btw I tried the AdamTF with the `py conf` version and it seems slightly better at beginning and worse in the end.

It reaches 3K reward quickly and then dropped and never rise again.

alf/examples/ppo_rnd_mrevenge_conf.py
```diff
@@ -0,0 +1,110 @@
...
+    mini_batch_size=1024,
+    num_iterations=0,
+    num_env_steps=50000000,  # = 200M frames / 4 (frame_skip)
+    debug_summaries=True,
+    summarize_grads_and_vars=False,
+    summary_interval=100,
+    num_checkpoints=10,
+    use_rollout_state=True,
+    update_counter_every_mini_batch=True,
+    replay_buffer_length=35)
```
hnyu: `replay_buffer_length` is not used by `OneTimeExperienceReplayer`

alf/examples/ppo_rnd_mrevenge_conf.py
```diff
@@ -0,0 +1,110 @@
...
+import torch
+
+import functools
+
+# from alf.algorithms.actor_critic_algorithm import ActorCriticAlgorithm
+from alf.algorithms.rnd_algorithm import RNDAlgorithm
+from alf.networks import EncodingNetwork, ActorDistributionNetwork, CategoricalProjectionNetwork, ValueNetwork
+from alf.tensor_specs import TensorSpec
+
+from alf.examples import ppo_conf
```
hnyu, line 32: Also due to my previous PR, ppo_conf by default uses EntropyTargetAlgorithm. I think we'd better change it back to not use it, by

``` python
alf.config('Agent', rl_algorithm_cls=PPOAlgorithm, 
                 #enforce_entropy_target=True
)

#alf.config('EntropyTargetAlgorithm', initial_alpha=1.)
```

breakds:
Converted python config with  `AdamTF`:

ppo_rnd_mrevenge-adamtf

alf/examples/ppo_rnd_mrevenge_conf.py
breakds: It does matter after I removed `OnetimeExperienceReplayer`.

breakds: I cannot decide yet whether it's better to set it implicitly or explicitly for `whole_replay_buffer_training` case, but in this PR I think it is better to let it stick just for safety.

alf/examples/ppo_rnd_mrevenge_conf.py
breakds: Sure. We can create a separate PR to address this by disabling it in `ppo_conf.py` by default.

breakds:
PR is updated with TODO. PTAL.

hnyu (approved):

PR #1022, opened 2021-10-05
Replay buffer respects num_earliest_frames_ignored in `gather_all()` and `clear()`

breakds:
The is a further improvement over #1015 

# Motivation

We would like to retire `OnetimeExperienceReplayer` so that everything is done by using the replay buffer, even when the case `whole_replay_buffer_training = True`. Previously, when `whole_replay_buffer_training` is set to `True`, experiences from `unroll()` are collected as transformed while in the other cases experiences are collected as untransformed (raw).

If we want to unify the behavior, the more general way to handle this is to ask all experiences to be collected as raw, and apply transformation when they are taken out of the buffer for training. The problem is that one of the data transformation `FrameStacker` will need the replay buffer to respect `num_earliest_frames_ignored`.

Normally the replay buffer respect `num_earliest_frames_ignored` when experiences are taken out of it via `get_batch()`. However, for `whole_replay_buffer_training` this is done via `gather_all()` and `clear()` - neither of them respect `num_earliest_frames_ignored` at this moment.

# Solution

In this PR I made a few related changes to earn the respect for `num_earliest_frames_ignore`:

1. `gather_all()`, similar to `get_batch()`, will now return `batch_info` as well
2. `gather_all()`, when specified explicitly, will respect `num_earliest_frames_ignored` and drop such number of experiences.
3. `clear()`, when specified explicitly, will respect `num_earliest_frames_ignored` and keep such number of experiences (in addition to `keep_last_exp`) for the consumption of `FrameStacker` in the next round.

# Testing

Added a test case simulating the above scenario in the unit test, and make sure it passes.

# NOTE

This PR does not change the default behavior of `clear()` and `gather_all()`, and therefore is NOOP logically.

### DIFF
--- a/alf/experience_replayers/replay_buffer.py
+++ b/alf/experience_replayers/replay_buffer.py
@@ -557,14 +557,31 @@ class ReplayBuffer(RingBuffer):
         return first_step_pos
 
     @atomic
-    def gather_all(self):
+    def gather_all(self, ignore_earliest_frames: bool = False):
         """Returns all the items in the buffer.
 
+        Args:
+
+            ignore_earliest_frames: if set to ``True``, gather_all() will
+                respect ``num_earliest_frames_ignored`` and for each environment
+                it will return the trajectory with the first
+                ``num_earliest_frames_ignored`` experiences removed.
+
         Returns:
-            Tensors of shape [B, T, ...], B=num_environments, T=current_size
+            tuple:
+                - nested Tensors of shape [B, T, ...], where
+                  B=num_environments, T=current_size
+                - BatchInfo: Information about the batch. Its shapes are [B], where
+                  B=num_environments
+                    - env_ids: [0, 1, 2, ..., num_envs - 1]
+                    - positions: starting position in the replay buffer for each
+                      of the sequences. For gather_all() all starting positions will
+                      be the same.
+
         Raises:
             AssertionError: if the current_size is not same for all the
             environments.
+
         """
         size = self._current_size.min()
         max_size = self._current_size.max()
@@ -585,13 +602,27 @@ class ReplayBuffer(RingBuffer):
         # buffer whose data can start from the middle, so this is limited
         # to the case where clear() is the only way to remove data from
         # the buffer.
-        if size == self._max_length:
+        if size == self._max_length and not ignore_earliest_frames:
             result = self._buffer
+            info = BatchInfo(
+                env_ids=torch.arange(self._num_envs),
+                positions=torch.tensor([0] * self._num_envs))
         else:
+            if ignore_earliest_frames:
+                actual_start = self._num_earliest_frames_ignored
+            else:
+                actual_start = 0
             # Assumes that non-full buffer always stores data starting from 0
-            result = alf.nest.map_structure(lambda buf: buf[:, :size, ...],
-                                            self._buffer)
-        return convert_device(result)
+            result = alf.nest.map_structure(
+                lambda buf: buf[:, actual_start:size, ...], self._buffer)
+            info = BatchInfo(
+                env_ids=torch.arange(self._num_envs),
+                positions=torch.tensor([actual_start] * self._num_envs))
+            if alf.get_default_device() != self._device:
+                result, info = convert_device((result, info))
+            info = info._replace(replay_buffer=self)
+
+        return result, info
 
     def dequeue(self, env_ids=None):
         raise NotImplementedError(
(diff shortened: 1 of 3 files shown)


alf/experience_replayers/replay_buffer.py
```diff
@@ -651,15 +720,20 @@ def clear(self, keep_last_exp: bool = False) -> None:
...
 
-            # The index of the last experience in the ring buffer will be the
-            # modulus remainder of pos - 1.
-            last_exp_idx = self.circular(pos - 1)
-            last_exp_in_buffer = alf.nest.map_structure(
-                lambda buf: buf[:, last_exp_idx, ...], self._buffer)
-
-            # Clear the ring buffer and put the last experience back into it.
+            kept = []
+            for i in range(num_to_keep):
```
emailweixu, line 735: num_to_keep = min(pos, num_to_keep)

alf/experience_replayers/replay_buffer.py
```diff
@@ -585,13 +602,27 @@ def gather_all(self):
         # buffer whose data can start from the middle, so this is limited
         # to the case where clear() is the only way to remove data from
         # the buffer.
-        if size == self._max_length:
+        if size == self._max_length and not ignore_earliest_frames:
             result = self._buffer
+            info = BatchInfo(
+                env_ids=torch.arange(self._num_envs),
+                positions=torch.tensor([0] * self._num_envs))
```
emailweixu: torch.zeros(self._num_envs, dtype=torch.int64)

alf/experience_replayers/replay_buffer.py
```diff
@@ -622,26 +653,64 @@ def get_field(self, field_name, env_ids, positions):
...
+            num_to_keep += 1
+        # keep_as_earliest_frames requires keeping num_earliest_frames_ignored
+        # extra experiences from the end.
+        if keep_as_earliest_frames:
+            num_to_keep += self._num_earliest_frames_ignored
+
+        # TODO(breakds): At this moment I am not sure whether retaining
+        # experiences after clearning will interfere in a bad way with
+        # keep_episodic_info. Therefore this is prohibited explicitly for now.
+        # Lift this in the future if we find a solution.
```
le-horizon: Good point, even in the case num_to_keep == 0, clear() should also clear the two _indexed_pos buffers.  Let's either expand the assert to `assert not self._keep_episodic_info` or add a `clear()` for this class to properly clear the _indexed_pos buffers?

alf/experience_replayers/replay_buffer.py
```diff
@@ -651,15 +720,20 @@ def clear(self, keep_last_exp: bool = False) -> None:
...
+                kept.append(
+                    alf.nest.map_structure(lambda buf: buf[:, exp_idx, ...],
+                                           self._buffer))
+
+            # Clear the ring buffer and put the last several kept
+            # experiences back into it.
             super().clear(env_ids=None)
-            self.add_batch(last_exp_in_buffer)
+            for exp in reversed(kept):
+                self.add_batch(exp)
```
le-horizon, line 747: using for loops is ok if num_to_keep is small.  Maybe add an assert num_to_keep < 1000?
for large deletes, you can use RingBuffer.remove_up_to(n), we have to also deal with the _headless_indexed_pos buffer.  It's a bit tricky because some of the envs could contain headless episodes due to the mass removal.  We'll need to first identify the newest removed FIRST steps in each env and then update _headless_indexed_pos, so probably add as a TODO for now.  (_indexed_pos is easier, just set to 0 is probably sufficient.)

alf/experience_replayers/replay_buffer_test.py
```diff
@@ -531,6 +533,61 @@ def test_clear_with_keep_last_exp(self):
...
+        # are ignoring the earliest 2 experiences, the result of
+        # gather_all will be t = 8 through t = 14.
+        experience, batch_info = replay_buffer.gather_all(
+            ignore_earliest_frames=True)
+
+        self.assertEqual(torch.tensor([0, 1, 2, 3]), batch_info.env_ids)
+        self.assertEqual(torch.tensor([2, 2, 2, 2]), batch_info.positions)
+        self.assertEqual(
+            torch.tensor([[8, 9, 10, 11, 12, 13, 14]] * 4), experience.t)
+
```
le-horizon, line 590: Maybe also add test that _indexed_pos is properly cleared, if you plan to take this route (instead of more strict assert).

le-horizon (commented):
Thanks for finding out and improving for this tricky case, Break.

Mostly good.  Only some suggestions and TODOs.

Le

alf/experience_replayers/replay_buffer.py
breakds: I am going to use `num_to_keep = min(num_to_keep, self.total_size)` to make it slightly more readable. The logic should be identical.

breakds: Actually after thinking about it, what we need here is that if the total size is smaller than `num_to_keep`, we just do nothing for `clear()`.

alf/experience_replayers/replay_buffer_test.py
breakds: Extend the `assert` instead. Thanks for pointing this out.

alf/experience_replayers/replay_buffer.py
breakds: Added an `assert` to prevent `num_to_keep` becomes too large as suggested. Currently added a `TODO` for `_indexed_pos` and `headless_indexed_pos`, but I plan to have a discussion with @le-horizon to make the decision here.

alf/experience_replayers/replay_buffer.py
emailweixu: good point

alf/experience_replayers/replay_buffer.py
```diff
@@ -622,26 +654,73 @@ def get_field(self, field_name, env_ids, positions):
...
+        # keep_last_exp requires keeping 1 extra experience from the end.
+        if keep_last_exp:
+            num_to_keep += 1
+        # keep_as_earliest_frames requires keeping num_earliest_frames_ignored
+        # extra experiences from the end.
+        if keep_as_earliest_frames:
+            num_to_keep += self._num_earliest_frames_ignored
+
+        # Do nothing if there is only so many experiences in the buffer.
+        if num_to_keep >= self.total_size:
```
emailweixu: total_size is sum of self._current_size. num_to_keep cannot be compared with it.

breakds: Thanks for pointing this out. I updated this to `num_to_keep >= self._current_size.max()`.

emailweixu (approved):

breakds:
There was another problem though: 

```python
        if num_earliest_frames_ignored > 0:
            if keep_episodic_info is None:
                keep_episodic_info = True

```

`keep_episodic_info` is automatically on when `num_earliest_frames_ignored` is specified, unless we **explicitly** disable `keep_episodic_info`. I am not sure why we want the above logic but this means that simply assert `not num_earliest_frames_ignored` is not working.

alf/experience_replayers/replay_buffer.py
```diff
@@ -622,26 +654,74 @@ def get_field(self, field_name, env_ids, positions):
...
+            num_to_keep += 1
+        # keep_as_earliest_frames requires keeping num_earliest_frames_ignored
+        # extra experiences from the end.
+        if keep_as_earliest_frames:
+            num_to_keep += self._num_earliest_frames_ignored
+
+        # Do nothing if there is only so many experiences (per environment) in
+        # the buffer.
+        if num_to_keep >= self._current_size.max():
+            return
```
le-horizon, line 714: instead of silently return and potentially causing problem with RingBuffer, maybe we should just assert num_to_keep < self._current_size.min()?  Sorry for the delayed comment.

PR #1025, opened 2021-10-07
Convert gin file to conf for fixed_pid_carla method

Haichao-Zhang:


alf/examples/fixed_pid_carla_conf.py
```diff
@@ -0,0 +1,79 @@
...
+alf.config(
+    'suite_carla.Player',
+    # camera is on for visualization purpose during play
+    with_camera_sensor=True,
+    with_gnss_sensor=False,
+    with_imu_sensor=False,
+    with_radar_sensor=False,
+    sparse_reward=False,
+    allow_negative_distance_reward=True,
+    max_collision_penalty=20.,
```
hnyu: Not completely the same with the gin file?

alf/examples/fixed_pid_carla_conf.py
```diff
@@ -0,0 +1,79 @@
...
+    mini_batch_size=64,
+    num_updates_per_train_iter=1,
+    num_iterations=100,
+    num_checkpoints=20,
+    evaluate=False,
+    debug_summaries=True,
+    summarize_grads_and_vars=False,
+    summary_interval=100,
+    num_summaries=1000,
+    replay_buffer_length=10,  # smaller replay buffer
```
hnyu: Why this is only 10?

hnyu (commented):
I think it will be better if this could be based on sac_carla_conf.py

alf/examples/fixed_pid_carla_conf.py
Haichao-Zhang: Now commented out with comments.

alf/examples/fixed_pid_carla_conf.py
Haichao-Zhang: Only use a small replay buffer since there is no need for training since it is a ```hand-crafted``` algorithm.

Haichao-Zhang:
> I think it will be better if this could be based on sac_carla_conf.py

As the algorithm is relatively simple, its conf is also very simple. Including sac_carla_conf will include many unused components and makes the dependency relationship less clear.

hnyu (approved):

PR #1026, opened 2021-10-08
 [Refactor] Replay buffer's gather_all() now respects ring buffer order

breakds:
# Motivation

Previously the `gather_all()` method of `ReplayBuffer` assumes the experiences in the buffer are stored **in order** from index 0 to index `size - 1`. Such assumption simplifies the implementation of `gather_all()` without considering the general case of ring buffer.

The above assumption is **no longer going to hold** as we plan to retire replay buffer's `clear()`. The original `clear()` can be effectively done by pushing enough number of experiences to flush the old experiences in the buffer, which will definitely fall into the general ring buffer case when `gather_all()` is called. Thanks to @emailweixu's inspiration!

# Solution

1. Update the implementation of `gather_all()` so that it handles the general ring buffer order as well. This is a backward compatible change.
2. Remove the specialized `clear()` of `ReplayBuffer` since we will not need it in the future. As a bonus, we do not need to handle `keep_episodic_info` issue in the `clear()` as well.

# Testing

Adapt the unit tests so it properly tests the case when the replay buffer is not full, and after the replay buffer circulates.

### DIFF
--- a/alf/experience_replayers/replay_buffer.py
+++ b/alf/experience_replayers/replay_buffer.py
@@ -651,13 +651,10 @@ class ReplayBuffer(RingBuffer):
 
         info = BatchInfo(
             env_ids=torch.arange(self._num_envs),
-            positions=(
-                torch.ones(self._num_envs, dtype=torch.int64) * start_pos))
+            positions=(torch.full((self._num_envs, ),
+                                  start_pos,
+                                  dtype=torch.int64)))
 
-        # TODO(breakds): Maybe remove this in a future PR. It is not desired to
-        # move everything to the GPU memory yet. It will be more flexiable (and
-        # GPU memory friendly) to do this in a later stage where we only need to
-        # store a mini batch in the GPU memory at a atime.
         if alf.get_default_device() != self._device:
             result, info = convert_device((result, info))
 
--- a/alf/experience_replayers/replay_buffer.py
+++ b/alf/experience_replayers/replay_buffer.py
@@ -588,40 +588,80 @@ class ReplayBuffer(RingBuffer):
         assert size == max_size, (
             "Not all environments have the same size. min_size: %s "
             "max_size: %s" % (size, max_size))
-        if size < self._max_length:
-            pos = self._current_pos.min()
-            max_pos = self._current_pos.max()
-            assert pos == max_pos, (
-                "Not all environments have the same ending position. "
-                "min_pos: %s max_pos: %s" % (pos, max_pos))
-            assert size == pos, (
-                "When buffer not full, ending position of the data in the "
-                "buffer current_pos coincides with current_size")
 
-        # NOTE: this is not the proper way to gather all from a ring
-        # buffer whose data can start from the middle, so this is limited
-        # to the case where clear() is the only way to remove data from
-        # the buffer.
-        if size == self._max_length and not ignore_earliest_frames:
-            result = self._buffer
-            info = BatchInfo(
-                env_ids=torch.arange(self._num_envs),
-                positions=torch.zeros(self._num_envs, dtype=torch.int64))
-        else:
-            if ignore_earliest_frames:
-                actual_start = self._num_earliest_frames_ignored
-            else:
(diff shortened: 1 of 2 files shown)


alf/experience_replayers/replay_buffer.py
```diff
@@ -588,40 +588,80 @@ def gather_all(self, ignore_earliest_frames: bool = False):
...
+            env_ids=torch.arange(self._num_envs),
+            positions=(
+                torch.ones(self._num_envs, dtype=torch.int64) * start_pos))
+
+        # TODO(breakds): Maybe remove this in a future PR. It is not desired to
+        # move everything to the GPU memory yet. It will be more flexiable (and
+        # GPU memory friendly) to do this in a later stage where we only need to
+        # store a mini batch in the GPU memory at a atime.
+        if alf.get_default_device() != self._device:
+            result, info = convert_device((result, info))
```
emailweixu, line 659: If the caller really want the data on CPU, it can always call this function under CPU context using `with alf.device("cpu")`

alf/experience_replayers/replay_buffer.py
```diff
@@ -588,40 +588,80 @@ def gather_all(self, ignore_earliest_frames: bool = False):
...
-                result, info = convert_device((result, info))
-            info = info._replace(replay_buffer=self)
+                lambda buf: torch.cat(
+                    (buf[:, start_idx:, ...], buf[:, :end_idx, ...]), dim=1),
+                self._buffer)
+
+        info = BatchInfo(
+            env_ids=torch.arange(self._num_envs),
+            positions=(
+                torch.ones(self._num_envs, dtype=torch.int64) * start_pos))
```
emailweixu: Use torch.full() instead:
```
torch.full((self._num_envs,), start_pos, dtype=torch.int64)
```

alf/experience_replayers/replay_buffer.py
breakds: Makes a lot of sense. Did not know this convention.
I removed the `TODO` comments.

alf/experience_replayers/replay_buffer.py
breakds: Learned a new trick. Thanks!

breakds:
Updated and PTAL.

alf/experience_replayers/replay_buffer.py
```diff
@@ -588,40 +588,77 @@ def gather_all(self, ignore_earliest_frames: bool = False):
...
+            # This is the more complicated case when
+            #
+            # |#########o---o*******************| env 0
+            # |#########o---o*******************| env 1
+            # |             .                   |
+            # |             .                   |
+            # |             .                   |
+            # |#########o---o*******************| env n-1
+            #           ^   ^
+            #     end_idx   start_idx
```
le-horizon, line 641: nice illustrations!

le-horizon (dismissed):
Nice implementation!  We don't actually need remove_up_to(n), which will cause extra problems with episodic info stored in _headless_indexed_pos.

LGTM.

Le

alf/experience_replayers/replay_buffer.py
```diff
@@ -651,13 +651,10 @@ def gather_all(self, ignore_earliest_frames: bool = False):
 
         info = BatchInfo(
             env_ids=torch.arange(self._num_envs),
-            positions=(
-                torch.ones(self._num_envs, dtype=torch.int64) * start_pos))
+            positions=(torch.full((self._num_envs, ),
```
emailweixu: It seems that the '()' surrounding torch.full is not necessary.

breakds: Right. I have them removed.

breakds (commented):
> Nice implementation! We don't actually need remove_up_to(n), which will cause extra problems with episodic info stored in _headless_indexed_pos.
> 
> LGTM.
> 
> Le

Yep realized them after discussion with Wei. Thanks!

emailweixu (approved):

PR #1027, opened 2021-10-08
[Refactor] Retire one time experience replayer

breakds:
# Motivation

This the second attempt to retry the refactor proposed by #1012. The previous one did not handle the frame stacker case for whole replay buffer training properly and therefore caused breakage.

Currently there is a concept distinguishing 2 types of experience replayer, `one_time` and `uniform`, where the former does not count as "using replay buffer" and the latter counts. We would like to simplify this by having only "the experience replayer with replay buffer" because:

1. One less concept to learn for the new users, especially when such behavior is implicitly configured before
2. Less confusing as "uniform" does not directly imply "it can be used similar to one time experience replayer as well", i.e. always "replay_all"
3. When updating logic of replayers, we now have only one thing to worry about

# Solution

Remove `OnetimeExperienceReplayer` and experience replayer type related logic entirely. Several thoughts about this decision:

1. This will increase the time consumption of the training, but that increase is negligible. Such increase mainly comes from saving the experience in cpu during observation and copying them to GPU during updating. In procgen environment training (whose image observation is huge), this adds about 3ms to each training iteration.
2. To ensure `observe_for_replay` is called after `set_exp_replayer` is called, it will raise exception upon such incorrect calls.
3. In `whole_replay_buffer_training` case, `batch_info` is now returned by `replay_all()`, and it participates the data transformation. Unlike previously, now even for `whole_replay_buffer_training`, the data transformation is done at the time when the experiences are taken out of the replay buffer.
4. There is no replay buffer `clear()` any more. For `whoe_replay_buffer_training`, we rely on correctly set the replay buffer's size to effectively do the "clear with a few experiences kept" behavior.
5. Also, the `initial_collect_steps` will be updated automatically for whole replay buffer training if the given value is not big enough.

# Testing

Tested `ppo_cart_pole` and `ppo_rnd_mrevenge` where both used to use `one-time` experience replayer.

### DIFF
--- a/alf/algorithms/algorithm.py
+++ b/alf/algorithms/algorithm.py
@@ -1306,9 +1306,9 @@ class Algorithm(AlgorithmInterface):
             batch_info = BatchInfo(
                 env_ids=batch_info.env_ids.repeat_interleave(
                     num_mini_batches_per_original_traj),
-                positions=(batch_info.positions.repeat_interleave(
+                positions=batch_info.positions.repeat_interleave(
                     num_mini_batches_per_original_traj) + torch.arange(
-                        0, length, mini_batch_length).repeat(num_envs)),
+                        0, length, mini_batch_length).repeat(num_envs),
                 replay_buffer=batch_info.replay_buffer)
 
             # Treatment 2: Adjust the mini_batch_size.
--- a/alf/algorithms/algorithm.py
+++ b/alf/algorithms/algorithm.py
@@ -248,8 +248,9 @@ class Algorithm(AlgorithmInterface):
             # Do not even create the experience replayer if the
             # required parameters are not set by set_exp_replayer
             common.warning_once(
-                'Experience replayer must be initialized first by calling set_exp_replayer() before observe_for_replay() is called! Skipping ...'
-            )
+                'Experience replayer must be initialized first by calling '
+                'set_exp_replayer() before observe_for_replay() is called! '
+                'Skipping ...')
             return
 
         self._experience_spec = dist_utils.extract_spec(sample_exp, from_dim=1)
@@ -283,12 +284,7 @@ class Algorithm(AlgorithmInterface):
                     exp.state, self.train_state_spec, value_to_match=()))
 
         if self._exp_replayer is None:
-            # Do not even create the experience replayer if the
-            # required parameters are not set by set_exp_replayer
-            if (self._exp_replayer_num_envs is not None
-                    and self._exp_replayer_length is not None
-                    and self._prioritized_sampling is not None):
-                self._set_exp_replayer(exp)
+            self._set_exp_replayer(exp)
 
         exp = dist_utils.distributions_to_params(exp)
         for observer in self._observers:
--- a/alf/algorithms/algorithm.py
+++ b/alf/algorithms/algorithm.py
@@ -27,8 +27,7 @@ from torch.nn.modules.module import _IncompatibleKeys, _addindent
 
 import alf
 from alf.data_structures import AlgStep, LossInfo, StepType, TimeStep, experience_to_time_step
-from alf.experience_replayers.experience_replay import (
-    OnetimeExperienceReplayer, SyncExperienceReplayer)
+from alf.experience_replayers.experience_replay import SyncExperienceReplayer
 from alf.utils.checkpoint_utils import is_checkpoint_enabled
 from alf.utils import common, dist_utils, spec_utils, summary_utils
 from alf.utils.summary_utils import record_time
@@ -132,7 +131,11 @@ class Algorithm(AlgorithmInterface):
         self._observers = []
         self._metrics = []
         self._exp_replayer = None
-        self._exp_replayer_type = None
+
+        # These 3 parameters are only set when ``set_exp_replayer()`` is called.
+        self._exp_replayer_num_envs = None
+        self._exp_replayer_length = None
+        self._prioritized_sampling = None
 
         self._use_rollout_state = False
         if config:
@@ -215,24 +218,18 @@ class Algorithm(AlgorithmInterface):
         self._set_children_property('use_rollout_state', flag)
 
     def set_exp_replayer(self,
-                         exp_replayer: str,
                          num_envs,
                          max_length: int,
                          prioritized_sampling=False):
         """Set experience replayer.
 
         Args:
-            exp_replayer (str): type of experience replayer. One of ("one_time",
-                "uniform")
             num_envs (int): the total number of environments from all batched
                 environments.
             max_length (int): the maximum number of steps the replay
                 buffer store for each environment.
             prioritized_sampling (bool): Use prioritized sampling if this is True.
         """
-        assert exp_replayer in ("one_time", "uniform"), (
-            "Unsupported exp_replayer: %s" % exp_replayer)
-        self._exp_replayer_type = exp_replayer
         self._exp_replayer_num_envs = num_envs
         self._exp_replayer_length = max_length
         self._prioritized_sampling = prioritized_sampling
@@ -245,24 +242,28 @@ class Algorithm(AlgorithmInterface):
         Args:
             sample_exp (nested Tensor):
         """
+        if (self._exp_replayer_num_envs is None
+                or self._exp_replayer_length is None
+                or self._prioritized_sampling is None):
+            # Do not even create the experience replayer if the
+            # required parameters are not set by set_exp_replayer
+            common.warning_once(
+                'Experience replayer must be initialized first by calling set_exp_replayer() before observe_for_replay() is called! Skipping ...'
+            )
+            return
+
         self._experience_spec = dist_utils.extract_spec(sample_exp, from_dim=1)
         self._exp_contains_step_type = ('step_type' in dict(
             alf.nest.extract_fields_from_nest(sample_exp)))
 
-        if self._exp_replayer_type == "one_time":
-            self._exp_replayer = OnetimeExperienceReplayer()
-        elif self._exp_replayer_type == "uniform":
-            exp_spec = dist_utils.to_distribution_param_spec(
-                self._experience_spec)
-            self._exp_replayer = SyncExperienceReplayer(
-                exp_spec,
-                self._exp_replayer_num_envs,
-                self._exp_replayer_length,
-                prioritized_sampling=self._prioritized_sampling,
-                num_earliest_frames_ignored=self._num_earliest_frames_ignored,
-                name="exp_replayer")
-        else:
-            raise ValueError("invalid experience replayer name")
+        exp_spec = dist_utils.to_distribution_param_spec(self._experience_spec)
+        self._exp_replayer = SyncExperienceReplayer(
+            exp_spec,
+            self._exp_replayer_num_envs,
+            self._exp_replayer_length,
+            prioritized_sampling=self._prioritized_sampling,
+            num_earliest_frames_ignored=self._num_earliest_frames_ignored,
+            name="exp_replayer")
         self._observers.append(self._exp_replayer.observe)
 
     def observe_for_replay(self, exp):
@@ -281,8 +282,13 @@ class Algorithm(AlgorithmInterface):
                 state=alf.nest.prune_nest_like(
                     exp.state, self.train_state_spec, value_to_match=()))
 
-        if self._exp_replayer is None and self._exp_replayer_type:
-            self._set_exp_replayer(exp)
+        if self._exp_replayer is None:
+            # Do not even create the experience replayer if the
+            # required parameters are not set by set_exp_replayer
+            if (self._exp_replayer_num_envs is not None
+                    and self._exp_replayer_length is not None
+                    and self._prioritized_sampling is not None):
+                self._set_exp_replayer(exp)
 
         exp = dist_utils.distributions_to_params(exp)
         for observer in self._observers:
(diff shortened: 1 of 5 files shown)


breakds:
The base branch of this PR does not seem to be right. I'll try to close #1026 first.

breakds:
I'll work on resolving the unit test failure first.

alf/bin/train_play_test.py
```diff
@@ -124,6 +124,9 @@ def _to_conf_params(parameters):
     'TrainerConfig.num_checkpoints=1',
     # disable evaluate
     'TrainerConfig.evaluate=False',
+    # Do not get stuck there and asking for confirmation upon crash.
+    # Die and give feedback immediately for unit tests.
+    'confirm_checkpoint_upon_crash=False',
```
breakds, line 127: Adding this because otherwise running the unit tests that are supposed to crash immediately will hang forever locally.

breakds:
Sorry that this PR is bigger than my previous PRs. It is ready for review now. 

Please read the description on what is changed and let me know if I can make the description clearer. Thanks!

alf/algorithms/algorithm.py
```diff
@@ -245,24 +242,28 @@ def _set_exp_replayer(self, sample_exp):
         Args:
             sample_exp (nested Tensor):
         """
+        if (self._exp_replayer_num_envs is None
+                or self._exp_replayer_length is None
+                or self._prioritized_sampling is None):
+            # Do not even create the experience replayer if the
+            # required parameters are not set by set_exp_replayer
+            common.warning_once(
+                'Experience replayer must be initialized first by calling set_exp_replayer() before observe_for_replay() is called! Skipping ...'
```
emailweixu: make the line shorter?

alf/algorithms/algorithm.py
```diff
@@ -281,8 +282,13 @@ def observe_for_replay(self, exp):
...
 
-        if self._exp_replayer is None and self._exp_replayer_type:
-            self._set_exp_replayer(exp)
+        if self._exp_replayer is None:
+            # Do not even create the experience replayer if the
+            # required parameters are not set by set_exp_replayer
+            if (self._exp_replayer_num_envs is not None
+                    and self._exp_replayer_length is not None
+                    and self._prioritized_sampling is not None):
+                self._set_exp_replayer(exp)
```
emailweixu: line 288-290 seems unnecessary given line 245-245

alf/algorithms/algorithm.py
```diff
@@ -1157,27 +1160,45 @@ def train_from_replay_buffer(self, update_global_counter=False):
...
+        experience = alf.data_structures.add_batch_info(
+            experience, batch_info, self._exp_replayer.replay_buffer)
+        experience = self.transform_experience(experience)
+
+        if whole_replay_buffer_training:
+            # For whole replay buffer training, explicitly set
+            # batch_info to None as an indicator for the following
+            # code to skip certain logics specifically designed for
+            # sampled replay buffer training.
+            batch_info = None
```
emailweixu: I think we don't need to do this for whole_replay_buffer_training. line 1204 may still use it. Seems that batch_info is needed for line 1220. However, line 1220 seems unnecessary.

alf/algorithms/rl_algorithm.py
```diff
@@ -204,12 +258,23 @@ def __init__(self,
...
-                                  config.replay_buffer_length,
+            replay_buffer_length = adjust_replay_buffer_length(
+                config, self._num_earliest_frames_ignored)
+
+            if config.whole_replay_buffer_training:
+                # For whole replay buffer training, we would like to be sure
+                # that the replay buffer have enough samples in it to perform
+                # the training, which will most likely happen in the 2nd
+                # iteration. The minimum_initial_collect_steps guarantees that.
+                minimum_initial_collect_steps = replay_buffer_length * self._env.batch_size
```
emailweixu, line 269: should this be `(unroll_length + num_earliest_frames_ignored) * self._env.batch_size` ?

breakds: Here the `replay_buffer_length` is already adjusted, which should have included `num_earliest_frames_ignored` if it is set.

alf/algorithms/algorithm.py
breakds: Removed.

alf/algorithms/algorithm.py
breakds: Done.

alf/algorithms/algorithm.py
breakds: It is also used in line 1297 below. Did you mean we should not rely on `batch_info` in those places for different behaviors?

breakds:
Before and after the change, the training behavior of `ppo_rnd_mrevenge` changed:

ppo_rnd_mrevenge_000

Bule = after the change
Orange = before the change

I am not quite sure whether the difference is important but they are pretty consistent - I trained twice for each of the settings. Will take a closer look tomorrow. Meanwhile I am still waiting for `ppo_procgen`'s training result. Will update once it is out.

hnyu:
> Before and after the change, the training behavior of `ppo_rnd_mrevenge` changed:
> 
> ppo_rnd_mrevenge_000
> 
> Bule = after the change Orange = before the change
> 
> I am not quite sure whether the difference is important but they are pretty consistent - I trained twice for each of the settings. Will take a closer look tomorrow. Meanwhile I am still waiting for `ppo_procgen`'s training result. Will update once it is out.

Is this change because of AdamTF we last discussed, or something else?

breakds:
> > Before and after the change, the training behavior of `ppo_rnd_mrevenge` changed:
> > ppo_rnd_mrevenge_000
> > Bule = after the change Orange = before the change
> > I am not quite sure whether the difference is important but they are pretty consistent - I trained twice for each of the settings. Will take a closer look tomorrow. Meanwhile I am still waiting for `ppo_procgen`'s training result. Will update once it is out.
> 
> Is this change because of AdamTF we last discussed, or something else?

Something else. Both "before change" and "after change" are using `AdamTF`.

hnyu:
> > > Before and after the change, the training behavior of `ppo_rnd_mrevenge` changed:
> > > ppo_rnd_mrevenge_000
> > > Bule = after the change Orange = before the change
> > > I am not quite sure whether the difference is important but they are pretty consistent - I trained twice for each of the settings. Will take a closer look tomorrow. Meanwhile I am still waiting for `ppo_procgen`'s training result. Will update once it is out.
> > 
> > 
> > Is this change because of AdamTF we last discussed, or something else?
> 
> Something else. Both "before change" and "after change" are using `AdamTF`.

I think last time according to the discussion, AdamTF can get about 3k in the end even before the change?

hnyu:
> > > Before and after the change, the training behavior of `ppo_rnd_mrevenge` changed:
> > > ppo_rnd_mrevenge_000
> > > Bule = after the change Orange = before the change
> > > I am not quite sure whether the difference is important but they are pretty consistent - I trained twice for each of the settings. Will take a closer look tomorrow. Meanwhile I am still waiting for `ppo_procgen`'s training result. Will update once it is out.
> > 
> > 
> > Is this change because of AdamTF we last discussed, or something else?
> 
> Something else. Both "before change" and "after change" are using `AdamTF`.

The orange curve seems still worse than 3k as we got the last time, but I don't know if it is because of randomness or not.

breakds:
> > > > Before and after the change, the training behavior of `ppo_rnd_mrevenge` changed:
> > > > ppo_rnd_mrevenge_000
> > > > Bule = after the change Orange = before the change
> > > > I am not quite sure whether the difference is important but they are pretty consistent - I trained twice for each of the settings. Will take a closer look tomorrow. Meanwhile I am still waiting for `ppo_procgen`'s training result. Will update once it is out.
> > > 
> > > 
> > > Is this change because of AdamTF we last discussed, or something else?
> > 
> > 
> > Something else. Both "before change" and "after change" are using `AdamTF`.
> 
> The orange curve seems still worse than 3k as we got the last time, but I don't know if it is because of randomness or not.

Let me run more experiments.

alf/algorithms/algorithm.py
emailweixu: I think we can always provide batch_info so that user don't need to worry about whether it is available or not.

alf/algorithms/rl_algorithm.py
emailweixu: replay_buffer_length can be very big if clear_replay_buffer is False.

alf/algorithms/algorithm.py
breakds: Sure. I added this because originally for `whole_replay_buffer_training` it is set to `None`, and when not set to `None`, training will crash (I did not investigate into it). Let me dig into it.

breakds:
> The orange curve seems still worse than 3k as we got the last time, but I don't know if it is because of randomness or not.

Trained it again without this change and it falls to 1.2k. The training is not as consistent as we think.

ppo_rnd_mrevenge_2

Red = before change
Orange = after the change

On the other hand, both scores are still less than 3k, even though the historical high score is 6k.

breakds:
The comparison for `ppo_procgen_bossfight` looks like below. 

ppo_procgen_bossfight_2


light blue: before change
dark blue: after change

There are difference but both of them fall to zero at the end of the training (7000 iterations), while (4k iterations = PPG's experiement's steps).

## Proposed Explanation

I am now seeing that 

1. Before the change and after the change, `ppo_cart_pole`, `ppo_rnd_mrevenge` and `ppo_procgen_bossfight` having the same or **very similar** training behavior.
2. Although, `ppo_rnd_mrevenge` and `ppo_procgen_bossfight` are indeed **worse** than they used to be, but I think those are **not a direct consequence** of this PR.

## Plan

1. I am running experiment to see whether `ppo_rnd_mreveng` can consistently achieve 3k or even better in a previous commit of alf.
2. `ppo_procgen_bossfight` will be further tested and tuned because it serves as the baseline for PPG.

breakds:
This PR is updated with special treatment @emailweixu and I discussed:

1. Update the `batch_info` for `whole_replay_buffer_training` so that it does not crash when we adjust the `batch_info` after the random permutation.
2. If the `batch_size` is not divisible by `mini_batch_size`, squeeze the remainder into the "big" mini batches to avoid having a tiny mini batch in the end.

PTAL. Thanks!

## Side notes

I found the original code on handling special cases here to be not as readable and not super clear. My new special treatment only makes readability **worse**. The decision here is to not further complicate the PR (that would not be fair to the reviewers), but I will have a follow-up PR clean-up the logic here with well designed unit tests to make sure that the logic works as designed.

alf/algorithms/algorithm.py
```diff
@@ -1275,11 +1281,65 @@ def _train_experience(self,
...
             return alf.nest.map_structure(lambda x: x.transpose(0, 1), nest)
 
+        print(
+            f'batch_size = {batch_size} vs mini_batch_size = {mini_batch_size}'
+        )
         for u in range(num_updates):
+            # TODO(breakds): Currently the ``if`` below will also be executed
+            # for the cases when whole_replay_buffer_training = False and
+            # config.num_updates_per_train_iter > 0. This is not necessary and
+            # not desired because in that case the sampled trajectories are
```
emailweixu: 1. sampled trajectories are not guaranteed to be in random order (e.g. experiences from gather_all, even for it is not even guaranteed for get_batch())
2. even if it is, it is still better to randomize the order between updates

alf/algorithms/algorithm.py
```diff
@@ -1275,11 +1281,65 @@ def _train_experience(self,
...
+            #
+            # env_ids =   [0,  0,  0,  0,  1,  1,  1,  1]
+            # positions = [7, 10, 13, 16,  7, 10, 13, 16] (before circular())
+            num_envs = batch_info.env_ids.shape[0]
+            replay_buffer = batch_info.replay_buffer
+            num_mini_batches_per_original_traj = length // mini_batch_length
+            batch_info = BatchInfo(
+                env_ids=batch_info.env_ids.repeat_interleave(
+                    num_mini_batches_per_original_traj),
+                positions=replay_buffer.circular(
```
emailweixu: circular should not be applied to positions

alf/algorithms/algorithm.py
breakds: Ah I see. I can see the reason 1. For No.2 in non `whole_replay_buffer_training` case, there is effectively only one update.

Removed the `TODO` here.

alf/algorithms/algorithm.py
breakds: Right. This is now fixed.

alf/algorithms/algorithm.py
```diff
@@ -1300,18 +1300,16 @@ def _train_experience(self,
...
             num_mini_batches_per_original_traj = length // mini_batch_length
             batch_info = BatchInfo(
                 env_ids=batch_info.env_ids.repeat_interleave(
                     num_mini_batches_per_original_traj),
-                positions=replay_buffer.circular(
-                    batch_info.positions.repeat_interleave(
-                        num_mini_batches_per_original_traj) + torch.arange(
-                            0, length, mini_batch_length).repeat(num_envs)),
-                replay_buffer=replay_buffer)
+                positions=(batch_info.positions.repeat_interleave(
```
emailweixu: unnecessary '(' after '='

breakds: Hmm I was trying to make it format better. Will remove it.

breakds: Updated.

emailweixu (dismissed):

emailweixu (approved):

PR #1028, opened 2021-10-11
A tool to conveniently create a docker-based environment that resembles the CI env

breakds:
# Motivation

There are certain cases when you want to debug code and run unit tests or sanity check while

1. CI failed in a way that we do not anticipate because the CI environment is different from our dev environment
2. You might not have an already setup alf dev environment on some other laptop/workstation

In those cases we probably want to create  a development environment that resembles the CI environment.

# Solution

In this PR a tool `box.sh` is implemented to quickly throw such an environment to you. It is a bash script and the only dependency is `docker`.

# Testing

See the video:

alf-dev-box


le-horizon (commented):
Nice tool to have.  Thanks for doing it Break!

The gif file is helpful, but maybe a bit large, and not very easy to control viewing.  Maybe upload to youtube and add a link here is better?

Thanks,
Le

breakds:
> Nice tool to have. Thanks for doing it Break!
> 
> The gif file is helpful, but maybe a bit large, and not very easy to control viewing. Maybe upload to youtube and add a link here is better?
> 
> Thanks, Le

Ah yep it is huge ... Sorry about that. I have removed the gif from the commit. It it not that necessary anyway :)

le-horizon (approved):
LG

breakds:
> LG

Ty for the prompt review!

PR #1029, opened 2021-10-12
[Refactor] Retire experience replayer

breakds:
This is a follow-up PR of #1027 and mostly straight-forward.

# Motivation

After the deprecation of `OneTimeExperienceReplayer`, the remaining `ExperienceReplayer` is merely a thin wrapper over `ReplayBuffer`, and it is a natural move to remove such thin wrapper as well to make our code base a bit more succinct.

Note that in the future if we find ourselves in the situation of needing a similar wrapper again, we can always have it back by revisiting this commit.

# Solution

Replace all the usage of `ExperienceReplayer` with `ReplayBuffer`.

# Testing

Ran `ppo_rnd_mrevenge` and this time it surprisingly reproduced the 3k score.

ppo_rnd_mrevenge_3


breakds:
This is now ready for review. 

Also the experiment for this PR on `ppo_rnd_mrevenge` reaches 3k (instead of 1.2k) which reproduces the previously high score (but now the **historical** high score 6k). This seems to suggest that `ppo_rnd_mrevenge` is not that stable.

emailweixu (approved):

PR #1030, opened 2021-10-13
PPG full implementation

breakds:
# Motivation

This is to implement #965. 

ppg

# Review Guide

Various depended changes has been split to other PRs and submitted. Even though this final PR is big.

Please be advised that the main changes are

1. `alf/algorithms/ppg/ppg_aux_phase_loss.py` - This adds the auxiliary phase loss implementation. PPG has two phases (See graph above) where the main phase use the PPO Loss, the auxiliary phase needs its own loss implementation, and this file implements that.
2. `alf/algorithms/ppo_loss.py` - Added an option to compute the advantages inside the loss. This is actually the default behavior of actor critic loss.
3. `alf/algorithms/ppg_algorithm.py` - The main PPG algorithm implementation. It is specifically designed to have a context manager that can switch the mode to "Auxiliary Phase". Auxliliary phase update is called periodically in `after_train_iter`, and when activated, certain parameters such as "which replay buffer to use" is temporarily modified until the auxliliary update is done.
4. There are two sample configurations for PPG algorithm added: `ppg_procgen_bossfight_conf.py` and `ppg_cart_pole`.

# Testing

Mainly on procgen bossfight environment, where the perfect score is `15`.

ppg_bossfight_1

Both curve above comes from running the PPG algorithm, and the difference is

1. The orange curve is PPG algorithm with aux phase update every 32 iteration
2. The green curve is PPG algorithm without any aux phase update, meaning that it is **effectively just PPO**.

We can see that 

- PPG reaches a relatively high score pretty quickly
- At the same number of (main) training iterations, PPG has a higher reward.
- PPO reaches 9 in the end (6000 iterations), which took it 15 hours. PPG reaches 9 in 780 iterations, which took 3 hours.

This roughly resembles the experiment in the paper (though my settings is slightly different from the paper as I am training on 1 card while the paper is on 4 cards)

ppg_paper_bossfight


breakds:
This PR is now ready for review. Thanks a lot!

alf/algorithms/ppg_algorithm.py
```diff
@@ -0,0 +1,468 @@
...
+       an extra auxiliary value head in addition to the policy head and value
+       head. In the current implementation, the auxiliary value head also tries
+       to estimate the value function, similar to the (actual) value head.
+
+    2. It does PPO update in normal iterations. However, after every specified
+       number of iterations, it will perform auxiliary phase updates based on
+       auxiliary phase losses (different from PPO loss, see
+       algorithms/ppg/ppg_aux_phase_loss.py for details). Auxiliary phase
+       updates does not require new rollouts. Instead it is performed on all of
+       the experience collected since the last auxiliary phase update.
```
emailweixu, line 51: add a reference to ppg paper

alf/algorithms/ppg_algorithm.py
```diff
@@ -0,0 +1,468 @@
...
+            self._active_phase = self._policy_phase
+            # Restore the _replay_buffer
+            self._replay_buffer = previous_replay_buffer
+            self._config.mini_batch_length = original_mini_batch_length
+            self._config.mini_batch_size = original_mini_batch_size
+            self._config.num_updates_per_train_iter = original_num_updates_per_train_iter
+
+    @property
+    def on_policy(self) -> bool:
+        return False
```
emailweixu: this is already implemented in OffPolicyAlgorithm

alf/algorithms/ppg_algorithm.py
```diff
@@ -0,0 +1,468 @@
...
+                aux_phase_interval iterations
+            env (Environment): The environment to interact with. env is a
+                batched environment, which means that it runs multiple
+                simulations simultateously. env only needs to be provided to the
+                root Algorithm. NOTE: env will default to None if PPGAlgorithm
+                is run via Agent.
+            config (TrainerConfig): config for training. config only needs to be
+                provided to the algorithm which performs ``train_iter()`` by
+                itself.
+            encoding_network_ctor (Callable[..., Network]): Function to
```
emailweixu: Comment how is this function called. And I think kernel_initializer should not be its required argument since it can always be configured if necessary.

alf/algorithms/ppg_algorithm.py
```diff
@@ -0,0 +1,468 @@
...
+            network=dual_actor_value_network,
+            optimizer=policy_optimizer,
+            loss=PPOLoss(debug_summaries=debug_summaries),
+            replay_buffer=None)
+
+        # The auxiliary phase uses an extra replay buffer with the loss
+        # specific to the auxiliary update phase.
+        self._aux_phase = PPGPhaseContext(
+            ppg_algorithm=self,
+            network=dual_actor_value_network.copy(),
```
emailweixu: It's not obvious why copy() is needed for everyone. Perhaps add some comment. (i.e. In ALF, one parameter cannot not be handled by multiple optimizers)

alf/algorithms/ppg_algorithm.py
```diff
@@ -0,0 +1,468 @@
...
+            num_envs = exp.env_id.shape[0]
+            max_length = self._config.unroll_length * self._aux_options.interval
+            max_length += 1 + self._num_earliest_frames_ignored
+            self._aux_phase._replay_buffer = ReplayBuffer(
+                data_spec=exp_spec,
+                num_environments=exp.env_id.shape[0],
+                max_length=max_length,
+                prioritized_sampling=False,
+                num_earliest_frames_ignored=self._num_earliest_frames_ignored,
+                name=f'aux_replay_buffer')
```
emailweixu: no need to have 'f'

alf/algorithms/ppg_algorithm.py
```diff
@@ -0,0 +1,468 @@
...
+    def rollout_step(self, inputs: TimeStep, state) -> AlgStep:
+        """Rollout step for PPG algorithm
+
+        Besides running the network prediction, it does one extra thing to store
+        the experience in the auxiliary replay buffer so that it can be consumed
+        by the auxiliary phase updates.
+
+        """
+        policy_step = self._policy_phase.network_forward(inputs, state)
+        self._observe_for_aux_replay(inputs.untransformed.cpu(), state,
```
emailweixu: To really integrate with Agent, it should store `inputs` instead of untransformed to the replay buffer. This is because Agent may change the original time step (e.g., by representation learner, or goal generator) before passing it to RL algorithm.

alf/algorithms/ppo_loss.py
```diff
@@ -31,6 +31,7 @@ def __init__(self,
                  td_error_loss_fn=element_wise_squared_loss,
                  td_lambda=0.95,
                  normalize_advantages=True,
+                 compute_advantages_internally=False,
```
emailweixu, line 34: doc string for this arg

alf/algorithms/ppo_loss.py
```diff
@@ -129,4 +131,6 @@ def _pg_loss(self, info, advantages):
         return policy_gradient_loss
 
     def _calc_returns_and_advantages(self, info, value):
-        return info.returns, info.advantages
+        if not self._compute_advantages_internally:
+            return info.returns, info.advantages
+        return super()._calc_returns_and_advantages(info, value)
```
emailweixu: should use the value from rollout to calculate the advantage and return

alf/algorithms/ppg_algorithm.py
```diff
@@ -0,0 +1,468 @@
...
+
+        """
+        previous_replay_buffer = None
+
+        original_mini_batch_length = self._config.mini_batch_length
+        original_mini_batch_size = self._config.mini_batch_size
+        original_num_updates_per_train_iter = self._config.num_updates_per_train_iter
+
+        try:
+            self._aux_phase.on_context_switch_from(self._policy_phase)
```
emailweixu: Putting line 167 here and 358 directly can make the code a bit easier to read.

emailweixu (changes requested):

alf/algorithms/ppg_algorithm.py
breakds: Reference added.

alf/algorithms/ppg_algorithm.py
breakds: Removed this.

alf/algorithms/ppg_algorithm.py
breakds: Removed `kernel_intializer` from the code, and updated the argument doc string to explicit state how this function should be called.

alf/algorithms/ppg_algorithm.py
breakds: Documentation added.

alf/algorithms/ppo_loss.py
breakds: Done.

alf/algorithms/ppo_loss.py
breakds: Updated this (and also that in `ppg_aux_loss`) so that they use the value target based on the rollout time value prediction.

alf/algorithms/ppg_algorithm.py
breakds: Agree, updated.

alf/algorithms/ppg_algorithm.py
breakds: This is previously required but not needed now after #1036. Removed it.

I was not aware that agent may change the original time step. Thanks for mentioning that!

breakds:
Updated by addressing the comments and integrated with #1036. PTAL.

Really appreciate your thorough review as well as the discussions @emailweixu . I did learn a lot for this!

alf/algorithms/ppo_loss.py
```diff
@@ -133,4 +133,6 @@ def _pg_loss(self, info, advantages):
     def _calc_returns_and_advantages(self, info, value):
         if not self._compute_advantages_internally:
             return info.returns, info.advantages
-        return super()._calc_returns_and_advantages(info, value)
+        value_to_use = info.rollout_value if hasattr(
```
emailweixu: value_to_use = getattr(info, 'rollout_value', info.value)

alf/algorithms/ppg_algorithm.py
```diff
@@ -0,0 +1,461 @@
...
+            # Make the policy phase network's parameters identical to the aux
+            # phase's.
+            self._policy_phase._network.load_state_dict(
+                self._aux_phase._network.state_dict())
+            self._active_phase = self._policy_phase
+            # Restore the _replay_buffer
+            self._replay_buffer = previous_replay_buffer
+            self._config.mini_batch_length = original_mini_batch_length
+            self._config.mini_batch_size = original_mini_batch_size
+            self._config.num_updates_per_train_iter = original_num_updates_per_train_iter
```
emailweixu: The code might be cleaner if aux-phase training is put in a sub-algorithm (say PPGAuxAlgorithm) so that we don't need to have this unnatural and potentially unsafe context switch (as it depends on too much detail of `train_from_replay_buffer()`) .
This PPGAuxAlgorithm will hold aux copy of the model, its own replay buffer and optimizer. We can all the PPGAuxAlgorithm.train_from_replay_buffer() to do the training without context switch. At PPGAlgorithm.rollout_step, we can call PPGAuxAlgorithm.observe_for_replay to store the experience.

alf/algorithms/ppo_loss.py
emailweixu: value_to_use = getattr(info, 'rollout_value', info.value)

alf/algorithms/ppg_algorithm.py
emailweixu: Another possible benefit is that we may be able to get rid of the problem of maintaining two copy of the model. This can be done by putting PPGAuxAlgorithm into _trainable_attributes_to_ignore of PPGAlgorithm

breakds: Good point. I think I understand how this is supposed to work now. Let me give it a try.

breakds: Algorithm is updated with a separate `PPGAuxAlgorithm`. Much cleaner!

breakds:
The PR is updated with a separate auxiliary update. PTAL.

Did another round of training to verify that it has the similar performance:

ppg_new

In the above graph, 

1. green = the refactored implementation
2. light blue = the previous implementation before the refactor

The performance is very close.

alf/algorithms/ppg_algorithm.py
```diff
@@ -0,0 +1,205 @@
...
+
+        if self._aux_algorithm:
+            self._aux_algorithm.observe_for_aux_replay(inputs, state,
+                                                       policy_step)
+
+        return policy_step
+
+    def train_step(self, inputs: TimeStep, state,
+                   plain_rollout_info: PPGRolloutInfo) -> AlgStep:
+        """Phase context dependent evaluation on experiences for training
```
emailweixu: outdated comment

alf/algorithms/ppg_algorithm.py
```diff
@@ -0,0 +1,205 @@
...
+        train_info = PPGTrainInfo(
+            action=plain_rollout_info.action,
+            rollout_value=plain_rollout_info.value,
+            rollout_action_distribution=plain_rollout_info.
+            action_distribution).absorbed(alg_step.info)
+
+        return alg_step._replace(info=train_info)
+
+    def calc_loss(self, info: PPGTrainInfo) -> LossInfo:
+        """Phase context dependent loss function evaluation
```
emailweixu: outdated comment

alf/algorithms/ppg_algorithm.py
```diff
@@ -0,0 +1,205 @@
...
+        after_train_iter() hook currently.
+
+        """
+        if not self._aux_algorithm:
+            return
+
+        if alf.summary.get_global_counter(
+        ) % self._aux_algorithm.interval == 0:
+            # NOTE(breakds): The auxiliary steps ``aux_steps`` is
+            # currently not used.
```
emailweixu: so you don't need to assign to aux_steps

alf/algorithms/ppg/ppg_aux_algorithm.py
```diff
@@ -0,0 +1,184 @@
...
+            dual_actor_value_network: the underlying network for PPG algorithm.
+                PPGAuxAlgorithm does not own the network. Instead, this should
+                be shared reference from the parent PPGAlgorithm.
+            aux_options: Options that controls the auxiliary phase training.
+            name (str): Name of this algorithm.
+
+        """
+        # Use a config with overriden fields as auxiliary phase updates has a
+        # bunch of its own options.
+        updated_config = copy.deepcopy(config)
```
emailweixu: why not copy.copy()?

alf/algorithms/ppg/ppg_aux_algorithm.py
```diff
@@ -0,0 +1,184 @@
...
+                 config: Optional[TrainerConfig] = None,
+                 optimizer: torch.optim.Optimizer = None,
+                 dual_actor_value_network=None,
+                 aux_options: PPGAuxOptions = PPGAuxOptions(),
+                 name: str = 'PPGAuxAlgorithm'):
+        """Construct a PPGAuxAlgorithm instances.
+
+        Args:
+
+            observation_spec (nested TensorSpec): representing the observations.
```
hnyu, line 85: If type hints are already used, the types in docstrings can be omitted.

alf/algorithms/ppg/ppg_aux_algorithm.py
```diff
@@ -0,0 +1,184 @@
...
+        """
+        # Note that we need to release the ``untransformed`` from the time step
+        # (inputs) before constructing the experience.
+        lite_time_step = inputs._replace(untransformed=())
+        exp = make_experience(lite_time_step, policy_step, state)
+        if not self._use_rollout_state:
+            exp = exp._replace(state=())
+        # Set the experience spec explicitly if it is not set, based on this
+        # (sample) experience
+        if not self._experience_spec:
```
hnyu, line 152: Can this code block just reuse _set_exp_replayer defined in algorithm.py?

alf/algorithms/ppg/ppg_aux_algorithm.py
```diff
@@ -0,0 +1,184 @@
...
+            self._experience_spec = dist_utils.extract_spec(exp, from_dim=1)
+        exp = dist_utils.distributions_to_params(exp)
+
+        if self._replay_buffer is None:
+            exp_spec = dist_utils.to_distribution_param_spec(
+                self._experience_spec)
+            num_envs = exp.env_id.shape[0]
+            max_length = self._config.unroll_length
```
hnyu, line 163: If PPGAuxOptions.interval > 1 and we only store unroll_length steps in the buffer, will some steps (within the interval) be discarded ?

alf/algorithms/ppg/ppg_aux_phase_loss.py
```diff
@@ -0,0 +1,135 @@
...
+        loss = td_loss_actual + td_loss_aux + self._policy_kl_loss_weight * policy_kl_loss
+
+        return LossInfo(
+            loss=loss,
+            extra=PPGAuxPhaseLossInfo(
+                td_loss_actual=td_loss_actual,
+                td_loss_aux=td_loss_aux,
+                policy_kl_loss=policy_kl_loss))
+
+    def _calc_returns_with_advantages(self, info, value):
```
hnyu: The name is misleading. Can just be "_calc_returns" because you are not returning the advantages. Maybe can directly derive from ActorCriticLoss and save some code of this function?

alf/algorithms/ppg_algorithm.py
```diff
@@ -0,0 +1,205 @@
...
+        self._network = dual_actor_value_network
+        self._loss = PPOLoss(debug_summaries=debug_summaries)
+
+        if epsilon_greedy is None:
+            epsilon_greedy = alf.get_config_value(
+                'TrainerConfig.epsilon_greedy')
+        self._predict_step_epsilon_greedy = epsilon_greedy
+
+    def _trainable_attributes_to_ignore(self):
+        return ['PPGAuxAlgorithm']
```
hnyu: Should this be the actual variable name "_aux_algorithm" instead of the class name?

alf/algorithms/ppo_loss.py
```diff
@@ -68,6 +68,11 @@ def __init__(self,
             normalize_advantages (bool): If True, normalize advantage to zero
                 mean and unit variance within batch for caculating policy
                 gradient.
+            compute_advantages_internally (bool): Normally PPOLoss does not
+                compute the adavantage and it expects the info to carry the
+                already-computed advantage. If this flag is set to True, PPOLoss
+                will instead compute the advantage internally without depending
+                on the input info.
```
hnyu: Should additionally explain in what scenarios we need to compute it internally because normally PPO's advantages stay fixed during the updates, so we tend to only compute once before training.

alf/examples/ppg_conf.py
```diff
@@ -0,0 +1,32 @@
...
+# limitations under the License.
+"""This file constains basic configurations for PPO. An entropy target is
+automatically enforced for PPO's policy.
+"""
+
+import alf
+from alf.algorithms.agent import Agent
+from alf.algorithms.ppg_algorithm import PPGAlgorithm
+from alf.algorithms.ppo_algorithm import PPOLoss
+
```
hnyu, line 22: Can you also change the ppo_conf.py so that by default the enforce_entropy_target=False?

hnyu (changes requested):

alf/algorithms/ppg/ppg_aux_algorithm.py
```diff
@@ -0,0 +1,184 @@
...
+                max_length=max_length,
+                prioritized_sampling=False,
+                num_earliest_frames_ignored=self._num_earliest_frames_ignored,
+                name='ppg_aux_replay_buffer')
+
+        self._replay_buffer.add_batch(exp, exp.env_id)
+
+    def train_step(self, inputs: TimeStep, state,
+                   plain_rollout_info: PPGRolloutInfo) -> AlgStep:
+        """Phase context dependent evaluation on experiences for training
```
emailweixu: no more phase context

alf/algorithms/ppg/ppg_aux_algorithm.py
```diff
@@ -0,0 +1,184 @@
...
+        train_info = PPGTrainInfo(
+            action=plain_rollout_info.action,
+            rollout_value=plain_rollout_info.value,
+            rollout_action_distribution=plain_rollout_info.
+            action_distribution).absorbed(alg_step.info)
+
+        return alg_step._replace(info=train_info)
+
+    def calc_loss(self, info: PPGTrainInfo) -> LossInfo:
+        """Phase context dependent loss function evaluation
```
emailweixu: no more phase context

alf/algorithms/ppg/ppg_utils.py
```diff
@@ -0,0 +1,140 @@
...
+
+        This function generate a new PPGTrainInfo instead of updating
+        ``self`` in place.
+
+        In ``train_step()`, we would like to keep the derived information in
+        PPGTrainInfo while updating most of the shared fields (with
+        PPGRolloutInfo) from evaluation of the updated network. This function
+        makes it easy to do that.
+
+        It is also used in ``preprocess_experience()`` where we need to combine
```
emailweixu: there is no preprocess_experience

alf/algorithms/ppg/ppg_utils.py
```diff
@@ -0,0 +1,140 @@
...
+        namedtuple(
+            'PPGTrainInfo',
+            PPGRolloutInfo._fields + ('rollout_action_distribution',
+                                      'rollout_value'),
+            default_value=())):
+    """Data structure that stores extra derived information for training
+    in addition to the original rollout information.
+
+    Such extra information is derived during training updates, in
+    ``preprocess_experience()`` and used across calls to
```
emailweixu: there is no preprocess_experience

alf/algorithms/ppg/ppg_aux_algorithm.py
breakds: Good point. I removed the type hints here because it should be nested TensorSpec. However, we do not have a type hint for nested `TensorSpec` yet.

alf/algorithms/ppg/ppg_aux_algorithm.py
breakds: Yes, but it does not do much to avoid code duplication - we still need to compute the replay buffer parameters explicitly. Besides, `_set_replay_buffer` does extra thing to append to `self._observers`, which does no harm but is not necessary.

alf/algorithms/ppg/ppg_aux_algorithm.py
breakds: Here the `unroll_length` has an updated value because Aux phase use an updated `TrainerConfig`. Sorry that this is not as clear - I added a comment here to make it clearer.

alf/algorithms/ppg/ppg_aux_phase_loss.py
breakds: Updated the name. Probably not important, but one of the reasons to duplicate part of the code is to save the part where `returns` is calculated but not used for `gae` case.

alf/algorithms/ppg_algorithm.py
breakds: Sure. Updated the name to be `_aux_algorithm`.

alf/examples/ppg_conf.py
breakds: Will do in a separate PR. Sorry it seems that I forgot about this previously.

alf/algorithms/ppg_algorithm.py
breakds: Removed.

alf/algorithms/ppg_algorithm.py
breakds: Removed.

alf/algorithms/ppg_algorithm.py
breakds: Removed. Originally I was thinking the note and the assignment can be left here to remind the reader that `train_from_replay_buffer` does return the number of steps.

alf/algorithms/ppg/ppg_aux_algorithm.py
breakds: Removed.

alf/algorithms/ppg/ppg_aux_algorithm.py
breakds: Removed.

alf/algorithms/ppg_algorithm.py
hnyu: > Sure. Updated the name to be `_aux_algorithm`.

Does this mean that you were not actually excluding the parameters from its parent?

alf/algorithms/ppg/ppg_utils.py
breakds: Good catch. Removed.

alf/algorithms/ppg_algorithm.py
breakds: Did you mean `_trainable_attributes_to_ignore`? If so I have updated both places.

alf/algorithms/ppo_loss.py
breakds: Good point. The direct reason is that we want to avoid having `preprocess_experience()`, but I am not sure this still holds after splitting the algorithms. Let me think and confirm.

alf/algorithms/ppg/ppg_aux_algorithm.py
breakds: I was being over-cautious :) Apparently if it is only updating POD fields we don't have to do `deepcopy`. Updated.

alf/algorithms/ppg_algorithm.py
hnyu: > Did you mean `_trainable_attributes_to_ignore`? If so I have updated both places.

Sorry I didn't explain it clearly. The reason why we have this function is to make sure that the parameters of a child won't be modified by its parent. So if originally the attribute names were wrong, does it mean that it's possible when you were training the algorithm, the child's parameters were modified by its parent in an unexpected way?

breakds: Ah I think I misunderstood your comments previously. Let me explain:

1. My understanding is that here we use the `name` field of the `PPGAuxAlgorithm` instance to exclude it, instead of the variable name.
2. Originally the `name` field is forced to set to `PPGAuxAlgorithm`, which is also the class name. And therefore originally I was excluding it using `PPGAuxAlgorithm`.
3. I mistook your comments as "it is better to name it `_aux_algorithm` to avoid being identical to the class name". Therefore, I have updated both places to be `_aux_algorithm`, which I think it should still work.

hnyu: From what I know, for this function we need to provide the variable name. For example, if you have a class member called `self._vars` then the list should be ["_vars"]. It has nothing to do with the name you assign to that class instance. See the example of SAC's "_target_critic_networks". In other words, it's possible that you weren't actually excluded its parameters from the parent, because the wrong name was provided.

breakds: Ah Yes you are right. I should have carefully read the documentation. Thanks for pointing this out.

And yes, previously it was not correctly ignored.

alf/algorithms/ppg/ppg_aux_algorithm.py
```diff
@@ -0,0 +1,183 @@
...
+        # Use a config with overriden fields as auxiliary phase updates has a
+        # bunch of its own options.
+        updated_config = copy.copy(config)
+        updated_config.unroll_length = config.unroll_length * aux_options.interval
+        updated_config.whole_replay_buffer_training = True
+        updated_config.clear_replay_buffer = True
+        updated_config.mini_batch_length = (aux_options.mini_batch_length
+                                            or config.unroll_length)
+        updated_config.mini_batch_size = aux_options.mini_batch_size
+        updated_config.num_updates_per_train_iter = aux_options.num_updates_per_train_iter
```
emailweixu, line 109: should clear data_transformer too. data_transformer is used to process the untransformed data. The data stored in the replay buffer of PPGAuxAlgorithm is already processed by Agent.

alf/algorithms/ppg_algorithm.py
emailweixu: Since your result is based on using "PPGAuxAlgorithm", which does not have any effect, how does it affect the result?

alf/algorithms/ppg/ppg_aux_algorithm.py
breakds: Right. Fixed by setting it to `None` and adding documentation explaining why.

breakds:
> Since your result is based on using "PPGAuxAlgorithm", which does not have any effect, how does it affect the result?

Not sure yet. Running the experiment to confirm this.

alf/algorithms/ppo_loss.py
breakds: Added documentation that explains the reason in the docstring.

breakds:
> > Since your result is based on using "PPGAuxAlgorithm", which does not have any effect, how does it affect the result?
> 
> Not sure yet. Running the experiment to confirm this.

The experiment shows the new refactor results in the same training dynamics and performance (i.e. previously the bug in `_trainnable_attributes_to_ignore` does not make a difference in the current experiment setting for procgen bossfight).

updated_ppg

emailweixu (dismissed):

alf/algorithms/ppo_loss.py
```diff
@@ -129,4 +138,12 @@ def _pg_loss(self, info, advantages):
...
     def _calc_returns_and_advantages(self, info, value):
-        return info.returns, info.advantages
+        if not self._compute_advantages_internally:
+            return info.returns, info.advantages
+        # If rollout_value is present in ``info``, we use it to compute the
+        # advantage. This is mainly for algorithms like PPG where at this time
+        # info.value is the newly computed value which is different from the
+        # rollout time value prediction. The rollout time value prediction is
+        # preserved in info.rollout_value.
+        value_to_use = getattr(info, 'rollout_value', info.value)
```
hnyu: The default value "info.value" seems problematic if 'rollout_value' doesn't exist? I suggest just asserting to False

breakds: Sure. Updated this with an assert to make it more explicit to the user/reader.

hnyu (approved):

PR #1031, opened 2021-10-15
Fix a bug of Algorithm.optimizers()

emailweixu:
Previously, the result is not correct when include_ignored_attributes=True.
When include_ignored_attributes=True, it uses Module.children() to get the children.
However, some sub-algorithms may be held by ModuleList or ModuleDict (e.g. SequentialAlg._nets).
But optimizers() will only recurse into Algorithm, so those algorithms in ModuleList
will be skipped.

This bug does not affect training. It only affects the reporting of unoptimized_parameters.

Solution: Modified Algorithm._get_children() to handle include_ignored_attributes=True and use it in optimizers()

### DIFF
--- a/alf/algorithms/algorithm.py
+++ b/alf/algorithms/algorithm.py
@@ -412,15 +412,19 @@ class Algorithm(AlgorithmInterface):
         """
         return []
 
-    def _get_children(self):
+    def _get_children(self, include_ignored_attributes=False):
         children = []
+        if include_ignored_attributes:
+            to_ignore = []
+        else:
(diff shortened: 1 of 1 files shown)


breakds (approved):

Haichao-Zhang:
> This bug does affect training. It only affects the reporting of unoptimized_parameters.

Can modify this comment: this does not affect training?

Haichao-Zhang (approved):

emailweixu:
> > This bug does affect training. It only affects the reporting of unoptimized_parameters.
> 
> Can modify this comment: this does not affect training?

Changed.

PR #1033, opened 2021-10-15
Fix a path issue in taac_fetch_conf

hnyu:
alf.examples no longer has "fetch_conf.py". It has been moved to "examples/benchmarks/fetch/"

### DIFF
--- a/alf/examples/taac_fetch_conf.py
+++ b/alf/examples/taac_fetch_conf.py
@@ -12,9 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import sys
-sys.path.append("./benchmarks/fetch")
-
 from functools import partial
 
 import alf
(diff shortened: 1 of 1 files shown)


alf/examples/taac_fetch_conf.py
```diff
@@ -22,7 +22,7 @@
 from alf.utils import dist_utils
 
 from alf.examples import sac_conf
-from alf.examples import fetch_conf
+import fetch_conf
```
emailweixu: Does this mean that this needs to run under alf/examples/benchmarks/fetch?
Should we do `from alf.examples.benchmarks import fetch_conf` instead?

hnyu: > Does this mean that this needs to run under alf/examples/benchmarks/fetch? Should we do `from alf.examples.benchmarks import fetch_conf` instead?

"benchmarks/fetch" has already been added to system path in the conf. But I've changed to not do in that way.

emailweixu (dismissed):

emailweixu (approved):

PR #1036, opened 2021-10-16
Move image scaling transformation to Impala CNN

breakds:
# Motivation

Previously, the `ImpalaCNN` assumes the input images are already converted to float in range [0.0, 1.0]`. However, there are cases when the input pixel are actually `uint8` (one of the motivations to store experiences in `uint8` pixels is to save memory, as `uint8` takes only a quarter of the memory consumption than `float32`).

In this case, we should make `ImpalaCNN` more flexible to be able to apply the image scaling `unit8` -> `float32[0.0, 1.0]` on the fly.

# Solution

Add an option to the factory function of `ImpalaCNN` to allow for applying image scale transformation on the fly. Besides that:

1. Also updated `procgen` configuration to not apply `ImageScaleTransformer` by default.
2. In `ppo_procgen`, now it needs to explicitly state that it requires `ImageScaleTransformer`.

This prepares for the upcoming `ppg_procgen`, where it is desirable to save more memory and do image scale transformation on the fly.

# Testing

1. Made sure that `ppo_procgen` still run training correctly.
2. Although not submitted yet, trained a PPG procgen (bossfight) with the graph below, which is based on this PR.

ppg_bossfight

Light blue = PPG that depends on this PR
dark blue = PPO

### DIFF
--- a/alf/examples/networks/impala_cnn_encoder.py
+++ b/alf/examples/networks/impala_cnn_encoder.py
@@ -29,7 +29,7 @@ def _create_residual_cnn_block(
     """Create a CNN block with 2 Conv2D plus a residual connection
 
     It essentially performs
-    
+
     .. math::
 
         f(x) = x + \textrm{Conv2D}(\textrm{Conv2D}(\textrm{ReLU}(x)))
@@ -39,13 +39,13 @@ def _create_residual_cnn_block(
     channels (i.e. the dimension of :math:`x` remains the same).
 
     Args:
-        
+
         input_tensor_spec (nested TesnorSpec): The spec of the input tensor,
             denotated as ``x`` in the above formula. The shape of the input
             tensor omitting batch size should be (C, W, H).
         kernel_initializer: initializer for the Conv2D and FC layers in the
             module.
-    
+
     Returns:
 
         A module that perofrms the described operations.
@@ -79,7 +79,7 @@ def _create_downsampling_cnn_stack(
         num_residual_blocks: int = 2,
         kernel_initializer: Callable[[Tensor], None] = xavier_uniform_):
     """Create a stack of Convolutional layers that also does downsampling
-    
+
     It essentially performs one ``Conv2D`` that changes the number of channels,
     a max pooling filter that downsamples the image by half, followed by a
     series of residual CNN blocks. The number of CNN blocks in the series is
@@ -100,7 +100,7 @@ def _create_downsampling_cnn_stack(
             module.
 
     Returns:
-        
+
         A module that perofrms the described operations.
 
     """
@@ -133,7 +133,8 @@ def create(input_tensor_spec,
            cnn_channel_list: List[int] = (16, 32, 32),
            num_blocks_per_stack: int = 2,
            output_size: int = 256,
-           kernel_initializer: Callable[[Tensor], None] = xavier_uniform_):
+           kernel_initializer: Callable[[Tensor], None] = xavier_uniform_,
+           apply_image_scale_transform: bool = False):
     """Create the Impala CNN Encoder
 
     Here the so called Impala CNN Encoder is essentially a series of CNN
@@ -162,14 +163,28 @@ def create(input_tensor_spec,
             image in the mini batch.
         kernel_initializer: initializer for the Conv2D and FC layers in the
             network.
-    
+        apply_image_scale_transform: when set to ``True``, the input image will
+            first go through a scale transformation so that the pixels are
+            converted from 0 - 255 integer to 0.0 - 1.0 float.
+
     Returns:
-        
+
         A module representing the Impala CNN encoding network.
 
     """
     stacks = []
-    last_output_spec = input_tensor_spec
+    if apply_image_scale_transform:
+        scale_factor = 1.0 / 255.0
+        stacks.append(alf.layers.Cast(dtype=torch.float32))
+        stacks.append(lambda x: x * scale_factor)
+        last_output_spec = alf.BoundedTensorSpec(
+            input_tensor_spec.shape,
+            dtype=torch.float32,
+            minimum=0.0,
+            maximum=1.0)
+    else:
+        last_output_spec = input_tensor_spec
+
     for output_channels in cnn_channel_list:
         stacks.append(
             _create_downsampling_cnn_stack(
@@ -179,6 +194,7 @@ def create(input_tensor_spec,
                 kernel_initializer=kernel_initializer))
         last_output_spec = stacks[-1].output_spec
     stack_output_spec = stacks[-1].output_spec
+
     return alf.nn.Sequential(
         *stacks,
         # Flatten the images
@@ -188,4 +204,5 @@ def create(input_tensor_spec,
             input_size=stack_output_spec.numel,
             output_size=output_size,
             activation=torch.relu_,
-            kernel_initializer=kernel_initializer))
+            kernel_initializer=kernel_initializer),
+        input_tensor_spec=input_tensor_spec)
--- a/alf/examples/networks/impala_cnn_encoder.py
+++ b/alf/examples/networks/impala_cnn_encoder.py
@@ -133,8 +133,7 @@ def create(input_tensor_spec,
            cnn_channel_list: List[int] = (16, 32, 32),
            num_blocks_per_stack: int = 2,
            output_size: int = 256,
-           kernel_initializer: Callable[[Tensor], None] = xavier_uniform_,
-           apply_image_scale_transform: bool = False):
+           kernel_initializer: Callable[[Tensor], None] = xavier_uniform_):
     """Create the Impala CNN Encoder
 
     Here the so called Impala CNN Encoder is essentially a series of CNN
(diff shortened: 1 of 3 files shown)


hnyu:
I believe it's better to just use ImageScaleTransformer in the conf file? Note that even if you use the data transformer, the replay buffer still stores the untransformed data type (in your case uint8).

alf/examples/ppo_procgen/base_conf.py
```diff
@@ -30,10 +30,19 @@
...
 from alf.examples import procgen_conf
 from alf.examples.networks import impala_cnn_encoder
+from alf.algorithms.data_transformer import ImageScaleTransformer
 
 # Environment Configuration
 alf.config('create_environment', num_parallel_environments=96)
 
+# Explicitly apply data transformer here so that the network does not
+# have to do this. This incurs more GPU memory consumption as we are
+# effectively storing float32 images instead of int8 images (4x GPU
```
hnyu: I don't understand why after applying the data transformer we are storing float32?  As in 
https://github.com/HorizonRobotics/alf/blob/pytorch/alf/algorithms/rl_algorithm.py#L563
the exp is made from untransformed timestep.

breakds:
> I believe it's better to just use ImageScaleTransformer in the conf file? Note that even if you use the data transformer, the replay buffer still stores the untransformed data type (in your case uint8).

You are right. Since now everything goes into the replay buffer, everything is stored in its untransformed version. We do not need this PR. Closing it.

breakds:
After running into an OOM I realized that this PR is still needed to reduce the memory footprint. This is because:

1. For auxiliary phase training, the replay buffer is huge
2. The image scale transformer changes the dtype from `uint8` to `float32`. The memory footprint becomes x4.
3. Therefore, we do not want to apply the transformer on the whole replay buffer (via `gather_all`) in one shot. A better way is to do this on each mini batch.

The best way to achieve No. 3 above is to do it on the `forward` of each mini batch, hence putting it in the network would do.

PTAL.

hnyu:
Also only adding the option to impalacnn is not a general solution. Potentially we will have this issue for many other CNN networks.

alf/examples/ppo_procgen/base_conf.py
emailweixu: comment is not correct.

alf/examples/networks/impala_cnn_encoder.py
```diff
@@ -162,14 +163,28 @@ def create(input_tensor_spec,
...
+
     Returns:
-        
+
         A module representing the Impala CNN encoding network.
 
     """
     stacks = []
-    last_output_spec = input_tensor_spec
+    if apply_image_scale_transform:
```
emailweixu: Perhaps derive apply_image_scale_transform from input_tensor_spec? (i.e. if spec.dtype is uint8

breakds:
> Also only adding the option to impalacnn is not a general solution. Potentially we will have this issue for many other CNN networks.

Agree that this is not a general solution - we can optimize the `train_from_replay_buffer` by at least

1. Sample from replay buffer one mini batch at each time
2. Apply data transformation one mini batch at each time

The above is going to be bigger surgery but we should address that to save resources.

> For this PR you still keep imagescaletransformer in the conf. `apply_image_scale_transform: bool = False` is always False. This is inconsistent with what you want to achieve explained in the comment.

The conf is for PPO which does not have an auxiliary phase with huge replay buffer. Therefore it can keep using the trivial solution of `ImageScaleTransformer`. The added option is actually used for PPG configuration in the other PR.

alf/examples/networks/impala_cnn_encoder.py
breakds: Good idea. Let me update it.

hnyu:
> > Also only adding the option to impalacnn is not a general solution. Potentially we will have this issue for many other CNN networks.
> 
> Agree that this is not a general solution - we can optimize the `train_from_replay_buffer` by at least
> 
> 1. Sample from replay buffer one mini batch at each time
> 2. Apply data transformation one mini batch at each time
> 
> The above is going to be bigger surgery but we should address that to save resources.
> 
> > For this PR you still keep imagescaletransformer in the conf. `apply_image_scale_transform: bool = False` is always False. This is inconsistent with what you want to achieve explained in the comment.
> 
> The conf is for PPO which does not have an auxiliary phase with huge replay buffer. Therefore it can keep using the trivial solution of `ImageScaleTransformer`. The added option is actually used for PPG configuration in the other PR.

Just curious, for the auxiliary phase, does it matter that every time you sample the entire replay buffer? Potentially if the replay buffer is really huge, you can train lots of iterations in each auxiliary phase, each iteration only sampling a small batch, which avoids the memory issue.

breakds:
> > > Also only adding the option to impalacnn is not a general solution. Potentially we will have this issue for many other CNN networks.
> > 
> > 
> > Agree that this is not a general solution - we can optimize the `train_from_replay_buffer` by at least
> > 
> > 1. Sample from replay buffer one mini batch at each time
> > 2. Apply data transformation one mini batch at each time
> > 
> > The above is going to be bigger surgery but we should address that to save resources.
> > > For this PR you still keep imagescaletransformer in the conf. `apply_image_scale_transform: bool = False` is always False. This is inconsistent with what you want to achieve explained in the comment.
> > 
> > 
> > The conf is for PPO which does not have an auxiliary phase with huge replay buffer. Therefore it can keep using the trivial solution of `ImageScaleTransformer`. The added option is actually used for PPG configuration in the other PR.
> 
> Just curious, for the auxiliary phase, does it matter that every time you sample the entire replay buffer? Potentially if the replay buffer is really huge, you can train lots of iterations in each auxiliary phase, each iteration only sampling a small batch, which avoids the memory issue.

Totally agree with you that we can just sample a small batch every time. This comes to the point that we should

> 1. Sample from replay buffer one mini batch at each time

Because currently `train_from_replay_buffer` will sample all the `num_updates_per_iteration` at once so that we still have the memory problem.

hnyu:
> > > > Also only adding the option to impalacnn is not a general solution. Potentially we will have this issue for many other CNN networks.
> > > 
> > > 
> > > Agree that this is not a general solution - we can optimize the `train_from_replay_buffer` by at least
> > > 
> > > 1. Sample from replay buffer one mini batch at each time
> > > 2. Apply data transformation one mini batch at each time
> > > 
> > > The above is going to be bigger surgery but we should address that to save resources.
> > > > For this PR you still keep imagescaletransformer in the conf. `apply_image_scale_transform: bool = False` is always False. This is inconsistent with what you want to achieve explained in the comment.
> > > 
> > > 
> > > The conf is for PPO which does not have an auxiliary phase with huge replay buffer. Therefore it can keep using the trivial solution of `ImageScaleTransformer`. The added option is actually used for PPG configuration in the other PR.
> > 
> > 
> > Just curious, for the auxiliary phase, does it matter that every time you sample the entire replay buffer? Potentially if the replay buffer is really huge, you can train lots of iterations in each auxiliary phase, each iteration only sampling a small batch, which avoids the memory issue.
> 
> Totally agree with you that we can just sample a small batch every time. This comes to the point that we should
> 
> > 1. Sample from replay buffer one mini batch at each time
> 
> Because currently `train_from_replay_buffer` will sample all the `num_updates_per_iteration` at once so that we still have the memory problem.

What I really mean is that you still have num_updates_per_iteration=1. Instead, you just call the train_from_replay_buffer() multiple times in each auxiliary phase. Each call of train_from_replay_buffer only samples a small batch and converts that small batch at one time.

breakds:
This PR is updated. PTAL.

breakds:
> > > > > Also only adding the option to impalacnn is not a general solution. Potentially we will have this issue for many other CNN networks.
> > > > 
> > > > 
> > > > Agree that this is not a general solution - we can optimize the `train_from_replay_buffer` by at least
> > > > 
> > > > 1. Sample from replay buffer one mini batch at each time
> > > > 2. Apply data transformation one mini batch at each time
> > > > 
> > > > The above is going to be bigger surgery but we should address that to save resources.
> > > > > For this PR you still keep imagescaletransformer in the conf. `apply_image_scale_transform: bool = False` is always False. This is inconsistent with what you want to achieve explained in the comment.
> > > > 
> > > > 
> > > > The conf is for PPO which does not have an auxiliary phase with huge replay buffer. Therefore it can keep using the trivial solution of `ImageScaleTransformer`. The added option is actually used for PPG configuration in the other PR.
> > > 
> > > 
> > > Just curious, for the auxiliary phase, does it matter that every time you sample the entire replay buffer? Potentially if the replay buffer is really huge, you can train lots of iterations in each auxiliary phase, each iteration only sampling a small batch, which avoids the memory issue.
> > 
> > 
> > Totally agree with you that we can just sample a small batch every time. This comes to the point that we should
> > > 1. Sample from replay buffer one mini batch at each time
> > 
> > 
> > Because currently `train_from_replay_buffer` will sample all the `num_updates_per_iteration` at once so that we still have the memory problem.
> 
> What I really mean is that you still have num_updates_per_iteration=1. Instead, you just call the train_from_replay_buffer() multiple times in each auxiliary phase. Each call of train_from_replay_buffer only samples a small batch and converts that small batch at one time.

Good point, this does sound like effectively equivalent to what is implemented now. I give this a try later.

alf/examples/ppo_procgen/base_conf.py
breakds: I reverted the change in the configuration. The original comment is misleading.

emailweixu (approved):

PR #1041, opened 2021-10-27
[Bug] Treat parameters that are assigned to child algorithm's default optimizer as child_handled

breakds:
# Motivation

This bug fix explicitly seek to handle the following scenario correctly:

> An algorithm `A` has a sub algorithm `B`. A parameter `c` is handled by `B`'s default optimizer, and `c` also appear in another `Module` under `A`.
>
> We would like `A` to refrain from handling `c` with its default optimizer. This requires `B`'s `_setup_optimizer_()` call return `c` as "handled by child".

There is a bug in the current implementation that `new_params` assigned to a default optimizer are not returned by `_setup_optimizer_()`. This will break the intended behavior above.

# Solution

1. Fix the bug by returning parameters that are assigned to the default optimizer.
2. Slightly change the local variable name and documentation to improve readability.

# Testing

Tested with the PPG implementation (with `_trainable_attributes_to_ignore` disabled) and observed the desired behavior.

# Review guide

Please help check that the newly added documentation is clear and not misleading.

### DIFF
--- a/alf/algorithms/algorithm.py
+++ b/alf/algorithms/algorithm.py
@@ -491,11 +491,11 @@ class Algorithm(AlgorithmInterface):
 
         def _add_params_to_optimizer(params, opt):
             existing_params = set(_get_optimizer_params(opt))
-            added_param_names = list(
+            added_param_list = list(
                 filter(lambda p: p not in existing_params, params))
-            if added_param_names:
-                opt.add_param_group({'params': added_param_names})
-            return added_param_names
+            if added_param_list:
+                opt.add_param_group({'params': added_param_list})
+            return added_param_list
 
         # Iterate over all the child modules and add their parameters
         # into either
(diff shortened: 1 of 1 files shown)


alf/algorithms/algorithm.py
```diff
@@ -491,11 +491,24 @@ def _setup_optimizers_(self):
 
         def _add_params_to_optimizer(params, opt):
             existing_params = set(_get_optimizer_params(opt))
-            params = list(filter(lambda p: p not in existing_params, params))
-            if params:
-                opt.add_param_group({'params': params})
-            return params
-
+            added_param_names = list(
```
emailweixu: They are not names. They are parameters.

alf/algorithms/algorithm.py
```diff
@@ -491,11 +491,24 @@ def _setup_optimizers_(self):
...
+        # into either
+        #
+        # 1. ``handled``: when the module is already assigned to an
+        #    optimizer to handle in this algorithm.
+        #
+        # 2. or ``new_params``: when the module does not have an
+        #    assigned optimizer in this algorithm (yet).
+        #
+        # For case 2, after the loop the ``new_params`` will be assigned to the
+        # default optimizer is there is one. Otherwise, they will be left as
```
emailweixu: "is there is one" => if there is one

alf/algorithms/algorithm.py
breakds: Ah yep. Updated it as `added_param_list`.

alf/algorithms/algorithm.py
breakds: Fixed.

emailweixu (approved):

PR #1049, opened 2021-10-29
Sampled MuZero

emailweixu:
Implements Sampled Muzero described in the following paper:

Hubert et. al. "Learning and Planning in Complex Action Spaces" arXiv:2104.06303

There are following main changes:

1. Changing the prior used for MCTS from original probability to 1/num_sampled_actions
2. Expand all the children of the root node immediately. This is described in Appendix A
of the paper: "At the root of the search tree only, we evaluated all sampled actions before
the start of the search and used those to initialise the Q(s; a) quantities in the PUCT formula".
It turns out that this is very important for the performance.
3. Inspired by 2, I also implemented an option to evaluate all the children of the leaf of a
selected searc path.
4. MCTSAlgorithmTest is modified to test different variants.

Test plan:

1. Change mcts_algorithm_test.py to test all variants
2. muzero_pendulum.gin


alf/algorithms/mcts_models.py
```diff
@@ -426,16 +447,23 @@ def prediction_model(self, state):
             actions = action_distribution.rsample(
                 (self._num_sampled_actions, ))
             # [B, num_sampled_actions]
-            log_probs = action_distribution.log_prob(actions).transpose(0, 1)
-            action_probs = F.softmax(log_probs, dim=1)
+            # log_probs = action_distribution.log_prob(actions).transpose(0, 1)
+            # action_probs = F.softmax(log_probs, dim=1)
             actions = actions.transpose(0, 1)
+            # Accordint to the following paper, we should use 1/K as action_probs
```
Haichao-Zhang: Accordint -> According

alf/algorithms/mcts_models.py
```diff
@@ -426,16 +447,23 @@ def prediction_model(self, state):
...
+            # for sampled actions.
+            # Huber et. al. Learning and Planning in Complex Action Spaces, 2021
```
Haichao-Zhang: Huber -> Hubert

alf/algorithms/mcts_algorithm.py
```diff
@@ -232,6 +243,17 @@ def __init__(
...
                 the reward(s).
+            expand_all_children (bool): If True, when a new leaf is selected, immediately
+                expand all its children.
+            expand_all_root_children (bool): whether to expand all root children
+                before search. This is described in Appendix A of "Learning and
+                Planning in Complex Action Spaces". However, our implementation is
+                different from the paper's. The paper initializes Q(s, a) for
+                root s for all the action being sampled. We expand all sampled
+                action for root s. Our implementation depends on setting
+                (act/learn)_with_exploration_policy=True as the visit counts
```
Haichao-Zhang: Does this mean we need to set `(act/learn)_with_exploration_policy=True` when `expand_all_root_children` is set as `True`?  If this is the case, it might be useful to added a check on this condition to avoid mis-use.

alf/algorithms/mcts_algorithm.py
```diff
@@ -626,7 +800,11 @@ def _set_tree_state(ts, s):
         trees.prior[:, n] = prior
 
 
-def calculate_exploration_policy(value, prior, c, tol=1e-6):
+def calculate_exploration_policy(value,
+                                 prior,
+                                 c,
+                                 tol=1e-6,
+                                 handle_zero_prior=False):
```
Haichao-Zhang: handle_zero_prior is not used in the function

emailweixu: removed

alf/algorithms/mcts_models.py
emailweixu: fixed

Haichao-Zhang (dismissed):

alf/algorithms/mcts_models.py
```diff
@@ -426,16 +447,23 @@ def prediction_model(self, state):
             actions = action_distribution.rsample(
                 (self._num_sampled_actions, ))
             # [B, num_sampled_actions]
-            log_probs = action_distribution.log_prob(actions).transpose(0, 1)
-            action_probs = F.softmax(log_probs, dim=1)
+            # log_probs = action_distribution.log_prob(actions).transpose(0, 1)
+            # action_probs = F.softmax(log_probs, dim=1)
```
breakds: What is the reason of keeping this commented instead of having this replaced? Might be good to add some comments if we want to show "This is how we used to do it".

alf/algorithms/mcts_models.py
```diff
@@ -426,16 +447,23 @@ def prediction_model(self, state):
...
+            action_probs = torch.ones(
+                actions.shape[:2]) / self._num_sampled_actions
         else:
             action_probs = action_distribution.probs
             if self._num_sampled_actions is None:
                 actions = ()
             else:
                 action_probs, actions = action_probs.topk(
                     self._num_sampled_actions, sorted=False)
+                action_probs = action_probs / action_probs.sum(
```
breakds, line 465: Is this also a result of sampled-muzero's action probability rebalancing? In the Top-K approach, I have a hard time figuring out what is a "effectively equivalent" sampling distribution from the original action space.

alf/algorithms/mcts_models.py
```diff
@@ -37,6 +37,7 @@
         'actions',
 
         # [B, K], probabilities of the candidate actions. prob of 0 indicates invalid action
+        # action_probs[i, :] should sum to 1
```
breakds:
```suggestion
In the case when the candidate actions are sampled from the original action space, 
the ``action_probs`` will be normalized over the sampled candidate action set.
 i.e. ``action_probs[i, :]`` should sum to 1.
```

alf/algorithms/mcts_algorithm.py
```diff
@@ -317,8 +351,15 @@ def predict_step(self, time_step: TimeStep, state: MCTSState):
...
-                          self._known_value_bounds)
+        branch_factor = model_output.action_probs.shape[1]
+        if self._expand_all_children:
+            tree_size = (self._num_simulations + 1) * branch_factor + 1
+        elif self._expand_all_root_children:
+            tree_size = self._num_simulations + 1 + branch_factor
+        else:
+            tree_size = self._num_simulations + 1
+
+        trees = _MCTSTree(tree_size, model_output, self._known_value_bounds)
```
breakds: Not related to this PR, but suggest renaming to `_MCTSTrees` or `_MCTSTreeBatch` will be slightly more friendly to the new reader.

alf/algorithms/mcts_algorithm.py
```diff
@@ -232,6 +244,18 @@ def __init__(
                 shape as ``steps``.
             reward_spec (TensorSpec): a rank-1 or rank-0 tensor spec representing
                 the reward(s).
+            expand_all_children (bool): If True, when a new leaf is selected, immediately
+                expand all its children. With this option, the visit
+                count does not truly refect the quality of a node. Hence it
```
breakds:
```suggestion
                count does not truly reflect the quality of a node. Hence it
```

alf/algorithms/mcts_algorithm.py
```diff
@@ -258,6 +283,15 @@ def __init__(
         self._root_dirichlet_alpha = root_dirichlet_alpha
         self._root_exploration_fraction = root_exploration_fraction
         self._visit_softmax_temperature_fn = visit_softmax_temperature_fn
+        self._expand_all_children = expand_all_children
+        self._expand_all_root_children = expand_all_root_children
+
+        if expand_all_children or expand_all_root_children and not (
```
breakds:
```suggestion
        if (expand_all_children or expand_all_root_children) and not (
```

alf/algorithms/mcts_algorithm.py
```diff
@@ -232,6 +244,18 @@ def __init__(
                 shape as ``steps``.
             reward_spec (TensorSpec): a rank-1 or rank-0 tensor spec representing
                 the reward(s).
+            expand_all_children (bool): If True, when a new leaf is selected, immediately
+                expand all its children. With this option, the visit
+                count does not truly refect the quality of a node. Hence it
+                should be used with (act/learn)_with_exploration_policy=True
+            expand_all_root_children (bool): whether to expand all root children
```
breakds, line 261: Is `expand_all_root_children = True` covered by `expand_all_children = True`?

alf/algorithms/mcts_algorithm.py
```diff
@@ -232,6 +244,18 @@ def __init__(
                 shape as ``steps``.
             reward_spec (TensorSpec): a rank-1 or rank-0 tensor spec representing
                 the reward(s).
+            expand_all_children (bool): If True, when a new leaf is selected, immediately
+                expand all its children. With this option, the visit
```
breakds, line 258: Here when we say "expand" does it also implicitly include "back-up"?

alf/algorithms/mcts_algorithm.py
```diff
@@ -232,6 +244,18 @@ def __init__(
...
+                count does not truly refect the quality of a node. Hence it
+                should be used with (act/learn)_with_exploration_policy=True
+            expand_all_root_children (bool): whether to expand all root children
+                before search. This is described in Appendix A of "Learning and
+                Planning in Complex Action Spaces". However, our implementation is
+                different from the paper's. The paper initializes Q(s, a) for
+                root s for all the action being sampled. We expand all sampled
+                action for root s. With this option, the visit
+                count does not truly refect the quality of a node. Hence it
+                should be used with (act/learn)_with_exploration_policy=True.
```
breakds, line 268: It is not clear to me why `act_with_exploration_policy` alleviate the problem of "visit count being off". Should this be lifted to the "There are several differences:" above and elaborate a bit?

breakds:
Still working on the review of `mcts_algorithm.py` second half.

alf/algorithms/mcts_algorithm.py
```diff
@@ -580,15 +703,74 @@ def _backup(self, trees: _MCTSTree, search_paths, path_lengths, values):
...
         valid = depth < path_lengths
         nodes = (B.expand(T, -1)[valid], search_paths[valid])
-        trees.visit_count[nodes] += 1
+        trees.visit_count[nodes] += count
         trees.value_sum[nodes] = value_sum[valid]
         trees.update_value_stats((B, search_paths), valid)
         self._update_best_child(trees, nodes)
 
+    def _expand_children(self, trees: _MCTSTree, nodes, n, to_plays):
+        """Expand all children of nodes.
```
breakds, line 727: Also state that the sum of the value due to the expansion will be returned (for computing the backing up)?

alf/algorithms/mcts_algorithm.py
```diff
@@ -338,9 +379,35 @@ def predict_step(self, time_step: TimeStep, state: MCTSState):
...
+        """Build the tree by evaluating one node a time"""
+        roots = (trees.B, trees.root_indices)
+        sim0 = 1
+        if self._expand_all_root_children:
+            v = self._expand_children(trees, roots, 1, to_plays)
+            branch_factor = trees.branch_factor
+            trees.visit_count[roots] += branch_factor
+            trees.value_sum[roots] += branch_factor * v
+            children = trees.get_children(roots)
+            nodes = torch.broadcast_tensors(*roots)
```
breakds: I do not understand immediately why we need to do `broadcast_tensors`. If `roots` is something like 

```python
(tensor([0, 1, 2, 3]), tensor([0, 0, 0, 0]))
```

Assuming 4 batches. How is `nodes` different from `roots` after applying the `broadcast_tensor`?

alf/algorithms/mcts_algorithm.py
```diff
@@ -338,9 +379,35 @@ def predict_step(self, time_step: TimeStep, state: MCTSState):
...
+            output=action, state=MCTSState(steps=state.steps + 1), info=info)
+
+    def _build_tree1(self, trees, to_plays):
+        """Build the tree by evaluating one node a time"""
+        roots = (trees.B, trees.root_indices)
+        sim0 = 1
+        if self._expand_all_root_children:
+            v = self._expand_children(trees, roots, 1, to_plays)
+            branch_factor = trees.branch_factor
+            trees.visit_count[roots] += branch_factor
```
breakds, line 409: Suggest adding a comments here saying that we are effectively backing up the expansion result (of the roots' children)?

alf/algorithms/mcts_algorithm.py
```diff
@@ -366,9 +433,40 @@ def predict_step(self, time_step: TimeStep, state: MCTSState):
...
-        return AlgStep(
-            output=action, state=MCTSState(steps=state.steps + 1), info=info)
+    def _build_tree2(self, trees, to_plays):
+        """Build the tree by evaluating all the children of one node a time"""
+        branch_factor = trees.branch_factor
+        roots = (trees.B, trees.root_indices)
+        v = self._expand_children(trees, roots, 1, to_plays)
+        trees.visit_count[roots] += branch_factor
+        trees.value_sum[roots] += branch_factor * v
+        self._update_best_child(trees, roots)
```
breakds, line 454: The logic of "expanding the roots' children and backing up" are identical to the `_build_tree1` case. I do not think the code duplication is a problem, but similar to above adding a comment line summarizing this code block can be helpful to the reader.

alf/algorithms/mcts_algorithm.py
emailweixu: yes

alf/algorithms/mcts_algorithm.py
emailweixu: fixed.

alf/algorithms/mcts_algorithm.py
emailweixu: updated

alf/algorithms/mcts_algorithm.py
emailweixu: Good suggestion.

alf/algorithms/mcts_algorithm.py
emailweixu: added.

alf/algorithms/mcts_algorithm.py
emailweixu: You are right, in this case, broadcast does not do anything

alf/algorithms/mcts_algorithm.py
emailweixu: added

alf/algorithms/mcts_algorithm.py
emailweixu: added.

alf/algorithms/mcts_models.py
emailweixu: updated

alf/algorithms/mcts_models.py
emailweixu: removed.

alf/algorithms/mcts_models.py
emailweixu: This is based on the requirement the action_probs sum to one.

breakds (approved):

Haichao-Zhang (approved):

0xJchen:
@emailweixu Hi, wei. In the original Sampled Muzero paper, the authors claim that the action representation is:
```
For the policy prediction, we chose the factored policy representation introduced by (Tang & Agrawal, 2020), representing each
dimension by a categorical distribution
```
image

However in the test `examples\muzero_pendulum_conf.py`, it is using a `StableNormalProjectionNetwork` which makes policy a Gaussian instead of the Categorical. Moreover, the `CategoricalProjectionNetwork` does not support continuous action (it is only for discrete action space) through the above categorical trick.
May I ask what's the concern here and if I can help with it.

emailweixu:
> 

If you want to use factorized discrete distribution, you can try to `alf.environments.alf_wrappers.DiscreteActionWrapper` to change the action space and then use `CategoricalProjectionNetwork`.

emailweixu:
> > 
> 
> If you want to use factorized discrete distribution, you can try to `alf.environments.alf_wrappers.DiscreteActionWrapper` to change the action space and then use `CategoricalProjectionNetwork`.

@PeppaCat Sorry, my previous answer is not correct. `alf.environments.alf_wrappers.DiscreteActionWrapper` does not do what you want. The original `DiscreteActionWrapper` flattens the discretized action into one dimensional discrete action. It can be slightly modified so that the new action space is a multi-dimensional discrete space. Then `CategoricalProjectionNetwork` will output a factorized categorical distribution.

0xJchen:
Thanks for pointing out that! I noticed this and create customized wrapper for it.

PR #1055, opened 2021-11-06
Stabilized PPO

Haichao-Zhang:
Background: encountered unstable training behavior when using PPO for CARLA task (Break also encountered a similar issue on meta drive training), e.g. loss has inf value, actions eventually become nan.

This PR improves the stability of PPO algorithm, especially for continuous control task, by better handling the log probability values.

### DIFF
--- a/alf/algorithms/actor_critic_algorithm.py
+++ b/alf/algorithms/actor_critic_algorithm.py
@@ -29,8 +29,8 @@ ActorCriticState = namedtuple(
 
 ActorCriticInfo = namedtuple(
     "ActorCriticInfo", [
-        "step_type", "discount", "reward", "action", "action_distribution",
-        "value", "reward_weights"
+        "step_type", "discount", "reward", "action", "log_prob",
+        "action_distribution", "value", "reward_weights"
     ],
     default_value=())
 
@@ -151,7 +151,8 @@ class ActorCriticAlgorithm(OnPolicyAlgorithm):
         action_distribution, actor_state = self._actor_network(
             inputs.observation, state=state.actor)
 
-        action = dist_utils.sample_action_distribution(action_distribution)
+        action, log_prob = dist_utils.sample_action_distribution(
+            action_distribution, return_log_prob=True)
 
         if self.has_multidim_reward():
             reward_weights = tensor_utils.tensor_extend_new_dim(
@@ -163,6 +164,7 @@ class ActorCriticAlgorithm(OnPolicyAlgorithm):
             state=ActorCriticState(actor=actor_state, value=value_state),
             info=ActorCriticInfo(
                 action=common.detach(action),
+                log_prob=log_prob,
                 value=value,
                 step_type=inputs.step_type,
                 reward=inputs.reward,
--- a/alf/algorithms/actor_critic_algorithm.py
+++ b/alf/algorithms/actor_critic_algorithm.py
@@ -164,7 +164,7 @@ class ActorCriticAlgorithm(OnPolicyAlgorithm):
             state=ActorCriticState(actor=actor_state, value=value_state),
             info=ActorCriticInfo(
                 action=common.detach(action),
-                log_prob=log_prob,
+                log_prob=common.detach(log_prob),
                 value=value,
                 step_type=inputs.step_type,
                 reward=inputs.reward,
--- a/alf/examples/ppo_conf.py
+++ b/alf/examples/ppo_conf.py
@@ -24,14 +24,7 @@ alf.config(
 
 alf.config('EntropyTargetAlgorithm', initial_alpha=1.)
 
-# use proper positive value for stablization, especially for continuous control
-log_prob_clipping = 100
-
-alf.config(
-    'PPOLoss',
-    entropy_regularization=None,
-    normalize_advantages=True,
-    log_prob_clipping=log_prob_clipping)
+alf.config('PPOLoss', entropy_regularization=None, normalize_advantages=True)
 
 alf.config('PPOAlgorithm', loss_class=PPOLoss)
 
--- a/alf/examples/ppo_conf.py
+++ b/alf/examples/ppo_conf.py
@@ -24,7 +24,14 @@ alf.config(
 
 alf.config('EntropyTargetAlgorithm', initial_alpha=1.)
 
-alf.config('PPOLoss', entropy_regularization=None, normalize_advantages=True)
+# use proper positive value for stablization, especially for continuous control
+log_prob_clipping = 100
+
+alf.config(
+    'PPOLoss',
+    entropy_regularization=None,
+    normalize_advantages=True,
+    log_prob_clipping=log_prob_clipping)
 
 alf.config('PPOAlgorithm', loss_class=PPOLoss)
 
--- a/alf/test/case.py
+++ b/alf/test/case.py
@@ -53,6 +53,6 @@ class TestCase(unittest.TestCase):
                               'First argument is not a Tensor')
         self.assertIsInstance(t2, torch.Tensor,
                               'Second argument is not a Tensor')
-        if not (torch.min(torch.abs(t1 - t2)) > epsilon):
+        if not (torch.max(torch.abs(t1 - t2)) > epsilon):
             standardMsg = '%s is actually close to %s' % (t1, t2)
             self.fail(self._formatMessage(msg, standardMsg))
--- a/alf/test/case.py
+++ b/alf/test/case.py
@@ -53,6 +53,6 @@ class TestCase(unittest.TestCase):
                               'First argument is not a Tensor')
         self.assertIsInstance(t2, torch.Tensor,
                               'Second argument is not a Tensor')
-        if not (torch.max(torch.abs(t1 - t2)) > epsilon):
+        if torch.max(torch.abs(t1 - t2)) < epsilon:
             standardMsg = '%s is actually close to %s' % (t1, t2)
             self.fail(self._formatMessage(msg, standardMsg))
--- a/alf/utils/dist_utils.py
+++ b/alf/utils/dist_utils.py
@@ -836,10 +836,22 @@ def rsample_action_distribution(nested_distributions, return_log_prob=False):
     Args:
         nested_distributions (nested Distribution): action distributions.
         return_log_prob (bool): whether to compute and return the log
-            probability of the sampled actions, in addition to the rsampled
-            actions.
+            probability of the sampled actions, in addition to the sampled
+            actions. In some cases, it is useful to compute the log probability
+            immediately after the actions are sampled, as some subsequent
+            operations might makes the cache mechanism (if turned on) invalid.
+            Some example scenarios include 1) additional sampling operation
+            applied on ``nested_distributions``, 2) some operations applied to
+            the actions sampled from ``nested_distributions`` (e.g., cloning).
+            This which could cause numerical issues if we want to compute the
+            log probability for actions sampled at an early stage,
+            especially for actions that are close to action bounds.
+            For more details on PyTorch Transform, its cache mechanism, and its
+            impacts on RL algorithms, please check
+            `<https://alf.readthedocs.io/en/latest/notes/pytorch_notes.html#transform-bijector>`_.
     Returns:
-        rsampled actions
+        - sampled actions if return_log_prob is False
+        - sampled actions and log_prob if return_log_prob is True
     """
     assert all(nest.flatten(nest.map_structure(lambda d: d.has_rsample,
                 nested_distributions))), \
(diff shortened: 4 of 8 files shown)


alf/examples/ppo_conf.py
```diff
@@ -24,7 +24,14 @@
...
 
-alf.config('PPOLoss', entropy_regularization=None, normalize_advantages=True)
+# use proper positive value for stablization, especially for continuous control
+log_prob_clipping = 100
+
+alf.config(
+    'PPOLoss',
+    entropy_regularization=None,
+    normalize_advantages=True,
+    log_prob_clipping=log_prob_clipping)
```
breakds: This may have impact on all configurations that includes `ppo_conf.py`, may need to update those as well.

breakds:
Looking at this PR and I think it is highly possible that my NaN problem is related to this as well. I patched this and ran training again but still hit the crash on, therefore it is likely that the problem on my side is different.

To summarize, the problem I had is at some point the whole network (encoder + stable normal projection net) has all their parameter updated to NaN, and therefore it is producing an ill-formed action distribution (with NaN as loc/scale). Still looking into it.

alf/utils/value_ops.py
```diff
@@ -66,7 +66,14 @@ def action_importance_ratio(action_distribution, collect_action_distribution,
...
     action_log_prob = dist_utils.compute_log_probability(
         current_policy_distribution, action)
+
     if log_prob_clipping > 0.0:
+        # here we need to clamp the sample_action_log_probs as it is computed
+        # using actions that not sampled from the collect_action_distribution
+        # itself but from another distribution and is more prone to have
+        # numerical issues
+        sample_action_log_probs = sample_action_log_probs.clamp(
+            -log_prob_clipping, log_prob_clipping)
```
hnyu: I think in this case we should just *turn off* cache and force the distribution to compute the inverse, e.g., using Beta with cache_size=0? Or unsquashed Normal distribution (as in the very original PPO)?

Just applying this kind of clipping is sort of hacky in my opinion.

Haichao-Zhang: Cache does not play a role in the computation in this case, and that's the reason for unstable computation.
I have already used beta which seems to be more stable.

breakds:
I got another repro with debugging info. In my case the problem happened in computing pg loss in `ActorCriticLoss`. To be more specific:

```python
    def _pg_loss(self, info, advantages):
        action_log_prob = dist_utils.compute_log_probability(
            info.action_distribution, info.action)
        return -advantages * action_log_prob
```

Basically it is the same `compute_log_probability` issue. If my understanding is correct, this is also due to `action_distribution` and `action` are inconsistent so that we get 0 probability density evaluated. What confuses me is why it is more error prone in `StableGaussian` than `Beta`?

Is `grad_clip` hook the reason?

alf/utils/value_ops.py
hnyu: > Cache does not play a role in the computation in this case, and that's the reason for unstable computation. I have already using beta which seems to be more stable.

Right. If the 'y' object is not the same with the cached one, then it will compute.

hnyu:
> I got another repro with debugging info. In my case the problem happened in computing pg loss in `ActorCriticLoss`. To be more specific:
> 
> ```python
>     def _pg_loss(self, info, advantages):
>         action_log_prob = dist_utils.compute_log_probability(
>             info.action_distribution, info.action)
>         return -advantages * action_log_prob
> ```
> 
> Basically it is the same `compute_log_probability` issue. If my understanding is correct, this is also due to `action_distribution` and `action` are inconsistent so that we get 0 probability density evaluated. What confuses me is why it is more error prone in `StableGaussian` than `Beta`?
> 
> Is `grad_clip` hook the reason?

A conventional Gaussian won't have this issue. Usually our NormalProjectionNetwork have a squashing transform applied to the sampled 'x' so that 'x' is in [-1,1]. This transform is done by tanh. Given a 'y' that's close to either -1 or 1, its inverse 'x' can be ill-defined. The inverse is needed to compute the log likelihood of 'y' (change-of-variable). One way to solve this numerical issue is to cache a (x, y) pair and when doing inverse, just look up the corresponding 'x'. But if the distribution and sampled action are not matched, this lookup method will be skipped and an inverse will be computed. 

For Beta, it natively supports samples in [-1,1] and there is no squashing used. But if its two parameters are too special, there still could be huge gradients near -1 or 1.

alf/utils/value_ops.py
hnyu: > Cache does not play a role in the computation in this case, and that's the reason for unstable computation. I have already using beta which seems to be more stable.

You could try BetaProjectionNetwork with min_concentration=1. or Beta with eps=1e-5.

breakds:
> > I got another repro with debugging info. In my case the problem happened in computing pg loss in `ActorCriticLoss`. To be more specific:
> > ```python
> >     def _pg_loss(self, info, advantages):
> >         action_log_prob = dist_utils.compute_log_probability(
> >             info.action_distribution, info.action)
> >         return -advantages * action_log_prob
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Basically it is the same `compute_log_probability` issue. If my understanding is correct, this is also due to `action_distribution` and `action` are inconsistent so that we get 0 probability density evaluated. What confuses me is why it is more error prone in `StableGaussian` than `Beta`?
> > Is `grad_clip` hook the reason?
> 
> A conventional Gaussian won't have this issue. Usually our NormalProjectionNetwork have a squashing transform applied to the sampled 'x' so that 'x' is in [-1,1]. This transform is done by tanh. Given a 'y' that's close to either -1 or 1, its inverse 'x' can be ill-defined. The inverse is needed to compute the log likelihood of 'y' (change-of-variable). One way to solve this numerical issue is to cache a (x, y) pair and when doing inverse, just look up the corresponding 'x'. But if the distribution and sampled action are not matched, this lookup method will be skipped and an inverse will be computed.
> 
> For Beta, it natively supports samples in [-1,1] and there is no squashing used. But if its two parameters are too special, there still could be huge gradients near -1 or 1.

Thanks for the explanation Haonan. And though it happens rather late, the Beta version did have the same problem after thounsands of iterations.

hnyu:
> > > I got another repro with debugging info. In my case the problem happened in computing pg loss in `ActorCriticLoss`. To be more specific:
> > > ```python
> > >     def _pg_loss(self, info, advantages):
> > >         action_log_prob = dist_utils.compute_log_probability(
> > >             info.action_distribution, info.action)
> > >         return -advantages * action_log_prob
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Basically it is the same `compute_log_probability` issue. If my understanding is correct, this is also due to `action_distribution` and `action` are inconsistent so that we get 0 probability density evaluated. What confuses me is why it is more error prone in `StableGaussian` than `Beta`?
> > > Is `grad_clip` hook the reason?
> > 
> > 
> > A conventional Gaussian won't have this issue. Usually our NormalProjectionNetwork have a squashing transform applied to the sampled 'x' so that 'x' is in [-1,1]. This transform is done by tanh. Given a 'y' that's close to either -1 or 1, its inverse 'x' can be ill-defined. The inverse is needed to compute the log likelihood of 'y' (change-of-variable). One way to solve this numerical issue is to cache a (x, y) pair and when doing inverse, just look up the corresponding 'x'. But if the distribution and sampled action are not matched, this lookup method will be skipped and an inverse will be computed.
> > For Beta, it natively supports samples in [-1,1] and there is no squashing used. But if its two parameters are too special, there still could be huge gradients near -1 or 1.
> 
> Thanks for the explanation Haonan. And though it happens rather late, the Beta version did have the same problem after thounsands of iterations.

Can you try BetaProjectionNetwork with min_concentration=1 and/or Beta.eps=1e-5?

breakds:
> > > > I got another repro with debugging info. In my case the problem happened in computing pg loss in `ActorCriticLoss`. To be more specific:
> > > > ```python
> > > >     def _pg_loss(self, info, advantages):
> > > >         action_log_prob = dist_utils.compute_log_probability(
> > > >             info.action_distribution, info.action)
> > > >         return -advantages * action_log_prob
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > Basically it is the same `compute_log_probability` issue. If my understanding is correct, this is also due to `action_distribution` and `action` are inconsistent so that we get 0 probability density evaluated. What confuses me is why it is more error prone in `StableGaussian` than `Beta`?
> > > > Is `grad_clip` hook the reason?
> > > 
> > > 
> > > A conventional Gaussian won't have this issue. Usually our NormalProjectionNetwork have a squashing transform applied to the sampled 'x' so that 'x' is in [-1,1]. This transform is done by tanh. Given a 'y' that's close to either -1 or 1, its inverse 'x' can be ill-defined. The inverse is needed to compute the log likelihood of 'y' (change-of-variable). One way to solve this numerical issue is to cache a (x, y) pair and when doing inverse, just look up the corresponding 'x'. But if the distribution and sampled action are not matched, this lookup method will be skipped and an inverse will be computed.
> > > For Beta, it natively supports samples in [-1,1] and there is no squashing used. But if its two parameters are too special, there still could be huge gradients near -1 or 1.
> > 
> > 
> > Thanks for the explanation Haonan. And though it happens rather late, the Beta version did have the same problem after thounsands of iterations.
> 
> Can you try BetaProjectionNetwork with min_concentration=1 and/or Beta.eps=1e-5?

With that parameters it gets 5000 iterations training done without problem (previously it usually crashed between 1000 - 2000 iterations). However I found `BetaProjection`'s performance really bad. I'll train with a temporary fix on the `StableNormal` to see how it works. Thanks for the suggestions and explanation - really helpful!

hnyu:
> > > > > I got another repro with debugging info. In my case the problem happened in computing pg loss in `ActorCriticLoss`. To be more specific:
> > > > > ```python
> > > > >     def _pg_loss(self, info, advantages):
> > > > >         action_log_prob = dist_utils.compute_log_probability(
> > > > >             info.action_distribution, info.action)
> > > > >         return -advantages * action_log_prob
> > > > > ```
> > > > > 
> > > > > 
> > > > >     
> > > > >       
> > > > >     
> > > > > 
> > > > >       
> > > > >     
> > > > > 
> > > > >     
> > > > >   
> > > > > Basically it is the same `compute_log_probability` issue. If my understanding is correct, this is also due to `action_distribution` and `action` are inconsistent so that we get 0 probability density evaluated. What confuses me is why it is more error prone in `StableGaussian` than `Beta`?
> > > > > Is `grad_clip` hook the reason?
> > > > 
> > > > 
> > > > A conventional Gaussian won't have this issue. Usually our NormalProjectionNetwork have a squashing transform applied to the sampled 'x' so that 'x' is in [-1,1]. This transform is done by tanh. Given a 'y' that's close to either -1 or 1, its inverse 'x' can be ill-defined. The inverse is needed to compute the log likelihood of 'y' (change-of-variable). One way to solve this numerical issue is to cache a (x, y) pair and when doing inverse, just look up the corresponding 'x'. But if the distribution and sampled action are not matched, this lookup method will be skipped and an inverse will be computed.
> > > > For Beta, it natively supports samples in [-1,1] and there is no squashing used. But if its two parameters are too special, there still could be huge gradients near -1 or 1.
> > > 
> > > 
> > > Thanks for the explanation Haonan. And though it happens rather late, the Beta version did have the same problem after thounsands of iterations.
> > 
> > 
> > Can you try BetaProjectionNetwork with min_concentration=1 and/or Beta.eps=1e-5?
> 
> With that parameters it gets 5000 iterations training done without problem (previously it usually crashed between 1000 - 2000 iterations). However I found `BetaProjection`'s performance really bad. I'll train with a temporary fix on the `StableNormal` to see how it works. Thanks for the suggestions and explanation - really helpful!

Not sure about your tasks, but I've been using Beta as the default policy distribution for quite a long time. On some tasks like safety gym I compared, Beta is no worse than Normal.

alf/utils/value_ops.py
```diff
@@ -66,7 +66,14 @@ def action_importance_ratio(action_distribution, collect_action_distribution,
 
     action_log_prob = dist_utils.compute_log_probability(
         current_policy_distribution, action)
+
     if log_prob_clipping > 0.0:
+        # here we need to clamp the sample_action_log_probs as it is computed
+        # using actions that not sampled from the collect_action_distribution
+        # itself but from another distribution and is more prone to have
+        # numerical issues
```
emailweixu: For PPO, `action` is sampled from `collect_dictribution`.

Haichao-Zhang: > For PPO, `action` is sampled from `collect_dictribution`.

Right, my comment was incorrect. Will update it. Working on an alternative solution by computing the log_prob while sampling.

alf/algorithms/actor_critic_algorithm.py
```diff
@@ -151,7 +151,8 @@ def rollout_step(self, inputs: TimeStep, state: ActorCriticState):
         action_distribution, actor_state = self._actor_network(
             inputs.observation, state=state.actor)
 
-        action = dist_utils.sample_action_distribution(action_distribution)
+        action, log_prob = dist_utils.sample_action_distribution(
```
breakds, line 154: A documentation briefly explaining why the corresponding `log_prob` is stored can help the future the reader.

breakds (dismissed):

alf/test/case.py
```diff
@@ -53,6 +53,6 @@ def assertTensorNotClose(self, t1, t2, epsilon=1e-6, msg=None):
                               'First argument is not a Tensor')
         self.assertIsInstance(t2, torch.Tensor,
                               'Second argument is not a Tensor')
-        if not (torch.min(torch.abs(t1 - t2)) > epsilon):
+        if not (torch.max(torch.abs(t1 - t2)) > epsilon):
```
Haichao-Zhang: There seems to be a bug for checking the `closeness` of two tensors.

alf/algorithms/actor_critic_algorithm.py
Haichao-Zhang: > A documentation briefly explaining why the corresponding `log_prob` is stored can help the future the reader.

Good point. Added more documentation in the sampling function. Also added related unit tests.

alf/test/case.py
hnyu: Suggest directly checking `if torch.max(torch.abs(t1 - t2)) < epsilon` for readability.

alf/utils/dist_utils.py
```diff
@@ -828,32 +828,76 @@ def _compute_log_prob(single_distribution, single_action):
...
-        sampled actions
+        - sampled actions if return_log_prob is False
+        - sampled actions and log_prob if return_log_prob is True
     """
-    return nest.map_structure(lambda d: d.sample(), nested_distributions)
+    sample = nest.map_structure(lambda d: d.sample(), nested_distributions)
+    if return_log_prob:
+        log_prob = nest.map_structure(
+            lambda d, s: compute_log_probability(d, s), nested_distributions,
+            sample)
```
emailweixu: compute_log_probability can handle nested distribution

emailweixu (approved):

PR #1056, opened 2021-11-09
Support for truncated distributions

emailweixu:
Squashing transformations seem to be a problem in training continuous action distributions.
Truncated distributions may be helpful.

### DIFF
--- a/alf/networks/projection_networks.py
+++ b/alf/networks/projection_networks.py
@@ -689,8 +689,8 @@ class TruncatedProjectionNetwork(Network):
             std_bias_initializer_value (float): Initial value for the bias of the
                 ``std_projection_layer``.
             state_dependent_scale (bool): If True, std will be generated depending
-                on the current state; otherwise a global std will be generated
-                regardless of the current state.
+                on the current state (i.e. inputs); otherwise a global scale will
+                be generated regardless of the current state.
             scale_transform (Callable): Transform to apply to the std, on top of
                 `activation`.
             dist_ctor(Callable): constructor for the distribution called as:
@@ -735,7 +735,7 @@ class TruncatedProjectionNetwork(Network):
         action_means = (self._action_high + self._action_low) / 2
         action_magnitudes = (self._action_high - self._action_low) / 2
 
-        # Although the TruncatedDistribution will ensure the actions are wihtin
+        # Although the TruncatedDistribution will ensure the actions are within
         # the bound, we still make sure the loc parameter to be within the bound
         # for better numerical stability
         self._loc_transform = (
--- a/alf/networks/projection_networks.py
+++ b/alf/networks/projection_networks.py
@@ -519,7 +519,9 @@ class CauchyProjectionNetwork(NormalProjectionNetwork):
                  scale_bias_initializer_value=0.0,
                  state_dependent_scale=False,
                  scale_transform=nn.functional.softplus,
-                 **kwargs):
+                 scale_distribution=False,
+                 dist_squashing_transform=dist_utils.StableTanh(),
+                 name='CauchyProjectionNetwork'):
         """Similar to ``NormalProjectionNetwork`` except that the output
         distribution is a ``DiagMultivariateCauchy``. Also since Cauchy doesn't
         have mean or std, we provide parameters for its median and scale instead.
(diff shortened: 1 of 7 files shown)


alf/networks/projection_networks.py
```diff
@@ -648,3 +659,95 @@ def forward(self, inputs, state=()):
...
+            input_size (int): input vector dimension
+            action_spec (TensorSpec): a tensor spec containing the information
+                of the output distribution.
+            activation (Callable): activation function to use in
+                dense layers.
+            projection_output_init_gain (float): Output gain for initializing
+                action means and std weights.
+            std_bias_initializer_value (float): Initial value for the bias of the
+                ``std_projection_layer``.
+            state_dependent_scale (bool): If True, std will be generated depending
```
breakds, line 691: Here the word `state` is a bit confusing as it does not appear in the context. I suppose it means the input, in which case an annotation on `the current state (input)` would be helpful.

alf/utils/distributions.py
```diff
@@ -0,0 +1,318 @@
...
+class TruncatedDistribution(td.Distribution):
+    r"""The base class of truncated distributions.
+
+    A truncated distribution :math:`q(x)` is defined as a standard base distribution :math:`p(x)` and
+    location :math:`\mu`, scale parameters :math:`s`, lower bound :math:`l` and
+    upper bound :math:`u`
+
+    .. math::
+
+        q(x) = \frac{1}{s (P(u) - P(l))}p(\frac{x-\mu}{s})
```
hnyu: Should it also have q(x)=0, x > u or x < l?

alf/utils/distributions.py
```diff
@@ -0,0 +1,318 @@
...
+        'scale': td.constraints.positive
+    }
+    has_rsample = True
+
+    def __init__(self, loc: Tensor, scale: Tensor, lower_bound: Tensor,
+                 upper_bound: Tensor, its: InverseTransformSampling):
+        event_shape = torch.broadcast_shapes(lower_bound.shape,
+                                             upper_bound.shape)
+        batch_shape = torch.broadcast_shapes(scale.shape, loc.shape,
+                                             event_shape)
```
breakds, line 166: This seems to imply `scale` and `loc`, unlike `lower_bound` and `upper_bound`, may have batch dimension. If so, should we explicitly state this in the docstring?

Also it is not entirely clear that `lower_bound` and `upper_bound` ('s broadcast) defines the shape of the event from the docstring.

alf/utils/distributions.py
```diff
@@ -0,0 +1,318 @@
...
+# limitations under the License.
+
+import math
+import torch
+from torch import Tensor
+
+import torch.distributions as td
+
+
+class InverseTransformSampling(object):
```
breakds, line 22: If I understand it correctly, the usage of such interface in `TruncatedDistribution` implies "zero mean" or more strictly the "even function" of the p.d.f. - should we document that here to help when new `*ITS` are added?

alf/utils/distributions.py
```diff
@@ -0,0 +1,318 @@
...
+            sample_shape: sample shape
+        Returns:
+            Tensor of shape ``sample_shape + batch_shape + event_shape``
+        """
+        r = torch.rand(sample_shape + self._batch_shape + self._event_shape)
+        r = (1 - r) * self._cdf_lb + r * self._cdf_ub
+        r = r.clamp(0.001, 0.999)
+        x = self._its.icdf(r) * self._scale + self._loc
+        assert torch.isfinite(x).all()
+        # because of the r.clamp() above, x may be out of bound
```
hnyu, line 221: r.clamp will make sure that we won't have x out of bound?

alf/utils/distributions.py
```diff
@@ -0,0 +1,318 @@
...
+        are batched.
+
+        Args:
+            sample_shape: sample shape
+        Returns:
+            Tensor of shape ``sample_shape + batch_shape + event_shape``
+        """
+        r = torch.rand(sample_shape + self._batch_shape + self._event_shape)
+        r = (1 - r) * self._cdf_lb + r * self._cdf_ub
+        r = r.clamp(0.001, 0.999)
```
hnyu, line 219: Is 1e-3 the standard way of doing this clipping? It seems a somewhat arbitrary decision.

alf/networks/projection_networks.py
```diff
@@ -648,3 +659,95 @@ def forward(self, inputs, state=()):
...
+        self._action_high = torch.as_tensor(action_spec.maximum).broadcast_to(
+            action_spec.shape)
+        self._action_low = torch.as_tensor(action_spec.minimum).broadcast_to(
+            action_spec.shape)
+        self._dist_ctor = dist_ctor
+
+        action_means = (self._action_high + self._action_low) / 2
+        action_magnitudes = (self._action_high - self._action_low) / 2
+
+        # Although the TruncatedDistribution will ensure the actions are wihtin
```
Haichao-Zhang: wihtin -> within

alf/utils/distributions.py
```diff
@@ -0,0 +1,318 @@
...
+
+    where :math:`P` is the cdf of :math:`p`.
+
+    Args:
+        loc: the location parameter
+        scale: the scale parameter
+        lower_bound: the lower bound
+        upper_bound: the upper bound
+        its: the standard distribution to be used.
+    """
```
Haichao-Zhang: Maybe we can to move Line 146-151 to ``__init__`` function? And then it might be more useful to add more comments on the expected type/shape of these parameters.

breakds (approved):

PR #1057, opened 2021-11-11
conf and SAC changes to reproduce reported locomotion results

hnyu:
Running a performance verification job on our cluster. The tensorboard is at http://tensorboard-1013392.train.hogpu.cc/ (three random seeds for each of the four tasks).

No uniform action sampling: http://tensorboard-1017452.train.hogpu.cc/

Summary of modifications:
1. Uniformly sampling actions from [-1,1] before reaching `init_collect_steps`.
2. Policy and critic losses have different masks (offset by 1). 

`locomotion_conf.py` is the standard configuration used by the SAC paper.

### DIFF
--- a/alf/algorithms/sac_algorithm.py
+++ b/alf/algorithms/sac_algorithm.py
@@ -169,6 +169,7 @@ class SacAlgorithm(OffPolicyAlgorithm):
                  critic_optimizer=None,
                  alpha_optimizer=None,
                  debug_summaries=False,
+                 reproduce_locomotion=False,
                  name="SacAlgorithm"):
         """
         Args:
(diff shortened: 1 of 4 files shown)


hnyu:
@runjerry I replaced `common.is_eval()` with a local variable. For `self._training_started`, so far I've kept the current way (basically two lines in total) because there will be more work doing this in a base class. If there are more usage cases in the future, I can define it as a common function.

runjerry:
> @runjerry I replaced `common.is_eval()` with a local variable. For `self._training_started`, so far I've kept the current way (basically two lines in total) because there will be more work doing this in a base class. If there are more usage cases in the future, I can define it as a common function.

Looks good to me now, thanks.

runjerry (approved):

PR #1060, opened 2021-11-12
Episodic metrics support averaging over steps adaptively

breakds:
# Motivation

As part of effort to integrate metadrive environment into alf, I would like to show metrics on tensorboard about the average velocity, average steering and success rate (reaching goal). All of them comes from the `env_info` of the environment. However, currently alf will just aggregate those information by summing over the whole episode. This makes sense for the success rate, but not so much for velocity and steering.

# Solution

There are many trade-offs between readability, flexibility and complexity. In the end the winning solution is to use the the name of the metric to indicate a different behavior. In detail, for metrics managed by `AverageEpisodicSumMetric`:

1. If the metric ends with "@step", the value will be averaged by episode length to get a per-step average value. Works for `velocity@step` and `steering@step`.
2. Otherwise the normal behavior will be active, where the sum of the value over the whole episode is reported. E.g. `reach_goal` where a `1.0` is reported at the end of the episode if the car reaches goal.

Note that as a side effect I moved `self._current_step` to the parent class `AverageEpisodicSumMetric`.

## Why not ... ?

* A mode option of the constructor of `AverageEnvInfoMetric`? - This requires extra configuration of this for each training configuration but I think such behavior is more of a environment specific thing. The above solution frees the user from worrying about this, while put the burden on the right shoulder - the environment's author.
* A per field nest specifying the aggregation type of each of the field in env info? This is too complicated for readability consideration. Besides, the documentation cost is going to be very high.
* Only apply this to `AverageEnvInfoMetric` instead of `AverageEpisodicSumMetric`? This is to make the solution more flexible so that other types of `AverageEpisodicSumMetric` can enjoy this feature as well.
* Change the name of `AverageEpisodicSumMetric`? That will be make the change too big. I think The original name of the class still make sense as averaging still requires a "sum" aggregation first.

# Testing

* The original tests passed.
* Added a new unit test and make sure the new feature works as intended.

Tested with MetaDrive training:

metrics

### DIFF
--- a/alf/metrics/metrics.py
+++ b/alf/metrics/metrics.py
@@ -132,10 +132,12 @@ class AverageEpisodicSumMetric(metric.StepMetric):
     """A base metric to sum up quantities over an episode. It supports accumulating
     a nest of scalar values.
 
-    NOTE: normally this class report metrics by summing values over the whole
-    episode. However, if a metric is named with a postfix of "@step", the value
-    will be modified by averaging it over the whole episode length, so that a
-    per-step average value is reported.
+    NOTE: normally this class and its sub-classes report metrics by summing values
+    over the whole episode. However, there is a special treatment: if
+    ``_extract_metric_values()`` returns a nested structure in which a dictionary or
+    namedtuple has a field with postfix "@step", the corresponding value will be
+    averaged instead of summed over the whole episode length, so that a per-step
+    average value is reported.
 
     """
 
--- a/alf/metrics/metrics.py
+++ b/alf/metrics/metrics.py
@@ -128,8 +128,8 @@ class NumberOfEpisodes(metric.StepMetric):
         self._number_episodes.fill_(0)
 
 
-class AverageEpisodicSumMetric(metric.StepMetric):
-    """A base metric to sum up quantities over an episode. It supports accumulating
+class AverageEpisodicAggregationMetric(metric.StepMetric):
+    """A base metric to aggregate quantities over an episode. It supports accumulating
     a nest of scalar values.
 
     NOTE: normally this class and its sub-classes report metrics by summing values
@@ -142,7 +142,7 @@ class AverageEpisodicSumMetric(metric.StepMetric):
     """
 
     def __init__(self,
-                 name="AverageEpisodicSumMetric",
+                 name="AverageEpisodicAggregationMetric",
                  prefix='Metrics',
                  dtype=torch.float32,
                  batch_size=1,
@@ -160,7 +160,7 @@ class AverageEpisodicSumMetric(metric.StepMetric):
             example_metric_value (nest): an example of metric value to be summarized;
                 if ``None``, a zero scalar is used.
         """
-        super(AverageEpisodicSumMetric, self).__init__(
+        super(AverageEpisodicAggregationMetric, self).__init__(
             name=name, dtype=dtype, prefix=prefix)
         if example_metric_value is None:
             example_metric_value = torch.zeros((), device='cpu')
@@ -267,7 +267,7 @@ class AverageEpisodicSumMetric(metric.StepMetric):
         alf.nest.map_structure(lambda acc: acc.fill_(0), self._accumulator)
 
 
-class AverageReturnMetric(AverageEpisodicSumMetric):
+class AverageReturnMetric(AverageEpisodicAggregationMetric):
     """Metric for computing the average return."""
 
     def __init__(self,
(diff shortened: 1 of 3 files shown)


hnyu:
Generally I like the flexibility of adding "@step" into the metric fields without explicitly configuring the metric class. But we need to document this usage very clearly. 

Actually, I think we need to make it clear that in order to use this option, when defining a new metric class, in its `_extract_metric_values`, we need to return a dict or namedtuple with fields as "xxx@step". Other nest types won't support this feature because currently you're using map_structure's path info to decide.

alf/metrics/metrics_test.py
```diff
@@ -123,6 +123,79 @@ def testMetric(self, metric_class, num_trajectories, expected_result,
             self.assertEqual([0.0] * 2 if vector_reward else 0.0,
                              metric.result())
 
+    def test_average_per_step(self):
+        metric = AverageEnvInfoMetric(
```
hnyu: Suggest defining a new example metric class in this test case where `_extract_metric_values` return other custom dict or namedtuple other than `env_info`.

breakds:
> Generally I like the flexibility of adding "@step" into the metric fields without explicitly configuring the metric class. But we need to document this usage very clearly.
> 
> Actually, I think we need to make it clear that in order to use this option, when defining a new metric class, in its `_extract_metric_values`, we need to return a dict or namedtuple with fields as "xxx@step". Other nest types won't support this feature because currently you're using map_structure's path info to decide.

Agree with the comment. Let me improve the documentation about this.

alf/metrics/metrics_test.py
breakds: Good point. Updated the unit tests with a cusotimzed metric class with customized value extraction.

breakds:
The PR is updated according to the comments. 

@hnyu and I had a brief discussion on how to make `example_metric_value` more consistent. I decided to not add that into this PR so that this PR can be more concentrated and simple.

alf/metrics/metrics_test.py
```diff
@@ -45,6 +46,44 @@ def timestep_last(reward, env_id, env_info):
...
+            name=name,
+            dtype=dtype,
+            prefix=prefix,
+            batch_size=batch_size,
+            buffer_size=buffer_size,
+            example_metric_value={
+                'velocity@step': 1.0,
+                'acceleration@step': 0.1,
+                'success': 0.0
+            })
```
hnyu, line 74: Right now this example_metric_value usage is really inconvenient. It's much more natural to directly provide an example timestep as an argument of AverageEpisodicMetric and call _extract_metric_values on it.

alf/metrics/metrics.py
```diff
@@ -213,17 +230,24 @@ def _update_accumulator_(acc, val):
 
         alf.nest.map_structure(_update_accumulator_, self._accumulator, values)
```
hnyu, line 227: Actually I feel that self._accumulator should be plain tensors instead of buffers, because if we run from a checkpoint, although the metric buffers should be restored, the accumulators should be reset to zero instead of being restored, since all envs start from new episodes.

alf/metrics/metrics.py
```diff
@@ -131,6 +131,14 @@ def reset(self):
 class AverageEpisodicSumMetric(metric.StepMetric):
     """A base metric to sum up quantities over an episode. It supports accumulating
     a nest of scalar values.
+
```
Haichao-Zhang, line 134: Just a thought. The added note is clear. But the name of the class ``AverageEpisodicSumMetric`` can not accurately reflect its functionalities anymore and might be a bit mis-leading. Maybe can remove ``Sum``?

alf/metrics/metrics.py
breakds: Fixed in #1083

alf/metrics/metrics_test.py
breakds: Sure. Will implement this in a separate PR.

alf/metrics/metrics.py
hnyu: I think it's OK. We have average option for NestSum().

breakds: Good point. Updated this to `AverageEpisodicAggregationMetric`.

breakds: Have updated the name to `AverageEpisodicAggregationMetric` to make it more general.

hnyu (approved):

Haichao-Zhang (approved):

PR #1062, opened 2021-11-12
Add BatchedTensorWrapper for environments

breakds:
# Motivation

The upcoming `MetaDriveWrapper` of Meta Drive environments cannot be put into a thread (and therefore `ThreadEnvironment` wrapper), but still need to be wrapped in `ProcessEnvironment`. The latter prevents it from having tensor-based I/O (i.e. `step()` and `reset()` should only use numpy for interfaces), and the former prevents it from being used in `alf.bin.play`.

# Solution

1. Use Meta Drive environment directly without a `ThreadEnvironment` wrapper.
2. To make No.1 possible, we will need another wrapper to achieve one thing that `ThreadEnvironment` provide, which is to conver the numpy-based non-batch output to torch tensor-based batch output, for both `reset()` and `step()`

This PR implements that thin wrapper.

# Testing

Used jointly with the meta drive environment for training and confirmed that the training went properly.

### DIFF
--- a/alf/metrics/metrics.py
+++ b/alf/metrics/metrics.py
@@ -165,16 +165,12 @@ class AverageEpisodicSumMetric(metric.StepMetric):
         raise NotImplementedError()
 
     def _initialize(self, example_metric_value):
-        counter = [0]
-
         def _init_buf(val):
             return MetricBuffer(max_len=self._buffer_size, dtype=self._dtype)
 
         def _init_acc(val):
             accumulator = torch.zeros(
                 self._batch_size, dtype=self._dtype, device='cpu')
-            self.register_buffer('_accumulator%d' % counter[0], accumulator)
-            counter[0] += 1
             return accumulator
 
         self._buffer = alf.nest.map_structure(_init_buf, example_metric_value)
--- a/alf/environments/alf_wrappers.py
+++ b/alf/environments/alf_wrappers.py
@@ -831,6 +831,34 @@ class CarlaActionWrapper(AlfEnvironmentBaseWrapper):
         return self._time_step_spec
 
 
+class BatchedTensorWrapper(AlfEnvironmentBaseWrapper):
+    """Wrapper that converts non-batched numpy-based I/O to batched tensors.
+    """
+
+    def __init__(self, env):
+        assert not env.batched, (
+            'BatchedTensorWrapper can only be used to wrap non-batched env')
+        super().__init__(env)
+
+    @staticmethod
+    def _to_batched_tensor(raw):
+        """Conver the structured input into batched (batch_size = 1) tensors
+        of the same structure.
+        """
+        return nest.map_structure(
+            lambda x: (torch.as_tensor(x).unsqueeze(dim=0) if isinstance(
+                x, (np.ndarray, np.number, float, int)) else x), raw)
+
+    def _step(self, action):
+        numpy_action = nest.map_structure(
+            lambda x: x.squeeze(dim=0).cpu().numpy(), action)
+        return BatchedTensorWrapper._to_batched_tensor(
+            super()._step(numpy_action))
+
+    def _reset(self):
+        return BatchedTensorWrapper._to_batched_tensor(super()._reset())
+
+
 @alf.configurable
 class DiscreteActionWrapper(AlfEnvironmentBaseWrapper):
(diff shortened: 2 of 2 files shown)


emailweixu:
> # Motivation
> The upcoming `MetaDriveWrapper` of Meta Drive environments cannot be put into a thread (and therefore `ThreadEnvironment` wrapper), but still need to be wrapped in `ProcessEnvironment`. The latter prevents it from having tensor-based I/O (i.e. `step()` and `reset()` should only use numpy for interfaces), and the former prevents it from being used in `alf.bin.play`.

Why not wrap it with ParallelAlfEnvironment?

breakds:
> > # Motivation
> > The upcoming `MetaDriveWrapper` of Meta Drive environments cannot be put into a thread (and therefore `ThreadEnvironment` wrapper), but still need to be wrapped in `ProcessEnvironment`. The latter prevents it from having tensor-based I/O (i.e. `step()` and `reset()` should only use numpy for interfaces), and the former prevents it from being used in `alf.bin.play`.
> 
> Why not wrap it with ParallelAlfEnvironment?

Sorry I forgot to explain this in the description. Mainly 2 reasons:

1. `ParallelAlfEnvironment` is a bit heavier weight
2. The follow-up call to `seed()` will also need to consider both cases as `ParalellAlfEnvironment` need to add a batch dimension.

Although I do not hold strong opinion on the above 2.

emailweixu:
> > Why not wrap it with ParallelAlfEnvironment?
> 
> Sorry I forgot to explain this in the description. Mainly 2 reasons:
> 
> 1. `ParallelAlfEnvironment` is a bit heavier weight

This does not seem to be a big issue given that this is only for evaluation.

> 2. The follow-up call to `seed()` will also need to consider both cases as `ParalellAlfEnvironment` need to add a batch dimension.

I don't understand this

breakds:
> > > Why not wrap it with ParallelAlfEnvironment?
> > 
> > 
> > Sorry I forgot to explain this in the description. Mainly 2 reasons:
> > 
> > 1. `ParallelAlfEnvironment` is a bit heavier weight
> 
> This does not seem to be a big issue given that this is only for evaluation.
> 
> > 1. The follow-up call to `seed()` will also need to consider both cases as `ParalellAlfEnvironment` need to add a batch dimension.
> 
> I don't understand this

Sorry about not making it clear. About point no.2, I was saying if we need to go for `ParalellAlfEnvironment` or `ThreadEnvironment` here: https://github.com/HorizonRobotics/alf/blob/pytorch/alf/environments/utils.py#L105-L108

we will need to add an `if` to each of the branches here, so that we can feed `[seed]` instead of `seed` to the `ParallelAlfEnvironment` case. Not a big issue still. I can change and update the PR to use `ParallelAlfEnvironment` instead.

breakds:
Reopen as I think it is still better to have this. 

https://github.com/HorizonRobotics/alf/blob/pytorch/alf/environments/utils.py#L97-L104

One of the use case of the above code block is to play the environment with a trained agent, possibly with `VideoRecorder` enabled. Using `ParallelAlfEnvironment` for `VideoRecorder` is currently not supported, and I would argue `ParallelAlfEnvironment` is not meant for play + video recording. Compared to updating `ParallelAlfEnvironment` I think adding a wrapper here is a simpler solution and has minimum (no) impact on other situations.

What do you think @emailweixu ?

Haichao-Zhang (approved):

alf/metrics/metrics.py
```diff
@@ -165,16 +165,12 @@ def _extract_metric_values(self, time_step):
...
     def _initialize(self, example_metric_value):
-        counter = [0]
-
         def _init_buf(val):
             return MetricBuffer(max_len=self._buffer_size, dtype=self._dtype)
 
         def _init_acc(val):
             accumulator = torch.zeros(
                 self._batch_size, dtype=self._dtype, device='cpu')
-            self.register_buffer('_accumulator%d' % counter[0], accumulator)
```
emailweixu, line 176: why do you remove this?
This is needed so that accumulator can be checkpointed.

breakds: This is actually from #1083, and apparently this PR has not been rebased to the latest master. 

As suggested by @hnyu , the accumulator is better not check-pointed, so that when a training is resumed from a checkpoint, the accumulating restarts.

PR #1063, opened 2021-11-12
Initial implementation of the MetaDrive environment wrapper

breakds:
# Motivation

To integrate [MetaDrive](https://metadrive-simulator.readthedocs.io/en/latest/) driving simulator into alf. This is the main PR to address #1084 

# Solution

This PR implements `suite_metadrive.load` on top of `AlfMetaDriveWrapper`.

One of the special requirement for meta drive is that its API (specifically `step()`) has to run inside the main thread. Therefore in this PR we support:

1. All `numpy` interfaces so that it can be used to create `ProcessEnvironment` and `ParallelEnvironment`.
2. Introduced a new property `no_thread_env` so that it can properly work with alf's evaluation and play as well.

# Testing

This has been used for running MetaDrive experiments successfully.

### DIFF
--- a/alf/environments/suite_metadrive.py
+++ b/alf/environments/suite_metadrive.py
@@ -168,7 +168,8 @@ class AlfMetaDriveWrapper(AlfEnvironment):
         return canvas
 
     def _acquire_next_frame(self, action):
-        """Returns the TimeStep give the input action.
+        """Returns the TimeStep given the input action.
+
         This is the underlying implementation of both _step() and _reset()
         1. In _step(), normally it just delegates to this method unless a reset
            needs to be performed.
@@ -220,13 +221,14 @@ class AlfMetaDriveWrapper(AlfEnvironment):
 
     def seed(self, seed: Optional[int] = None):
         """Reset the underlying MetaDrive environment with a specified seed.
-        MetaDrive uses a slightly different mechanism for seeds. Upon
-        construction of a MetaDrive environment, the user needs to specify a
-        seed range [start_seed, start_seed + scenario_num]. When being forced
-        to reset with a specific seed, that seed must be within the predefined
-        range.
-        Args:
 
+        MetaDrive uses a slightly different mechanism for seeds. Upon
+        construction of a MetaDrive environment, the user needs to
+        specify a seed range [start_seed, start_seed + scenario_num].
+        When being forced to reset with a specific seed, that seed
+        must be within the predefined range.
+
+        Args:
             seed: the seed that the environment will be reset with. If it is
                 specified as None, a random seed within the range will be
                 selected by the underlying MetaDrive environment.
--- a/alf/environments/suite_metadrive.py
+++ b/alf/environments/suite_metadrive.py
@@ -12,9 +12,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from typing import Tuple, Optional, Union
+
 import numpy as np
 import gym
 import torch
+
+from matplotlib.backends.backend_agg import FigureCanvasAgg as FigureCanvas
+from matplotlib.figure import Figure
 
 import alf
 from alf.environments.alf_environment import AlfEnvironment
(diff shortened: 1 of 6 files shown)


alf/environments/utils.py
```diff
@@ -97,11 +97,19 @@ def create_environment(env_name='CartPole-v0',
     elif nonparallel:
         # Each time we can only create one unwrapped env at most
 
-        # Create and step the env in a separate thread. env `step` and `reset` must
-        #   run in the same thread which the env is created in for some simulation
-        #   environments such as social_bot(gazebo)
-        alf_env = thread_environment.ThreadEnvironment(lambda: env_load_fn(
-            env_name))
+        if hasattr(env_load_fn, 'no_thread_env') and env_load_fn.no_thread_env:
```
emailweixu: `getattr(env_load_fn, 'no_thread_env', False)`

alf/environments/suite_metadrive.py
```diff
@@ -0,0 +1,201 @@
...
+    def _reset(self) -> ds.TimeStep:
+        _ = self._env.reset()
+
+        # Zero action means do nothing in both longitudinal and lateral
+        first_time_step = self._acquire_next_frame(
+            self._action_spec.zeros().cpu().numpy())
+
+        return first_time_step._replace(step_type=ds.StepType.FIRST)
+
+    def seed(self, seed):
```
emailweixu: just to confirm that metadrive does not have random seed?.

alf/environments/suite_metadrive.py
```diff
@@ -0,0 +1,201 @@
...
+        traffic_density: float = 0.1,
+        environment_num: int = 200,
+        start_seed: int = np.random.randint(10000),
+        decision_repeat: int = 5,  # 0.02 * 5 = 0.1 seconds per action
+        map_size: int = 4):
+    """The loader for metadrive environment.
+
+    See https://metadrive-simulator.readthedocs.io/en/latest/config_system.html
+
+    for details about the configurations.
```
emailweixu: map_name and batch_size is not used anywhere and need some comment. Also env_id.

alf/environments/suite_metadrive.py
```diff
@@ -0,0 +1,201 @@
...
+        return self._observation_spec
+
+    def action_spec(self):
+        return self._action_spec
+
+    def render(self, mode):
+        return self._env.render(mode='top_down')
+
+    def _acquire_next_frame(self, action):
+        """Returns the TimeStep give the input action.
```
Haichao-Zhang: give -> given

alf/environments/utils.py
breakds: Updated. Thanks!

alf/environments/suite_metadrive.py
breakds: It does, but in a slightly different way. Updated with an adaptive implementation.

alf/environments/suite_metadrive.py
breakds: `map_name` and `env_id` are used at this moment. Removed `batch_size`.

breakds:
Updated the PR with 

1. comments addressed
2. visualization logic
3. `seed()`
4. customization parameters
5. example PPG configuration to train on metadrive
6. documentations

PTAL. Thanks!

alf/environments/suite_metadrive.py
```diff
@@ -0,0 +1,322 @@
...
+
+    def seed(self, seed: Optional[int] = None):
+        """Reset the underlying MetaDrive environment with a specified seed.
+        MetaDrive uses a slightly different mechanism for seeds. Upon
+        construction of a MetaDrive environment, the user needs to specify a
+        seed range [start_seed, start_seed + scenario_num]. When being forced
+        to reset with a specific seed, that seed must be within the predefined
+        range.
+        Args:
+
```
Haichao-Zhang, line 230: unnecessary empty line

alf/environments/suite_metadrive.py
breakds: Thanks. Fixed.

alf/environments/suite_metadrive.py
breakds: Removed.

Haichao-Zhang (approved):
LGTM

PR #1064, opened 2021-11-12
Expanding multiple leaves at a time for MCTS

emailweixu:
MCTS is slow because it needs to do many rounds of tree expansion. This PR tries to alleviate this my doing multiple expansions simultaneously.

Also added more summaries for MCTS

Test plan:
1. mcts_algorithm_test.py
2. with muzero_pendulum_conf, got similar performance with num_parallel_sim=1 or 2
  num_parallel_sims=1: mean=157.0, std=24.0
  num_parallel_sims=2: mean=164.2, std=36.8

### DIFF
--- a/alf/algorithms/mcts_algorithm.py
+++ b/alf/algorithms/mcts_algorithm.py
@@ -217,10 +217,15 @@ class MCTSAlgorithm(OffPolicyAlgorithm):
     * search_with_exploration_policy=False. The best child is same the case k=1.
       The second best child is found by assuming the visit count of the best
       child and the parent are increased by 1 and applying the UCB criterion
-      again. This is repeated k times to get k best children. During the process
-      of traversing from the root to contruct k search paths, if the current end
-      node of several (let's say k') paths are the same, we will use best k'
-      children of that node to extend the paths.
+      again. This is repeated k times to get k best children. Note that this is
+      different from directly selecting the best k childrens based on the original
+      UCB scores. The reason of not doing that is that if the highest score is
+      much bigger than the second highest score, we want to both paths to select
+      the same child. During the process of traversing from the root to contruct
+      k search paths, if several (let's say k') paths are exactly same so far,
+      we will use best k' children of the last node of these k' paths to extend
+      the paths so that The k' children (may contains duplicates) being selected to
+      extend these k' paths are most promising according to the UCB scores.
     """
 
     def __init__(
@@ -285,7 +290,8 @@ class MCTSAlgorithm(OffPolicyAlgorithm):
                 shape as ``steps``.
             learn_policy_temperature (float): transform the policy p found by
                 MCTS by :math:`p^{1/learn_policy_temperature} / Z` as policy
-                target for model learning.
+                target for model learning, where Z is a normalization factor so
+                that the resulting probabilities sum to one.
             reward_spec (TensorSpec): a rank-1 or rank-0 tensor spec representing
                 the reward(s).
             expand_all_children (bool): If True, when a new leaf is selected, immediately
@@ -314,9 +320,11 @@ class MCTSAlgorithm(OffPolicyAlgorithm):
             learn_with_exploration_policy (bool): If True, a policy calculated
                 using reverse KL divergence will be used for learning.
             max_unroll_length (int): maximal allowed unroll steps when building
-                the search tree.
+                the search tree. If ``expand_all_children`` is False, the maximal
+                allowed tree depth will be ``max_unroll_length``. Otherwise, the
+                maximal allowed tree depth will be ``max_unroll_length-1``
             num_parallel_sims (int): expanding so many leaves at a time for one
-                tree.
+                tree. ``num_simulations`` must be divisable by ``num_parallel_sims``.
             name (str): the name of the algorithm.
         """
         assert not nest.is_nested(
@@ -338,9 +346,9 @@ class MCTSAlgorithm(OffPolicyAlgorithm):
         self._num_parallel_sims = num_parallel_sims
         self._parallel = num_parallel_sims > 1
         if expand_all_children:
-            self._max_depth = max_unroll_length - 1
+            self._max_allowed_depth = max_unroll_length - 1
         else:
-            self._max_depth = max_unroll_length
+            self._max_allowed_depth = max_unroll_length
 
         assert not expand_all_children or num_parallel_sims == 1, (
             "num_parallel_sim > 1 is not supported when expand_all_children=True"
@@ -457,7 +465,7 @@ class MCTSAlgorithm(OffPolicyAlgorithm):
             output=action, state=MCTSState(steps=state.steps + 1), info=info)
 
     def _build_tree1(self, trees, to_plays):
-        """Build the tree by evaluating one node a time"""
+        """Build the tree by evaluating one node a time."""
         roots = (trees.B, trees.root_indices)
         sim0 = 1
         if self._expand_all_root_children:
@@ -577,7 +585,7 @@ class MCTSAlgorithm(OffPolicyAlgorithm):
         return model_output
 
     def _build_tree2(self, trees, to_plays):
-        """Build the tree by evaluating all the children of one node a time"""
+        """Build the tree by evaluating all the children of one node a time."""
         # expand the root node and backup the values of the children to root.
         branch_factor = trees.branch_factor
         roots = (trees.B, trees.root_indices)
@@ -647,14 +655,17 @@ class MCTSAlgorithm(OffPolicyAlgorithm):
         best_child_index = trees.best_child_index
         game_over = trees.game_over
         psims = self._num_parallel_sims
-        parallel = self._parallel
         search_paths = []
         B = trees.B
         nodes = trees.root_indices
-        calc_i = False
-        if parallel:
+        calc_i = False  # see comment below
+        if self._parallel:
             B = B.unsqueeze(-1)
             nodes = trees.root_indices.unsqueeze(-1).expand(-1, psims)
+            # If the p-th search path of the b-th tree is at node n, it will be
+            # extended by the child indicated by best_child_index[b, n, i[p]].
+            # We only need to caculate i if not search_with_exploration_policy.
+            # See the doc string of MCTSAlgorithm for more detail.
             if self._search_with_exploration_policy:
                 i = (torch.arange(psims), )
             else:
@@ -671,7 +682,7 @@ class MCTSAlgorithm(OffPolicyAlgorithm):
         else:
             done = torch.zeros_like(nodes, dtype=torch.bool)
         depth = 0
-        while not torch.all(done) and depth < self._max_depth:
+        while not torch.all(done) and depth < self._max_allowed_depth:
             depth += 1
             if calc_i:
                 # If several nodes are same, we should select actions from different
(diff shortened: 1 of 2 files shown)


alf/algorithms/mcts_algorithm.py
```diff
@@ -279,6 +313,11 @@ def __init__(
                 using reverse KL divergence will be used for tree search.
             learn_with_exploration_policy (bool): If True, a policy calculated
                 using reverse KL divergence will be used for learning.
+            max_unroll_length (int): maximal allowed unroll steps when building
+                the search tree.
+            num_parallel_sims (int): expanding so many leaves at a time for one
```
breakds, line 326: Suggest adding a note on that `num_simulations % num_parallel_sims = 0`. This also implies a bit on how `num_parallel_sims` is being used and can help the reader, and guide the user to set proper value.

alf/algorithms/mcts_algorithm.py
```diff
@@ -252,6 +283,9 @@ def __init__(
                 where ``steps`` is a vector representing the number of steps in
                 the games. And it is expected to return a float vector of the same
                 shape as ``steps``.
+            learn_policy_temperature (float): transform the policy p found by
+                MCTS by :math:`p^{1/learn_policy_temperature} / Z` as policy
```
breakds, line 292: It does not immediately occur to me that `Z` is a normalization factor, probably worth noting it.

alf/algorithms/mcts_algorithm.py
```diff
@@ -194,6 +200,27 @@ class MCTSAlgorithm(OffPolicyAlgorithm):
...
+      When contructing the search paths, the i-th search path is based on the
+      i-th best child of each node.
+
+    * search_with_exploration_policy=False. The best child is same the case k=1.
+      The second best child is found by assuming the visit count of the best
+      child and the parent are increased by 1 and applying the UCB criterion
+      again. This is repeated k times to get k best children. During the process
+      of traversing from the root to contruct k search paths, if the current end
+      node of several (let's say k') paths are the same, we will use best k'
+      children of that node to extend the paths.
```
breakds: When I reading this paragraph without reading the code yet, I had a few questions:

1. Is this procedure seems "not parallel" across `k`?
2. "second best" is a bit ambiguous here, it can mean "the one that is only worse to the best one" or "the best that is picked in the second sampling".
3. Had some hard time understanding this "if the current end node of several (let's say k') paths are the same, we will use best k' children of that node to extend the paths". Some Ascii graph might help. Also I am confused whether this behavior is
   * a special treatment or,
   * is implied from the procedure but highlighted

alf/algorithms/mcts_algorithm.py
```diff
@@ -279,6 +313,11 @@ def __init__(
                 using reverse KL divergence will be used for tree search.
             learn_with_exploration_policy (bool): If True, a policy calculated
                 using reverse KL divergence will be used for learning.
+            max_unroll_length (int): maximal allowed unroll steps when building
+                the search tree.
```
breakds: Would be better if we state the connection between `max_unroll_length` and the maximum tree depth. This would help clarify to the reader how it affects the tree search.

alf/algorithms/mcts_algorithm.py
```diff
@@ -453,9 +586,13 @@ def _build_tree2(self, trees, to_plays):
         trees.value_sum[roots] += branch_factor * v
         self._update_best_child(trees, roots)
 
+        max_depth = torch.zeros(trees.B.shape[0], dtype=torch.int64)
```
breakds, line 597: There is already `self._max_depth` and this naming seems to have a conflict. Would something like `longest_path` (and `avg_path_length`) or `deepest` work?

alf/algorithms/mcts_algorithm.py
```diff
@@ -505,19 +646,59 @@ def _search(self, trees, to_plays):
         children_index = trees.children_index
         best_child_index = trees.best_child_index
         game_over = trees.game_over
+        psims = self._num_parallel_sims
+        parallel = self._parallel
```
breakds: This shorthand seems unnecessary.

alf/algorithms/mcts_algorithm.py
```diff
@@ -505,19 +646,59 @@ def _search(self, trees, to_plays):
...
         search_paths = []
         B = trees.B
         nodes = trees.root_indices
-        path_lengths = torch.ones_like(B)
+        calc_i = False
```
breakds: I think it is worth explaining what `i` is here, since this part of code is already hundreds lines away from where "parallel sims" is explained. Or give a pointer to where "parallel sims" is explained above (or below).

alf/algorithms/mcts_algorithm.py
```diff
@@ -505,19 +646,59 @@ def _search(self, trees, to_plays):
...
+        if parallel:
+            B = B.unsqueeze(-1)
+            nodes = trees.root_indices.unsqueeze(-1).expand(-1, psims)
+            if self._search_with_exploration_policy:
+                i = (torch.arange(psims), )
+            else:
+                calc_i = True
```
breakds, line 672: In this case `i` is not initialized (I know this is not c++, but habits :) ) - Maybe we can just put the `i = ()` in the else branch right before the `if parallel` and remove the `else` branch below?

alf/algorithms/mcts_algorithm.py
```diff
@@ -505,19 +646,59 @@ def _search(self, trees, to_plays):
...
+            if calc_i:
+                # If several nodes are same, we should select actions from different
+                # entries in best_child_index. So we need to find out which entry
+                # of best_child_index a node should use. For search_with_exploration_policy,
+                # the entries in best_child_index are sampled independently from
+                # the calculated policy, so it is ok to use i'th best_child_index
+                # for i-th node. But for UCB based selection, we want to use
+                # the top entries in best_child_index when ever they are not used.
+                #
+                # For x>y, i[b, x, y] indicates where nodes[b, x] and nodes[b, y] are same
```
breakds: It is a bit confusing here when `i` is explained as 3 dimensional here but becomes (a tuple of) 2-d tensors after the `sum` below. Shall we use a different variable?

alf/algorithms/mcts_algorithm.py
```diff
@@ -505,19 +646,59 @@ def _search(self, trees, to_plays):
...
+                # For example, suppose nodes[0] = [2,3,2,3,3], i[0] will be
+                # [[0, 0, 0, 0, 0],
+                #  [0, 0, 0, 0, 0],
+                #  [1, 0, 0, 0, 0],
+                #  [0, 1, 0, 0, 0],
+                #  [0, 1, 0, 1, 0]]
+                # [B, psims, psims]. `None` stands for unsqueezing at that dim.
+                i = (nodes[:, :, None] == nodes[:, None, :]).tril(diagonal=-1)
+                # With the above example, i[0] is [0, 0, 1, 1, 2]
+                # [B, psims]
```
breakds, line 719: A further example on how this is affecting the `torch.where` below would be very helpful. I have to admit that this alone takes quite some time to digest.

breakds (commented):
First half of review comments.

alf/algorithms/mcts_algorithm.py
emailweixu: added.

alf/algorithms/mcts_algorithm.py
emailweixu: added.

alf/algorithms/mcts_algorithm.py
emailweixu: added.

alf/algorithms/mcts_algorithm.py
emailweixu: Added explanation for 2 and 3. 
For 1,  although the part for calculating the k children is not parallel, its time is little compared to other part (which are parallel).

alf/algorithms/mcts_algorithm.py
emailweixu: changed self._max_depth to self._max_allowed_depth

alf/algorithms/mcts_algorithm.py
emailweixu: fixed

alf/algorithms/mcts_algorithm.py
emailweixu: added.

alf/algorithms/mcts_algorithm.py
emailweixu: Doesn't seem to be a problem to me. After all, this is python. And that extra step of i=() do incur a small cost.

alf/algorithms/mcts_algorithm.py
emailweixu: Added comment about i at line 665.

alf/algorithms/mcts_algorithm.py
emailweixu: changed to `same_node`

alf/algorithms/mcts_algorithm.py
```diff
@@ -508,19 +654,63 @@ def _search(self, trees, to_plays):
...
+                # For example, suppose nodes[0] = [2,3,2,3,3], same[0] will be
+                # [[0, 0, 0, 0, 0],
+                #  [0, 0, 0, 0, 0],
+                #  [1, 0, 0, 0, 0],
+                #  [0, 1, 0, 0, 0],
+                #  [0, 1, 0, 1, 0]]
+                # [B, psims, psims]. `None` stands for unsqueezing at that dim.
+                same_node = (nodes[:, :, None] == nodes[:, None, :]).tril(
+                    diagonal=-1)
+                # With the above example, i[0] is [0, 0, 1, 1, 2]
```
breakds: I think it would be better to also explain the meaning of the result in `i` w.r.t. the example, since it took me a while to fully under stand what this is doing.

> suppose nodes[0] = [2, 3, 2, 3, 3], it means that the current 5 parallel searches on tree 0 are at tree 0's node 2, node 3, node 2, node 3 and node 3 respectively. According to our algorithm, even under USB-based selection strategy, we want the 3 searches at node 3 to go to different "next nodes" (i.e. the top 3 best children) respectively. The same goes for the 2 searches at node 2 (they should go to the top 2 best children respectively). Therefore we want to use `i` to map nodes index [2, 3, 2, 3, 3] to children index [0, 0, 1, 1, 2].
> ...

alf/algorithms/mcts_algorithm.py
```diff
@@ -420,34 +483,109 @@ def _build_tree1(self, trees, to_plays):
...
+        B = trees.B
+        if parallel:
+            B = B.unsqueeze(-1)
+            i = (torch.arange(psims), )
+        else:
+            i = ()
+        max_depth = torch.zeros(B.shape[0], dtype=torch.int64)
+        avg_depth = 0.
+        new_nodes = []
+        for sim in range(sim0, self._num_simulations + sim0, psims):
```
breakds, line 497: Suggest renaming the loop variable `sim` to `starting_sim_index`, as the name `sim` will be used below for the variable that holdes `[start_sim_index, start_sim_index + 1, ..., start_sim_index + psims - 1]`

emailweixu: For psims=1, `sim`  is still sim.

alf/algorithms/mcts_algorithm.py
emailweixu: updated

alf/algorithms/mcts_algorithm.py
breakds: Sounds good.

breakds (approved):

PR #1065, opened 2021-11-12
Fixed the bug that causes PPO performance gap

### LINKED ISSUE #1059: PPO cartpole unittest fails sometimes
Two issues:
1) the unittest is not deterministic, different runs may produce different results. One example is that the ppo_cart_pole test fails sometimes.
2) verify if there is still a performance gap after #1055

Haichao-Zhang:
This PR fixed a bug introduced in  #1055 that causes the behavior difference for PPO .

### DIFF
--- a/alf/algorithms/ppo_algorithm.py
+++ b/alf/algorithms/ppo_algorithm.py
@@ -52,7 +52,6 @@ class PPOAlgorithm(ActorCriticAlgorithm):
                 reward=alg_step.info.reward,
                 discount=alg_step.info.discount,
                 action_distribution=alg_step.info.action_distribution,
-                rollout_log_prob=alg_step.info.log_prob,
                 value=alg_step.info.value,
                 reward_weights=alg_step.info.reward_weights))
 
--- a/alf/utils/dist_utils.py
+++ b/alf/utils/dist_utils.py
@@ -853,9 +853,7 @@ def compute_log_probability(distributions, actions):
     return total_log_probs
 
 
-def rsample_action_distribution(nested_distributions,
-                                return_log_prob=False,
-                                disable_cache=True):
+def rsample_action_distribution(nested_distributions, return_log_prob=False):
     """Sample actions from distributions with reparameterization-based sampling.
 
     It uses ``Distribution.rsample()`` to do the sampling to enable backpropagation.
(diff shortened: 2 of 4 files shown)


hnyu:
So you mean cache (the more accurate result) actually led to worse results in our test case?

alf/utils/dist_utils.py
```diff
@@ -853,7 +853,9 @@ def _compute_log_prob(single_distribution, single_action):
     return total_log_probs
 
 
-def rsample_action_distribution(nested_distributions, return_log_prob=False):
+def rsample_action_distribution(nested_distributions,
+                                return_log_prob=False,
+                                disable_cache=True):
```
emailweixu: I feel we should default it to False. And config it to True for ppo_cart_pole.

Haichao-Zhang:
> So you mean cache (the more accurate result) actually led to worse results in our test case?

Still under testing. Will push more updates.

alf/utils/dist_utils.py
Haichao-Zhang: > I feel we should default it to False. And config it to True for ppo_cart_pole.

Will do. 
Current test shows that even if we disable the cache, the ppo_cartpole test can still fail, which is strange. 
I'm still doing more testing on this.

alf/algorithms/ppo_algorithm.py
```diff
@@ -52,7 +52,6 @@ def train_step(self, inputs: TimeStep, state, rollout_info):
                 reward=alg_step.info.reward,
                 discount=alg_step.info.discount,
                 action_distribution=alg_step.info.action_distribution,
-                rollout_log_prob=alg_step.info.log_prob,
```
Haichao-Zhang, line 55: Introduced a bug here in the previous change. The ``rollout_info.rollout_log_prob`` are extracted from replay buffer by the ``preprocess_experience`` function, and should not be modified here.

hnyu (approved):

alf/utils/dist_utils.py
Haichao-Zhang: Found the root cause and pushed a new commit fixing the bug.

breakds (approved):

emailweixu (approved):

PR #1071, opened 2021-11-15
Dynymic TD horizon for MuZero

emailweixu:
Implement the "dynamic horizon" trick described in paper  Mastering Atari Games with Limited Data https://arxiv.org/abs/2111.00210v1

Tested on half cheetah and is useful.

### DIFF
--- a/alf/algorithms/muzero_algorithm.py
+++ b/alf/algorithms/muzero_algorithm.py
@@ -117,7 +117,7 @@ class MuzeroAlgorithm(OffPolicyAlgorithm):
                 useful if priority replay is enabled.
             train_game_over_function (bool): whether train game over function.
             reanalyze_mcts_algorithm_ctor (Callable): will be called as
-                ``mcts_algorithm_ctor(observation_spec=?, action_spec=?, debug_summaries=?, name=?)``
+                ``reanalyze_mcts_algorithm_ctor(observation_spec=?, action_spec=?, debug_summaries=?, name=?)``
                 to construct an ``MCTSAlgorithm`` instance. If not provided,
                 ``mcts_algorithm_ctor`` will be used.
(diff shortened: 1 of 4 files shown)


alf/utils/data_buffer.py
```diff
@@ -409,6 +409,14 @@ def get_earliest_position(self, env_ids):
         """
         return self._current_pos[env_ids] - self._current_size[env_ids]
 
+    def get_current_position(self):
+        """Get the current posistion for each evnironment.
```
Haichao-Zhang: evnironment -> environment

alf/algorithms/muzero_algorithm.py
```diff
@@ -109,13 +112,24 @@ def __init__(self,
...
             train_reward_function (bool): whether train reward function. If
                 False, reward should only be given at the last step of an episode.
-            calculate_priority (bool): whether to calculate priority. This is
-                only useful if priority replay is enabled.
+            calculate_priority (bool): whether to calculate priority. If not provided,
+                will be same as ``TrainerConfig.priority_replay``. This is only
+                useful if priority replay is enabled.
             train_game_over_function (bool): whether train game over function.
+            reanalyze_mcts_algorithm_ctor (Callable): will be called as
+                ``mcts_algorithm_ctor(observation_spec=?, action_spec=?, debug_summaries=?, name=?)``
```
Haichao-Zhang: We can probably do the following minor change, although they are the same by default:
mcts_algorithm_ctor(...) -> reanalyze_mcts_algorithm_ctor(...)

alf/utils/data_buffer.py
Haichao-Zhang: posistion -> position

Haichao-Zhang (approved):

PR #1072, opened 2021-11-15
Support entropy_target for MCTSModel

emailweixu:
When training with continuous action distributions, the entropy may become very small and hurt training performance. So similar to SAC, add the entropy target to prevent policy being too deterministic.

### DIFF
--- a/alf/algorithms/mcts_models.py
+++ b/alf/algorithms/mcts_models.py
@@ -19,12 +19,14 @@ import torch
 from torch import nn
 import torch.nn.functional as F
 import torch.distributions as td
+from typing import Callable
 
 import alf
 from alf.data_structures import LossInfo, namedtuple
 from alf.networks import EncodingNetwork, StableNormalProjectionNetwork, CategoricalProjectionNetwork
 from alf.utils import dist_utils, tensor_utils, summary_utils
 from alf.utils.losses import element_wise_squared_loss
+from alf.utils.schedulers import ConstantScheduler
 
 ModelOutput = namedtuple(
     'ModelOutput',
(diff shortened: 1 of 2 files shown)


alf/algorithms/mcts_models.py
```diff
@@ -174,9 +191,15 @@ def calc_loss(self, model_output: ModelOutput,
...
-            loss = loss - self._entropy_regularization * entropy_for_gradient.sum(
+        if self._log_alpha is not None:
+            alpha = self._log_alpha.exp().detach()
+            loss = loss - alpha * (loss_scale * entropy_for_gradient).sum(
                 dim=1)
+            if self._target_entropy is not None:
+                # For some unknown reason, there are memory leaks for not using
+                # detach()
+                self._log_alpha -= self._alpha_adjust_rate * (
+                    entropy.mean() - self._target_entropy()).sign().detach()
```
Haichao-Zhang, line 202: How important is it to use ``sign`` here?

alf/algorithms/mcts_models.py
```diff
@@ -379,15 +406,19 @@ def __init__(self,
                 as input and output the prediction for (value, reward, action_distribution, game_over_logit).
             game_over_logit_thresh (float): the threshold of treating the
                 state as game over if the logit for game is greater than this.
-            entropy_regularization (float): Coefficient for entropy regularization
-                loss term for policy.
+            initial_alpha (float): initial value for the weight of entropy regulariation
```
Haichao-Zhang: regulariation -> regularization

Haichao-Zhang (approved):

alf/algorithms/mcts_models.py
emailweixu: I used sign() so the update is invariant to the magnitude of entropy, which is similar to adam.

PR #1080, opened 2021-11-20
Implement KL divergence for truncated normal distribution

breakds:
# Motivation

When KL divergence is needed in computing loss and when the projection network uses `TruncatedNormal`, an implementation of KL Divergence for `TruncatedNormal` is needed. We currently do not have it, and therefore we need to have it implemented for those cases (e.g. PPG Auxiliary Loss).

# Solution

Computed the KL Divergence closed form solution by hand and implemented it (turns out that it is **much harder** than I thought to get it right ...).

# Testing

Added an unit test to compare the solution with a numerical solution, and the result are consistent.

### DIFF
--- a/alf/utils/distributions.py
+++ b/alf/utils/distributions.py
@@ -270,6 +270,54 @@ class TruncatedNormal(TruncatedDistribution):
         super().__init__(loc, scale, lower_bound, upper_bound, NormalITS())
 
 
+@td.kl.register_kl(TruncatedNormal, TruncatedNormal)
+def _kl_truncated_normal_trucated_normal(p, q):
+    """Registered KL Divergence computation specialized for TruncatedNormal
+
+    It is closed form w.r.t. torch.erf.
+
+    """
+    assert torch.all(
+        torch.logical_and(
+            torch.isclose(p.lower_bound, q.lower_bound),
+            torch.isclose(p.upper_bound, q.upper_bound)))
+
+    delta = (p.loc - q.loc)
+    delta2 = delta * delta
+
+    sigma_p2 = p.scale * p.scale
+    # Pad sigma_q2 as it is positive will only be served as denominator
+    sigma_q2 = q.scale * q.scale + 1e-30
+
+    c1 = 0.5 * (torch.log(q.scale) - torch.log(p.scale)) + 0.25 * (
+        delta2 + sigma_p2) / sigma_q2 - 0.25
+
+    # 1 / sqrt(2 pi) = 0.3989422804014327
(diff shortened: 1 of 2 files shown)


breakds:
As requested, you can find [the deduction](https://files.breakds.org/wl/?id=T5RbjCi4tUmYLvYEEZPZgrhTZIGJZsc4) here.

alf/utils/distributions.py
```diff
@@ -267,6 +267,54 @@ def __init__(self, loc, scale, lower_bound, upper_bound):
...
+
+    It is closed form w.r.t. torch.erf.
+
+    """
+    assert torch.all(
+        torch.logical_and(
+            torch.isclose(p.lower_bound, q.lower_bound),
+            torch.isclose(p.upper_bound, q.upper_bound)))
+
+    delta = (p.loc - q.loc)
```
emailweixu: () is not necessary

alf/utils/distributions.py
```diff
@@ -267,6 +267,54 @@ def __init__(self, loc, scale, lower_bound, upper_bound):
...
+    delta2 = delta * delta
```
emailweixu: delta ** 2 might be actually faster because of special handling of **2 by pytorch (https://github.com/pytorch/pytorch/blob/master/aten/src/ATen/native/cuda/PowKernel.cu)

alf/utils/distributions.py
```diff
@@ -267,6 +267,54 @@ def __init__(self, loc, scale, lower_bound, upper_bound):
...
+    # original gaussian (but both p and q are not normalized by area_p
+    # and area_q respectively). This will be handled below.
+    before_normalization = upper - lower
+
+    # Pad area_p as it is positive will only be served as denominator
+    area_p = p._cdf_ub - p._cdf_lb + 1e-30
+    area_q = q._cdf_ub - q._cdf_lb
+
+    return (torch.log(area_q / area_p) + before_normalization / area_p).sum(
+        dim=-1)
```
emailweixu: `sum(dim=list(range(-len(self._event_shape), 0)))`

alf/utils/distributions.py
breakds: Removed.

alf/utils/distributions.py
breakds: Good to learn. Updated all square computation to `** 2`.

alf/utils/distributions.py
breakds: Good catch, thanks!

emailweixu (dismissed):

breakds:
Rebased `pytorch`.

emailweixu (approved):

PR #1082, opened 2021-11-23
[Small] Update TruncatedNormal.mode to support batches

breakds:
# Motivation

Currently we are using `torch.clamp` in computing the mode of the `TruncatedNormal` distribution. However when the distribution is batched, `torch.clamp` is limited in a sense that it requires both `upper_bound` and `lower_bound` to be a number (instead of compatible tensors).

# Solution

Although `torch.clamp` is extended in the newer version of pytorch, for pytorch 1.8 we will use `torch.where` to implement it. This is slightly more verbose but can work with batched distribution `loc`.

# Testing

1. Added unit tests and passed.
2. Greedy policy trained with truncated normal projection can be played after applying this patch.

### DIFF
--- a/alf/utils/distributions.py
+++ b/alf/utils/distributions.py
@@ -201,7 +201,11 @@ class TruncatedDistribution(td.Distribution):
     @property
     def mode(self):
         """Mode of this distribution."""
-        return self._loc.clamp(self._lower_bound, self._upper_bound)
+        result = torch.where(self._loc < self._lower_bound, self._lower_bound,
+                             self._loc)
+        result = torch.where(self._loc > self._upper_bound, self._upper_bound,
+                             result)
+        return result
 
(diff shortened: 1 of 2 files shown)


alf/utils/distributions.py
```diff
@@ -201,7 +201,11 @@ def upper_bound(self):
     @property
     def mode(self):
         """Mode of this distribution."""
-        return self._loc.clamp(self._lower_bound, self._upper_bound)
+        result = torch.where(self._loc < self._lower_bound, self._lower_bound,
```
emailweixu: using torch.maximum() and minimum() will be better

breakds: Definitely. Fixed.

emailweixu (approved):

emailweixu (approved):

PR #1085, opened 2021-11-24
Episodic metrics should create buffer based on example time step for consistency

breakds:
# Motivation

Currently the episodic metrics class requires a sample metric value to be initialized. However, when a metric value is generated, it is actually extracted from a `TimeStep`. This inconsistency can be harmful as it forces the caller to give different evidence of specification at construction time and usage time of such a metric.

This change is suggested by @hnyu from #1060 

# Solution

The initialization will also use a sample `TimeStep`. The related sub classes are modified accordingly.

# Testing

1. All unit tests passed
2. Using `ac_cart_pole` to verify that the training, evaluation and play are the same. Tensorboard shows the same curve.

tensorboard

### DIFF
--- a/alf/metrics/metrics.py
+++ b/alf/metrics/metrics.py
@@ -20,6 +20,7 @@ import torch
 
 import alf
 import alf.utils.data_buffer as db
+from alf.data_structures import TimeStep, StepType
 
 from alf.utils import common
 from . import metric
@@ -146,33 +147,35 @@ class AverageEpisodicAggregationMetric(metric.StepMetric):
                  name="AverageEpisodicAggregationMetric",
                  prefix='Metrics',
                  dtype=torch.float32,
-                 batch_size=1,
                  buffer_size=10,
-                 example_metric_value=None):
+                 example_time_step=None):
         """
         Args:
             name (str):
             prefix (str): a prefix indicating the category of the metric
             dtype (torch.dtype): dtype of metric values. Should be floating types
                 in order to be averaged.
-            batch_size (int): the number of metric values generated in parallel
             buffer_size (int): number of episodes the metric value will be averaged
                 across
-            example_metric_value (nest): an example of metric value to be summarized;
-                if ``None``, a zero scalar is used.
+            example_time_step (nest): an example of the time step where the metric
+               values are extracted from. If ``None``, a zero scalar is used as the
+               example metric value.
         """
         super(AverageEpisodicAggregationMetric, self).__init__(
             name=name, dtype=dtype, prefix=prefix)
-        if example_metric_value is None:
+        if example_time_step is None:
             example_metric_value = torch.zeros((), device='cpu')
-        self._batch_size = batch_size
+        else:
+            example_metric_value = self._extract_metric_values(
+                alf.nest.utils.convert_device(example_time_step, device='cpu'))
+        self._batch_size = alf.nest.get_nest_batch_size(example_time_step)
         self._buffer_size = buffer_size
         self._initialize(example_metric_value)
 
         # ``self._current_step`` will be set to zero for the first step, and is
         # added by one otherwise. Therefore, at the episode end, its value
         # equals to episode length - 1.
-        self._current_step = torch.zeros(batch_size, device='cpu')
+        self._current_step = torch.zeros(self._batch_size, device='cpu')
 
     def _extract_metric_values(self, time_step):
         """Extract metrics from the time step. The return can be a nest."""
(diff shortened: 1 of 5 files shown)


alf/metrics/metrics.py
```diff
@@ -145,33 +146,35 @@ def __init__(self,
...
         """
         super(AverageEpisodicAggregationMetric, self).__init__(
             name=name, dtype=dtype, prefix=prefix)
-        if example_metric_value is None:
+        if example_time_step is None:
             example_metric_value = torch.zeros((), device='cpu')
-        self._batch_size = batch_size
+        else:
+            example_metric_value = self._extract_metric_values(
+                alf.nest.utils.convert_device(example_time_step, device='cpu'))
```
hnyu: Not sure why the original example_metric_value is created on cpu. It's not necessary as I see. Only its structure (not its values) is used.

alf/metrics/metrics.py
```diff
@@ -145,33 +146,35 @@ def __init__(self,
...
+        self._batch_size = alf.nest.get_nest_batch_size(example_time_step)
```
hnyu, line 171: Do we really need the batch size as a class member?

alf/metrics/metrics.py
```diff
@@ -415,8 +400,9 @@ def __init__(self,
             name=name,
             dtype=dtype,
             prefix=prefix,
-            batch_size=batch_size,
-            buffer_size=buffer_size)
+            buffer_size=buffer_size,
+            example_time_step=TimeStep(
+                step_type=torch.full((batch_size, ), StepType.FIRST)))
```
hnyu: Maybe also pass in the example_time_step instead of creating a dummy one here? Also batch_size will no longer be needed as others. This will unify the constructors of all Metrics.

hnyu (changes requested):

alf/metrics/metrics.py
breakds: I think it is because of `AverageDicountedReturnMetric`'s `accumulated_discount`.

alf/metrics/metrics.py
breakds: It is used 2 times below. I think it does no harm to be kept here.

alf/metrics/metrics.py
hnyu: Can directly use `example_time_step.cpu()`

breakds: Fixed.

alf/metrics/metrics.py
breakds: Sounds good. Originally I was thinking about trading off consistency for simplicity, but either way is fine.

alf/metrics/metrics_test.py
```diff
@@ -137,16 +132,12 @@ def testMetric(self, metric_class, num_trajectories, expected_result,
...
+                AverageReturnMetric, AverageDiscountedReturnMetric,
+                AverageEnvInfoMetric
         ]:
-
-            metric = metric_class(
-                batch_size=2, reward_shape=(2, ) if vector_reward else ())
-        elif metric_class == AverageEnvInfoMetric:
-            metric = metric_class(
-                batch_size=2, example_env_info=dict(x=0, y=0))
+            metric = metric_class(example_time_step=trajectories[0])
```
hnyu, line 138: Don't need if elif else anymore.

alf/metrics/metrics_test.py
```diff
@@ -166,7 +157,17 @@ def testMetric(self, metric_class, num_trajectories, expected_result,
                              metric.result())
 
     def test_average_per_step(self):
-        metric = AverageDrivingMetric(batch_size=2)
+        metric = AverageDrivingMetric(
```
hnyu: Can create this metric after `trajectories` is created, using `trajectories[0]` for metric

hnyu (changes requested):

alf/metrics/metrics_test.py
breakds: Good point. Updated.

alf/metrics/metrics_test.py
breakds: Updated with `if .. else ..`. The first two branches are merged.

alf/metrics/metrics_test.py
```diff
@@ -214,6 +212,9 @@ def test_average_per_step(self):
                     },
                     'success': to_tensor([1.0, 0.0])
                 }))
+
+        metric = AverageDrivingMetric(example_time_step=trajectories[0])
```
hnyu, line 204: Should delete the metric in the beginning of this function?

breakds: Thought I removed it. Sorry.

hnyu (approved):

PR #1086, opened 2021-11-25
Allow AverageDiscountedReturnMetric to take reward transformer

emailweixu:
Sometime an RL algorithm takes tranformed reward as its input,
if we want to compare the discounted return to the value estimate
of the RL algorithm, the metric compuation also should use transformed reward.


le-horizon (dismissed):
LG

Haichao-Zhang (dismissed):

emailweixu:
> To avoid encountering similar issues in the future, would it be useful to move this to `setUp` function of `alf.TestCase`?

good point

Haichao-Zhang (approved):

le-horizon (approved):
LG

PR #1088, opened 2021-11-25
Fix priority calculation for MuzeroAlgorithm

emailweixu:
Previously the priority is calculated using value from replay buffer and target value. Should be the newly calculated value and target value, which is the value loss.

Also change reward_normalizer to reward_transformer to handle generic reward transformation.


breakds (dismissed):
LGTM.

Haichao-Zhang (approved):

PR #1091, opened 2021-11-25
Enhance explained_variance to do it for each dimension

emailweixu:
Sometimes, it's useful to calculate the explained variance for each dimension of the data. Hence adding support for this.

### DIFF
--- a/alf/utils/tensor_utils.py
+++ b/alf/utils/tensor_utils.py
@@ -115,7 +115,7 @@ def tensor_prepend_zero(x):
     return torch.cat((torch.zeros(1, *x.shape[1:], dtype=x.dtype), x))
 
 
-def explained_variance(ypred, y, valid_mask=None):
+def explained_variance(ypred, y, valid_mask=None, dim=None):
     """Computes fraction of variance that ypred explains about y.
 
     Adapted from baselines.ppo2 explained_variance()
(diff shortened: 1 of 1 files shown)


alf/utils/tensor_utils.py
```diff
@@ -130,23 +130,32 @@ def explained_variance(ypred, y, valid_mask=None):
         ypred (Tensor): prediction for y
         y (Tensor): target
         valid_mask (Tensor): an optional
+        dim (None|int): the dimension to reduce. If not provided, the explained
+            variance is calculated for dimensitions.
```
hnyu: for all dimensions

alf/utils/tensor_utils.py
```diff
@@ -253,7 +262,7 @@ def clip_by_global_norm(tensors, clip_norm, use_norm=None, in_place=False):
     assert clip_norm.ndim == 0, "clip_norm must be a scalar!"
     assert clip_norm > 0, "clip_norm must be positive!"
 
-    scale = clip_norm / torch.max(clip_norm, use_norm)
+    scale = clip_norm / use_norm
```
hnyu, line 265: This will change the behavior. Currently if clip_norm is greater than the vector norm, no scaling will be performed, which means clip_norm is the upper bound of the norm and it doesn't have to be satisfied if the vector norm is already smaller than it.

hnyu (changes requested):

alf/utils/tensor_utils.py
emailweixu: line 275 ensures that nothing happens if clip_norm > use_norm

alf/utils/tensor_utils.py
emailweixu: fixed.

hnyu (approved):

PR #1092, opened 2021-11-25
Check elementwise for _numerical_calculate_gain

emailweixu:
Current _numerical_calculate_gain only works for elementwise ops. Use gain of 1 and give a warning message if not


alf/initializers.py
```diff
@@ -21,6 +22,15 @@
 import alf.utils.math_ops as math_ops
 
 
+def _is_elementwise_op(op):
+    """Check whether ``op`` is an elementwise operation."""
```
hnyu, line 26: By elementwise, do you include operations like normalizing e.g. sum-to-one normalize?

emailweixu: Things like sum-to-one or softmax are not elementwise.

hnyu: > 

Then sum-to-one will not be correctly checked by this function? Because reshaping or not won't change the normalization result.

hnyu (dismissed):

hnyu (approved):

Issue #1096, opened 2021-11-30
Multi-GPU Training with DDP


breakds:
This is a follow-up to #913 

# Motivation

Add full support for multi-process and multi-GPU training in alf with pytorch's [DDP](https://pytorch.org/docs/stable/notes/ddp.html).

# Goals

- [x] A general DDP wrapper that can convert a function to DDP module, where calling the `forward` of the wrapped DDP module is equivalent to calling the original function, with distributed hooks added to the result #1098 
- [x] Update single machine multiple GPU training for on policy branch (e.g. standard Actor Critic) #1098 
- [x] Support single machine multiple GPU training on off policy branch (e.g. PPO) #1099 
- [ ] `unroll()` **should not** go through DDP in off-policy branch #1114 
- [ ] Investigate how reducer's bucket size affects the performance
- [ ] Compare the performance with [ZeroRedundancyOptimizer](https://pytorch.org/tutorials/recipes/zero_redundancy_optimizer.html) turned on and off

While achieving the main goals above, we should also make sure that the following specific use cases are considered.

- [x] In a composite algorithm such PPG, all the networks involved in training should be properly distributed.
- [x] If evaluation is turned on, it should only be turned on for process with rank = 0
- [ ] There can be parameters that are not updated via `backward` and `optimizer` (e.g. target updater in SAC). Make sure that the behavior is consistent with the non-distributed version.
- [ ] There can be other variables that can affect training but not synchronized via DDP (e.g. batch normalization). This can introduce inconsistency with the non-distributed version. Investigate whether such inconsistency can be a problem.
- [x] Find a reasonable way to adapt the metrics
- [x] Find a reasonable way to reinterpret the termination conditions such as `num_env_steps`
- [ ] Resolve the problem when terminated by `SIGINT`, there are defunct zombie processes left 
- [ ] Run on 4 - 8 cards on Cluster

## Blockers and Issues:
- [ ] DDP + PPG + MetaDrive with default configuration may get stuck, and the performance is really bad. However, it is verified that DDP + PPG + Procgen reaches the performance parity (and even slightly better).

breakds:
## What exactly should DDP wrap?

Essentially, a DDP module wraps **a set of parameters**, and a function `f` (as the `forward()` of the DDP module). After the wrapping, when `forward()` (i.e. `f`) is called, all the result and intermediate result that depends on the parameters will be marked and **autograd hooks** are injected for those results.

[...]

PR #1098, opened 2021-12-01
Implement @data_distributed Decorator

breakds:
# Motivation

Part of the effort to #1096 .

Originally we only "DDP"-fied the `unroll` of `RLAlgorithm`, but later we found that such DDP wrapping can be made more general. To be more specific, we want to have a generally applicable, simple and transparent utility for this. It is almost always the case that we want to wrap a method of a torch Module. Therefore, a decorator for module methods is desirable.

# Solution

Implemented the `@data_distributed` decorator that can be applied on Module (such as `RLAlogirhtm`) methods. It will convert the method to be DDP-capable. A DDP-capable method will run with DDP wrapper automatically if the module it belongs to has DDP activated. A module is DDP-acitvated if its `_ddp_activated_rank` is set to a non-negative value. For `Algorithm` and its derived classes, this can be done by calling `activate_ddp(rank)`.

Also removed `_UnrollPerformer`, and use `@data_distributed` to wrap `unroll` instead. The code should be much cleaner.

*NOTE*: this still only enables DDP for the on-policy training. Will follow up on off-policy training too, which is also based on this PR.

# Testing

Run `ac_breakout` with both DDP turned on and off. 
ddp_ac_breakout

Note that with and without DDP, the training curve almost coincides.

### DIFF
--- a/alf/algorithms/algorithm.py
+++ b/alf/algorithms/algorithm.py
@@ -216,9 +216,9 @@ class Algorithm(AlgorithmInterface):
         return self._use_rollout_state
 
     def activate_ddp(self, rank: int):
-        """Prepare the RLAlgorithm with DistributedDataParallel wrapper
+        """Prepare the Algorithm with DistributedDataParallel wrapper
 
-        Note that RLAlgorithm does not need to remember the rank of the device.
+        Note that Algorithm does not need to remember the rank of the device.
 
         Args:
             rank (int): DDP wrapper needs to know on which GPU device this
--- a/alf/algorithms/algorithm.py
+++ b/alf/algorithms/algorithm.py
@@ -133,6 +133,8 @@ class Algorithm(AlgorithmInterface):
         self._metrics = []
         self._replay_buffer = None
 
+        self._ddp_activated_rank = -1
+
         # These 3 parameters are only set when ``set_replay_buffer()`` is called.
         self._replay_buffer_num_envs = None
         self._replay_buffer_max_length = None
@@ -212,6 +214,17 @@ class Algorithm(AlgorithmInterface):
         be a subset of the ``rollout_state_spec``.
         """
         return self._use_rollout_state
+
+    def activate_ddp(self, rank: int):
+        """Prepare the RLAlgorithm with DistributedDataParallel wrapper
+
+        Note that RLAlgorithm does not need to remember the rank of the device.
+
+        Args:
+            rank (int): DDP wrapper needs to know on which GPU device this
+                module's parameters and buffers are supposed to be.
+        """
+        self._ddp_activated_rank = rank
 
     @use_rollout_state.setter
     def use_rollout_state(self, flag):
--- a/alf/utils/distributed.py
+++ b/alf/utils/distributed.py
@@ -1,0 +1,132 @@
+# Copyright (c) 2021 Horizon Robotics and ALF Contributors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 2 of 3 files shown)


alf/utils/distributed.py
```diff
@@ -0,0 +1,132 @@
...
+        """
+        super().__init__()
+
+        self._wrapped_module = module  # Register and inherit the parameters
+        self._perform = functools.partial(perform, self._wrapped_module)
+
+        # DDP will panic if the wrapped module has member in its state_dict()
+        # that is not a Tensor. Here such state_dict members are picked and
+        # thrown into _ddp_params_and_buffers_to_ignore. By contract this
+        # implicitly instruct DDP wrapper to not include them in its
```
emailweixu: instruct => instructs

alf/utils/distributed.py
```diff
@@ -0,0 +1,132 @@
...
+    def forward(self, *args, **kwargs):
+        return self._perform(*args, **kwargs)
+
+
+def data_distributed(method):
+    """This decorator makes a target method of a module capable of being data
+    distributed via DDP.
+
+    This is to provide a simple and transparent way to enable DDP for specific
+    code logics.
```
emailweixu, line 71: It will be helpful for the user to use this correctly by noting what is synchronized at when

breakds: Good point. I will update the documentation.

alf/utils/distributed.py
breakds: Will fix.

breakds:
PR updated based on the comments.

alf/algorithms/algorithm.py
```diff
@@ -213,6 +215,17 @@ def use_rollout_state(self):
         """
         return self._use_rollout_state
 
+    def activate_ddp(self, rank: int):
+        """Prepare the RLAlgorithm with DistributedDataParallel wrapper
```
hnyu: Why do you call it "RLAlgorithm" when this is in the base class "Algorithm"?

alf/utils/distributed.py
```diff
@@ -0,0 +1,142 @@
...
+
+    When the method is wrapped by @data_distributed, the outputs (tensors) of
+    this method will have gradient synchronization hooks attached to them. Later
+    when those outputs are used in ``backward()`` to compute gradients, the
+    hooks will be called to synchronize across all processes. As a result, the
+    corresponding parameters does not only receive the gradients from this
```
hnyu: does not only receive -> receive not only

alf/utils/distributed.py
```diff
@@ -0,0 +1,142 @@
...
+    .. code-block:: python
+
+        class A(nn.Module):
+            # ...
+            @data_distributed
+            def compute_something(self, input):
+              return self._network1(input), self._network2(input)
+            # ...
+        variance_scaling_init(layer.weight.data,
+                              nonlinearity=nn.functional.leaky_relu)
```
hnyu: Why do you need these two lines?

alf/utils/distributed.py
```diff
@@ -0,0 +1,142 @@
...
+        for name, value in self.state_dict().items():
+            if type(value) is not torch.Tensor:
+                self._ddp_params_and_buffers_to_ignore.append(name)
+
+    def forward(self, *args, **kwargs):
+        return self._perform(*args, **kwargs)
+
+
+def data_distributed(method):
+    """This decorator makes a target method of a module capable of being data
```
hnyu, line 67: It seems that this decorator can only apply to methods of ``Algorithm``? If so, the docstring should clarify this.

hnyu (changes requested):

alf/algorithms/algorithm.py
breakds: Copy-paste error. Fixed.

alf/utils/distributed.py
breakds: Updated.

alf/utils/distributed.py
breakds: Another copy-paste error, but I did not remember how ... Thanks!

alf/utils/distributed.py
breakds: It is just for `Algorithm`, but actually all `nn.Module` derived classes. One of the prerequisite is that it needs to have `self._ddp_activated_rank`, which is stated in the docstring as a contract.

hnyu (approved):

PR #1099, opened 2021-12-01
Enable DDP for Off-Policy training branch

breakds:
# Motivation

Previously DDP (data distributed parallel) is only enabled for the on-policy training branch (i.e. `train_from_unroll()`). However, most of the algorithms actually runs on the off-policy training branch (i.e. `train_from_replay_buffer()`)  which does not enjoy DDP yet.

# Solution

After #1098 we are ready to make this work for non-composite off-policy branch training, (e.g. standard PPO). According to [this](https://github.com/HorizonRobotics/alf/issues/1096#issuecomment-982180390), we need wrap the computation that involves all trainable parameters. Therefore, the change is mainly:

1. Put calls to `train_step()` (i.e. the `collection_train_info...()`) and computation of loss into a standalone method of `Algorithm`, called `_compute_train_info_and_loss_info`.
2. Wrap `_compute_train_info_and_loss_info` with `@data_distributed`.

Also, a special treatment is done to ignore replay buffer from participating DDP's synchronization.

# Testing

Tested with

```
python -m alf.bin.train --conf alf/examples/ppo_procgen/bossfight_conf.py --root_dir ~/tmp/alf_sessions/ppo_procgen/ddp --distributed multi-gpu
```

(Note that `mini_batch_size` is halved)

Still under training but the throughput is almost doubled (13000 vs 7000) on a machine with two 3080s.

Also verified on `ppo_cart_pole` as a small experiment, see [here](https://github.com/HorizonRobotics/alf/issues/1096#issuecomment-984014459).

### DIFF
--- a/alf/utils/distributed.py
+++ b/alf/utils/distributed.py
@@ -68,18 +68,18 @@ class _MethodPerformer(torch.nn.Module):
         # Those buffers are not registered in the state_dict() because of Alf's
         # special treatment but can be found under named_buffers(). We do not
         # want DDP to synchronize replay buffers.
-        ignored_named_buffers = []
+        ignored_named_buffers = set()
         for sub_module in module.modules():
             if isinstance(sub_module, ReplayBuffer):
                 for _, buf in sub_module.named_buffers():
                     # Find all the buffers that are registered under a
                     # ReplayBuffer submodule.
-                    ignored_named_buffers.append(buf)
+                    ignored_named_buffers.add(buf)
 
         for name, buf in self.named_buffers():
             # If the buffer is in the ignored_named_buffers (address-wise equal,
             # i.e. ``is``), add its name to DDP's ignore list.
-            if any([buf is x for x in ignored_named_buffers]):
+            if buf in ignored_named_buffers:
                 self._ddp_params_and_buffers_to_ignore.append(name)
 
     def forward(self, *args, **kwargs):
--- a/alf/utils/distributed.py
+++ b/alf/utils/distributed.py
@@ -18,6 +18,8 @@ import functools
 import torch
 from torch.nn.parallel import DistributedDataParallel as DDP
 
+from alf.experience_replayers.replay_buffer import ReplayBuffer
+
 
 class _MethodPerformer(torch.nn.Module):
     """A nn.Module wrapper whose forward() performs a specified method of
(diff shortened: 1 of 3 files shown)


alf/utils/distributed.py
```diff
@@ -59,6 +59,16 @@ def __init__(self, module: torch.nn.Module, perform: Callable[..., Any]):
             if type(value) is not torch.Tensor:
                 self._ddp_params_and_buffers_to_ignore.append(name)
 
+        # We also need to ignore all the buffers that is under the replay buffer
+        # of the module (e.g. when the module is an Algorithm) for DDP, because
+        # we do not want DDP to synchronize replay buffers across processes.
+        # Those buffers are not registered in the state_dict() because of Alf's
+        # special treatment but can be found under named_buffers(). We do not
+        # want DDP to synchronize replay buffers.
+        for name, _ in self.named_buffers():
```
emailweixu: This seems a little unreliable to me. How about something like the following:
```python
for module in self.modules():
    if isinstance(module, ReplayBuffer):
       ignore buffers in module
```

breakds: Updated the PR so that it hunts for `ReplayBuffer` module and ignores the names of all the buffers that is under the `ReplayBuffer` module. Tested that the undesired buffers are correctly removed.

breakds:
Addressed comments. PTAL.

alf/utils/distributed.py
```diff
@@ -62,11 +64,22 @@ def __init__(self, module: torch.nn.Module, perform: Callable[..., Any]):
...
+            if isinstance(sub_module, ReplayBuffer):
+                for _, buf in sub_module.named_buffers():
+                    # Find all the buffers that are registered under a
+                    # ReplayBuffer submodule.
+                    ignored_named_buffers.append(buf)
+
+        for name, buf in self.named_buffers():
+            # If the buffer is in the ignored_named_buffers (address-wise equal,
+            # i.e. ``is``), add its name to DDP's ignore list.
+            if any([buf is x for x in ignored_named_buffers]):
```
emailweixu: you can make ignore_named_buffers a set and here will be `if x in ignored_named_buffers`

breakds: Agree and that was how I originally implement this, then I realized `in set` does not explicitly check for address equality, and I would like make sure to use `is` for address equality here.

breakds: I think there might be ways to make `set` use address (a.k.a `id`) to check for equality, but that would be at the cost of readability ... so I compromised and chose to use this.

breakds: Turns out I was over-thinking. I verified that `set` actually test membership in the desired way. Updated the PR as suggested.

emailweixu (approved):

PR #1100, opened 2021-12-02
AutoShapeImageDeconvNetwork

Haichao-Zhang:
Image Deconv Network with auto-shape inference.

Instead of specifying an initial start shape for image deconv as in ``ImageDecodingNetwork``, this class only needs to specify the desired output shape for the image and will automatically calculate the desired shape to start decoding based on the specified parameters for the deconv layers.

Therefore, it has different usage scenarios compared to ``ImageDecodingNetwork``.

### DIFF
--- a/alf/networks/encoding_networks.py
+++ b/alf/networks/encoding_networks.py
@@ -231,6 +231,149 @@ class ImageDecodingNetwork(_Sequential):
 
 
 @alf.configurable
+class ImageDeconvNetwork(_Sequential):
+    """
+    A general template class for creating image deconv (transposed convolutional)
+        networks with auto-shape inference.
+    """
+
+    def __init__(self,
+                 input_size,
+                 transconv_layer_params,
+                 output_shape,
+                 start_decoding_channels,
+                 preprocess_fc_layer_params=None,
+                 activation=torch.relu_,
+                 kernel_initializer=None,
+                 output_activation=torch.tanh,
+                 name="ImageDeconvNetwork"):
+        """
+        Instead of specifying an initial start shape for image deconv, this
+        class only needs to specify the desired output shape for the image
+        and will automatically calculate the desired shape to start
+        decoding based on the specified ``transconv_layer_params``
+        and uses a FC layer to map the to the desired start shape.
+
+        Args:
+            input_size (int): the size of the input latent vector
+            transconv_layer_params (tuple[tuple]): a non-empty
+                tuple of tuple (num_filters, kernel_size, strides, padding),
+                where ``padding`` is optional.
+            output_shape (tuple): the complete output size would be
+                output_shape = (c, h, w).
+            start_decoding_channels (int): the initial number of channels we'd
+                like to have for the feature map. Note that we always first
+                project an input latent vector into a vector of an appropriate
+                length so that it can be reshaped into (``start_decoding_channels``,
+                ``start_decoding_height``, ``start_decoding_width``).
+            preprocess_fc_layer_params (tuple[int]): a tuple of fc
+                layer units. These fc layers are used for preprocessing the
+                latent vector before transposed convolutions.
+            activation (nn.functional): activation for hidden layers
+            kernel_initializer (Callable): initializer for all the layers.
+            output_activation (nn.functional): activation for the output layer.
+                Usually our image inputs are normalized to [0, 1] or [-1, 1],
+                so this function should be ``torch.sigmoid`` or
+                ``torch.tanh``.
+            name (str):
+        """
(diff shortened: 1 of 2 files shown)


alf/networks/encoding_networks.py
```diff
@@ -230,6 +230,149 @@ def __init__(self,
...
+            transconv_layer_params (tuple[tuple]): a non-empty
+                tuple of tuple (num_filters, kernel_size, strides, padding),
+                where ``padding`` is optional.
+            output_shape (tuple): the complete output size would be
+                output_shape = (c, h, w).
+            start_decoding_channels (int): the initial number of channels we'd
+                like to have for the feature map. Note that we always first
+                project an input latent vector into a vector of an appropriate
+                length so that it can be reshaped into (``start_decoding_channels``,
+                ``start_decoding_height``, ``start_decoding_width``).
```
hnyu: There are no longer input ``start_decoding_height`` and ``start_decoding_width``.

alf/networks/encoding_networks.py
```diff
@@ -230,6 +230,149 @@ def __init__(self,
         super().__init__(nets, input_tensor_spec=input_tensor_spec, name=name)
 
 
+@alf.configurable
+class ImageDeconvNetwork(_Sequential):
```
hnyu: Suggest renaming it to a more informative name differentiating from ``ImageDecodingNetwork``, e.g., ImageDeconvAutoCalcNetwork

alf/networks/encoding_networks.py
```diff
@@ -230,6 +230,149 @@ def __init__(self,
...
+                    kernel_initializer=kernel_initializer,
+                    strides=strides,
+                    padding=padding,
+                    output_padding=output_padding))
+            in_channels = filters
+
+        super().__init__(nets, input_tensor_spec=input_tensor_spec, name=name)
+
+    def _conv_out(self, h_in, padding, kernel_size, stride):
+        return int((h_in + 2. * padding - (kernel_size - 1.) - 1.) / stride +
```
hnyu: Why not just -kernel_size?

alf/networks/encoding_networks.py
```diff
@@ -230,6 +230,149 @@ def __init__(self,
...
+                   1.)
+
+    def _output_padding(self, h_in, conv_out, padding, kernel_size, stride):
```
hnyu: Suggest just merging this function into _conv_padding_shape to avoid creating more functions. Same for _conv_out.

hnyu (changes requested):

alf/networks/encoding_networks.py
```diff
@@ -230,6 +230,149 @@ def __init__(self,
...
 
+@alf.configurable
+class ImageDeconvNetwork(_Sequential):
+    """
+    A general template class for creating image deconv (transposed convolutional)
+        networks with auto-shape inference.
+    """
+
+    def __init__(self,
+                 input_size,
```
hnyu: I think now we start to use type hints gradually in the code?

alf/networks/encoding_networks.py
Haichao-Zhang: Changed.

alf/networks/encoding_networks.py
Haichao-Zhang: Changed name to ``AutoShapeImageDeconvNetwork``.

alf/networks/encoding_networks.py
Haichao-Zhang: Added comments on them.

alf/networks/encoding_networks.py
Haichao-Zhang: Merged.

alf/networks/encoding_networks.py
Haichao-Zhang: Added

alf/networks/encoding_networks.py
```diff
@@ -230,6 +232,162 @@ def __init__(self,
...
+            kernel_size = common.tuplify2d(kernel_size)
+
+            padding = paras[3] if len(paras) > 3 else 0
+            padding = common.tuplify2d(padding)
+            conv_shape = self._calc_conv_out_shape(out_shape, padding,
+                                                   kernel_size, stride)
+            out_padding = self._calc_output_padding_shape(
+                out_shape, conv_shape, padding, kernel_size, stride)
+            out_shape = conv_shape
+            conv_shapes.append(conv_shape)
```
hnyu: No need to keep all conv shapes because only the last one will be used?

Haichao-Zhang: Good point. Removed.

hnyu (approved):

PR #1101, opened 2021-12-02
Automatically adjust the the configurations when DDP is enabled

breakds:
# Motivation

In order to make training under DDP **more transparently done**, we want to achieve that for most of the configurations you just simply add the `--distributed multi-gpu` and the 

1. Effective number of environments over all the processes
2. Effective `mini_batch_size`
3. Effective `num_env_steps` as termination condition

the same. Also we want to make sure that there is only one process running evaluation if evaluation is enabled.

# Solution

At the very beginning of the training, automatically adjust the above option values. This is done at the configuration evaluation time.

# Testing

Verified on ppo_procgen_bossfight.

### DIFF
--- a/alf/config_helpers.py
+++ b/alf/config_helpers.py
@@ -115,7 +115,8 @@ def get_reward_spec():
 _is_parsing = False
 
 
-def adjust_config_by_multi_process_divider(multi_process_divider: int = 1):
+def adjust_config_by_multi_process_divider(ddp_rank: int,
+                                           multi_process_divider: int = 1):
     """Adjust specific configuration value in multiple process settings
     Alf assumes all configuration files geared towards single process training.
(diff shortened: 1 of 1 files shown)


alf/config_helpers.py
```diff
@@ -124,8 +125,10 @@ def adjust_config_by_multi_process_divider(multi_process_divider: int = 1):
     For example, if we run 64 environments in parallel for single process
     settings, the value needs to be overriden with 16 if there are 4 identical
     processes running DDP training to achieve parity.
+
```
emailweixu, line 128: Can you also list the configs which are adjusted in doc-string?

breakds: Good point. Updated.

emailweixu (approved):

PR #1103, opened 2021-12-02
Quantile regression for DRL

runjerry:
* Add quantile regression variants to td_loss and one_step_loss

* Extend a quantile regression sac from sac


alf/networks/projection_networks.py
```diff
@@ -142,6 +142,58 @@ def forward(self, inputs, state=()):
...
+class QuantileProjectionNetwork(Network):
+    def __init__(self,
+                 input_size,
+                 output_tensor_spec,
+                 activation=math_ops.identity,
+                 projection_output_init_gain=0.1,
+                 name="QuantileProjectionNetwork"):
+        """Creates an instance of QuantileProjectionNetwork.
+
+        Currently there seems no need for this class to handle nested inputs;
```
emailweixu: comment what this network does. 
And this network is just a FC. Seems to me we don't even need it.

It seems to me that having a `QuantileCriticNetwork`, which is analogous to `CriticNetwork`, is better.

alf/utils/losses.py
```diff
@@ -43,3 +43,18 @@ def element_wise_squared_loss(x, y):
...
+    """Huber function.
+
+    Args:
+        x (Tensor): difference between the observed and predicted values
+        delta (float): the threshold at which to change between delta-scaled 
+            L1 and L2 loss, must be positive. Default value is 1.0
+    Returns:
+        Huber function (Tensor)
+    """
+    return torch.where(x.abs() <= delta, 0.5 * x**2,
```
emailweixu, line 59: We can use  `torch.nn.SmoothL1Loss` instead.

alf/algorithms/one_step_loss.py
```diff
@@ -41,3 +44,139 @@ def __init__(self,
...
+            (cdf_midpoints - (diff.detach() < 0).float())) * huber_loss
+
+        if self._sum_over_quantiles:
+            loss = loss.mean(-2).sum(-1)
+        else:
+            loss = loss.mean(-2).mean(-1)
+
+        # The shape of the loss expected by Algorith.update_with_gradient is
+        # [T, B], so we need to augment it with additional zeros.
+        loss = tensor_utils.tensor_extend_zero(loss)
```
Ji4chenLi: Shall we add a configurable option `scale_final_loss` here to determine whether to scale the loss value by T / (T - 1), i.e., 
```
""" 
Args:
      scale_final_loss (bool): Determine whether to scale the final loss with T / (T - 1).
"""
if self.scale_final_loss:
    T = loss.shape[0]
    scale = T / (T - 1)
    loss = loss * scale
```

alf/algorithms/one_step_loss.py
```diff
@@ -41,3 +44,139 @@ def __init__(self,
...
+        Returns:
+            td_target
+        """
+        if info.reward.ndim == 3:
+            # [T, B, 1]
+            assert info.reward.shape[-1] == 1, (
+                "Cannot handle multi-dimensional reward for OneStepTDQRLoss.")
+
+        value_shape = target_value.shape
+        rewards = info.reward.unsqueeze(-1).expand(value_shape)
```
emailweixu: this will be wrong if reward.ndim==3. And no need to do expand here and at the next line. The broadcast semantics (https://pytorch.org/docs/stable/notes/broadcasting.html) should be able to handle it correctly.

alf/algorithms/one_step_loss.py
```diff
@@ -41,3 +44,139 @@ def __init__(self,
...
+                can be same as ``value``.
+        Returns:
+            LossInfo: with the ``extra`` field same as ``loss``.
+        """
+        assert value.ndim == 3 and target_value.ndim == 3, (
+            "input value and target_value should have shape (T, B, N)")
+        returns = self.compute_td_target(info, target_value)
+        value = value[:-1]
+
+        cdf_midpoints = self._cdf_midpoints.view(1, 1, 1, -1)
```
emailweixu: It would be better to comment the shape of the tensor to help reading the code.

alf/algorithms/one_step_loss.py
```diff
@@ -41,3 +44,139 @@ def __init__(self,
...
+                        tensor_utils.explained_variance(v, r, mask))
+                    safe_mean_hist_summary('values' + suffix, v, mask)
+                    safe_mean_hist_summary('returns' + suffix, r, mask)
+                    safe_mean_hist_summary("quantile_td" + suffix, qtd, mask)
+
+                for i in range(value.shape[-1]):
+                    for j in range(returns.shape[-1]):
+                        suffix = f'/{j}-{i}'
+                        _summarize(value[..., i], returns[..., j],
+                                   diff[..., j, i], suffix)
```
emailweixu: For num_quantiles, this will generate 4 * 2500 summaries. Too much!
Also, value[..., i] and returns[..., j] are summarized multiple times.

Since values are quantile estimations, it would be nice if we can summarize the percent of target that is smaller than the value so that we can get an idea whether the quantiles are estimated correctly.

alf/algorithms/qrsac_algorithm.py
```diff
@@ -0,0 +1,151 @@
...
+                         action,
+                         critics_state,
+                         replica_min=True,
+                         quantile_mean=True):
+        critic_inputs = (observation, action)
+        critic_quantiles, critics_state = critic_net(
+            critic_inputs, state=critics_state)
+
+        if replica_min:
+            critic_quantiles = critic_quantiles.min(dim=1)[0]
```
Ji4chenLi: Shall we add a configurable option `min_based_on_q_mean` to enable diverse implementations of `replica_min`
```
""" 
Args:
      min_based_on_q_mean(bool): If True, use the replica of quantiles with the lowest Q mean to be crtics;
          Otherwise, use the minimum across replicas for each quantile
"""
if self.min_based_on_q_mean:
    # shape [B, replicas]
    critic_mean = critics.mean(dim=-1)
    index = torch.min(critic_mean, dim=-1)[1]
    critics = critics[torch.arange(len(index)), index]
else:
    critics = critics.min(dim=1)[0]
```

alf/algorithms/qrsac_algorithm.py
```diff
@@ -0,0 +1,151 @@
...
+            fc_layer_params=(64, 64),
+            last_layer_size=64,
+            last_activation=alf.math.identity),
+        alf.nn.QuantileProjectionNetwork(
+            input_size=64, output_tensor_spec=TensorSpec((quantile_size, ))))
+
+
+@alf.configurable
+class QrsacAlgorithm(SacAlgorithm):
+    """Quantile regression actor critic algorithm. """
```
emailweixu: Need some documentation.

emailweixu (changes requested):

alf/algorithms/qrsac_algorithm.py
```diff
@@ -0,0 +1,151 @@
...
+import alf.nest.utils as nest_utils
+from alf.networks import ActorDistributionNetwork, CriticNetwork
+from alf.networks import QNetwork
+from alf.tensor_specs import TensorSpec, BoundedTensorSpec
+from alf.utils import dist_utils
+
+
+def critic_network_cls(input_tensor_spec, quantile_size=64):
+    return alf.nn.Sequential(
+        alf.nn.EncodingNetwork(
```
Ji4chenLi: Seems like the conventionally used `CriticNetwork` has separate `EncodingNetwork` for the `observation` and `action` https://github.com/HorizonRobotics/alf/blob/de2c6a1a57e5572246bbddc75d9e1595787ac813/alf/networks/critic_networks.py#L132

Should we also follow the convention here, though we actually did not use such an network in the config file `alf/examples/qrsac_halfcheetah_conf.py`?

alf/algorithms/one_step_loss.py
```diff
@@ -41,3 +44,139 @@ def __init__(self,
...
             td_lambda=0.0,
             name=name)
+
+
+@alf.configurable
+class OneStepTDQRLoss(OneStepTDLoss):
+    """One step temporal difference quantile regression loss. """
+
+    def __init__(self,
+                 num_quantiles=50,
```
Ji4chenLi: The `num_quantiles` can be inferred from input, seems like we do not need to pre-specify it in the initialization.

alf/algorithms/qrsac_algorithm.py
```diff
@@ -0,0 +1,151 @@
...
+            actor_optimizer=actor_optimizer,
+            critic_optimizer=critic_optimizer,
+            alpha_optimizer=alpha_optimizer,
+            debug_summaries=debug_summaries,
+            name=name)
+
+        assert self._act_type == ActionType.Continuous, (
+            "Only continuous action space is supported for qrsac algorithm.")
+
+        self._num_quantiles = self._critic_networks.output_spec.shape[-1]
```
Ji4chenLi: The `self._num_quantiles` seems to be unused.

alf/algorithms/one_step_loss.py
Ji4chenLi:
```
# For QR-DQN, quantiles have a shape
# (T-1, B, n_quantiles). Thus, make self._cdf_midpoints
# broadcastable to (T-1, B, n_quantiles, n_quantiles)
```

alf/utils/losses.py
runjerry: `torch.nn.SmoothL1Loss` takes two inputs `x` and `y`, wherein our case `x` has shape (..., 1, d) and `y` has shape (..., d, 1), while it works correctly for x-y in the loss, it does give a warning of different shapes between two inputs.

alf/algorithms/one_step_loss.py
runjerry: Thanks for catching this bug. I just realize a neat way to handle the unsqueeze of my case (vs current td_loss) by just adding one line in the `one_step_discounted_return` of `value_ops.py`.

alf/algorithms/one_step_loss.py
runjerry: I feel that the `num_quantiles` is fixed throughout our calls of the `OneStepTDQRLoss`, so I want to make the `cdf_midpoints` fixed at initialization, rather than recreating it at every calls.

alf/algorithms/qrsac_algorithm.py
runjerry: Removed.

alf/algorithms/qrsac_algorithm.py
runjerry: I just want a simple default `critic_network_cls` here in case it is not specified in the conf file, I will replace it with a `QuantileCriticNetwork` as suggested by Wei.

alf/algorithms/one_step_loss.py
Ji4chenLi: That make senses. We can also do the
`self._cdf_midpoints = self._cdf_midpoints.view(1, 1, 1, -1)`
in the initialization.

And add an assertion to make sure the number of quantiles of the input is the same as the pre-defined `num_quantiles`, i.e., `assert value.shape[-1] == self._num_quantiles`

alf/networks/projection_networks.py
runjerry: Updated to have a `QuantileCriticNetowor` instead.

alf/algorithms/one_step_loss.py
runjerry: Updated to support multi-dim reward and n-step td.

alf/algorithms/one_step_loss.py
runjerry: Added.

alf/algorithms/one_step_loss.py
runjerry: I added the assertion and left the unsqueezing of cdf_midpoint here to easily incorporate the multi-dim reward case.

alf/algorithms/one_step_loss.py
runjerry: Now only summarize a mean explained cdf of target by value quantile as you suggested.

alf/algorithms/qrsac_algorithm.py
runjerry: Updated.

alf/algorithms/one_step_loss.py
runjerry: I think this is a design choice more general than the current PR. Maybe you can submit another PR for it?

alf/algorithms/qrsac_algorithm.py
runjerry: Added.

alf/algorithms/one_step_loss.py
Ji4chenLi: Sure, I can do it.

alf/utils/value_ops.py
```diff
@@ -160,6 +160,7 @@ def discounted_return(rewards, values, step_types, discounts, time_major=True):
     is_lasts = (step_types == StepType.LAST).to(dtype=torch.float32)
     is_lasts = common.expand_dims_as(is_lasts, values)
     discounts = common.expand_dims_as(discounts, values)
+    rewards = common.expand_dims_as(rewards, values)
```
Ji4chenLi, line 165: The doc also needs to be updated. We should change Line 140 to 

`values (Tensor): When it represents values, the shape is [T,B] (or [T]). When it represents quantiles of a distribution, the shape is [T, B, num_quantiles] (or [T, num_quantiles]).`

alf/algorithms/td_loss.py
```diff
@@ -198,3 +214,121 @@ def _summarize(v, r, td, suffix):
...
+class TDQRLoss(TDLoss):
+    """Temporal difference quantile regression loss. 
+    Compared to TDLoss, GAE support has not been implemented. """
+
+    def __init__(self,
+                 num_quantiles=50,
+                 gamma=0.99,
+                 td_error_loss_fn=losses.huber_function,
+                 td_lambda=1.0,
+                 sum_over_quantiles=False,
```
Ji4chenLi: I think it is better to set the default value of `sum_over_quantiles ` to be `True` as per the QR-DQN paper (the last line of Algorithm 1, https://arxiv.org/pdf/1710.10044.pdf)

Ji4chenLi (dismissed):

alf/utils/value_ops.py
runjerry: Updated.

alf/algorithms/td_loss.py
runjerry: Updated.

alf/algorithms/one_step_loss.py
```diff
@@ -41,3 +43,37 @@ def __init__(self,
...
+        """
+        Args:
+            num_quantiles (int): the number of quantiles.
+            gamma (float|list[float]): A discount factor for future rewards. For
+                multi-dim reward, this can also be a list of discounts, each
+                discount applies to a reward dim.
+            td_error_loss_fn (Callable): A function for computing the TD errors
+                loss. This function takes as input the target and the estimated
+                Q values and returns the loss for each element of the batch.
+            sum_over_quantiles (bool): Whether to sum over the quantiles.
```
emailweixu: It's not clear what the effect of sum_over_quantiles is from the comment.

alf/algorithms/one_step_loss.py
```diff
@@ -41,3 +43,37 @@ def __init__(self,
...
             name=name)
+
+
+@alf.configurable
+class OneStepTDQRLoss(TDQRLoss):
+    """One step temporal difference quantile regression loss. """
+
+    def __init__(self,
+                 num_quantiles=50,
+                 gamma=0.99,
```
emailweixu: We are switching to use type hint to denote the type for new code. So it should be:
```python
gamma: Union[float, List[float]] = 0.99
```
You need to "from typing import Union, List" first. And you can omit the type in the Args section of the doc-string.

alf/algorithms/td_loss.py
```diff
@@ -198,3 +214,121 @@ def _summarize(v, r, td, suffix):
...
+        returns = self.compute_td_target(info, target_value)
+        value = value[:-1]
+
+        # for quantile regression TD, the value has shape
+        # (T-1, B, n_quantiles) for scalar reward and
+        # (T-1, B, reward_dim, n_quantiles) for multi-dim reward.
+        # Make cdf_midpoints broadcastable to quantile TD, which has shape
+        # (T-1, B, n_quantiles, n_quantiles) for scalar reward and
+        # (T-1, B, reward_dim, n_quantiles, n_quantiles) for multi-dim reward
+        cdf_midpoints = self._cdf_midpoints.view(*([1] * value.ndim), -1)
```
emailweixu: You only need to make the shape of cdf_midpoints be [n_quantiles, 1] in order to make it broadcastable with diff.

alf/algorithms/td_loss.py
```diff
@@ -198,3 +214,121 @@ def _summarize(v, r, td, suffix):
...
+            name (str): The name of this loss.
+        """
+        super().__init__(
+            gamma=gamma,
+            td_error_loss_fn=td_error_loss_fn,
+            td_lambda=td_lambda,
+            debug_summaries=debug_summaries,
+            name=name)
+
+        self._num_quantiles = num_quantiles
```
emailweixu, line 260: assert that td_lambda  is 0 or 1.

alf/networks/critic_networks.py
```diff
@@ -299,3 +299,100 @@ def make_parallel(self, n):
         to create the parallel network.
         """
         return super().make_parallel(n, True)
+
+
+@alf.configurable
+class QuantileCriticNetwork(CriticNetwork):
```
emailweixu: Seems that this class does not add anything meaningful on top of CriticNetwork. Perhaps we can just use CriticNetwork?

alf/algorithms/one_step_loss.py
runjerry: updated.

alf/algorithms/one_step_loss.py
runjerry: updated here and other places

alf/algorithms/td_loss.py
runjerry: updated.

alf/algorithms/td_loss.py
runjerry: Updated.

alf/networks/critic_networks.py
runjerry: Agreed, updated.

alf/algorithms/qrsac_algorithm.py
```diff
@@ -0,0 +1,189 @@
...
+                         action,
+                         critics_state,
+                         replica_min: bool = True,
+                         quantile_mean: bool = True):
+        critic_inputs = (observation, action)
+        critic_quantiles, critics_state = critic_net(
+            critic_inputs, state=critics_state)
+
+        # For multi-dim reward, do:
+        #   [B, replicas * reward_dim] -> [B, replicas, reward_dim]
```
emailweixu: n_quantiles should be part of the shape here

alf/algorithms/qrsac_algorithm.py
```diff
@@ -0,0 +1,189 @@
...
+                idx = torch.min(
+                    critic_mean, dim=1)[1]  # [B] or [B, reward_dim]
+                if self.has_multidim_reward():
+                    idx = torch.repeat_interleave(
+                        idx.unsqueeze(-1), critic_quantiles.shape[-1],
+                        dim=-1)  # [B, reward_dim, n_quantiles]
+                    idx = idx.unsqueeze(1)  # [B, 1, reward_dim, n_quantiles]
+                    # [B, reward_dim, n_quantiles]
+                    critic_quantiles = torch.gather(
+                        critic_quantiles, dim=1, index=idx).squeeze()
```
emailweixu: If familiar with array indexing (https://numpy.org/doc/stable/reference/arrays.indexing.html):
the following seems more understandable:
```python
B, replicas, reward_dim = critic_mean.shape
critic_quantiles = critic_quantiles[torch.arange(B)[:, None], idx, torch.arange(reward_dim)]
```

alf/algorithms/qrsac_algorithm.py
```diff
@@ -0,0 +1,189 @@
...
+        # For multi-dim reward, do:
+        #   [B, replicas * reward_dim] -> [B, replicas, reward_dim]
+        # For scalar reward, do nothing
+        if self.has_multidim_reward():
+            remaining_shape = critic_quantiles.shape[2:]
+            critic_quantiles = critic_quantiles.reshape(
+                -1, self._num_critic_replicas, *self._reward_spec.shape,
+                *remaining_shape)
+        if replica_min:
+            if self._min_critic_by_critic_mean:
```
emailweixu, line 144: need comment about what this does

alf/algorithms/qrsac_algorithm.py
runjerry: updated.

alf/algorithms/qrsac_algorithm.py
runjerry: Thanks for the reference, updated.

alf/algorithms/qrsac_algorithm.py
runjerry: Updated.

alf/algorithms/td_loss.py
```diff
@@ -198,3 +216,125 @@ def _summarize(v, r, td, suffix):
...
+                                   diff[..., i, :, :], suffix)
+
+        huber_loss = self._td_error_loss_fn(diff)
+        loss = torch.abs(
+            (self._cdf_midpoints - (diff.detach() < 0).float())) * huber_loss
+
+        if self._sum_over_quantiles:
+            loss = loss.mean(-2).sum(-1)
+        else:
+            loss = loss.mean(-2).mean(-1)
```
emailweixu: loss.mean(dim=(-2, -1))

runjerry: updated.

emailweixu (approved):

Ji4chenLi (approved):

PR #1105, opened 2021-12-03
Self-supevised consistency loss for MuZero

emailweixu:
As reported in [Ye et. al. Mastering Atari Games with Limited Data](https://arxiv.org/abs/2111.00210), the consistency loss for latent representation is useful. This PR implements it.

Tested on pendulum, does not show significant difference, possibly because pendulum is too simple.

### DIFF
--- a/alf/algorithms/mcts_models.py
+++ b/alf/algorithms/mcts_models.py
@@ -93,27 +93,41 @@ class MCTSModel(nn.Module, metaclass=abc.ABCMeta):
                  train_game_over_function,
                  train_repr_prediction=False,
                  predict_reward_sum=False,
-                 initial_alpha=0.0,
                  value_loss_weight=1.0,
                  reward_loss_weight=1.0,
                  policy_loss_weight=1.0,
                  game_over_loss_weight=1.0,
                  repr_prediction_loss_weight=1.0,
+                 initial_alpha=0.0,
                  target_entropy=None,
                  alpha_adjust_rate=0.001,
                  debug_summaries=False,
                  name="MCTSModel"):
         """
         Args:
-            entropy_regularization (float): Coefficient for entropy regularization
-                loss term for policy.
+            representation_net (Network): the network for generating initial
+                latent representation from observation. It is called as
+                ``representation_net(observation)``.
+            dynamics_net (Network): the network for generating the next latent
+                representation given the current latent representation and action.
+                It is called as ``dynamics_net((current_latent_representation, action))``
             train_reward_function (bool): whether to predict reward
             train_game_over_function (bool): whether to predict game over
+            train_repr_prediction (bool): whether to train to predict future
+                latent representation.
+            predict_reward_sum (bool): If True, the loss for reward is the mean
+                square error between the sum of predicted reward and the sum of
+                actual reward. If False, the loss for reward is the mean square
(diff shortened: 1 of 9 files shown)


alf/layers.py
```diff
@@ -2105,6 +2105,9 @@ def __init__(self, shape):
             shape (tuple): desired shape not including the batch dimension.
```
breakds: The type hint `(tuple)` here is not precise after the change.

alf/algorithms/mcts_models.py
```diff
@@ -75,36 +75,69 @@
...
+                ``representation_net(observation)``.
+            dynamics_net (Network): the network for generating the next latent
+                representation given the current latent representation and action.
+                It is called as ``dynamics_net((current_latent_representation, action))``
             train_reward_function (bool): whether to predict reward
             train_game_over_function (bool): whether to predict game over
+            train_repr_prediction (bool): whether to train to predict future
+                latent representation.
+            predict_reward_sum (bool): If True, the loss for reward is the mean
+                square error between the sum of predicted reward and the sum of
```
breakds: Might be good to explicitly state how (across what) do we compute the `sum` of reward to help the user understand.

alf/algorithms/mcts_models.py
```diff
@@ -412,8 +518,29 @@ def __init__(self,
             alpha_adjust_rate (float): the speed to adjust alpha
             train_reward_function (bool): whether to predict reward
             train_game_over_function (bool): whether to predict game over
+            train_repr_prediction (bool): whether to train to predict future
+                latent representation.  This implements the self-supervised
+                consistency loss described in `Ye et. al. Mastering Atari Games
+                with Limited Data <https://arxiv.org/abs/2111.00210>`_. The loss
+                is ``-cosine(prediction_net(projection_net(x), projection_net(y))``,
```
breakds: Seems that there is a missing `)` right after `projection_net(x)`.

breakds (dismissed):
LGTM. Thanks for making it easier to read and follow!

alf/algorithms/muzero_algorithm.py
```diff
@@ -343,6 +363,24 @@ def preprocess_experience(self, root_inputs: TimeStep,
...
+                    observation, step_type = replay_buffer.get_field(
+                        ('observation', 'step_type'), env_ids, positions)
+                    exp = alf.data_structures.make_experience(
+                        root_inputs, AlgStep(), state=())
+                    exp = exp._replace(
+                        step_type=step_type,
+                        observation=observation,
+                        batch_info=batch_info,
+                        replay_buffer=replay_buffer)
+                    exp = self._data_transformer.transform_experience(exp)
```
Haichao-Zhang, line 381: I might mis-understood this part. 
But I'm curious why do we need to ``transform_experience`` when ``type(self._data_transformer)`` is  an ``IdentityDataTransformer``? 
We should ``transform_experience`` when it is not an ``IdentityDataTransformer``?

emailweixu: Thanks for spotting it. This is introduced when I cleaned up for PR.

Haichao-Zhang (approved):

breakds (approved):

Issue #1106, opened 2021-12-03
map_structure() fails in a SAC + HER setting


JakobThumm:
Hi,

I'm trying to run SAC together with HER.
Unfortunately, I'm getting an error
```
E1203 17:38:06.614589 140193157813440 nest.py:54] map_structure() fails for (Experience(step_type=tensor([[0, 1],
        [0, 1],
        [0, 1],
        ...,
        [1, 1],
        [1, 1],
        [1, 1]], device='cuda:0', dtype=torch.int32), reward=tensor([[ 0., -1.],
        [ 0., -1.],
        [ 0., -1.],
        ...,
        [-1., -1.],
        [-1., -1.],
        [-1., -1.]], device='cuda:0'), discount=tensor([[1., 1.],
        [1., 1.],
        [1., 1.],
        ...,
        [1., 1.],
        [1., 1.],
        [1., 1.]], device='cuda:0'), observation=OrderedDict([('achieved_goal', tensor([[[-1.2300,  0.5380,  0.0699],
         [-1.2300,  0.5380,  0.0712]],

        [[-0.9312, -0.2614,  0.0699],
         [-0.9312, -0.2614,  0.0712]],

        [[-0.3025,  0.9560,  0.0699],
         [-0.3025,  0.9560,  0.0712]],

        ...,

        [[-0.2939, -0.9577,  0.0729],
         [-0.2939, -0.9577,  0.0729]],

        [[ 1.2256, -0.0098,  0.0729],
         [ 1.2256, -0.0098,  0.0729]],

        [[-0.8353, -0.3085,  0.0729],
         [-0.8353, -0.3085,  0.0729]]], device='cuda:0')), ('desired_goal', tensor([[[ 0.5009,  1.5851,  0.0030],
         [ 0.5009,  1.5851,  0.0030]],

        [[-1.0657,  1.5558,  0.0030],
         [-1.0657,  1.5558,  0.0030]],

        [[ 0.1514,  0.4314,  0.0030],
         [ 0.1514,  0.4314,  0.0030]],

        ...,

        [[-1.6355,  0.0183,  0.0030],
         [-1.6355,  0.0183,  0.0030]],

        [[-1.3232, -0.4962,  0.0030],
         [-1.3232, -0.4962,  0.0030]],

        [[ 0.0062,  1.4117,  0.0030],
[...]

PR #1107, opened 2021-12-03
Fix her sac (also fix Hindsight data transformer order)

le-horizon:
This fixes issue #1106, as well as the old fix for hindsight data transformer order that should've been in place long ago..

### DIFF
--- a/alf/algorithms/data_transformer.py
+++ b/alf/algorithms/data_transformer.py
@@ -760,13 +760,18 @@ class HindsightExperienceTransformer(DataTransformer):
         # buffer (ReplayBuffer) is needed for access to future achieved goals.
         buffer = info.replay_buffer
         assert buffer != (), "Hindsight requires replay_buffer to be populated"
+        accessed_fields = [
+            "batch_info", "reward", self._desired_goal_field,
+            self._achieved_goal_field
+        ]
         with alf.device(buffer.device):
-            experience = alf.data_structures.clear_batch_info(experience)
-            info = info._replace(replay_buffer=())
-            experience = experience._replace(batch_info=info)
-            experience = convert_device(experience)
-            info = experience.batch_info
+            experience = alf.nest.transform_nest(
+                experience, "batch_info.replay_buffer", lambda _: ())
+            for f in accessed_fields:
+                experience = alf.nest.transform_nest(
+                    experience, f, lambda t: convert_device(t))
             result = experience
+            info = experience.batch_info
 
             env_ids = info.env_ids
             start_pos = info.positions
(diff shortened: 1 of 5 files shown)


alf/algorithms/data_transformer.py
```diff
@@ -131,7 +131,7 @@ def __init__(self, data_transformer_ctors, observation_spec):
                 assert not has_non_frame_stacker, (
                     "FrameStacker need to be the "
                     "first data transformers if it is used.")
-            else:
+            elif not isinstance(obs_trans, HindsightExperienceTransformer):
```
hnyu: What if FrameStacker and Hindsight are used at the same time? Do we need some special check?

alf/algorithms/data_transformer.py
```diff
@@ -862,4 +869,7 @@ def create_data_transformer(data_transformer_ctor, observation_spec):
     if len(data_transformer_ctor) == 1:
         return data_transformer_ctor[0](observation_spec)
 
+    if HindsightExperienceTransformer in data_transformer_ctor:
+        assert HindsightExperienceTransformer == data_transformer_ctor[0], \
+            "Hindsight relabeling should happen before all other transforms."
```
hnyu: The check here should be better merged with the FrameStacker check in `SequentialDataTransformer`

docs/notes/knowledge_base.rst
```diff
@@ -196,6 +196,34 @@ Step type      Discount | Value used          | Value          Note
...
+
+A DataTransformer takes in data from rollout or replay, does some processing and
+returns the modified data.
+
+It is a useful abstraction to organize all kinds of data processing. For example,
+``ObservationNormalizer`` normalizes input data to be zero mean and one std.
+
+However, it is important to note that ``HindsightExperienceTransformer``,
+``FrameStacker`` or any data transformer that need to access the replay buffer
+for additional data need to be before all other data transformers.
```
hnyu: "that need to access the replay buffer for additional data" this sentence is still too vague to understand.

hnyu (changes requested):

alf/algorithms/data_transformer.py
le-horizon: Good point.  Reverted this line.

We should just throw error.  We haven't thought through this case yet.

Having Hindsight before FrameStack would not work, because FrameStack would try to retrieve data from ReplayBuffer directly.  Having Hindsight after FrameStack, then we need to stack future achieved_goal steps from ReplayBuffer to form stacked desired_goal.

alf/algorithms/data_transformer.py
le-horizon: Good point.  Removed here.

docs/notes/knowledge_base.rst
le-horizon: improved a bit and added more example.

le-horizon (commented):
@hnyu thanks for the comments, done all, PTAL.

alf/algorithms/data_transformer.py
```diff
@@ -122,17 +122,24 @@ def __init__(self, data_transformer_ctors, observation_spec):
...
+                    "FrameStacker needs to be the "
+                    "first data transformer if it is used.")
+            else:
                 has_non_frame_stacker = True
+            if isinstance(obs_trans, HindsightExperienceTransformer):
+                assert not has_non_hindsight, (
+                    "HindsightExperienceTransformer needs to be the "
+                    "first data transformer if it is used.")
+            else:
+                has_non_hindsight = True
```
hnyu: The logic here is right. But can be frustrating if a user tries switching the order of the two and finds both need to be the first. Should assert explicitly that they don't co-exist.

le-horizon: Good point.  Simplified a bit.  How about now?

hnyu (approved):

PR #1109, opened 2021-12-04
OnehotCategoricalProjection

Haichao-Zhang:
This PR implements the ``OnehotCategoricalProjectionNetwork`` that provides option to use the `straight-through` gradient estimator.

Straight through gradient estimator is proposed by Bengio et al., "Estimating or Propagating Gradients Through Stochastic Neurons for Conditional Computation", 2013.

This is used in some RL algorithms such as Dreamer.

### DIFF
--- a/alf/networks/projection_networks.py
+++ b/alf/networks/projection_networks.py
@@ -765,3 +765,68 @@ class TruncatedProjectionNetwork(Network):
             upper_bound=self._action_high)
 
         return dist, state
+
+
+@alf.configurable
+class OnehotCategoricalProjectionNetwork(Network):
+    def __init__(self,
+                 input_size,
+                 action_spec,
+                 logits_init_output_factor=0.1,
+                 use_straight_through_gradient=True,
+                 name="OnehotCategoricalProjectionNetwork"):
+        """Creates a onehot categorical projection network that outputs a
+        discrete distribution over a number of classes.
+
+        An option to use the `straight-through` estimator is provided for
+        this network, which is proposed by Bengio et al., "Estimating or
+        Propagating Gradients Through Stochastic Neurons for Conditional
+        Computation", 2013.
+
+        Args:
+            input_size (int): the input vector size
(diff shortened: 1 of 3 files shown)


alf/networks/projection_networks.py
```diff
@@ -765,3 +765,68 @@ def forward(self, inputs, state=()):
             upper_bound=self._action_high)
 
         return dist, state
+
+
+@alf.configurable
+class OnehotCategoricalProjectionNetwork(Network):
```
hnyu, line 771: This seems more appropriately a transformation/wrapper on top of the existing CategoricalProjectionNetwork's output dist: 
``` python
dist = CategoricalProjectionNetwork(...)
dist1 = OnehotCategoricalTransform(dist)
```

Haichao-Zhang: > This seems more appropriately a transformation/wrapper on top of the existing CategoricalProjectionNetwork:
> dist = OnehotCategorical(Categorical)
> 

The suggested one is a possible alternative approach, but my current thought is that it is unnecessary as it creates another layer of complication for the user.

alf/utils/dist_utils.py
```diff
@@ -560,6 +560,23 @@ def _get_categorical_builder(obj: td.Categorical):
...
+        return td.OneHotCategorical, {'logits': obj.logits}
+
+
+def _get_onehot_categorical_st_builder(
+        obj: td.OneHotCategoricalStraightThrough):
+    if 'probs' in obj.__dict__ and id(obj.probs) == id(obj._param):
+        # This means that obj is constructed using probs
+        return td.OneHotCategoricalStraightThrough, {'probs': obj.probs}
+    else:
+        return td.OneHotCategoricalStraightThrough, {'logits': obj.logits}
```
hnyu: You can probably merge the three functions above.

Haichao-Zhang: Good point. Merged.

hnyu (approved):

PR #1110, opened 2021-12-04
AtariTerminalOnLifeLossWrapper

emailweixu:
Wrapper to change discount to 0 upon life loss for Atari.

This can potentially make it easier for the learning agent to recognize the
signficance of losing a life.

Some papers report the results with this enabled (e.g. arXiv:2111.00210).

### DIFF
--- a/alf/environments/alf_wrappers.py
+++ b/alf/environments/alf_wrappers.py
@@ -933,3 +933,34 @@ class DiscreteActionWrapper(AlfEnvironmentBaseWrapper):
         # action: [B, action_dim] or [action_dim]
         time_step = self._env.step(action)
         return time_step._replace(prev_action=prev_action)
+
+
+class AtariTerminalOnLifeLossWrapper(AlfEnvironmentBaseWrapper):
+    """Wrapper to change discount to 0 upon life loss for Atari.
+
+    This can potentially make it easier for the learning agent to recognize the
+    signficance of losing a life.
+
+    Some papers report the results with this enabled (e.g. arXiv:2111.00210)
+    """
+
+    def __init__(self, env):
+        """
+        Args:
+            env: ALF env to be wrapped
+            actions_num: number of values to discretize each action dim into
+        """
+        super().__init__(env)
+        self._prev_lives = torch.zeros(env.batch_size, dtype=torch.long)
+
+    def _reset(self):
+        time_step = self._env.reset()
+        self._prev_lives = time_step.env_info['ale.lives']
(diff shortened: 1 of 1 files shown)


alf/environments/alf_wrappers.py
```diff
@@ -933,3 +933,34 @@ def _step(self, action):
...
+        self._prev_lives = torch.zeros(env.batch_size, dtype=torch.long)
+
+    def _reset(self):
+        time_step = self._env.reset()
+        self._prev_lives = time_step.env_info['ale.lives']
+        return time_step
+
+    def _step(self, action):
+        time_step = self._env.step(action)
+        lives = time_step.env_info['ale.lives']
```
hnyu, line 964: By default, 'ale.lives' might not exist in the env info?

alf/environments/alf_wrappers.py
```diff
@@ -933,3 +933,34 @@ def _step(self, action):
...
+        time_step.discount[lives < self._prev_lives] = 0.
```
le-horizon: This line needs to be the following, otherwise it fails when there is only one env:
```
        if len(time_step.discount.shape) > 0:
            time_step.discount[lives < self._prev_lives] = np.zeros((), dtype=np.float32)
        elif lives < self._prev_lives:
            time_step = time_step._replace(discount=np.zeros((), dtype=np.float32))
```

alf/environments/alf_wrappers.py
```diff
@@ -933,3 +933,34 @@ def _step(self, action):
         # action: [B, action_dim] or [action_dim]
         time_step = self._env.step(action)
         return time_step._replace(prev_action=prev_action)
+
+
```
le-horizon, line 937: Would be good to have
`@alf.configurable` so gin config can also use this class.

alf/environments/alf_wrappers.py
emailweixu: gym.envs.atari.atari_env always output "ale.lives".

alf/environments/alf_wrappers.py
emailweixu: Previously, this wrapper is intended to be used as create_environment.batched_wrappers. Now I changed it to use as suite_gym.load.alf_env_wrappers.

alf/environments/alf_wrappers.py
emailweixu: added

le-horizon (approved):
LG

PR #1111, opened 2021-12-04
Correctly handle TimeStep.discount for MuzeroAlgorithm

emailweixu:
Previously, the algorithm assumes that TimeStep.discount can be 0 only at
the end of an episode. This change is to remove this assumption. The purpose
is to support terminal on life loss for Atari, where the discount is set
to 0 upon life loss even though the epsiode does not end yet.

### DIFF
--- a/alf/algorithms/muzero_algorithm.py
+++ b/alf/algorithms/muzero_algorithm.py
@@ -416,13 +416,11 @@ class MuzeroAlgorithm(OffPolicyAlgorithm):
 
         values = replay_buffer.get_field(value_field, env_ids,
                                          bootstrap_positions)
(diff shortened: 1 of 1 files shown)


Haichao-Zhang:
> discount is set to 0 upon life loss even though the episode does not end yet.

With this PR, is the effect on training similar to adding an Env wrapper to terminate the game upon life loss?

Haichao-Zhang (approved):

emailweixu:
> > discount is set to 0 upon life loss even though the episode does not end yet.
> 
> With this PR, is the effect on training similar to adding an Env wrapper to terminate the game upon life loss?

The policy will be same. But the metrics will be different.

PR #1112, opened 2021-12-04
BatchNorm support for MCTSModel

emailweixu:
For RNN, the normalization statics can be dramatically different
for different step of RNN. Hence we need to maintain different running statistics
for different step of RNN.

The following example shows how to use it, assuming `rnn` is a `Network`
which contains some alf.layers.BatchNorm layers.

```python

    prepare_rnn_batch_norm(rnn)
    rnn.set_batch_norm_max_steps(5)

    for i in range(t):
        rnn.set_batch_norm_current_step(i)
        y, state = rnn(input[i], state)
```

### DIFF
--- a/alf/algorithms/mcts_models.py
+++ b/alf/algorithms/mcts_models.py
@@ -28,6 +28,13 @@ from alf.utils import dist_utils, tensor_utils, summary_utils
 from alf.utils.losses import element_wise_squared_loss
 from alf.utils.schedulers import ConstantScheduler
 
+ModelState = namedtuple(
+    'ModelState',
+    [
+        'state',  # the actual latent state of the model
+        'step',  # the current unroll step of the model
+    ])
+
 ModelOutput = namedtuple(
     'ModelOutput',
     [
@@ -173,14 +180,16 @@ class MCTSModel(nn.Module, metaclass=abc.ABCMeta):
         """
         state = self._representation_net(observation)[0]
         if not self._handle_bn:
-            return self.prediction_model(state)
+            model_output = self.prediction_model(state)
+            return model_output._replace(
+                state=ModelState(state=model_output.state))
         else:
             self._prediction_net.set_batch_norm_current_step(0)
             model_output = self.prediction_model(state)
             batch_size = model_output.value.shape[0]
             current_steps = torch.zeros(batch_size, dtype=torch.long)
             return model_output._replace(
-                state=(model_output.state, current_steps))
+                state=ModelState(model_output.state, current_steps))
 
     def recurrent_inference(self, state, action):
         """Generate prediction given state and action.
@@ -194,7 +203,9 @@ class MCTSModel(nn.Module, metaclass=abc.ABCMeta):
         """
         if not self._handle_bn:
             new_state = self._dynamics_net((state, action))[0]
-            return self.prediction_model(new_state)
+            model_output = self.prediction_model(state)
+            return model_output._replace(
+                state=ModelState(state=model_output.state))
         else:
             state, current_steps = state
             self._prediction_net.set_batch_norm_current_step(current_steps)
@@ -203,7 +214,7 @@ class MCTSModel(nn.Module, metaclass=abc.ABCMeta):
             self._dynamics_net.set_batch_norm_current_step(current_steps)
             model_output = self.prediction_model(new_state)
             return model_output._replace(
-                state=(model_output.state, current_steps))
+                state=ModelState(model_output.state, current_steps))
 
     def calc_loss(self, model_output: ModelOutput,
                   target: ModelTarget) -> LossInfo:
@@ -287,12 +298,8 @@ class MCTSModel(nn.Module, metaclass=abc.ABCMeta):
             observation = alf.nest.map_structure(_flatten, target.observation)
             with torch.no_grad():
                 target_repr = self._representation_net(observation)[0]
-            if self._handle_bn:
-                # [B*unroll_steps, ...]
-                repr = _flatten(model_output.state[0])
-            else:
-                # [B*unroll_steps, ...]
-                repr = _flatten(model_output.state)
+            # [B*unroll_steps, ...]
+            repr = _flatten(model_output.state.state)
             # [B*unroll_steps]
             repr_loss = self.calc_repr_prediction_loss(repr, target_repr)
             # [B, unroll_steps]
--- a/alf/algorithms/mcts_models.py
+++ b/alf/algorithms/mcts_models.py
@@ -33,7 +33,8 @@ ModelState = namedtuple(
     [
         'state',  # the actual latent state of the model
         'step',  # the current unroll step of the model
-    ])
+    ],
+    default_value=())
 
 ModelOutput = namedtuple(
     'ModelOutput',
@@ -202,16 +203,16 @@ class MCTSModel(nn.Module, metaclass=abc.ABCMeta):
             ModelOutput: the prediction
         """
         if not self._handle_bn:
-            new_state = self._dynamics_net((state, action))[0]
+            state = self._dynamics_net((state.state, action))[0]
             model_output = self.prediction_model(state)
             return model_output._replace(
                 state=ModelState(state=model_output.state))
         else:
             state, current_steps = state
-            self._prediction_net.set_batch_norm_current_step(current_steps)
+            self._dynamics_net.set_batch_norm_current_step(current_steps)
             new_state = self._dynamics_net((state, action))[0]
             current_steps = current_steps + 1
-            self._dynamics_net.set_batch_norm_current_step(current_steps)
+            self._prediction_net.set_batch_norm_current_step(current_steps)
             model_output = self.prediction_model(new_state)
             return model_output._replace(
                 state=ModelState(model_output.state, current_steps))
--- a/alf/algorithms/mcts_models.py
+++ b/alf/algorithms/mcts_models.py
@@ -87,8 +87,10 @@ class MCTSModel(nn.Module, metaclass=abc.ABCMeta):
     """The interface for the model used by MCTSAlgorithm."""
 
     def __init__(self,
+                 num_unroll_steps,
                  representation_net,
                  dynamics_net,
+                 prediction_net,
                  train_reward_function,
                  train_game_over_function,
                  train_repr_prediction=False,
@@ -133,6 +135,7 @@ class MCTSModel(nn.Module, metaclass=abc.ABCMeta):
         super().__init__()
         self._representation_net = representation_net
         self._dynamics_net = dynamics_net
+        self._prediction_net = prediction_net
         self._debug_summaries = debug_summaries
         self._name = name
         self._train_reward_function = train_reward_function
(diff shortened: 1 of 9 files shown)


alf/batch_norm.py
```diff
@@ -0,0 +1,330 @@
...
+    bns = set()
+    todo = [("", module)]
+    visited = set()
+    while len(todo) > 0:
+        path, m = todo.pop()
+        if isinstance(m, (BatchNorm1d, BatchNorm2d)):
+            bns.add(m)
+        elif isinstance(m, (nn.BatchNorm1d, nn.BatchNorm2d, nn.BatchNorm3d)):
+            raise ValueError(
+                "MCTSModel does not support torch.nn.BatchNorm layer "
```
breakds: Not sure we should mention `MCTSModel`, as I assume this function should be more generally applicable.

alf/batch_norm.py
```diff
@@ -0,0 +1,330 @@
...
+    The reason is that for RNN, the normalization statics can be dramatically different
+    for different step of RNN. Hence we need to maintain different running statistics
+    for different step of RNN.
+
+    The following example shows how to use it, assuming ``rnn`` is a ``Network``
+    which contains some alf.layers.BatchNorm layers.
+
+    .. code-block:: python
+
+        prepare_rnn_batch_norm(rnn)
```
breakds, line 191: It is not clear in this example that whether the user need to run `prepare` just one time in the lifetime of the `rnn`, or once for every pass of all steps. Probably worth mentioning it.

alf/batch_norm.py
```diff
@@ -0,0 +1,330 @@
...
+        """Set max steps to keeping running statistics.
+
+        Args:
+            max_steps: the maximum steps for which the batch norm running statistics
+                are maintained.
+        """
+        self._max_steps = max_steps
+        if not self._track_running_stats:
+            return
+        for i in range(len(self._running_means), max_steps):
```
breakds, line 60: Why this starts from `len(self._running_means)` instead of `0`?

alf/batch_norm.py
```diff
@@ -0,0 +1,330 @@
...
+    The following example shows how to use it, assuming ``rnn`` is a ``Network``
+    which contains some alf.layers.BatchNorm layers.
+
+    .. code-block:: python
+
+        prepare_rnn_batch_norm(rnn)
+        rnn.set_batch_norm_max_steps(5)
+
+        for i in range(t):
+            rnn.set_batch_norm_current_step(i)
```
breakds, line 195: This example gives the reader an illusion that we can make `step` an extra parameter in `forward()`, or transparently increase the internal stateful `step` so that the caller does not have to worry about it.

Perhaps we can add another example where the `step` is a tensor and each batch in the input can be at different steps (not sure whether this is useful outside the MCTS algorithms).

alf/batch_norm.py
```diff
@@ -0,0 +1,330 @@
...
+            self._num_batches_tracked.append(torch.zeros((), dtype=torch.long))
+            self.register_buffer('_num_batches_tracked%s' % i,
+                                 self._num_batches_tracked[i])
+
+    def set_current_step(self, current_step: Union[torch.Tensor, int]):
+        """Use and/or update the running statistics at current_step for normalization.
+
+        Args:
+            current_step: the current step. If it is a Tensor, it means that
+                the current step for each sample in a batch.
```
breakds: Slightly more specific about the shape of the input `current_step`? Also the constraint on "all steps must be the same for training" can be mentioned here.

alf/batch_norm.py
```diff
@@ -0,0 +1,330 @@
...
+                self._bias,
+                # whether the mini-batch stats should be used for normalization
+                # rather than the running stats.
+                # If current_step is out of limit, we will use the running stats
+                # for max_steps - 1 to normalize the batch so that we can keep
+                # training and eval consistent.
+                not self._clamped,
+                exponential_average_factor,
+                self._eps)
+        else:
```
breakds: Document that this `else` stands for "Not during training and tracking running stats"?

alf/batch_norm.py
```diff
@@ -0,0 +1,330 @@
...
+                self._running_means, dim=0)[self._current_step]
+            running_vars = torch.stack(
+                self._running_vars, dim=0)[self._current_step]
+            k = input.ndim - 2
+            if k > 0:
+                ones = [1] * k
+                running_means = running_means.reshape(*running_means.shape,
+                                                      *ones)
+                running_vars = running_vars.reshape(*running_vars.shape, *ones)
+            y = (input - running_means) * (running_vars + self._eps).rsqrt()
```
breakds, line 163: This is a bit different from my understanding of `normalization` - is this meant to be `/ (running_vars + self._eps)` insteadof `*`?

alf/algorithms/mcts_models.py
```diff
@@ -174,8 +192,18 @@ def recurrent_inference(self, state, action):
...
+            return self.prediction_model(new_state)
+        else:
+            state, current_steps = state
+            self._prediction_net.set_batch_norm_current_step(current_steps)
+            new_state = self._dynamics_net((state, action))[0]
+            current_steps = current_steps + 1
+            self._dynamics_net.set_batch_norm_current_step(current_steps)
+            model_output = self.prediction_model(new_state)
+            return model_output._replace(
+                state=(model_output.state, current_steps))
```
breakds: Such "tuple for RNN state where the first is the hidden and the 2nd is the steps" is implicitly used in a lot of places - maybe a good indicator to create a `NamedTuple` for this so that both fields are named?

alf/batch_norm.py
```diff
@@ -0,0 +1,330 @@
...
+
+
+def prepare_rnn_batch_norm(module: nn.Module) -> bool:
+    """Prepare an RNN network ``module`` to use alf.layers.BatchNorm layers.
+
+    It will report error if any nn.BatchNorm layer is found within ``module``
+
+    Returns:
+        True if alf.layers.BatchNorm layers have been found
+        False otherwise.
```
Haichao-Zhang, line 310: It seems that there is a mis-match between the description and the current implementation. There seems to be no boolean return values.  The return values have been used in places like``mcts_models.py``:
```
        found1 = alf.layers.prepare_rnn_batch_norm(self._dynamics_net)
        found2 = alf.layers.prepare_rnn_batch_norm(self._prediction_net)
```

alf/algorithms/mcts_models.py
```diff
@@ -174,8 +192,18 @@ def recurrent_inference(self, state, action):
...
-        return self.prediction_model(new_state)
+        if not self._handle_bn:
+            new_state = self._dynamics_net((state, action))[0]
+            return self.prediction_model(new_state)
+        else:
+            state, current_steps = state
+            self._prediction_net.set_batch_norm_current_step(current_steps)
+            new_state = self._dynamics_net((state, action))[0]
+            current_steps = current_steps + 1
+            self._dynamics_net.set_batch_norm_current_step(current_steps)
```
Haichao-Zhang: Is the first step BN of ``_dynamics_net``  is not used? And the effective number of BNs is ``num_unroll_steps - 1`` for ``_dynamics_net``?

alf/batch_norm.py
emailweixu: This code was previously in MCTSModel. Fixed.

alf/batch_norm.py
emailweixu: only once. Comment added.

alf/batch_norm.py
emailweixu: This is to make it behave correctly if set_max_steps is called multiple times. Although it only needs to be called once.

alf/batch_norm.py
emailweixu: If step is kept as a state, this Layer needs to be implemented as a Network which output the state. That will make the construction of many networks quite complicated.

Added comments about `step` as a Tensor.

alf/batch_norm.py
emailweixu: added

alf/batch_norm.py
emailweixu: added.

alf/batch_norm.py
emailweixu: Thanks for spotting this. This is weird. With this bug, alf.layers.BatchNorm should behave exactly same as torch.nn.BatchNorm. I wonder why the training is so dramatically different.

alf/algorithms/mcts_models.py
emailweixu: dynamics_net will be called  `num_unroll_steps` times. prediction_net will be called `num_unroll_steps+1` times.

alf/batch_norm.py
emailweixu: rsqrt is reciprocal of sqrt.

alf/algorithms/mcts_models.py
emailweixu: good suggestion

alf/algorithms/mcts_models.py
Haichao-Zhang: Yes, I understand that the dynamics_net will be called ``num_unroll_steps`` times and therefore we have set
``self._dynamics_net.set_batch_norm_max_steps(num_unroll_steps)``.

What I meant was that there seems to be an index offset.

It seems to me that there will be a clamp issue in the end of recurrent inference.
If the recurrent inference has a total length of ``num_unroll_steps``,  then in the last step, 
when we call ``self._dynamics_net.set_batch_norm_current_step(current_steps)``, the  ``current_steps`` now has the value of ``num_unroll_steps``, which will be clamped to ``num_unroll_steps - 1`` in ``set_batch_norm_max_steps``.

On the other hand, the 0-th BN statistics for ``dynamics_net`` does not seems to be involved here (as ``current_steps`` starts from 1 because of Line213).

alf/batch_norm.py
Haichao-Zhang: This is indeed wired. I'm also curious about the reason for the observed difference.

emailweixu: I am investigating. My initial positive result with BN used a correct implementation. The bug is introduced during a refactoring to make it more general and easier to use. However, when testing the new code, the optimizer was also changed from SGD to AdamTF. Perhaps it's the use of AdamTF that made it better. I'm switching to SGD to see what happens.

alf/algorithms/mcts_models.py
emailweixu: For the first call of  `self._dynamics_net.set_batch_norm_current_step(current_steps)`, `current_step` is 0. For the last call, `current_step` is `num_unroll_steps - 1`

Haichao-Zhang (approved):
LGTM

breakds (approved):

alf/batch_norm.py
emailweixu: Did more experiments. Here is what I found:
With torch.nn.BatchNorm, the training is not stable. Using SGD with a small learning rate (0.01), the value estimation quickly diverges. Ironically, using SGD with a larger learning rate (0.1), the value does not diverge. Instead, it has big long period fluctuation.  Using Adam with learning rate 2e-4, it does not diverge and may even get reasonable performance. However, the value still fluctuates.

PR #1117, opened 2021-12-05
Ignore unused parameters when using DDP + PPG

breakds:
# Motivation

To make DDP + PPG work, even with #1116, DDP will still panic about **unused** parameters. Note that there are 2 isolated problems:

1. Parameters contributes to output, but the output is not used in computing the loss
2. Parameters that does not even contribute to the output

#1116 aims to fix No.1, and we need to instruct DDP to proactively prune the unused parameters. Though this will adds a little overhead but it is necessary for PPG.

# Solution

Add a globally configurable parameters to find and ignore unused parameters for DDP. By default it is disabled but for PPG it is explicitly enabled.

# Testing

Tested together with #1116 on ac/ppo/ppg cart pole to make sure all of them trains correct. PPG procgen is trained and reported in #1099.

### DIFF
--- a/alf/examples/ppg_conf.py
+++ b/alf/examples/ppg_conf.py
@@ -30,3 +30,5 @@ alf.config(
     algorithm_ctor=Agent,
     whole_replay_buffer_training=True,
     clear_replay_buffer=True)
+
+alf.config('make_ddp_performer', find_unused_parameters=True)
--- a/alf/utils/distributed.py
+++ b/alf/utils/distributed.py
@@ -18,6 +18,7 @@ import functools
 import torch
 from torch.nn.parallel import DistributedDataParallel as DDP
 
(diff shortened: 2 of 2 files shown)


emailweixu (dismissed):

breakds:
Rebased `pytorch`.

breakds:
See the red line below, when the auxiliary phase is turned off (effectively PPO), the getting stuck problem did no reproduce and the training dynamics seemed normal (it is not as efficient as PPG which is in general a fact we know).

ddp_ppg_metadrive_issue_1

emailweixu:
> Rebased `pytorch`.

Please avoid rebase in the future. As it's unclear what is changed. Use pull without --rebase instead.

emailweixu (approved):

PR #1120, opened 2021-12-06
less aggressive sanity check for reward functions of hindsight and en…

### LINKED ISSUE #1118: Hindsight experience transformer warns incorrectly
Hi,

I'm using HER to train the `fetchpush` task.
I get a very long warning message at every training step.
The message is created here:
https://github.com/HorizonRobotics/alf/blob/ee1e020ee732468d2f94629b0ad29feb68090075/alf/algorithms/data_transformer.py#L823-L835
If I understand this correctly, the condition checks if not all new rewards are the same as the old rewards.
Since the task of HER is to create new rewards based on new goals, this check should be exactly the opposite
```
if torch.allclose(relabeled_rewards[non_her_indices],
                              result.reward[non_her_indices]):
 ** WARN **
```
instead of 
```
if not torch.allclose(relabeled_rewards[non_her_indices],
                              result.reward[non_her_indices]):
 ** WARN **
```

The bug should be reproducible on the mujoco docker using [this gin file](https://github.com/HorizonRobotics/alf/blob/pytorch/alf/examples/her_fetchpush.

Did I get this correctly?

Best,

Jakob

le-horizon:
Fixes #1118
This change doesn't impact performance, only logging warning.

### DIFF
--- a/alf/algorithms/data_transformer_test.py
+++ b/alf/algorithms/data_transformer_test.py
@@ -226,8 +226,7 @@ class HindsightExperienceTransformerTest(ReplayBufferTest):
             data_spec=self.data_spec,
             num_environments=num_envs,
             max_length=max_length,
-            keep_episodic_info=True,
-            step_type_field="t")
+            keep_episodic_info=True)
 
         transform = HindsightExperienceTransformer(
             self.data_spec,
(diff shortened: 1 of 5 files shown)


alf/algorithms/data_transformer_test.py
```diff
@@ -226,8 +226,7 @@ def test_compute_her_future_step_distance(self, end_prob):
             data_spec=self.data_spec,
             num_environments=num_envs,
             max_length=max_length,
-            keep_episodic_info=True,
-            step_type_field="t")
```
hnyu, line 230: Why did you have this `step_type_field` in the first place?

le-horizon: So the test can have a different field name for step_type, which is no longer used due to this PR.
Just removed it.

hnyu (approved):

PR #1121, opened 2021-12-07
CriticNetwork Visualizer

Haichao-Zhang:
A function that can be used for generating the response images wrt action changes, which is useful for visualization.

### DIFF
--- a/alf/utils/visualizer.py
+++ b/alf/utils/visualizer.py
@@ -16,26 +16,39 @@
 import torch
 
 
-def critic_network_visualizer(net, observation, origin, dx, dy, H=20, W=20):
-    """Generate a batched network response image within the range specified by
-    ``origin``, ``dx``, ``dy`` as shown below:
+def critic_network_visualizer(net,
+                              observation,
+                              action_upper_left,
+                              action_upper_right,
+                              action_lower_left,
+                              H=20,
+                              W=20):
+    """Generate a batched network response image within the rectangular range
+    of actions (referred to as probing region) specified by ``action_top_left``,
+    ``action_top_right``, ``action_bottom_left`` as shown below:
 
-    ``origin`` ---> ``dx``
-        |
-        v
-      ``dy``
+    ``action_upper_left`` -----> ``action_upper_right``
+            |                           |
+            |                           |
+            v                           |
+    ``action_lower_left``---- ``action_lower_right``
+
+    where ``action_lower_right`` is computed from the three provided points
+    as the following because of the rectangular assumption:
+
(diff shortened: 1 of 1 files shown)


alf/utils/visualizer.py
```diff
@@ -0,0 +1,121 @@
...
+        data = data.cpu().numpy()
+
+        val_img = alf.summary.render.render_heatmap(name="val_img", data=data)
+
+
+    Args:
+        net (Callable): a callable that is called as``net((obsevation, actions))``
+        observation (Tensor): [B, ...]
+        origin (tensor): tensor representing the start of the probing region,
+            with the shape of [d]
```
hnyu: d -> action_dim

alf/utils/visualizer.py
```diff
@@ -0,0 +1,121 @@
...
+        dx (tensor): a tensor representing the end position of the probing
```
hnyu: If it's the end position, then probably shouldn't be called dx.

alf/utils/visualizer.py
```diff
@@ -0,0 +1,121 @@
...
+        The network response image of the shape [B, K, H, W], where K denotes
+        the dimensionality of the network output for the non-batch dimension.
+    """
+
+    batch_size = observation.shape[0]
+    # total number of anchors for probing the network
+    num_anchors = H * W
+    # expand observation from [B, ...] to [B * num_anchors, ...]
+    ext_obs = torch.repeat_interleave(observation, num_anchors, dim=0)
+    dxy = dx + dy - origin
```
hnyu, line 85: why dxy is computed in this way?

alf/utils/visualizer.py
```diff
@@ -0,0 +1,121 @@
...
+
+    # create the mini-image: [B, d, 2, 2], where d denotes the action dimensions
+    # [origin, dx]
+    # [dy,     dxy]
+    mini_image = torch.stack((torch.stack(
```
hnyu: Why mini_image has a batch size of B here?

alf/utils/visualizer.py
Haichao-Zhang: changed

alf/utils/visualizer.py
Haichao-Zhang: Now changed the names to be more informative.

alf/utils/visualizer.py
Haichao-Zhang: There should be no ``B`` here. Removed.

hnyu (approved):

PR #1124, opened 2021-12-08
Change record_time to record the average time for one global step

emailweixu:
Previously, record_time records average time for each execution of a segment of code.
However, that segment of code might be executed multiple times within one global step.
In order to be able to compare the time spent on different operations more meaningfully,
I change it to record the average time per global step.

### DIFF
--- a/alf/utils/summary_utils.py
+++ b/alf/utils/summary_utils.py
@@ -25,7 +25,7 @@ import alf
 from alf.data_structures import LossInfo
 from alf.nest import is_namedtuple, is_nested, py_map_structure_with_path, map_structure
 from alf.utils import dist_utils
-from alf.summary import should_record_summaries
+from alf.summary import should_record_summaries, get_global_counter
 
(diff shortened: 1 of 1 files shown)


alf/utils/summary_utils.py
```diff
@@ -384,12 +384,14 @@ def __exit__(self, type, value, traceback):
         if self._sync and torch.cuda.is_available():
             torch.cuda.synchronize()
         self._counter['time'] += time.time() - self._t0
-        self._counter['n'] += 1
         if should_record_summaries():
-            alf.summary.scalar(self._tag,
-                               self._counter['time'] / self._counter['n'])
-            self._counter['time'] = .0
-            self._counter['n'] = 0
+            c0 = self._counter['c0']
```
hnyu, line 388: This actually records average execution time between two summaries? If the summary interval is very large, then the counter won't be changed for a long time. Is this behavior intended?

hnyu (approved):

PR #1128, opened 2021-12-08
Implement velocity encoding for MetaDrive

breakds:
# Motivation

One of the hypotheses of why the previous baseline policy on MetaDrive drive too aggressively is that the agent does not fully understand the concept of "velocity" after the training. A natural idea is to explicit provide the velocity in the observation (BEV).

# Solution

Adding an extra channel to the BEV that encodes the most recent several steps' velocities as stripes.

# Testing

compare_velocity_encoding

In the above graph, red line is velocity encoding with step = 1, and blue line is velocity encoding with step = 5. Notably, the red line reaches the same success rate and total return compared to the orange baseline. From the learned policy, we can confidently say both velocity encoding experiments alleviates the aggressiveness problem, suggesting the effectiveness of the velocity encoding.

However, velocity encoding with step = 5 seems "too effective" and in which the agent tends to go for steering for all the "crisis". Will keep on exploring by adding steering encoding. At this moment velocity encoding will be turned on by default with step = 1.

Visualized velocity can be found in the video below (with step = 5).

https://user-images.githubusercontent.com/1111035/145258091-8d5f205d-ead4-4712-9e49-644205921cf9.mp4

### DIFF
--- a/alf/environments/suite_metadrive.py
+++ b/alf/environments/suite_metadrive.py
@@ -35,6 +35,21 @@ except ImportError:
     # create 'metadrive' as a mock to not break python argument type hints
     metadrive = Mock()
 
+# This is for configuration of encoding velocity in the BEV. See load() below
+# for details about how encoding is done.
+VelocityEncoding = ds.namedtuple(
+    'VelocityEncoding',
+    [
+        # Velocity in [current - num_steps, current] will be encoded.
+        'num_steps',
+        # Velocity (in km/h) will be divided by this number before encoded.
+        'normalization_factor',
+    ],
+    default_values={
+        'num_steps': 1,
+        'normalization_factor': 100.0,
+    })
+
 
 def _space_to_spec(space: gym.spaces.box.Box):
     # NOTE: this is meta drive specific conversion function as it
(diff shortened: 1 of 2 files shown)


hnyu:
Why not directly use velocity as a floating vector?

breakds:
> Why not directly use velocity as a floating vector?

I did not try that - I was thinking it would be better if it is in the same scale as other layers.

alf/environments/suite_metadrive.py
```diff
@@ -74,10 +94,19 @@ def __init__(self,
         self._image_channel_first = image_channel_first
         self._observation_spec = _space_to_spec(self._env.observation_space)
 
+        self._vel_encoding = vel_encoding
+        if self._vel_encoding is not None:
+            h, w, c = self._observation_spec.shape
+            self._observation_spec = BoundedTensorSpec(
```
Haichao-Zhang, line 100: An alternative approach could be to fuse at the model side (e.g. by expanding the scalar velocity to an image and concat), which has the advantage of using less memory.

breakds:
Thanks @hnyu and @Haichao-Zhang. Let me update this.

breakds:
Temporarily close the PR. Will refine and re-open later.

PR #1129, opened 2021-12-08
Update CI docker to enable CheckoutV2

Haichao-Zhang:
Ubuntu 18.04 officially supports git version 2.17 (which was the version in our CI image), but Github CI checkout@v2 requires a higher version of git (>=2.18).

With git 2.17, checkout@v2 will fail and break the workflow.

This PR update the Dockerfile for CI image to install git 2.18.

The CI image at ``horizonrobotics/alf`` has been updated with the one built from this file this morning.

### DIFF
--- a/.ci-cd/Dockerfile.cpu
+++ b/.ci-cd/Dockerfile.cpu
@@ -23,6 +23,8 @@ RUN python3.7 -m pip install --upgrade pip
 RUN ln -sf /usr/bin/python3.7 /usr/bin/python \
     && ln -sf /usr/bin/python3.7 /usr/bin/python3
 
+RUN apt install -y wget
+
 # install code style tools
 RUN pip3 install  pre-commit==1.17.0 \
     cpplint==1.4.4 \
@@ -31,8 +33,25 @@ RUN pip3 install  pre-commit==1.17.0 \
     pylint==2.3.1 \
     yapf==0.28.0
 
+# Ubuntu 18.04 officially supports git version 2.17 (through apt-get),
+# but github CI checkout@v2 requires a higher version of git (>=2.18).
+# Here we install git 2.18 from source.
 
-RUN apt update && apt install -y git
+# install some pre-requisites for building git from source
+RUN apt-get update && apt-get install -y libssl-dev libghc-zlib-dev \
+    libcurl4-gnutls-dev libexpat1-dev gettext unzip
+
+RUN mkdir -p /opt/local_git \
+    && cd /opt/local_git \
+    && wget https://github.com/git/git/archive/v2.18.0.zip -O git.zip \
+    && unzip git.zip \
+    && cd git-* \
+    && make prefix=/usr/local all \
+    && make prefix=/usr/local install
+
+# should get "git version 2.18.0"
+RUN git --version
+
 
 RUN apt install -y \
         libsm6  \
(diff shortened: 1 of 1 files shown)


.ci-cd/Dockerfile.cpu
```diff
@@ -31,8 +33,25 @@ RUN pip3 install  pre-commit==1.17.0 \
...
 
+# Ubuntu 18.04 officially supports git version 2.17 (through apt-get),
+# but github CI checkout@v2 requires a higher version of git (>=2.18).
+# Here we install git 2.18 from source.
+
+# install some pre-requisites for building git from source
+RUN apt-get update && apt-get install -y libssl-dev libghc-zlib-dev \
+    libcurl4-gnutls-dev libexpat1-dev gettext unzip
+
+RUN mkdir -p /opt/local_git \
```
hnyu, line 44: Is it possible to directly install using apt-get by adding PPA?

https://itsfoss.com/install-git-ubuntu/

Haichao-Zhang: > Is it possible to directly install using apt-get by adding PPA?
> 
> https://itsfoss.com/install-git-ubuntu/

Tried this first. Does not work.

.ci-cd/Dockerfile.cpu
```diff
@@ -31,8 +33,25 @@ RUN pip3 install  pre-commit==1.17.0 \
     pylint==2.3.1 \
     yapf==0.28.0
 
+# Ubuntu 18.04 officially supports git version 2.17 (through apt-get),
+# but github CI checkout@v2 requires a higher version of git (>=2.18).
+# Here we install git 2.18 from source.
+
+# install some pre-requisites for building git from source
```
breakds, line 40: Can do it more formally from original PPA?

```bash
add-apt-repository ppa:git-core/ppa
apt update
apt install git
```

We probably only need to add the `ppa` if there is `apt install git` somewhere.

Haichao-Zhang: > Can do it more formally from original PPA?
> 
> ```shell
> add-apt-repository ppa:git-core/ppa
> apt update
> apt install git
> ```
> 
> We probably only need to add the `ppa` if there is `apt install git` somewhere.

This is the first solution I tried, as follows:
```
# for add-apt-repository
RUN apt-get install software-properties-common -y
RUN add-apt-repository ppa:git-core/ppa && apt update -y && apt install git
```
but does not work.

Eventually will get an ``apt_pkg`` related error, which is potentially related to our self-installed python3.7 on ubuntu 18.04.
Build from source takes some time but is cleaner.

.ci-cd/Dockerfile.cpu
Haichao-Zhang: What I tried was
```
# for add-apt-repository
RUN apt-get install software-properties-common -y
RUN add-apt-repository ppa:git-core/ppa && apt update -y && apt install git
```

Eventually will get an ``apt_pkg`` related error. 
It does not work in our case potentially because we installed python 3.7 in ubuntu 18.04. 

Therefore I switched to building from source.

.ci-cd/Dockerfile.cpu
breakds: Ah okay. Sorry I don't have Ubuntu to try. Yep building from source is more consistent and deterministic, so in this case I am supporting building from source.

hnyu (approved):

breakds (approved):

PR #1132, opened 2021-12-09
[Small] Several bug fixes for MuZero

emailweixu:
Two was introduced by PR 1105: config needs to be passed to subclass init, missing SequentialDataTransformer.members()
One was introduced by PR 1112: prediction_net was created twice

Also changed prepare_rnn_batch_norm() to do warning instead of raising when torch.nn.BatchNorm is found.

### DIFF
--- a/alf/batch_norm.py
+++ b/alf/batch_norm.py
@@ -318,7 +318,7 @@ def prepare_rnn_batch_norm(module: nn.Module) -> bool:
             bns.add(m)
         elif isinstance(m, (nn.BatchNorm1d, nn.BatchNorm2d, nn.BatchNorm3d)):
             warning_once(
-                "MCTSModel may not perform well with torch.nn.BatchNorm layer "
+                "RNN may not perform well with torch.nn.BatchNorm layer "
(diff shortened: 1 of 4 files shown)


alf/batch_norm.py
```diff
@@ -317,9 +317,9 @@ def prepare_rnn_batch_norm(module: nn.Module) -> bool:
         if isinstance(m, (BatchNorm1d, BatchNorm2d)):
             bns.add(m)
         elif isinstance(m, (nn.BatchNorm1d, nn.BatchNorm2d, nn.BatchNorm3d)):
-            raise ValueError(
-                "RNN does not support torch.nn.BatchNorm layer "
-                "(at %s). Please use alf.layers.BatchNorm instead." % path)
+            warning_once(
+                "MCTSModel may not perform well with torch.nn.BatchNorm layer "
```
breakds: I think it might be more general to mention RNN instead of `MCTSModel` when making this recommendation.

emailweixu: agree.

breakds (approved):

PR #1134, opened 2021-12-10
Standalone multi_quantile_huber_loss

emailweixu:
In PR #1103 implemented quantile regression for TD learning. However, multiple quantile regression may be found useful in other scenarios. This PR implements a standalone multi quantile regression loss.

The reason that I didn't change TDQRLoss to use this is TDQRLoss used some intermediate result for summary, which is not the result of this loss. But I plan to add a standalone summary function for multi-quantile regression to make the summary more meaningful. I will update TDQRLoss At that time.

### DIFF
--- a/alf/utils/losses.py
+++ b/alf/utils/losses.py
@@ -24,12 +24,12 @@ def element_wise_huber_loss(x, y):
     """Elementwise Huber loss.
 
     Args:
-        x (Tensor): prediction
-        y (Tensor): target
+        x (Tensor): label
+        y (Tensor): prediction
     Returns:
         loss (Tensor)
     """
-    return F.smooth_l1_loss(x, y, reduction="none")
+    return F.smooth_l1_loss(y, x, reduction="none")
 
 
 @alf.configurable
@@ -37,12 +37,12 @@ def element_wise_squared_loss(x, y):
     """Elementwise squared loss.
 
     Args:
-        x (Tensor): prediction
-        y (Tensor): target
+        x (Tensor): label
+        y (Tensor): prediction
     Returns:
         loss (Tensor)
     """
-    return F.mse_loss(x, y, reduction="none")
+    return F.mse_loss(y, x, reduction="none")
 
 
 @alf.configurable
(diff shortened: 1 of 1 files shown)


alf/utils/losses.py
```diff
@@ -24,25 +24,25 @@ def element_wise_huber_loss(x, y):
...
     Args:
-        x (Tensor): label
-        y (Tensor): prediction
+        x (Tensor): prediction
+        y (Tensor): target
     Returns:
         loss (Tensor)
     """
-    return F.mse_loss(y, x, reduction="none")
+    return F.mse_loss(x, y, reduction="none")
```
runjerry: This changes the logic of inputs' order when calling this loss, I am afraid that many algorithms that call this loss will be affected?

alf/utils/losses.py
```diff
@@ -51,10 +51,50 @@ def huber_function(x: torch.Tensor, delta: float = 1.0):
...
+
+    This loss is described in the following paper:
+
+    `Dabney et. al. Distributional Reinforcement Learning with Quantile Regression
+    <https://www.aaai.org/ocs/index.php/AAAI/AAAI18/paper/viewFile/17184/16590>`_
+
+    Args:
+        quantiles: batch_shape + [num_quantiles,]
+        target: batch_shape or batch_shape + [num_targets, ]
+        delta: the smoothness parameter for huber loss (larger means smoother).
```
runjerry, line 82: Do you want to briefly describe the rationale of setting the default delta to be 0.1 rather than the commonly used 1.0?

alf/utils/losses.py
emailweixu: Because mes_loss is symmetric with respect to x and y, it won't have any real effect. This change is to make it consistent with the various loss function of pytorch, which always has prediction as its first argument and target as its second argument.

But I'll revert this change.  And change it together will the call site in a separate PR so that the argument provided by other algorithms will be consistent with the comment.

alf/utils/losses.py
emailweixu: added.

runjerry (approved):

Ji4chenLi (approved):

PR #1135, opened 2021-12-10
Fix epsilon_greedy gin config during play

le-horizon:
Hi Wei,

This PR makes epsilon_greedy configuration backward compatible with gin config.

It uses the config object passed into the algorithm class, instead of alf.get_config_value().
Using the config object allows --gin_param command line flag to override the value, and also allows play.py to correctly get the value from gin file.

This shouldn't affect alf config setting the parameter.

Does this make sense?  If so, I'll also change all other places that use alf.get_config_value().

Thanks,
Le

### DIFF
--- a/alf/algorithms/actor_critic_algorithm.py
+++ b/alf/algorithms/actor_critic_algorithm.py
@@ -96,8 +96,7 @@ class ActorCriticAlgorithm(OnPolicyAlgorithm):
             name (str): Name of this algorithm.
         """
         if epsilon_greedy is None:
-            epsilon_greedy = alf.get_config_value(
-                'TrainerConfig.epsilon_greedy', config_override=config)
+            epsilon_greedy = alf.common.get_epsilon_greedy(config)
         self._epsilon_greedy = epsilon_greedy
         actor_network = actor_network_ctor(
             input_tensor_spec=observation_spec, action_spec=action_spec)
--- a/alf/algorithms/actor_critic_algorithm.py
+++ b/alf/algorithms/actor_critic_algorithm.py
@@ -96,7 +96,7 @@ class ActorCriticAlgorithm(OnPolicyAlgorithm):
             name (str): Name of this algorithm.
         """
         if epsilon_greedy is None:
-            epsilon_greedy = alf.common.get_epsilon_greedy(config)
+            epsilon_greedy = alf.utils.common.get_epsilon_greedy(config)
         self._epsilon_greedy = epsilon_greedy
         actor_network = actor_network_ctor(
             input_tensor_spec=observation_spec, action_spec=action_spec)
--- a/alf/algorithms/actor_critic_algorithm.py
+++ b/alf/algorithms/actor_critic_algorithm.py
@@ -97,7 +97,7 @@ class ActorCriticAlgorithm(OnPolicyAlgorithm):
         """
         if epsilon_greedy is None:
             epsilon_greedy = alf.get_config_value(
-                'TrainerConfig.epsilon_greedy', override=config.epsilon_greedy)
+                'TrainerConfig.epsilon_greedy', config_override=config)
         self._epsilon_greedy = epsilon_greedy
         actor_network = actor_network_ctor(
             input_tensor_spec=observation_spec, action_spec=action_spec)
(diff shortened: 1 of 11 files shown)


emailweixu:
This makes sense to me. Now TrainerConfig is always passed to rl algorithms even if it is created by Agent.

But why do still keep using gin-config?

emailweixu (dismissed):

hnyu:
This might have problems if DDPG or SAC is used as a subalgorithm (not just under Agent) and config is not passed to them.

le-horizon:
> This might have problems if DDPG or SAC is used as a subalgorithm (not just under Agent) and config is not passed to them.

Good point @hnyu .  Fixed.  How does this look?

hnyu:
> > This might have problems if DDPG or SAC is used as a subalgorithm (not just under Agent) and config is not passed to them.
> 
> Good point @hnyu . Fixed. How does this look?

I think it's strange to provide something as input to `get_config_value` when you already have it. A simpler way is just to first check `config` and then `get_config_value` (if config is None) in DDPG and SAC.

hnyu:
> > > This might have problems if DDPG or SAC is used as a subalgorithm (not just under Agent) and config is not passed to them.
> > 
> > 
> > Good point @hnyu . Fixed. How does this look?
> 
I think it's strange to change the API of `get_config_value` and provide something as input to it when you already have it as what you want. A simpler way is just to first check `config` and then `get_config_value` (if config is None) in DDPG and SAC.

alf/algorithms/ddpg_algorithm.py
```diff
@@ -140,7 +141,8 @@ def __init__(self,
         """
         self._calculate_priority = calculate_priority
         if epsilon_greedy is None:
-            epsilon_greedy = config.epsilon_greedy if config else 0.
+            epsilon_greedy = alf.get_config_value(
+                'TrainerConfig.epsilon_greedy', override=config.epsilon_greedy)
```
hnyu: When config is None, this still throws an error of attribute 'epsilon_greedy' error.

le-horizon: You are right.  This was caught by unit test and fixed.

le-horizon:
> I think it's strange to change the API of `get_config_value` and provide something as input to it when you already have it as what you want. A simpler way is just to first check `config` and then `get_config_value` (if config is None) in DDPG and SAC.

There are ~10 algorithms with the problem.  That's why I chose to put it in a central place, in case we want to change its logic in the future.

How does this look to you?

alf/algorithms/ddpg_algorithm.py
hnyu: > You are right. This was caught by unit test and fixed.

I might not be clear about my first comment. I think the problem with this is that when a user use either SAC or DDPG as a child algorithm (it can be nested anywhere under any algorithm) and forgets to pass the `config` object to it, then accessing `epsilon_greedy` in this child algorithm will report an error.

hnyu:
> 

I think some function like `common.get_epsilon_greedy` can be defined to wrap the two lines:

```python
def get_epsilon_greedy(config):
     if config is not None:
        return config.epsilon_greedy
     else:
        return alf.get_config_value('TrainerConfig.epsilon_greedy')
```
This avoids changing the current `get_config_value` in a strange way. But more importantly, gin config will some day be no longer supported, so it's better not to modify any existing API for it.

le-horizon:
> > 
> 
> I think some function like `common.get_epsilon_greedy` can be defined to wrap the two lines:
> 
> ```python
> def get_epsilon_greedy(config):
>      if config is not None:
>         return config.epsilon_greedy
>      else:
>         return alf.get_config_value('TrainerConfig.epsilon_greedy')
> ```
> 
> This avoids changing the current `get_config_value` in a strange way. But more importantly, gin config will some day be no longer supported, so it's better not to modify any existing API for it.

Done.

hnyu (approved):

alf/algorithms/actor_critic_algorithm.py
```diff
@@ -95,8 +96,7 @@ def __init__(self,
             name (str): Name of this algorithm.
         """
         if epsilon_greedy is None:
-            epsilon_greedy = alf.get_config_value(
-                'TrainerConfig.epsilon_greedy')
+            epsilon_greedy = alf.utils.common.get_epsilon_greedy(config)
```
emailweixu, line 99: It feels strange to have a function in common just for getting epsilon_greedy.
epsilon_greedy is algorithm specific. It's better to set as the argument of the algorithm.

PR #1136, opened 2021-12-10
Residue block

emailweixu:
This is the residual block used in ResNet-18 and ResNet-34 of the original
ResNet paper [Deep residual learning for image recognition](https://arxiv.org/abs/1512.03385.) 

 Compared to BottleneckBlock, it has one less conv layer and is more often used by RL models due to its less computation load.

Tested with breakout training using MuZero

### DIFF
--- a/alf/layers.py
+++ b/alf/layers.py
@@ -2180,7 +2180,7 @@ class ResidueBlock(nn.Module):
             kernel_size: the kernel size of middle layer at main path
             filters: the number of filters of the two conv layers at main path
             stride: stride for this block.
-            transpose: a indicate using ``Conv2D`` or ``Conv2DTranspose``.
+            transpose: whether use ``Conv2D`` or ``Conv2DTranspose``.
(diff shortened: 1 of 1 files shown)


alf/layers.py
```diff
@@ -2155,6 +2155,90 @@ def _conv_transpose_2d(in_channels,
...
+                 transpose: bool = False,
+                 with_batch_normalization: bool = True,
+                 bn_ctor: Callable[[int], nn.Module] = nn.BatchNorm2d):
+        """
+        Args:
+            in_channels: the number of channels of input
+            kernel_size: the kernel size of middle layer at main path
+            filters: the number of filters of the two conv layers at main path
+            stride: stride for this block.
+            transpose: a indicate using ``Conv2D`` or ``Conv2DTranspose``.
```
Haichao-Zhang: minor: ``a indicate using``

Haichao-Zhang (dismissed):

Haichao-Zhang (approved):

PR #1144, opened 2021-12-15
ALF customized dmc_gym_wrapper and an SAC training example

hnyu:
A customized dmc->gym wrapper. Removed unnecessary code from dmc2gym repo:
1. action normalization: this can be automatically handled by ALF's NormalizedAction wrapper in `suite_gym.wrap_env`
2. frame_skip: this can be handled by ALF's FrameSkip Wrapper
3. seed: ALF will set the seed after init
4. channels_first: we always return channels-first images


emailweixu (dismissed):

hnyu:
It turns out that the original dmc2gym doesn't support seeding after init. The random seed is only set during init, after which calling ``seed()`` will not change the environment seed:

https://github.com/denisyarats/dmc2gym/blob/master/dmc2gym/wrappers.py#L135

Unfortunately, I didn't find an API to do this in dm_control. So I did a hack by creating a new environment in the ``seed()`` of dmc2gym and updated a fork. See 

https://github.com/HorizonRobotics/dmc2gym/commit/f9bd361b2b25054aea8a8858249d504c484f7722

alf/environments/suite_dmc.py
```diff
@@ -11,121 +11,91 @@
...
-        return step_result
-
-    def get_info(self):
-        return {}
-
-    def env_info_spec(self):
-        return {}
+    def __init__(self, env):
+        super().__init__(env)
+        self.metadata.update(self._metadata)
```
emailweixu: Can this be merged with DMCGYMWrapper?

alf/environments/suite_dmc.py
```diff
@@ -11,121 +11,79 @@
...
-        visualize_reward=False,
+        visualize_reward=visualize_reward,
         from_pixels=from_pixels,
-        # The episode_length passed to dmc2gym is doubled because we want to
+        # The episode_length passed to DMCGYMWrapper is doubled because we want to
         # make sure that the reset is triggered only in alf_wrappers.TimeLimit.
         episode_length=max_episode_steps * 2,
         height=image_size,
-        width=image_size,
-        frame_skip=action_repeat)
```
emailweixu, line 120: episode_length may not be large enough if using FrameSkip wrapper to replace this. episode_length needs to be bigger than max_episode_steps * action_repeat

hnyu: Now always pass into episode_length=float('inf')

runjerry:
Looks good to me.

alf/environments/suite_dmc.py
```diff
@@ -11,121 +11,90 @@
...
+            this is greater than the agent's primitive physics timestep, then
+            multiple physics simulation steps might be performed between two actions.
+            The difference between multi-physics steps and "action repeats"/FrameSkip
+            is that the intermediate physics step won't need to render an observation
+            (which might save time if rendering is costly). However, this also
+            means that unlike "action repeats"/FrameSkip which accumulates rewards
+            of several repeated steps, only a single-step reward is obtained after
+            all the physics simulation steps are done. The total number of
+            physics simulation steps in an episode is
+            ``control_timestep / physics_timestep * frame_skip * max_episode_steps``.
```
emailweixu, line 67: where is the parameter physics_timestep?

hnyu: > where is the parameter physics_timestep?

It's not specified here, but hardcoded in every mujoco xml file.

emailweixu (approved):

PR #1150, opened 2021-12-22
Transform multiple nests with inter-nests interactions

Haichao-Zhang:
1) add support to transform multiple nests 
2) and it also supports the computation of the node value using values from all the nests (i.e. interactions).

### DIFF
--- a/alf/nest/nest.py
+++ b/alf/nest/nest.py
@@ -797,6 +797,67 @@ def transform_nest(nested, field, func):
         nested=nested, levels=field.split('.') if field else [])
 
 
+def transform_nests(nests, field, func):
+    """Transform the node of each of the nest in nests indicated by ``field``
+        using ``func``.
+
+    This function can be used to transform multiple nests, and perform
+        transformations with inter-nest interactions.
+
+        .. code-block:: python
+
+            res1, res2 = transform_nests([nest1, nest2], 'a.b',
+                                lambda x: (x[0] * x[1], x[0] + x[1]))
+
+        where ``x[0]`` denotes the value from ``nest1`` and ``x[1]`` is
+        from ``nest2``.
+    Args:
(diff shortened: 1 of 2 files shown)


alf/nest/nest.py
```diff
@@ -797,6 +797,67 @@ def _traverse_transform(nested, levels):
...
+    """Transform the node of each of the nest in nests indicated by ``field``
+        using ``func``.
+
+    This function can be used to transform multiple nests, and perform
+        transformations with inter-nest interactions.
+
+        .. code-block:: python
+
+            res1, res2 = transform_nests([nest1, nest2], 'a.b',
+                                lambda x: (x[0] * x[1], x[0] + x[1]))
```
breakds, line 810: My understanding is that this is a drop in replacement to

```python
map(lambda x: x[0] * x[1], x[0] + x[1], [obj.a['b'] for obj in [nest1, nest2]])
```

If so, my humble opinion is that using native python code above here is simpler and more readable.

Haichao-Zhang: > My understanding is that this is a drop in replacement to
> 
> ```python
> map(lambda x: x[0] * x[1], x[0] + x[1], [obj.a['b'] for obj in [nest1, nest2]])
> ```
> 
> If so, my humble opinion is that using native python code above here is simpler and more readable.

Thanks for the comment. If I understand correctly, the native python code does not produce a result that preserves the original nest structure?

One of the reason for using ``transform_xxx`` functions is to transform the specified field of the nest and keep others and the whole nest structure unchanged.

breakds: Ah sorry I did not notice that there is already `transform_nest` as an analogy. I understand now that without `_replace` a namedtuple cannot be altered.

breakds (approved):

PR #1151, opened 2021-12-28
automatically find out y_range when plotting tb curves, when y_range …

le-horizon:
…not specified.

### DIFF
--- a/alf/utils/plot_tb_curves.py
+++ b/alf/utils/plot_tb_curves.py
@@ -643,12 +643,22 @@ if __name__ == "__main__":
             smoothing=3) for t in tasks
     ] for m in methods]
 
+    # Specify y_range, or automatically find out y_range with margin
+    y_range = (0, 1.0)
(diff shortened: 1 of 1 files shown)


alf/utils/plot_tb_curves.py
```diff
@@ -643,12 +643,22 @@ def _get_curve_path(dir):
             smoothing=3) for t in tasks
     ] for m in methods]
 
+    # Specify y_range, or automatically find out y_range with margin
+    y_range = (0, 1.0)
+    _curves = [cr[0]() for cr in curve_readers]
+    if not y_range:
```
hnyu, line 649: This code block is just an example. perhaps it will be better to do this in the CurvesPlotter class?

hnyu (changes requested):

alf/utils/plot_tb_curves.py
le-horizon: Realized that just specifying None will automatically determine y_range.  No need for this PR..

PR #1152, opened 2022-01-02
Add param batchnorm layers

runjerry:
* Add 1d and 2d ParamBatchNorm layers

* Make configuration of param layers consistent with alf.nn.layers

* Update param layers and param networks accordingly.

### DIFF
--- a/alf/layers.py
+++ b/alf/layers.py
@@ -1760,7 +1760,7 @@ class ParamFC(nn.Module):
             activation (torch.nn.functional):
             use_bias (bool): whether use bias
             use_bn (bool): whether to use Batch Normalization.
-            n_groups (int): number of parallel groups.
+            n_groups (int): number of parallel groups, must be specified if ``use_bn``
             kernel_initializer (Callable): initializer for the FC layer kernel.
                 If none is provided a ``variance_scaling_initializer`` with gain as
                 ``kernel_init_gain`` will be used.
@@ -1992,7 +1992,7 @@ class ParamConv2D(nn.Module):
             padding (int or tuple):
             use_bias (bool|None): whether use bias. If None, will use ``not use_bn``
             use_bn (bool): whether use BatchNorm
-            n_groups (int): number of parallel groups.
+            n_groups (int): number of parallel groups, must be specified if ``use_bn``
             kernel_initializer (Callable): initializer for the conv layer kernel.
                 If None is provided a variance_scaling_initializer with gain as
                 ``kernel_init_gain`` will be used.
@@ -2234,6 +2234,9 @@ class ParamConv2D(nn.Module):
 
 @alf.configurable
 class ParamBatchNorm(nn.Module):
+    """Base class for ParamBatchNorm, adapted from ``torch.nn.modules._BatchNorm``
+    """
+
     def __init__(self,
                  num_features,
                  n_groups,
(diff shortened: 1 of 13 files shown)


alf/layers.py
```diff
@@ -2073,45 +2190,380 @@ def forward(self, img, keep_group_dim=True):
...
             # reshape back: [B, n*C', H', W'] -> [B, n, C', H', W']
-            res = res.reshape(res.shape[0], self._groups, self._out_channels,
+            res = res.reshape(res.shape[0], self._n_groups, self._out_channels,
                               res.shape[2], res.shape[3])
 
         return res
 
 
+@alf.configurable
+class ParamBatchNorm(nn.Module):
```
emailweixu: now we have a separate file batch_norm.py. You can put this there.

alf/layers.py
```diff
@@ -1757,25 +1759,42 @@ def __init__(self,
...
         else:
             self._bias_length = 0
             self._bias = None
+        if use_bn:
+            assert n_groups is not None, (
+                "n_groups has to be specified if use_bn")
+            self._bn = ParamBatchNorm1d(output_size, n_groups)
+            self._n_groups = n_groups
+        else:
+            n_groups = 1
```
emailweixu, line 1800: why is n_groups 1 if not use_bn?

alf/layers.py
```diff
@@ -1941,18 +2012,28 @@ def __init__(self,
...
         else:
             self._bias_length = 0
             self._bias = None
+        if use_bn:
+            assert n_groups is not None, (
+                "n_groups has to be specified if use_bn")
+            self._bn = ParamBatchNorm2d(out_channels, n_groups)
+            self._n_groups = n_groups
+        else:
+            n_groups = 1
```
emailweixu, line 2043: similar to ParamFC, n_groups = 1?

alf/networks/param_networks.py
```diff
@@ -183,20 +193,26 @@ def __init__(self,
                 tuple takes a format
                 ``(filters, kernel_size, strides, padding, pooling_kernel)``,
                 where ``padding`` and ``pooling_kernel`` are optional.
-            fc_layer_params (tuple[tuple]): a tuple of tuples where each tuple
-                takes a format ``(FC layer sizes. use_bias)``, where
-                ``use_bias`` is optional.
+            fc_layer_params (tuple[int]): a tuple of integers
```
emailweixu, line 196: Perhaps a more generalizable way of doing this is to implement ParamSequential

alf/layers.py
runjerry: Updated.

alf/layers.py
runjerry: If not ``use_bn``, then ``self._n_groups`` will be determined by input parameters when calling ``set_parameters``, i.e., we allow calling ParamFC with changing ``n_groups`` if not ``use_bn``. In this case, ``n_groups`` could be initialized as any integer. But we have to fix a ``self._n_groups`` if bn is used. 

Updated the docstring to explain it.

alf/layers.py
runjerry: yep, same reason as explained above.

alf/networks/param_networks.py
runjerry: Yeah, that would be ideal. I will think it over after the submission deadline.

runjerry:
Merged into PR #1205

PR #1156, opened 2022-01-12
Fix incomplete operative_config

### LINKED ISSUE #1147: operative_config not complete
Currently, Trainer logs the operative config after the first iteration. However, for some algorithms, model update has not started (e.g., it needs to wait for enough initial random experiences). Hence some relevant functions haven't been called at that point and the operative is not correctly inferred.

emailweixu:
Fixes #1147

To do this, instead of making the summary in the very beginning of training, we wait until training actual happens, which is indicated by train_steps returned from train_iter().

### DIFF
--- a/alf/algorithms/rl_algorithm.py
+++ b/alf/algorithms/rl_algorithm.py
@@ -215,7 +215,7 @@ class RLAlgorithm(Algorithm):
             replay_buffer_length = adjust_replay_buffer_length(
                 config, self._num_earliest_frames_ignored)
 
-            if config.whole_replay_buffer_training:
+            if config.whole_replay_buffer_training and config.clear_replay_buffer:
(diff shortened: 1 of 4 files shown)


hnyu (dismissed):

alf/algorithms/rl_algorithm.py
```diff
@@ -215,7 +215,7 @@ def __init__(self,
             replay_buffer_length = adjust_replay_buffer_length(
                 config, self._num_earliest_frames_ignored)
 
-            if config.whole_replay_buffer_training:
+            if config.whole_replay_buffer_training and config.clear_replay_buffer:
```
hnyu, line 218: If config.clear_replay_buffer=True, we won't care about the initial_collect_steps?

hnyu (approved):

PR #1157, opened 2022-01-12
Full reanalyze mode for MuZeroAlgorithm

emailweixu:
Previously, during reanalyze only the target of first `num_unroll_steps+1`
steps are calculated using MCTS, and the next `reanalyze_td_steps`
are calculated from the model directly.

This change allows to calculate all steps using MCTS, which may give better
target value estimation at the cost of more computation.

Tested on breakout. Didn't observe significant difference.


breakds (approved):

le-horizon (approved):
LG

PR #1158, opened 2022-01-12
Support quantile regression for value and reward prediction in MCTSModel

emailweixu:
Tested on breakout, found it is helpful.


le-horizon (approved):
LG

PR #1160, opened 2022-01-12
Add CARLA Agents and add support to offline data collection

### LINKED ISSUE #1213: Data collection in CARLA

Haichao-Zhang:
This PR 
1. introduces the notion of CARLA agent, which is a rule-based model (or possibly learned model) that can control the Player in the CARLA environment;
2. provides an implementation of a SimpleNavigationAgent;
3. uses the SimpleNavigationAgent for data collection.


breakds (dismissed):
LGTM

emailweixu (commented):
Perhaps add an example config for collecting data using the data agent.

Haichao-Zhang:
> Perhaps add an example config for collecting data using the data agent.

Pushed a new commit with example config for data collection using this agent.

alf/examples/data_collection_carla_conf.py
```diff
@@ -0,0 +1,68 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from functools import partial
+from alf.environments import suite_carla
+from alf.examples import sac_conf
+
+import alf
+
+# This is a example config file for data collection in CARLA.
```
emailweixu: a example =>an example

alf/examples/data_collection_carla_conf.py
```diff
@@ -0,0 +1,68 @@
...
+    'suite_carla.Player',
+    data_collection_mode=data_collection_mode,
+    additional_time=additional_time)
+
+from alf.examples import sac_carla_conf
+# training config for data collection:
+# initial_collect_steps, num_env_steps, replay_buffer_length
+alf.config(
+    'TrainerConfig',
+    initial_collect_steps=initial_collect_steps,
```
emailweixu, line 64: we can use a large value for unroll_length and small value for mini_batch_size to speed up the collection. But perhaps removing SAC from the collection process is even better.

emailweixu (dismissed):

alf/examples/data_collection_carla_conf.py
Haichao-Zhang: Fixed

alf/examples/data_collection_carla_conf.py
Haichao-Zhang: Updated this part and its related comments.  Since the actual training is not invoked, the mini_batch_size does not have a large impact. 
Agree that should remove SAC in the future.

Haichao-Zhang: > we can use a large value for unroll_length and small value for mini_batch_size to speed up the collection. But perhaps removing SAC from the collection process is even better.

Pushed a new comment  (d745d252646fe9c211a802880d0df9db975668ec) to remove SAC by using the SimpleCarlaAlgorithm.

Haichao-Zhang:
Rebased to master and resolved conflicts.

breakds (approved):

PR #1162, opened 2022-01-12
fix rtd empty pages

hnyu:
RTD's API doc crashes and shows nothing for environment files.

### DIFF
--- a/alf/environments/carla_spectator.py
+++ b/alf/environments/carla_spectator.py
@@ -36,7 +36,6 @@ import carla
 import functools
 import math
 import numpy as np
-import pygame
 import os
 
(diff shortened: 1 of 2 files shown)


emailweixu:
Perhaps explain the reason of the crash and fix?

hnyu:
> Perhaps explain the reason of the crash and fix?

We introduced OneHotCategoricalStraightThrough but a lower version of pytorch doesn't have this class:

https://github.com/HorizonRobotics/alf/blob/pytorch/alf/utils/dist_utils.py#L558

Upgrading pytorch to 1.8.1 solved basically all errors.

emailweixu (approved):

PR #1166, opened 2022-01-13
Refactor Experience data structure

### LINKED ISSUE #1165: Refactor Experience data structure
The idea was described here:
https://github.com/HorizonRobotics/alf/pull/858#discussion_r630462293

Basically, we want to make the Experience structure better organized by retaining a TimeStep structure inside.

Haichao-Zhang:
Refactored the ``Experience`` according to the idea proposed here: 
https://github.com/HorizonRobotics/alf/pull/858#discussion_r630462293.

List of changes:
1. change ``Experience`` to a tuple of time_step, action, rollout_info, state etc.
2. expose the typical set of fields of time_step (e.g. observation, reward etc) as attributes of Experience, functioning as another view of the same data. This makes accessing these data easier and in the same way as before.
3. also provide additional helper functions for handling more complex cases (e.g. HER).

### DIFF
--- a/alf/algorithms/data_transformer.py
+++ b/alf/algorithms/data_transformer.py
@@ -383,7 +383,7 @@ class SimpleDataTransformer(DataTransformer):
         Note that for TimeStep, the shapes are [B, ...].
 
         Args:
-            timestep (TimeStep): data to be transformed
+            timestep: data to be transformed
         Returns:
             transformed TimeStep
         """
@@ -395,7 +395,7 @@ class SimpleDataTransformer(DataTransformer):
         For Experience, the shapes are [B, T, ...]
 
         Args:
-            experience (Experience): data to be transformed
+            experience: data to be transformed
         Returns:
             transformed Experience
         """
@@ -406,9 +406,9 @@ class SimpleDataTransformer(DataTransformer):
         """Transform TimeStep.
         Note that this function is used by both ``transform_timestep``
         and ``transform_experience``.
-
         Args:
-            timestep (TimeStep): data to be transformed
+            timestep (TimeStep): data to be transformed. The shape is
+            [B, ...] or [B, T, ...].
         Returns:
             transformed TimeStep
         """
@@ -585,7 +585,6 @@ class RewardTransformer(SimpleDataTransformer):
 
     def _transform(self, timestep):
         return timestep._replace(reward=self.forward(timestep.reward))
-        return
 
 
 @alf.configurable
--- a/alf/algorithms/data_transformer.py
+++ b/alf/algorithms/data_transformer.py
@@ -408,9 +408,9 @@ class SimpleDataTransformer(DataTransformer):
         and ``transform_experience``.
 
         Args:
-            timestep_or_exp (TimeStep|Experience): data to be transformed
+            timestep (TimeStep): data to be transformed
         Returns:
-            transformed TimeStep of Experience
+            transformed TimeStep
         """
         raise NotImplementedError()
 
--- a/alf/algorithms/data_transformer.py
+++ b/alf/algorithms/data_transformer.py
@@ -363,31 +363,49 @@ class FrameStacker(DataTransformer):
         for i, field in enumerate(self._fields):
             observation = alf.nest.transform_nest(observation, field,
                                                   partial(_stack_frame, i=i))
-        return experience._replace(observation=observation)
+        return experience.update_time_step_field('observation', observation)
 
 
 class SimpleDataTransformer(DataTransformer):
     """Base class for simple data transformers.
 
     For simple data transformers, there is no state for ``transform_timestep`` and
-    ``transform_timestep`` and ``transform_experience`` use same function
-    ``_transform`` to do the transformation
+    ``transform_experience``. And ``transform_experience`` use the same function
+    ``_transform`` to do the transformation of the ``time_step`` field of the
+    experience.
     """
 
     def __init__(self, transformed_observation_spec):
         super().__init__(transformed_observation_spec, state_spec=())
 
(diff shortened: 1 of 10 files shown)


le-horizon:
Thanks for taking on this nice idea @Haichao-Zhang !  I'm afraid you'll see lots of problems due to explicit references of experience's fields in places like the replay buffer, hindsight data transformer etc..  Also need to change nest paths referenced during e.g. transform_nest.

Our unit tests should catch most of them.  A good sweep of the repo using "experience." or "exp." would likely help also.

Haichao-Zhang:
> Thanks for taking on this nice idea @Haichao-Zhang ! I'm afraid you'll see lots of problems due to explicit references of experience's fields in places like the replay buffer, hindsight data transformer etc.. Also need to change nest paths referenced during e.g. transform_nest.
> 
> Our unit tests should catch most of them. A good sweep of the repo using "experience." or "exp." would likely help also.

Thanks for the comments, @le-horizon !   I will check again on the places where experiences is used.

hnyu:
> Also need to change nest paths referenced during e.g. transform_nest.

I wonder where are the examples of this referring to nest paths? Indeed, the original path like "reward" now becomes "time_step.reward"

emailweixu:
Also need to examine code accessing replay buffer directly using ReplayBuffer.get_field, where the structure of the experience is assumed. This can happen in DataTransformer and preprocess_experience. For example, FrameStacker, MuzeroAlgorithm.preprocess_experience, MuzeroAlgorithm.preprocess_experience

Haichao-Zhang:
> Also need to examine code accessing replay buffer directly using ReplayBuffer.get_field, where the structure of the experience is assumed. This can happen in DataTransformer and preprocess_experience. For example, FrameStacker, MuzeroAlgorithm.preprocess_experience, MuzeroAlgorithm.preprocess_experience

In the PR, the typical set of fields of time_step (e.g. action, observation etc) are exposed as attributes of Experience, functioning as another view of the same data. This makes accessing these data easier and in the same way as before.
Because of this,  many of the existing preprocess_experience operations are still valid.

Also provides additional helper functions for handling more complex cases (e.g. HER).

le-horizon (dismissed):
Thanks for the big refactor.  LG as is.

As for protections for future changes, it may be worthwhile to add a unit test that checks that map_structure only changes time_step fields once (e.g. add 1 to each field value, and check final results indeed only increment by 1).  This should protect against potential map_structure changes.

alf/utils/data_buffer_test.py
```diff
@@ -22,16 +22,13 @@
 from absl.testing import parameterized
 
 import alf
+from alf.data_structures import Experience
```
emailweixu: DataBuffer is not restricted to use Experience as its item. We should test non-Experience structure as its element.

alf/algorithms/data_transformer.py
```diff
@@ -363,36 +363,54 @@ def _stack_frame(obs, i):
...
-    def _transform(self, timestep_or_exp):
-        """Transform TimeStep or Experience.
+        """Transform Experience.
 
-        Note that for TimeStep, the shapes are [B, ...].
         For Experience, the shapes are [B, T, ...]
 
         Args:
-            timestep_or_exp (TimeStep|Experience): data to be transformed
+            experience (Experience): data to be transformed
```
emailweixu: no need to comment type. It's in the function signature.

alf/algorithms/data_transformer.py
```diff
@@ -363,36 +363,54 @@ def _stack_frame(obs, i):
...
+        transformed_time_step = self._transform(experience.time_step)
+        return experience._replace(time_step=transformed_time_step)
+
+    def _transform(self, timestep):
+        """Transform TimeStep.
+        Note that this function is used by both ``transform_timestep``
+        and ``transform_experience``.
+
+        Args:
+            timestep (TimeStep): data to be transformed
```
emailweixu: add a comment on its shape. [B, ...] or [B, T, ...]

alf/algorithms/data_transformer.py
```diff
@@ -565,9 +583,9 @@ def __init__(self, observation_spec):
         """
         super().__init__(observation_spec)
 
-    def _transform(self, timestep_or_exp):
-        return timestep_or_exp._replace(
-            reward=self.forward(timestep_or_exp.reward))
+    def _transform(self, timestep):
+        return timestep._replace(reward=self.forward(timestep.reward))
+        return
```
emailweixu: remove the extra 'return'

Haichao-Zhang: Removed.

alf/algorithms/data_transformer.py
Haichao-Zhang: Comments added.

alf/algorithms/data_transformer.py
Haichao-Zhang: Removed comment

alf/utils/data_buffer_test.py
Haichao-Zhang: Now changed back to non-Experience structure based tests in data_buffer_test.

alf/algorithms/data_transformer.py
Haichao-Zhang: Removed

Haichao-Zhang:
> Thanks for the big refactor. LG as is.
> 
> As for protections for future changes, it may be worthwhile to add a unit test that checks that map_structure only changes time_step fields once (e.g. add 1 to each field value, and check final results indeed only increment by 1). This should protect against potential map_structure changes.

Good point. Now added the suggested test in commit 2dbe01725a29de1232b75381662886b3e0349957.

emailweixu (approved):

PR #1168, opened 2022-01-14
ElementwiseLayerBase in layers.py

emailweixu:
With this, layers derived from this don't need to implement make_paralell()

Also fixed a bug of make_parallel_net, where the parameters of replicas are not reinitialized if the module does not implement make_parallel().

### DIFF
--- a/alf/layers.py
+++ b/alf/layers.py
@@ -62,7 +62,24 @@ def normalize_along_batch_dims(x, mean, variance, variance_epsilon):
     return x
 
 
-class Identity(nn.Module):
+class ElementwiseLayerBase(nn.Module):
+    """Base class for the layers of parameterless elementwise operations."""
+
+    def make_parallel(self, n: int):
+        """Create a layer with same operation to handle parallel batch.
+
+        It is assumed that a parallel batch has shape [B, n, ...].
+
+        Args:
+            n (int): the number of replicas.
+        Returns:
+            a layer with same operation to handle parallel batch.
+        """
+        assert len(list(self.parameters())) == 0
+        return self
+
+
+class Identity(ElementwiseLayerBase):
     """A layer that simply returns its argument as result."""
 
     def __init__(self):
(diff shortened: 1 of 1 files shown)


alf/layers.py
```diff
@@ -2251,6 +2265,16 @@ class BottleneckBlock(nn.Module):
...
       `<https://ngc.nvidia.com/catalog/model-scripts/nvidia:resnet_50_v1_5_for_pytorch>`_.
+
+    TODO:
+
+    1. ResNet-D in `Bag of Tricks for Image Classification with Convolutional Neural Networks
+       <https://openaccess.thecvf.com/content_CVPR_2019/papers/He_Bag_of_Tricks_for_Image_Classification_with_Convolutional_Neural_Networks_CVPR_2019_paper.pdf>`_
+       Note: v1_5 is the ResNet-B in the above paper.
+    2. Squeeze-and-Excitation (SE) in `Squeeze-and-Excitation Networks
+       <https://openaccess.thecvf.com/content_cvpr_2018/papers/Hu_Squeeze-and-Excitation_Networks_CVPR_2018_paper.pdf>`_
+       SE is also shown to be uesfule in
```
hnyu: uesfule -> useful

alf/layers.py
```diff
@@ -2833,8 +2841,23 @@ def __init__(self):
...
+class Scale(ElementwiseLayerBase):
+    def __init__(self, scale):
+        super().__init__()
+        self._scale = scale
+
+    def forward(self, input):
+        return self._scale * input
+
+
+class ScaleGradient(ElementwiseLayerBase):
```
hnyu, line 2863: docstring is missing

alf/layers.py
```diff
@@ -3131,6 +3154,8 @@ def __init__(self, module: Union[nn.Module, Callable], n: int):
         if isinstance(module, nn.Module):
             self._networks = nn.ModuleList(
                 [copy.deepcopy(module) for i in range(n)])
+            for net in self._networks:
+                reset_parameters(net)
```
hnyu, line 3175: Was this a bug?

hnyu (changes requested):

alf/layers.py
emailweixu: Yes, this is a bug. It can be triggered if a module does not implement make_parallel but make_parallel_net is called.

hnyu (approved):

PR #1170, opened 2022-01-21
Offline Training Pipeline

### LINKED ISSUE #1145: Offline RL

Haichao-Zhang:


alf/trainers/policy_trainer.py
```diff
@@ -521,6 +521,262 @@ def _eval(self):
         self._algorithm.train()
 
 
+class OfflineRLTrainer(Trainer):
+    """Trainer for reinforcement learning. """
+
+    def __init__(self, config: TrainerConfig, ddp_rank: int = -1):
```
emailweixu, line 527: Looks very similar to RLTrainer. Is it possible to reuse some code?

Haichao-Zhang: > Looks very similar to RLTrainer. Is it possible to reuse some code?

Yes, that's true. The current ``OfflineRLTrainer`` is indeed modified from ``RLTrainer`` with some of its own control logic embedded. 

More sharing will be better and I will push it more toward this goal.

Haichao-Zhang:
Closed with #1206.

PR #1174, opened 2022-01-25
[Small] Upgrade alf Nix development environment to use the latest MetaDrive

breakds:
# What is changed?

This moves development environment to point to the latest https://github.com/HorizonRobotics/alf-nix-devenv, so that it uses the up-to-date [MetaDrive](https://github.com/decisionforce/metadrive).

# What is new in the upgraded MetaDrive?

1. Better visualization for top-down view. It used to visualize the whole map so that it fails when the map is too big. It is updated to show a local view instead to avoid that issue.
2. More sophisticated model for the other agents, so that they can perform actions such as lane change, cut-in and aggressive merge.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -9,11 +9,11 @@
         "utils": "utils"
       },
       "locked": {
-        "lastModified": 1635208657,
-        "narHash": "sha256-grL7u4LQADbXzELwPwtpQZHIr4KW4ftgatCzepdNjhA=",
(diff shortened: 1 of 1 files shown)


le-horizon (approved):
Just curious, is there no content change, only hash changes?

breakds:
> Just curious, is there no content change, only hash changes?

Yes, only hash changes about the commits to use.

PR #1177, opened 2022-01-26
MetaDrive Vectorized Encoding Part II - Implement transformations on Polyline

breakds:
# Motivation

This is the 2nd part of the effort to create vectorized input for MetaDrive. After the road map and navigation is converted to `Polyline`, for every frame we will need to crop the polylines within the field of view of the ego car, and rotate/translate so that they are in ego car's body coordinate frame. This PR is here to implement those utilities to support such operations.

# Solution

Added utilities to perform the following transformation in batch:

1. Convert polylines to car body coordinate frame given reference car's pose (position and heading)
2. Filter out the polylines that are out of the filed of view of the reference car
3. Filter out the farthest polylines when the number of polylines is beyond a threshold

# Testing

1. Visualized the transformed polylines in Jupyter notebook to verify, see below.
2. Together with the other PRs (yet to come), it produces successful training result (on par with BEV baseline).

polyline_visualize

### DIFF
--- a/alf/environments/metadrive/__init__.py
+++ b/alf/environments/metadrive/__init__.py
@@ -1,0 +1,17 @@
+# Copyright (c) 2021 Horizon Robotics and ALF Contributors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from .geometry import FieldOfView
+
+__all__ = ['FieldOfView']
--- a/alf/environments/metadrive/__init__.py
+++ b/alf/environments/metadrive/__init__.py
@@ -13,5 +13,3 @@
 # limitations under the License.
 
 from .geometry import FieldOfView
-
-__all__ = ['FieldOfView']
--- a/alf/environments/metadrive/geometry.py
+++ b/alf/environments/metadrive/geometry.py
@@ -240,17 +240,16 @@ class Polyline(NamedTuple):
             center: A 2D point denoting the origin of the new coordinate frame.
             orientation: orientation (x-axis direction) of the new coordinate in radian.
 
+        Returns:
+
+            A NEW Polyline instance where all the polylines within are transformed.
         """
         # Construct the 2D rotation matrix.
         cos = np.cos(orientation)
         sin = np.sin(orientation)
         rotation = np.array([[cos, -sin], [sin, cos]])
 
-        if self.batched:
-            transformed = np.einsum('ijk,kl->ijl', (self.point - center),
-                                    rotation)
-        else:
-            transformed = np.matmul(self.point - center, rotation)
+        transformed = np.matmul(self.point - center, rotation)
 
         return Polyline(point=transformed, category=self.category)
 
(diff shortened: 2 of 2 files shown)


alf/environments/metadrive/__init__.py
```diff
@@ -0,0 +1,17 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from .geometry import FieldOfView
+
+__all__ = ['FieldOfView']
```
emailweixu: this seems redundant

alf/environments/metadrive/geometry.py
```diff
@@ -163,6 +193,68 @@ def batched(self) -> bool:
...
+
+        """
+        # Construct the 2D rotation matrix.
+        cos = np.cos(orientation)
+        sin = np.sin(orientation)
+        rotation = np.array([[cos, -sin], [sin, cos]])
+
+        if self.batched:
+            transformed = np.einsum('ijk,kl->ijl', (self.point - center),
+                                    rotation)
```
emailweixu: this branch is unnecessary. matmul can handle both cases.

alf/environments/metadrive/geometry.py
```diff
@@ -163,6 +193,68 @@ def batched(self) -> bool:
...
+        car's position and heading, and filtered out the polylines that are not
+        within the field of view.
+
+        Returns a new Polyline instance as the result. The update is not inplace.
+
+        Args:
+
+            position: the position of the car serving as the observer.
+            heading: heading of the car serving as the observer.
+            fov: the field of view of the car defining the area that are visible.
```
emailweixu, line 287: comment "Returns"

alf/environments/metadrive/geometry.py
```diff
@@ -163,6 +193,68 @@ def batched(self) -> bool:
...
 
+    def transformed(self, center: np.ndarray, orientation: float) -> Polyline:
+        """Returns a transformed (set of) polyline(s) so that the resulting points are
+        in the coordinate frame defined by the pose (center and the orientation).
+
+        Args:
+
+            center: A 2D point denoting the origin of the new coordinate frame.
+            orientation: orientation (x-axis direction) of the new coordinate in radian.
+
```
emailweixu, line 263: "Returns:"

alf/environments/metadrive/geometry.py
```diff
@@ -201,6 +231,68 @@ def batched(self) -> bool:
...
+        if self.batched:
+            transformed = np.einsum('ijk,kl->ijl', (self.point - center),
+                                    rotation)
+        else:
+            transformed = np.matmul(self.point - center, rotation)
+
+        return Polyline(point=transformed, category=self.category)
+
+    def within_fov(self, position: np.ndarray, heading: float,
```
Haichao-Zhang: Suggest to use a function name that can reflect the contents of the function better.

alf/environments/metadrive/geometry.py
breakds: Fixed.

alf/environments/metadrive/geometry.py
breakds: Fixed.

alf/environments/metadrive/__init__.py
breakds: Removed.

alf/environments/metadrive/geometry.py
breakds: Yes you are right. For some reason I mistakenly remembered pytorch allows this but numpy does not ...

alf/environments/metadrive/geometry.py
```diff
@@ -56,6 +56,12 @@ def __init__(self,
     def bbox(self):
         return self._bbox
 
+    def test(self, points: np.ndarray):
```
emailweixu: doc-string and better name such as "within"?

breakds: Fixed. Also renamed the function as suggested.

alf/environments/metadrive/geometry.py
breakds: Agree the original name does not suggest "transformation". I have updated the name to `transformed_within_fov`.

emailweixu (dismissed):

Haichao-Zhang (dismissed):

breakds:
Just rebased `pytorch` and need a refreshed approval. Thanks!

Haichao-Zhang (approved):

PR #1179, opened 2022-01-26
MetaDrive Vectorized Encoding Part IV - The vectorized TopDownEnv for MetaDrive

breakds:
# Motivation

The hierarchy of the MetaDrive environment in alf looks like:

1. At the top level we will use a `AlfMetaDriveWrapper` to wrap a native `MetaDriveEnv`.
2. For vectorized input, we need a customized `MetaDriveEnv`, we call it `VectorizedTopDownEnv`, as a counterpart to the raster based `TopDownEnv`.
3. `VectorizedTopDownEnv` needs to have its own MetaDrive `Observation`.
4. The `Observation` consists of several parts such as submodule for map, navigation and as on

Once 2 is built, we can embed it to our own `AlfMetaDriveWrapper` directly.

# Solution

The previous PRs build the foundations in 4, and this PR implements 2 and 3 on top of them, following MetaDrive's convention.

# Testing

Together with the other PRs (yet to come), had successful training with result on par with the BEV baseline.

### DIFF
--- a/alf/environments/metadrive/environments.py
+++ b/alf/environments/metadrive/environments.py
@@ -1,0 +1,60 @@
+# Copyright (c) 2021 Horizon Robotics and ALF Contributors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import numpy as np
+import gym
+import torch
+
+from alf.tensor_specs import TensorSpec
+
+try:
+    import metadrive
+except ImportError:
+    from unittest.mock import Mock
+    # create 'metadrive' as a mock to not break python argument type hints
+    metadrive = Mock()
+
+from .geometry import FieldOfView
+from .sensors import VectorizedObservation
+
+
+class VectorizedTopDownEnv(metadrive.MetaDriveEnv):
+    """This is the counterpart of the TopDownEnv from MetaDrive with vectorized
+    input insead of raster input (BEV).
+
+    """
+
+    @classmethod
+    def default_config(cls) -> metadrive.utils.Config:
+        """The default config is identical to that of the raster TopDownEnv.
+
+        """
+        config = metadrive.MetaDriveEnv.default_config()
+        config["vehicle_config"]["lidar"] = {"num_lasers": 0, "distance": 0}
+        config.update({
+            "frame_skip": 5,
+            "frame_stack": 3,
+            "post_stack": 5,
+            "rgb_clip": True,
+            "resolution_size": 84,
+            "distance": 30
+        })
+        return config
+
+    def get_single_observation(self, _=None):
+        return VectorizedObservation(self.config["vehicle_config"])
+
+    @property
+    def observation_spec(self):
+        return self.get_single_observation().observation_spec
--- a/alf/environments/metadrive/environments.py
+++ b/alf/environments/metadrive/environments.py
@@ -19,6 +19,7 @@ from alf.tensor_specs import TensorSpec
 
 try:
     import metadrive
+    from metadrive.obs.observation_base import ObservationBase
 except ImportError:
     from unittest.mock import Mock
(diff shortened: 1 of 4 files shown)


alf/environments/metadrive/environments.py
```diff
@@ -0,0 +1,60 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import numpy as np
+import gym
+import torch
```
Haichao-Zhang: torch does not seem to be used

breakds: Removed.

alf/environments/metadrive/environments.py
```diff
@@ -0,0 +1,59 @@
...
+            "frame_skip": 5,
+            "frame_stack": 3,
+            "post_stack": 5,
+            "rgb_clip": True,
+            "resolution_size": 84,
+            "distance": 30
+        })
+        return config
+
+    def get_single_observation(self, _=None):
```
emailweixu: I understand that the name is inherited from MetaDrive. But some comment can help reviewer understand this. It seems that the returned value is not an observation, but a class for extracting observation.

alf/environments/metadrive/sensors.py
```diff
@@ -0,0 +1,158 @@
...
+                goes beyond this limit, farthest polylines (from the ego car) will be
+                filtered out until the limit is satisfied. If the actual number of
+                polylines is below this limit, zero padding will be employed to bring
+                fill the vacancies.
+            history_window_size: The past positions of the most recent this number of
+                frames will be recorded and used for the ego history feature.
+
+        """
+        super().__init__(vehicle_config)
+        self._map_perception = MapPolylinePerception(
```
emailweixu, line 88: what is the rationale of having a separate MapPolylinePerception class? It seems that VectorizedObservation is a very thin wrapper on top of MapPolylinePerception. I feel these two can be merged.

alf/environments/metadrive/environments.py
breakds: Agreed. Added documentation to help the reader understand the true behavior here. Thanks!

alf/environments/metadrive/sensors.py
breakds: If there were only MapPerception, yes it should be merged (and that is actually how this class look like before agent perception are added in, i.e., there used to be only `VectorizedObervation` class).

With the `AgentPerception` added, I felt it is more clear to have split the code into 2 classes (hence `MapPolylinePerception` and `AgentPerception`, where the latter is introduced in the [next PR](https://github.com/HorizonRobotics/alf/pull/1181)) because both of them is non-trivial. The ego perception, however, is trivial enough so that it does not deserve its own modular design.

emailweixu: I guess this is a problem of separating a change into too many parts. Although a giant PR is not desirable, the other extreme is not good either.

alf/environments/metadrive/map_perception.py
```diff
@@ -0,0 +1,191 @@
...
+                    for side in range(num_sides):
+                        # Ignore lane boundaries with a type of NONE, because
+                        # that means in the real world there will be no line
+                        # drawn on the ground for them, even though logically
+                        # they exist.
+                        if lane.line_types[side] == LineType.NONE:
+                            continue
+                        category = self._category_encoder.encode_line_type(
+                            lane.line_types[side])
+                        offset = -0.5 if side == 0 else 0.5
```
emailweixu: offset = side - 0.5

alf/environments/metadrive/map_perception.py
```diff
@@ -0,0 +1,191 @@
+# Copyright (c) 2022 Horizon Robotics and ALF Contributors. All Rights Reserved.
```
emailweixu: This file is from the previous PR, right?

emailweixu (dismissed):

alf/environments/metadrive/sensors.py
breakds: Yes, apparently everything comes at a cost. Sorry that my intention to make it easier for the reviewer did not achieve the goal here. Will try to strike a better trade off netxt time.

alf/environments/metadrive/map_perception.py
breakds: Updated.

alf/environments/metadrive/map_perception.py
breakds: Yes. Just did a rebase and fixed this.

emailweixu (approved):

PR #1180, opened 2022-02-02
add outer product nest combiner

hnyu:
This PR adds a new NestCombiner class that does outer product across a nest. This can be an alternative to NestConcat for fusing different tensors.

An illustration from the ICLR2017 paper

image

I tried the outer product combiner on CriticNetwork for combining the observation and action, and got better training curves with SAC.

Outer product:
sac_bipedalwalker_obsaction_outerprod

Concat:
image


alf/nest/utils.py
```diff
@@ -76,7 +90,7 @@ def __init__(self, nest_mask=None, dim=-1, name="NestConcat"):
         concatenating them along a specified axis. If nest_mask is None,
         then all the tensors from the nest will be selected.
         It assumes that all the selected tensors have the same tensor spec.
-        Can be used as a preprocessing combiner in ``EncodingNetwork``.
+        Can be used as a preprocessing combiner of a ``PreprocessorNetwork``.
```
emailweixu: EncodingNetwork is not a PreprocessorNetwork.

hnyu: Some preprocessor network like ActorNetwork also has combiner but they are not EncodingNetwork. So I've changed to 'network'.

hnyu: > EncodingNetwork is not a PreprocessorNetwork.

Some preprocessor network like ActorNetwork also has combiner but they are not EncodingNetwork. So I've changed to 'network'.

alf/nest/utils.py
```diff
@@ -176,6 +190,71 @@ def make_parallel(self, n):
...
+        if activation is None:
+            activation = lambda x: x
+        self._activation = activation
+        self._padding = padding
+
+    def _combine_flat(self, tensors):
+        batch_shapes = tensors[0].shape[:self._batch_dims]
+        B = int(np.prod(batch_shapes))
+
+        tensors = [t.reshape(B, -1) for t in tensors]
```
breakds, line 246: It would be better to add an assert to make sure all the tensors have the same "batch shape" to avoid a silent failure before this step, at the cost of `O(number of tensors)` (which I think is negligible).

hnyu: > It would be better to add an assert to make sure all the tensors have the same "batch shape" to avoid a silent failure before this step, at the cost of `O(number of tensors)` (which I think is negligible).

I think the batch shape can be different, as long as ``B`` is the same. If not, then reshape() will throw an error.

breakds (approved):
LGTM.

emailweixu (approved):

PR #1182, opened 2022-02-02
Special handling of reanalyze_ratio==1

emailweixu:
To reduce unnecessary computation.

Also fixed non-deterministic behavior of MCTSAlgorithm which makes reproducing experiments difficult.

### DIFF
--- a/alf/algorithms/mcts_algorithm.py
+++ b/alf/algorithms/mcts_algorithm.py
@@ -81,8 +81,10 @@ class _MCTSTrees(object):
         # reward for player 0
         self.reward = None
         if isinstance(model_output.reward, torch.Tensor):
+            # we use value.dtype because the reward.dtype from the intial_inference
+            # can be different from the reward.dtype from the recurrent_inference
             self.reward = torch.zeros(
-                parent_shape, dtype=model_output.reward.dtype)
+                parent_shape, dtype=model_output.value.dtype)
 
         self.action = None
         if isinstance(model_output.actions, torch.Tensor):
@@ -857,8 +859,8 @@ class MCTSAlgorithm(OffPolicyAlgorithm):
 
     def _ucb_score1(self, parent_visit_count, child_visit_count, prior,
                     value_score):
-        pb_c = torch.log((parent_visit_count + self._pb_c_base + 1.) /
-                         self._pb_c_base) + self._pb_c_init
+        pb_c = torch.log1p(
+            (parent_visit_count + 1) / self._pb_c_base) + self._pb_c_init
         pb_c = pb_c * torch.sqrt(parent_visit_count.to(
             torch.float32)) / (child_visit_count + 1.)
         prior_score = pb_c * prior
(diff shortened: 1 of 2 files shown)


alf/algorithms/muzero_algorithm.py
```diff
@@ -321,10 +321,16 @@ def preprocess_experience(self, root_inputs: TimeStep,
...
                 # Here we assume state and info have similar name scheme.
                 mcts_state_field = 'state' + info_path[len('rollout_info'):]
-                r = torch.rand(
-                    root_inputs.step_type.shape[0]) < self._reanalyze_ratio
-                r_candidate_actions, r_candidate_action_policy, r_values = self._reanalyze(
-                    replay_buffer, env_ids[r], positions[r], mcts_state_field)
+                if self._reanalyze_ratio < 1:
+                    batch_size = root_inputs.step_type.shape[0]
+                    r = torch.randperm(
+                        batch_size) < batch_size * self._reanalyze_ratio
```
emailweixu, line 327: The reason of changing from rand to randperm is to ensure that the batch size does not change, which is more memory friendly.

alf/algorithms/mcts_algorithm.py
```diff
@@ -1013,21 +1020,31 @@ def _backup(self,
...
+            # nodes_i = nodes[0] * trees.visit_count.shape[1] + nodes[1]
+            # # using scatter_add_ to handle possible duplicated nodes correctly
+            # trees.visit_count.view(-1).scatter_add_(
+            #     0, nodes_i,
+            #     torch.tensor([count],
+            #                  dtype=trees.visit_count.dtype).expand_as(nodes_i))
+            # trees.value_sum.view(-1).scatter_add_(0, nodes_i, value_sum_delta)
+            # The scatter_add_ above is not deterministic. If deterministic behavior
+            # is needed, replace the above code with the following:
+            depth = depth.squeeze(-1)
```
le-horizon, line 1040: is the commented block faster?
If there is a material impact to speed, then maybe when config.random_seed is set, use deterministic path, otherwise use the one above.  If this is done, then the test also should cover both paths, making sure they are equivalent.

alf/algorithms/mcts_algorithm.py
```diff
@@ -1118,7 +1135,8 @@ def _set_tree_state(ts, s):
         nest.map_structure(_set_tree_state, trees.model_state,
                            model_output.state)
         if trees.reward is not None:
-            trees.reward[:, n] = model_output.reward
+            # model_output.reward.dtype may be different from initial_inference
+            trees.reward[:, n] = model_output.reward  #.to(trees.reward.dtype)
```
le-horizon: Is this .to in the comment needed?

le-horizon (commented):
A couple small points if they make sense.

alf/algorithms/mcts_algorithm.py
emailweixu: Both path are equivalent except that the first path is not deterministic, which means that there is still can be numerical difference due different order of arithmetic operations. In theory, the first path should be faster because the second path is a loop. But since this part of computation is small compare to other part of computation, so either way does not really affect speed much.

le-horizon (approved):
LG

breakds (approved):

PR #1183, opened 2022-02-02
Several enhancements for optimization

emailweixu:
1. Parameter specific weight decay for AdamTF
2. Bias only option for BatchNorm
3. Support fixed_norm for optimizer
4. Put optimizer hyper parameters to optimizer.defaults so that their values can be shown in tensorboard summary of optimizers

### DIFF
--- a/alf/batch_norm.py
+++ b/alf/batch_norm.py
@@ -46,7 +46,7 @@ class _NormBase(nn.Module):
             self._weight = None
         if use_bias:
             if self._weight is None:
-                # pytorch has a bug which cannot handle the case that weihgt is
+                # pytorch has a bug which cannot handle the case that weight is
                 # None but bias is not. So we have to provide a fixed weight.
                 self._weight = nn.Parameter(
                     torch.Tensor(num_features), requires_grad=False)
--- a/alf/batch_norm.py
+++ b/alf/batch_norm.py
@@ -12,12 +12,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import math
 import torch
 import torch.nn as nn
 import torch.nn.functional as F
 from typing import List, Union
 import types
 
+import alf
 from alf.utils.common import warning_once
 
 
@@ -29,6 +31,7 @@ class _NormBase(nn.Module):
                  eps: float = 1e-5,
                  momentum: float = 0.1,
                  affine: bool = True,
+                 use_bias: bool = True,
                  track_running_stats: bool = True):
         super().__init__()
         self._num_features = num_features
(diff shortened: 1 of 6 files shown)


alf/batch_norm.py
```diff
@@ -38,7 +41,19 @@ def __init__(self,
         self._track_running_stats = track_running_stats
         if affine:
             self._weight = nn.Parameter(torch.Tensor(num_features))
+            use_bias = True
+        else:
+            self._weight = None
+        if use_bias:
+            if self._weight is None:
+                # pytorch has a bug which cannot handle the case that weihgt is
```
Haichao-Zhang: weihgt -> weight

alf/batch_norm.py
```diff
@@ -102,8 +117,9 @@ def reset_parameters(self):
                 self._running_means[i].zero_()
                 self._running_vars[i].fill_(1)
                 self._num_batches_tracked[i].zero_()
-        if self._affine:
+        if self._weight is not None:
```
Haichao-Zhang, line 120: We can perhaps skip initializing for _weight in the bias only case?

alf/optimizers/optimizers.py
```diff
@@ -188,6 +200,10 @@ def step(self, closure=None):
 
         super(NewCls, self).step(closure=closure)
 
+        for param in params:
+            if param.grad is not None and getattr(param, 'fixed_norm', False):
+                param.data.mul_(self._norms[param] / (param.norm() + 1e-30))
```
Haichao-Zhang, line 216: It's not quite clear what the operation this ``fixed_norm` option represents. Maybe we can add some explanations.

alf/networks/projection_networks.py
```diff
@@ -34,6 +34,8 @@ def __init__(self,
                  input_size,
                  action_spec,
                  logits_init_output_factor=0.1,
+                 weight_opt_args=None,
+                 bias_opt_args=None,
```
hnyu, line 38: Why are these arguments only added for this class but not for the other ones?

alf/batch_norm.py
emailweixu: fixed.

alf/batch_norm.py
emailweixu: It does not matter since it's one

alf/networks/projection_networks.py
emailweixu: No particular reason. It's just that I didn't use the others in experiments.

hnyu (approved):

PR #1185, opened 2022-02-03
MetaDrive Vectorized Encoding Part VI: Implement the BEV observation

breakds:
# Motivation

To make a better interface for the coexistence of both the vectorized and raster encoding for MetaDrive, the design is to use `create_environment`'s `env_name` to determine whether we want the vectorized version or the raster version.

i.e.

```
alf.config(
    'create_environment', env_name='Vectorized', num_parallel_environments=36)
```
or
```
alf.config(
    'create_environment', env_name='BirdEye', num_parallel_environments=36)
```

The two will spawn the same environment wrapper `AlfMetaDriveWrapper` (as defined in `suite_metadrive.py`), but with different underlying environment, where they differ mostly on having different `Observation` instance. We already have the `VectorizedObservation`, and this PR adds its raster (BEV) counter part, called the `BirdEyeObservation`.

# Solution

Implement `BirdEyeObservation`. It wraps MetaDrive's BEV observation, with an extra output on velocity history. It is expected that the model will combine the velocity into the BEV.

# Testing

Together with the next PR, verified that the training is still working for BEV + CNN. Also verified the output BEV with visualization to make sure it is the same as before.

### DIFF
--- a/alf/environments/metadrive/environments.py
+++ b/alf/environments/metadrive/environments.py
@@ -26,10 +26,51 @@ except ImportError:
     metadrive = Mock()
 
 from .geometry import FieldOfView
-from .sensors import VectorizedObservation
+from .sensors import VectorizedObservation, BirdEyeObservation
 
 
 class VectorizedTopDownEnv(metadrive.MetaDriveEnv):
+    """This is the counterpart of the TopDownEnv from MetaDrive with vectorized
+    input insead of raster input (BEV).
+
+    """
+
+    @classmethod
+    def default_config(cls) -> metadrive.utils.Config:
+        """The default config is identical to that of the raster TopDownEnv.
+
+        """
+        config = metadrive.MetaDriveEnv.default_config()
+        config["vehicle_config"]["lidar"] = {"num_lasers": 0, "distance": 0}
+        config.update({
+            "frame_skip": 5,
+            "frame_stack": 3,
+            "post_stack": 5,
+        })
+        return config
+
+    def get_single_observation(self, _=None) -> ObservationBase:
+        """Implements the get_single_observation for the base class MetaDriveEnv.
+
+        The base class is calling this function to acquire the sensor (typed
+        ObservationBase) that is used for generating observations. Unlike the
+        name may suggest, it is
+
+        1. actually only called once per environment
+        2. returning a sensor object instead of the actual observation
+
+        The sensor object is then used to produce the actual observation of each
+        frame.
+
+        """
+        return VectorizedObservation(self.config["vehicle_config"])
+
+    @property
+    def observation_spec(self):
+        return self.get_single_observation().observation_spec
+
+
+class BirdEyeTopDownEnv(metadrive.MetaDriveEnv):
     """This is the counterpart of the TopDownEnv from MetaDrive with vectorized
     input insead of raster input (BEV).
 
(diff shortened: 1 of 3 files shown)


alf/environments/metadrive/environments.py
```diff
@@ -35,6 +35,34 @@ class VectorizedTopDownEnv(metadrive.MetaDriveEnv):
...
+
+    def get_single_observation(self, _=None):
+        return VectorizedObservation(self.config["vehicle_config"])
+
+    @property
+    def observation_spec(self):
+        return self.get_single_observation().observation_spec
+
+
+class BirdEyeTopDownEnv(metadrive.MetaDriveEnv):
```
breakds, line 73: The diff is presented in an odd way. Actually the code starting from here to the end of the file is the newly added code. The reason it is presented in such a way is that I removed unrelated options from the `config.update({})` call above.

alf/environments/metadrive/sensors.py
```diff
@@ -167,3 +169,85 @@ def reset(self, env, vehicle=None):
...
+                 velocity_steps: int = 1,
+                 velocity_normalization: float = 100.0):
+        """Construct a BirdEyeObservation instance.
+
+        Args:
+            
+            env_config: MetaDrive's environment configuration.
+            velocity_steps: The number of historical steps for the velocity.
+            velocity_normalization: The velocities (in m/s) will be normalizaed
+                by this factor before producing the feature.
```
Haichao-Zhang, line 198: normalizaed

alf/environments/metadrive/sensors.py
```diff
@@ -167,3 +169,85 @@ def reset(self, env, vehicle=None):
...
+
+        Args:
+
+            vehicle: The MetaDrive vehicle object is the container for all the ego
+                car related information.
+
+        """
+        img = super().observe(vehicle)
+
+        self._velocity_history[:-1] = self._velocity_history[1:]
```
Haichao-Zhang, line 247: will it be useful to include a mask for indicating the padding for velocity_history?

Haichao-Zhang: thought about this again and the mask might be unnecessary in this case since the initial velocities can be safely assumed to be zero.

Haichao-Zhang (dismissed):

alf/environments/metadrive/sensors.py
breakds: Fixed.

breakds:
Addressed comments.

Haichao-Zhang:
> Addressed comments.

Need rebase. 
The next few PRs in this series are likely to have the same problem after one of them is merged.

breakds:
> > Addressed comments.
> 
> Need rebase. The next few PRs in this series are likely to have the same problem after one of them is merged.

Thanks. Actually it was rebased in a wrong way. It should be fixed now.

Haichao-Zhang:
> > > Addressed comments.
> > 
> > 
> > Need rebase. The next few PRs in this series are likely to have the same problem after one of them is merged.
> 
> Thanks. Actually it was rebased in a wrong way. It should be fixed now.

There is a new commit to the master fixing the broken tests.

breakds:
Rebased and re-tested. PTAL. Thanks!

Haichao-Zhang (approved):

PR #1193, opened 2022-02-08
Pr muzero categorical loss for reward value

emailweixu:


alf/utils/losses.py
```diff
@@ -102,3 +104,371 @@ def multi_quantile_huber_loss(quantiles: torch.Tensor,
...
+            loss with the same shape as target
+        """
+        raise NotImplementedError()
+
+    def calc_expectation(self, pred: torch.Tensor):
+        """Calculate the expected predition in the untransfomred domain from ``pred``.
+        """
+        raise NotImplementedError()
+
+    def initialize_bias(self, bias: torch.Tensor):
```
breakds, line 126: When reading `ScalarPredictionLoss` alone, I find it a bit clueless about why definition of the loss cares about initialization on the last FC layer of the prediction. Can we copy some of motivation from concrete class below here as example?

alf/utils/losses.py
```diff
@@ -102,3 +104,371 @@ def multi_quantile_huber_loss(quantiles: torch.Tensor,
...
+
+        Args:
+            pred: raw model prediction
+        """
+        if self._transform is not None:
+            pred = self._transform.inverse_transform(pred)
+        return pred
+
+
+def _get_indexer(shape):
```
breakds: It would be great to add documentation for this function.

Also I would have used "torch.meshgrid" without learning the broadcast trick from reading this implementation. It basically fulfills the same purpose as `torch.meshgrid` with linear memory consumption. Thanks!

alf/utils/losses.py
```diff
@@ -102,3 +104,371 @@ def multi_quantile_huber_loss(quantiles: torch.Tensor,
...
+                 inverse_after_mean=False):
+        super().__init__()
+        self._transform = transform
+        if self._transform is not None:
+            self._inverse_after_mean = inverse_after_mean
+        else:
+            self._inverse_after_mean = True
+        self._support = None
+
+    def _calc_support(self, n):
```
breakds: Nit: `n: int`.

le-horizon (dismissed):
LG

alf/utils/losses.py
```diff
@@ -102,3 +104,371 @@ def multi_quantile_huber_loss(quantiles: torch.Tensor,
...
+            return self._support
+        upper_bound = n // 2
+        lower_bound = -((n - 1) // 2)
+        x = torch.arange(lower_bound, upper_bound + 1, dtype=torch.float32)
+        if self._transform is not None and not self._inverse_after_mean:
+            x = self._transform.inverse_transform(x)
+        self._support = x
+        return x
+
+    def _calc_bin(self, logits, target):
```
breakds, line 229: Suggesting adding documentation for this core function.

emailweixu: added.

le-horizon (approved):
LG

breakds (approved):
Thanks!

PR #1195, opened 2022-02-11
Parallel play

emailweixu:
Sometime we want to evaluate a trained model for many episodes. Only using one
environment for it can take a lot of time. So adding parallel_play commandline
flag to support parallel environments during play.

### DIFF
--- a/alf/trainers/policy_trainer.py
+++ b/alf/trainers/policy_trainer.py
@@ -628,7 +628,10 @@ def _step(algorithm,
     if recorder:
         recorder.capture_frame(policy_step.info, time_step.is_last())
     elif render:
-        env.render(mode='human')
+        if env.batch_size > 1:
+            env.envs[0].render(mode='human')
+        else:
+            env.render(mode='human')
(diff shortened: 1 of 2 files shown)


breakds (approved):

alf/trainers/policy_trainer.py
```diff
@@ -690,13 +693,18 @@ def play(root_dir,
...
+    batch_size = env.batch_size
     recorder = None
     if record_file is not None:
+        assert batch_size == 1, 'video recording is not supported for parallel play'
         recorder = VideoRecorder(
             env, append_blank_frames=append_blank_frames, path=record_file)
     elif render:
-        # pybullet_envs need to render() before reset() to enable mode='human'
-        env.render(mode='human')
+        if batch_size > 1:
```
hnyu, line 703: Should instead assert batch_size==1? In this case, there is no need to create multiple envs.

emailweixu: I guess we can still allow rendering if the user still want.

PR #1196, opened 2022-02-12
@repr_wrapper

emailweixu:
When using alf.config to configure, sometimes we pass an object as value. In the tensorboard text summary, only the name of the object is shown in the config, while the arguments for constructing that object are not shown. This is quite inconvenient as we need to open the config file to see that actual value.

This change solve this problem by automatically injecting appropriate `__repr__` function for the class so its textual representation is more informative.

### DIFF
--- a/alf/config_util.py
+++ b/alf/config_util.py
@@ -609,6 +609,9 @@ def _decorate(fn_or_cls, name, whitelist, blacklist):
 def repr_wrapper(cls):
     """A wrapper for automatically generating readable repr for an object.
 
+    The presentation shows the arguments used to construct of object.
+    It does not include the default arguments, nor the class members.
+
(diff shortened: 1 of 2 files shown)


alf/config_util.py
```diff
@@ -596,6 +606,73 @@ def _decorate(fn_or_cls, name, whitelist, blacklist):
         return fn_or_cls
 
 
+def repr_wrapper(cls):
+    """A wrapper for automatically generating readable repr for an object.
```
hnyu, line 610: It's better to clarify that the repr shows how the object is created with non-default argument values (but not other information like class members)

hnyu (changes requested):

hnyu (approved):

PR #1197, opened 2022-02-12
A bug fix of MCTSAlgorithm

emailweixu:
This bug can be triggered when search_with_exploration_policy is False and num_parallel_sims > 1.
Under this candition, if there are two search paths with different second-to-the-last node,
the second path will use the non-optimal (in terms of UCB score) action to expand.

This change fixes this problem.

### DIFF
--- a/alf/algorithms/mcts_algorithm.py
+++ b/alf/algorithms/mcts_algorithm.py
@@ -706,14 +706,12 @@ class MCTSAlgorithm(OffPolicyAlgorithm):
         nodes = trees.root_indices
         i = ()
         if self._parallel:
-            B = B.unsqueeze(-1)
-            nodes = trees.root_indices.unsqueeze(-1).expand(-1, psims)
             # If the p-th search path of the b-th tree is at node n, it will be
             # extended by the child indicated by best_child_index[b, n, i[p]].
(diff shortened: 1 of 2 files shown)


alf/algorithms/mcts_algorithm.py
```diff
@@ -677,12 +714,8 @@ def _search(self, trees, to_plays):
             # See the doc string of MCTSAlgorithm for more detail.
             if self._search_with_exploration_policy:
                 i = (torch.arange(psims), )
```
le-horizon, line 712: Let's move this "i" assignment closer to the i = () initialization, as is done in line 535-539 (in the new code).

Also, the two places assign "i" differently for the case of (_parallel but not _search_with_exploration_policy).  Is this intended?

le-horizon (commented):
Almost good.  One comment and one question.

Thanks,
Le

alf/algorithms/mcts_algorithm.py
emailweixu: It is the correct behavior. For _calc_i does not need ``i`` for the case of (_parallel but not _search_with_exploration_policy). 
In build_tree1(), ``i`` is also used for other purpose when parallel is True.

le-horizon (approved):
LG

PR #1199, opened 2022-02-15
PR discouted episodic return in ReplayBuffer

le-horizon:
Extracted the code for computing episodic discounted return from my branch.

Cheers,
Le

### DIFF
--- a/alf/algorithms/agent.py
+++ b/alf/algorithms/agent.py
@@ -415,6 +415,16 @@ class Agent(RLAlgorithm):
 
         exp, rl_info = self._rl_algorithm.preprocess_experience(
             exp, rollout_info.rl, batch_info)
+
+        # Add batch_info into rollout_info.
+        if hasattr(rl_info, "batch_info"):
+            batch_info = batch_info._replace(
+                replay_buffer=(), importance_weights=())
+            batch_info = alf.nest.map_structure(
+                lambda x: x.unsqueeze(1).expand(exp.reward.shape[:2]),
+                batch_info)
+            rl_info = rl_info._replace(batch_info=batch_info)
+
         return exp, rollout_info._replace(rl=rl_info)
 
     def summarize_rollout(self, experience):
--- a/alf/algorithms/agent.py
+++ b/alf/algorithms/agent.py
@@ -416,8 +416,9 @@ class Agent(RLAlgorithm):
         exp, rl_info = self._rl_algorithm.preprocess_experience(
             exp, rollout_info.rl, batch_info)
 
-        # Add batch_info into rollout_info.
-        if hasattr(rl_info, "batch_info"):
+        # Add batch_info into rollout_info to populate discounted_return.
+        if hasattr(rl_info,
+                   "batch_info") and batch_info.discounted_return != ():
             batch_info = batch_info._replace(
                 replay_buffer=(), importance_weights=())
             batch_info = alf.nest.map_structure(
--- a/alf/algorithms/agent.py
+++ b/alf/algorithms/agent.py
@@ -416,15 +416,13 @@ class Agent(RLAlgorithm):
         exp, rl_info = self._rl_algorithm.preprocess_experience(
             exp, rollout_info.rl, batch_info)
 
-        # Add batch_info into rollout_info to populate discounted_return.
+        # Expand discounted_return in batch_info to the correct shape, and
+        # populate to rl_info.
         if hasattr(rl_info,
-                   "batch_info") and batch_info.discounted_return != ():
-            batch_info = batch_info._replace(
-                replay_buffer=(), importance_weights=())
-            batch_info = alf.nest.map_structure(
-                lambda x: x.unsqueeze(1).expand(exp.reward.shape[:2]),
-                batch_info)
-            rl_info = rl_info._replace(batch_info=batch_info)
+                   "discounted_return") and batch_info.discounted_return != ():
+            discounted_return = batch_info.discounted_return.unsqueeze(
+                1).expand(exp.reward.shape[:2])
+            rl_info = rl_info._replace(discounted_return=discounted_return)
 
         return exp, rollout_info._replace(rl=rl_info)
 
--- a/alf/algorithms/ddpg_algorithm.py
+++ b/alf/algorithms/ddpg_algorithm.py
@@ -360,15 +360,6 @@ class DdpgAlgorithm(OffPolicyAlgorithm):
             priority=priority,
             extra=DdpgLossInfo(critic=critic_loss, actor=actor_loss.extra))
 
-    def preprocess_experience(self, exp, rollout_info, batch_info):
-        """Add batch_info into rollout_info.
-        """
-        batch_info = batch_info._replace(
-            replay_buffer=(), importance_weights=())
-        batch_info = alf.nest.map_structure(
-            lambda x: x.unsqueeze(1).expand(exp.reward.shape[:2]), batch_info)
-        return exp, rollout_info._replace(batch_info=batch_info)
-
     def after_update(self, root_inputs, info: DdpgInfo):
         self._update_target()
 
--- a/alf/algorithms/ddpg_algorithm.py
+++ b/alf/algorithms/ddpg_algorithm.py
@@ -42,7 +42,7 @@ DdpgState = namedtuple("DdpgState", ['actor', 'critics'])
 DdpgInfo = namedtuple(
     "DdpgInfo", [
         "reward", "step_type", "discount", "action", "action_distribution",
-        "actor_loss", "critic"
+        "actor_loss", "critic", "batch_info"
     ],
     default_value=())
 DdpgLossInfo = namedtuple('DdpgLossInfo', ('actor', 'critic'))
@@ -332,7 +332,8 @@ class DdpgAlgorithm(OffPolicyAlgorithm):
                 discount=inputs.discount,
                 action_distribution=policy_step.output,
                 critic=critic_info,
-                actor_loss=policy_step.info))
+                actor_loss=policy_step.info,
+                batch_info=rollout_info.batch_info))
 
     def calc_loss(self, info: DdpgInfo):
         critic_losses = [None] * self._num_critic_replicas
@@ -358,6 +359,15 @@ class DdpgAlgorithm(OffPolicyAlgorithm):
             loss=critic_loss + actor_loss.loss,
             priority=priority,
             extra=DdpgLossInfo(critic=critic_loss, actor=actor_loss.extra))
+
+    def preprocess_experience(self, exp, rollout_info, batch_info):
+        """Add batch_info into rollout_info.
+        """
+        batch_info = batch_info._replace(
+            replay_buffer=(), importance_weights=())
+        batch_info = alf.nest.map_structure(
+            lambda x: x.unsqueeze(1).expand(exp.reward.shape[:2]), batch_info)
+        return exp, rollout_info._replace(batch_info=batch_info)
 
     def after_update(self, root_inputs, info: DdpgInfo):
         self._update_target()
--- a/alf/algorithms/ddpg_algorithm.py
+++ b/alf/algorithms/ddpg_algorithm.py
@@ -42,7 +42,7 @@ DdpgState = namedtuple("DdpgState", ['actor', 'critics'])
 DdpgInfo = namedtuple(
     "DdpgInfo", [
         "reward", "step_type", "discount", "action", "action_distribution",
-        "actor_loss", "critic", "batch_info"
+        "actor_loss", "critic", "discounted_return"
     ],
     default_value=())
 DdpgLossInfo = namedtuple('DdpgLossInfo', ('actor', 'critic'))
@@ -333,7 +333,7 @@ class DdpgAlgorithm(OffPolicyAlgorithm):
                 action_distribution=policy_step.output,
                 critic=critic_info,
                 actor_loss=policy_step.info,
-                batch_info=rollout_info.batch_info))
+                discounted_return=rollout_info.discounted_return))
 
     def calc_loss(self, info: DdpgInfo):
         critic_losses = [None] * self._num_critic_replicas
--- a/alf/algorithms/td_loss.py
+++ b/alf/algorithms/td_loss.py
@@ -146,6 +146,16 @@ class TDLoss(nn.Module):
                 td_lambda=self._lambda)
             returns = advantages + target_value[:-1]
 
+        disc_ret = info.batch_info.discounted_return
+        if disc_ret != ():
+            with alf.summary.scope(self._name):
+                episode_ended = disc_ret > self._default_return
+                alf.summary.scalar("episodic_discounted_return_all",
+                                   torch.mean(disc_ret[episode_ended]))
+                alf.summary.scalar(
+                    "value_episode_ended_all",
+                    torch.mean(value[:-1][:, episode_ended[0, :]]))
+
         return returns
 
     def forward(self, info: namedtuple, value: torch.Tensor,
--- a/alf/algorithms/td_loss.py
+++ b/alf/algorithms/td_loss.py
@@ -146,7 +146,10 @@ class TDLoss(nn.Module):
                 td_lambda=self._lambda)
             returns = advantages + target_value[:-1]
 
-        disc_ret = info.batch_info.discounted_return
+        disc_ret = ()
+        if hasattr(info, "batch_info") and hasattr(info.batch_info,
+                                                   "discounted_return"):
+            disc_ret = info.batch_info.discounted_return
         if disc_ret != ():
             with alf.summary.scope(self._name):
                 episode_ended = disc_ret > self._default_return
--- a/alf/algorithms/td_loss.py
+++ b/alf/algorithms/td_loss.py
@@ -147,9 +147,8 @@ class TDLoss(nn.Module):
             returns = advantages + target_value[:-1]
 
         disc_ret = ()
-        if hasattr(info, "batch_info") and hasattr(info.batch_info,
-                                                   "discounted_return"):
-            disc_ret = info.batch_info.discounted_return
+        if hasattr(info, "discounted_return"):
+            disc_ret = info.discounted_return
         if disc_ret != ():
             with alf.summary.scope(self._name):
                 episode_ended = disc_ret > self._default_return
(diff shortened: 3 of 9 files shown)


alf/algorithms/td_loss.py
```diff
@@ -146,6 +146,19 @@ def compute_td_target(self, info: namedtuple, target_value: torch.Tensor):
                 td_lambda=self._lambda)
             returns = advantages + target_value[:-1]
 
+        disc_ret = ()
+        if hasattr(info, "batch_info") and hasattr(info.batch_info,
```
emailweixu: `info` does not have `batch_info` attribute.

alf/experience_replayers/replay_buffer.py
```diff
@@ -291,6 +329,13 @@ def add_batch(self, batch, env_ids=None, blocking=False):
                 self._indexed_pos[(env_ids[epi_first],
                                    self.circular(overwriting_pos[epi_first])
                                    )] = overwriting_pos[epi_first]
+                epi_last, = torch.where(step_types == ds.StepType.LAST)
```
emailweixu: move this into if block

alf/experience_replayers/replay_buffer.py
```diff
@@ -291,6 +329,13 @@ def add_batch(self, batch, env_ids=None, blocking=False):
...
+                if self._keep_episodic_info and self._record_episodic_return:
+                    # Backfill episodic returns for episodes which ended
+                    self._compute_store_episodic_return(
```
emailweixu: no need to do this if epi_last is empty.

alf/experience_replayers/replay_buffer.py
```diff
@@ -30,8 +30,13 @@
...
 
 BatchInfo = namedtuple(
-    "BatchInfo",
-    ["env_ids", "positions", "importance_weights", "replay_buffer"],
+    "BatchInfo", [
+        "env_ids",
+        "positions",
+        "importance_weights",
+        "replay_buffer",
+        "discounted_return",
```
emailweixu, line 38: I cant find any code that set a value for this.
Also, I feel BatchInfo is not the right place for this value since most of our code does not have easy access to it.
rollout_info may be a better place.

emailweixu (changes requested):

alf/algorithms/td_loss.py
le-horizon: Populated in Sac and Ddpg.

alf/experience_replayers/replay_buffer.py
le-horizon: Good catch, populated in get_batch, and then in preprocess_experience to rollout_info, and then train_step to policy_step.info, so td_loss can have access.

alf/experience_replayers/replay_buffer.py
le-horizon: done

alf/experience_replayers/replay_buffer.py
le-horizon: done

le-horizon (commented):
Thanks for the quick review.  Updated the PR for all comments, PTAL.

alf/algorithms/ddpg_algorithm.py
```diff
@@ -359,6 +360,15 @@ def calc_loss(self, info: DdpgInfo):
             priority=priority,
             extra=DdpgLossInfo(critic=critic_loss, actor=actor_loss.extra))
 
+    def preprocess_experience(self, exp, rollout_info, batch_info):
```
Haichao-Zhang: Is it possible to move this to a centralized place (e.g. Agent), so we do not need to add this for each algorithm?

le-horizon: Good point.  I've thought about it, but don't have a good way at the moment.  1) We could put it in the OffPolicyAlgorithm parent class.  But subclasses may override it and accidentally skip this preprocessing.  2) Put in Agent.  But then DdpgAlgo standalone won't work.  3) AlgorithmInfo namedtuples are algorithm specific.  Unless we extract some common parts out and put inside say OffPolicyAlgo, there will always be some algorithm specific change involved to utilize discounted_return.

Maybe let's think more about it.

le-horizon: A cleaner way is to extract the preprocess part into a Algorithm wrapper, containing the ddpg or sac algorithm as a sub-algorithm.  This will make it easy to configure any new algorithm.  Repurposing any existing config file with just command line args would be more difficult, and we'd probably have to create a separate config file for any runs using discounted_return.

@Haichao-Zhang If you are Ok with this, I can handle it perhaps in another PR.

emailweixu: I think it's ok to put it in Agent. The intention of Agent is an integration holder for different algorithms.

le-horizon: Sounds good.  Done (only when AlgoInfo has batch_info field).

Haichao-Zhang (dismissed):

Haichao-Zhang (dismissed):

alf/algorithms/agent.py
```diff
@@ -415,6 +415,17 @@ def preprocess_experience(self, root_inputs, rollout_info, batch_info):
 
         exp, rl_info = self._rl_algorithm.preprocess_experience(
             exp, rollout_info.rl, batch_info)
+
+        # Add batch_info into rollout_info to populate discounted_return.
+        if hasattr(rl_info,
+                   "batch_info") and batch_info.discounted_return != ():
```
emailweixu: how about just hasattr(rl_info, "discounted_return") and only pass discounted_return to rl_info to simply code?

alf/experience_replayers/replay_buffer.py
```diff
@@ -466,6 +517,51 @@ def _pad(self, x, env_ids):
         return ((self._current_pos[env_ids] - x - 1) //
                 self._max_length) * self._max_length + x
 
+    def _set_default_return(self, env_ids):
+        mask = torch.ones_like(env_ids, dtype=torch.bool)
+        if torch.any(mask):
+            current_pos = self._current_pos[env_ids]
+            current_pos -= 1
+            ind = (env_ids[mask], self.circular(current_pos[mask]))
```
emailweixu: Isn't env_ids[mask] same as env_ids since mask is an all-one tensor?
Same question for current_pos[mask]

alf/experience_replayers/replay_buffer.py
```diff
@@ -466,6 +517,51 @@ def _pad(self, x, env_ids):
         return ((self._current_pos[env_ids] - x - 1) //
                 self._max_length) * self._max_length + x
 
+    def _set_default_return(self, env_ids):
+        mask = torch.ones_like(env_ids, dtype=torch.bool)
+        if torch.any(mask):
```
emailweixu: Basically, this condition is True if env_ids is not empty.

alf/experience_replayers/replay_buffer.py
```diff
@@ -90,6 +99,23 @@ def __init__(self,
...
+                NOTE:
+                1) Reward transformations like ``RewardClipping.minmax=(-1, 1)'' set in
+                    ``TrainerConfig.data_transformer_ctor'' have to be set manually for
+                    the ReplayBuffer to be consistent:
+                    ``ReplayBuffer.reward_clip=(-1,1)''.
+                2) Discount ``gamma'' needs to be set consistent with ``TDLoss.gamma''.
+                3) Assumes ``keep_episodic_info'' to be True.
+            default_return (float): The default values of ``discounted_return''
+                when the episode has not ended.  It should be a very small value for
+                value target lower bounding.
```
emailweixu: Comment that its value should not be bigger than the smallest possible discounted return.
Also note that it can be 0 if reward is always non-negative.

alf/experience_replayers/replay_buffer.py
```diff
@@ -466,6 +517,51 @@ def _pad(self, x, env_ids):
...
+            ind = (env_ids[mask], self.circular(current_pos[mask]))
+            self._episodic_discounted_return[ind] = self._default_return
+
+    def _compute_store_episodic_return(self, env_ids, discount):
+        # Always pass in env_ids whose step_type is LAST, to save computation.
+        # overall mask
+        mask = torch.ones_like(env_ids, dtype=torch.bool)
+        # mask that shrinks as we are done with some env_ids
+        discounted_return = torch.zeros_like(env_ids, dtype=torch.float32)
+        if torch.any(mask):
```
emailweixu: this is always true if env_ids is not empty

alf/experience_replayers/replay_buffer.py
```diff
@@ -466,6 +517,51 @@ def _pad(self, x, env_ids):
...
+            mask &= discount == 0
+            env_ids = env_ids[mask]
+            current_pos = self._current_pos[env_ids]
+            current_pos -= 1
+            # Only store episodic return when game ends, not ending with timelimit.
```
emailweixu: We may consider bootstrapping the the episodic return with the default_return if game ends with timelimit. For example, the episode length of Atari breakout is limited to 4500, which can be reached quite often.

alf/experience_replayers/replay_buffer.py
```diff
@@ -466,6 +517,51 @@ def _pad(self, x, env_ids):
...
+        while torch.any(mask):
```
emailweixu: When num_parallel_environments is large, this can cost a lot of time since there will be lots of game endings.
It is possible to do this without using loop, similar to what MCTSAlgorithm._backup() does.

alf/algorithms/agent.py
le-horizon: There will be more to be passed, such as goal_distance in HER relabeling.

alf/experience_replayers/replay_buffer.py
le-horizon: Done

alf/experience_replayers/replay_buffer.py
le-horizon: Removed

alf/experience_replayers/replay_buffer.py
le-horizon: Removed.  Most likely copy pasted but forgot to remove.

alf/experience_replayers/replay_buffer.py
le-horizon: Removed

alf/experience_replayers/replay_buffer.py
le-horizon: Commented.

alf/experience_replayers/replay_buffer.py
le-horizon: Good point.  Done.

le-horizon (commented):
@emailweixu Addressed all comments, could you take another look?

Thanks,
Le

alf/experience_replayers/replay_buffer.py
```diff
@@ -518,49 +523,50 @@ def _pad(self, x, env_ids):
...
+                     **torch.arange(max_len, dtype=torch.float32)).unsqueeze(0)
+        epi_discounts = discounts * self._buffer.discount[all_ind]
+        # rewards = [r0, r1, r2, r3]
+        rewards = self._buffer.reward[all_ind]
+        rewards[~valid] = 0
+        epi_discounts[~valid] = 0
+        # reward[t] * discount[t - 1] is the correct discounted_return.
+        discounted_rewards = rewards[:, 1:] * epi_discounts[:, :-1]
+        # Then flip and cumsum from back of episode.
+        discounted_return = discounted_rewards.flip(1).cumsum(dim=1).flip(1)
```
emailweixu, line 582: There is a problem with this if some TimeStep.discount is 0 in the middle of the episode. cumsum will add the reward after the zero discount towards the steps before it.
Is this tested in unittest?

alf/experience_replayers/replay_buffer.py
```diff
@@ -518,49 +523,50 @@ def _pad(self, x, env_ids):
...
+        future_discounted_return = discounted_return / discounts[:, :-1]
+        valid_backfill = valid[:, :-1]
+        valid_ind = (env_ids.unsqueeze(1).expand(-1,
+                                                 max_len - 1)[valid_backfill],
+                     self.circular(epi_pos[:, :-1]).expand(
```
emailweixu: `epi_pos[:, :-1]` already has the right shape, no need to do `expand()`. And `circular(epi_pos)` was already calculated at line 546

alf/algorithms/agent.py
emailweixu: Isn't HER relabeling a DataTransformer? It can populate rollout_info by itself.

alf/experience_replayers/replay_buffer.py
le-horizon: Fixed and added tests.

le-horizon (commented):
Addressed all comments, PTAL @emailweixu 

Thanks

alf/experience_replayers/replay_buffer.py
```diff
@@ -90,6 +99,24 @@ def __init__(self,
             allow_multiprocess (bool): whether multiprocessing is supported.
             keep_episodic_info (bool): index episode start and ending positions.
                 If None, its value will be set to True if ``num_earliest_frames_ignored``>0
+            record_episodic_return (bool): If True, computes and stores episodic return for
+                every step in the episode upon episode completion.  The field
+                ``discounted_return'' stores the information.  When episodes are
```
Haichao-Zhang: '' -> ``. This is a minor issue that appeared multiple times in this section.

alf/experience_replayers/replay_buffer.py
```diff
@@ -549,6 +694,17 @@ def get_episode_begin_position(self, pos, env_ids):
...
 
+    def get_disc_0_begin_position(self, pos, env_ids):
+        """
+        Note that the discount 0 step may no longer be in the replay buffer.
+        """
+        indices = (env_ids, self.circular(pos))
+        first_step_pos = self._prev_disc_0_pos[indices]
+        # buffer_step_types = self._buffer.step_type
+        # is_first = buffer_step_types[indices] == ds.StepType.FIRST
+        # first_step_pos[is_first] = pos[is_first]
```
Haichao-Zhang: these lines can be removed now?

alf/test/main.py
```diff
@@ -16,6 +16,8 @@
 import torch
 import unittest
 
+import alf
+
```
Haichao-Zhang: These two lines can be removed?

alf/experience_replayers/replay_buffer.py
le-horizon: Ah, good catch.  Got mixed up with LaTex.

alf/experience_replayers/replay_buffer.py
le-horizon: Done

alf/test/main.py
le-horizon: Good catch.  It was from a git pick from an outdated branch..

le-horizon (commented):
Done all, thanks for the careful review Haichao.

alf/experience_replayers/replay_buffer.py
```diff
@@ -282,15 +330,32 @@ def add_batch(self, batch, env_ids=None, blocking=False):
...
-                self._store_episode_end_pos(non_first, overwriting_pos,
+                self._store_episode_end_pos(~epi_first, overwriting_pos,
                                             env_ids)
                 # 3.3. initialize episode beginning positions to itself
-                epi_first, = torch.where(step_types == ds.StepType.FIRST)
                 self._indexed_pos[(env_ids[epi_first],
                                    self.circular(overwriting_pos[epi_first])
                                    )] = overwriting_pos[epi_first]
+                if self._record_episodic_return:
+                    f_0 = (batch.discount == 0) | (
```
emailweixu: `f_0 = (batch.discount == 0) |  epi_first`

alf/experience_replayers/replay_buffer.py
```diff
@@ -535,7 +549,15 @@ def _compute_store_episodic_return(self, env_ids):
         # TODO: bootstrapping the the episodic return with the default_return
         # if game ends with timelimit. For example, the episode length of
         # Atari games is limited to 4500, which can be reached quite often.
-        first_step_pos = self.get_episode_begin_position(current_pos, env_ids)
+
+        # This is for computing from episode start, which shouldn't be used
+        # when computing episodic return every time discount = 0 step is added:
+        # first_step_pos = self.get_episode_begin_position(current_pos, env_ids)
```
emailweixu: Is this comment still valid?

alf/experience_replayers/replay_buffer.py
le-horizon: Done

alf/experience_replayers/replay_buffer.py
le-horizon: It's valid.  Let me remove this since it's no longer used.

le-horizon (commented):
Done all.  PTAL.

Thanks,
Le

Haichao-Zhang (approved):
LGTM

emailweixu (approved):

alf/algorithms/td_loss.py
```diff
@@ -146,6 +146,18 @@ def compute_td_target(self, info: namedtuple, target_value: torch.Tensor):
                 td_lambda=self._lambda)
             returns = advantages + target_value[:-1]
 
+        disc_ret = ()
+        if hasattr(info, "discounted_return"):
+            disc_ret = info.discounted_return
+        if disc_ret != ():
+            with alf.summary.scope(self._name):
+                episode_ended = disc_ret > self._default_return
```
hnyu, line 154: Where is this `self._default_return` set? By default it's not in `__init__`. If it's only set in a special case, then I think this discounted return thing shouldn't appear here in a general `TDLoss`. Maybe should drive a subclass from `TDLoss` for your purpose, or put this summary code somewhere else outside `TDLoss`.

le-horizon: Good catch.  PR#1311 will fix this, and also create LowerBoundedTDLoss as a subclass of TDLoss.
https://github.com/HorizonRobotics/alf/pull/1311/files/87b86c1a819136137ddd13bf945af8e2e9dad23b#diff-615ff6b0217f23a46d9383ae639dafa53d3ef7a47c2266d10d3656f3fd1865f3

PR #1200, opened 2022-02-15
Support fractional unroll_length

emailweixu:
Sometimes, we want to use a large value for num_updates_per_train_iter. But memory
is not large enough for that. With this change, we can achieve the same
effect by using unroll_length < 1.

### DIFF
--- a/alf/algorithms/algorithm.py
+++ b/alf/algorithms/algorithm.py
@@ -163,6 +163,7 @@ class Algorithm(AlgorithmInterface):
         if config:
             self._rl_train_every_update_steps = config.rl_train_every_update_steps
             self._rl_train_after_update_steps = config.rl_train_after_update_steps
+        self._num_updates_per_train_iter = 0
 
     def forward(self, *input):
         raise RuntimeError("forward() should not be called")
@@ -1290,6 +1291,9 @@ class Algorithm(AlgorithmInterface):
                 ``True``, it will affect the counter only if
                 ``config.update_counter_every_mini_batch=True``.
         """
+        if self._replay_buffer is None:
+            return 0
+
         config: TrainerConfig = self._config
 
         # returns 0 if haven't started training yet, when ``_replay_buffer`` is
@@ -1312,17 +1316,19 @@ class Algorithm(AlgorithmInterface):
                 mini_batch_size = config.mini_batch_size
                 if mini_batch_size is None:
                     mini_batch_size = self._replay_buffer.num_environments
+                self._num_updates_per_train_iter += config.num_updates_per_train_iter
+                num_updates_per_train_iter = int(self._num_updates_per_train_iter)
+                self._num_updates_per_train_iter -= num_updates_per_train_iter
                 if config.whole_replay_buffer_training:
                     experience, batch_info = self._replay_buffer.gather_all(
                         ignore_earliest_frames=True)
-                    num_updates = config.num_updates_per_train_iter
+                    num_updates = num_updates_per_train_iter
                 else:
                     assert config.mini_batch_length is not None, (
                         "No mini_batch_length is specified for off-policy training"
                     )
                     experience, batch_info = self._replay_buffer.get_batch(
-                        batch_size=(mini_batch_size *
-                                    config.num_updates_per_train_iter),
+                        batch_size=mini_batch_size * num_updates_per_train_iter,
                         batch_length=config.mini_batch_length)
                     num_updates = 1
             return experience, batch_info, num_updates, mini_batch_size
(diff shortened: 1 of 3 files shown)


alf/algorithms/config.py
```diff
@@ -190,8 +195,12 @@ def __init__(self,
             initial_collect_steps (int): if positive, number of steps each single
                 environment steps before perform first update. Only used
                 by ``OffPolicyAlgorithm``.
-            num_updates_per_train_iter (int): number of optimization steps for
-                one iteration. Only used by ``OffPolicyAlgorithm``.
+            num_updates_per_train_iter (int|float): number of optimization steps for
```
breakds: It might also be useful to document the motivation i.e. why people would want to use a fractional `num_updates_per_train_iter` (to overcome the GPU memory limit).

alf/algorithms/rl_algorithm.py
```diff
@@ -255,6 +258,7 @@ def __init__(self,
 
         self._original_rollout_step = self.rollout_step
         self.rollout_step = self._rollout_step
+        self._unroll_length = 0
```
breakds: Based on my understanding, this variable actually stores the "remaining fractional unroll length carried over to the next iteration", is that right? If so can we rename it so that it is more readable?

alf/algorithms/algorithm.py
```diff
@@ -1183,17 +1187,19 @@ def train_from_replay_buffer(self, update_global_counter=False):
             mini_batch_size = config.mini_batch_size
             if mini_batch_size is None:
                 mini_batch_size = self._replay_buffer.num_environments
+            self._num_updates_per_train_iter += config.num_updates_per_train_iter
+            num_updates_per_train_iter = int(self._num_updates_per_train_iter)
+            self._num_updates_per_train_iter -= num_updates_per_train_iter
```
breakds: Especially to new ALF users/readers, this change makes the logic much less readable to them. The original `num_updates = config.num_updates_per_train_iter` is very straight-forward. 

I think a more verbose but also clearer way is "if num_updates_per_train_iter is fractional, do this, otherwise do that". What do you think?

Also I do not actually understand how fractional `num_updates_per_train_iter` here effectively break the memory limit constraint (didn't follow during lunch meeting). Can you help elaborate a bit? Thanks!

hnyu: > Especially to new ALF users/readers, this change makes the logic much less readable to them. The original `num_updates = config.num_updates_per_train_iter` is very straight-forward.
> 
> I think a more verbose but also clearer way is "if num_updates_per_train_iter is fractional, do this, otherwise do that". What do you think?
> 
> Also I do not actually understand how fractional `num_updates_per_train_iter` here effectively break the memory limit constraint (didn't follow during lunch meeting). Can you help elaborate a bit? Thanks!

I think it's essentially looping from 0 to 1 with the step size of `config.num_updates_per_train_iter`. But I'm also not sure how the memory limit is avoided when `self._num_updates_per_train_iter` is close to 1 for a particular batch.

alf/algorithms/algorithm.py
```diff
@@ -1183,17 +1187,19 @@ def train_from_replay_buffer(self, update_global_counter=False):
...
-                num_updates = config.num_updates_per_train_iter
+                num_updates = num_updates_per_train_iter
             else:
                 assert config.mini_batch_length is not None, (
                     "No mini_batch_length is specified for off-policy training"
                 )
                 experience, batch_info = self._replay_buffer.get_batch(
-                    batch_size=(
-                        mini_batch_size * config.num_updates_per_train_iter),
+                    batch_size=(mini_batch_size * num_updates_per_train_iter),
```
hnyu: what if `num_updates_per_train_iter` is 0 here?

hnyu (changes requested):

alf/algorithms/algorithm.py
le-horizon: The motivation is to keep the number of training updates per rollout timestep fixed while increasing num_parallel_environments.  Without this PR, after we increase num_envs, we'll have to increase the number of updates per training iter proportionally.  This dramatically increases GPU memory consumption because there will be more mini_batches trained in one training iteration (assuming mini_batch_size is fixed as well).

hnyu: > The motivation is to keep the number of training updates per rollout timestep fixed while increasing num_parallel_environments. Without this PR, after we increase num_envs, we'll have to increase the number of updates per training iter proportionally. This dramatically increases GPU memory consumption because there will be more mini_batches trained in one training iteration (assuming mini_batch_size is fixed as well).

I think the scenario you described applies to fractional unroll_length but not fractional num_updates_per_iter here. For this case, the reason why we have fractional num_updates_per_iter must be with integer numbers, the memory is already exploded? So my question is, ``self._num_updates_per_train_iter`` will periodically reach 1 for some particular batches, which will still trigger the memory issue.

breakds (approved):

hnyu (approved):

Issue #1198, opened 2022-02-13
Implementing real NVP


hnyu:
As in the latent space policies paper ("Latent Space Policies for Hierarchical Reinforcement Learning"), they use normalizing flow sub-networks ("DENSITY ESTIMATION USING REAL NVP") to map from a higher-level action space to a lower-level action space in a hierarchical network. This enables modeling multiple action distributions in the network with just one stochastic variable, avoid marginal probability, and potentially address the "dead" (completely ignore some action subspace) lower-level action policy  otherwise.

PR #1201, opened 2022-02-16
normalizing flow networks

hnyu:
Implemented a normalizing flow realNVP in issue #1198 

See the docstring screenshot below for a brief introduction to realNVP. So far we assume a flattened rank-1 tensor to be transformed by realNVP. However, the conditional input can be very complex.

image

Also based on this, implemented a LatentActorDistributionNetwork class which can wrap any actor distribution network with a normalizing flow network. 

I tried defining a latent actor distribution network used by SAC and ran it on Pendulum, and got reasonable performance: 

sac_pendulum_latent_actor


alf/networks/normalizing_flow_networks.py
```diff
@@ -0,0 +1,408 @@
...
+        if self._conditional_inputs:
+            y, z = yz
+        else:
+            y, z = yz, None
+        transform = self.make_invertible_transform(z)
+        return transform.inv(y), ()
+
+
+@alf.configurable
+class realNVPNetwork(NormalizingFlowNetwork):
```
emailweixu: real => Real ?

alf/networks/normalizing_flow_networks.py
```diff
@@ -0,0 +1,408 @@
...
+        r"""Express the network forward computation as an invertible PyTorch
+        ``Transform``. This overall transformation can be a composed one chaining
+        many transformation layers.
+
+        Args:
+            conditional_inputs: an optional nested conditional inputs that
+                condition the mapping :math:`x \rightarrow y`.
+
+        Returns:
+            torch.td.Transform: an invertible transform
```
emailweixu: return type can be annotated: "def func_name() -> return_type:"

alf/networks/normalizing_flow_networks.py
```diff
@@ -0,0 +1,408 @@
...
+        """
+        if not self._use_transform_cache:
+            return self._make_invertible_transform(conditional_inputs)
+        old_t, old_z = self._cached_transform
+        if conditional_inputs is old_z and old_t is not None:
```
emailweixu, line 105: How is it possible that conditional_inputs is old_z and old_t is None?

alf/networks/normalizing_flow_networks.py
```diff
@@ -0,0 +1,408 @@
...
+            return old_t
+        t = self._make_invertible_transform(conditional_inputs)
+        self._cached_transform = (t, conditional_inputs)
+        return t
+
+    def _make_invertible_transform(
+            self, conditional_inputs: alf.nest.NestedTensor = None):
```
emailweixu, line 112: This function should be dcumented since subclass needs to implement this function.

emailweixu (commented):
One general issue is that it is not compatible with: `extract_spec()`, `distributions_to_params` and `params_to_distributions()`.
For example, in RLAlgorithm.unroll(), rollout_info_spec is extracted from the result of one rollout_step(), where the Transform object is stored in the spec. And the same Transform object is used when params_to_distributions() is used. But in the case of NF, different Transform object is created at different rollout_step.

Similar problem for `Algorithm_collect_train_info_sequentially()`

hnyu:
> One general issue is that it is not compatible with: `extract_spec()`, `distributions_to_params` and `params_to_distributions()`. For example, in RLAlgorithm.unroll(), rollout_info_spec is extracted from the result of one rollout_step(), where the Transform object is stored in the spec. And the same Transform object is used when params_to_distributions() is used. But in the case of NF, different Transform object is created at different rollout_step.
> 
> Similar problem for `Algorithm_collect_train_info_sequentially()`

Right, the problem is that currently the transform objects are always using the conditional inputs at the first step, which is equivalent to not condition on anything.

emailweixu:
> > One general issue is that it is not compatible with: `extract_spec()`, `distributions_to_params` and `params_to_distributions()`. For example, in RLAlgorithm.unroll(), rollout_info_spec is extracted from the result of one rollout_step(), where the Transform object is stored in the spec. And the same Transform object is used when params_to_distributions() is used. But in the case of NF, different Transform object is created at different rollout_step.
> > Similar problem for `Algorithm_collect_train_info_sequentially()`
> 
> Right, the problem is that currently the transform objects are always using the conditional inputs at the first step, which is equivalent to not condition on anything.

`distributions_to_params` needs to be modified to extract the conditional inputs as params.

alf/networks/normalizing_flow_networks.py
hnyu: When conditional_inputs is None and the default value of old_z is also None

hnyu: > How is it possible that conditional_inputs is old_z and old_t is None?

When conditional_inputs is None and the default value of old_z is also None

alf/networks/normalizing_flow_networks.py
hnyu: > real => Real ?

'realNVP' is what the paper uses to call their method.

Haichao-Zhang:
> maybe we can change the implementation of AC to compute all probs-related values in rollout_step()?

We already support prob computation at rollout_step: #1055 ?

alf/networks/normalizing_flow_networks.py
emailweixu: This is an interface which the subclass needs to implement. So its expected behavior should be documented.

alf/networks/normalizing_flow_networks.py
```diff
@@ -0,0 +1,538 @@
...
+        xy: torch.Tensor,
+        z_spec: alf.nest.NestedTensorSpec = None,
+        z: alf.nest.NestedTensor = None
+) -> Tuple[alf.nest.NestedTensor, alf.utils.tensor_utils.BatchSquash]:
+    """A general function for adjusting the shapes of inputs and conditional inputs
+    of a conditional flow, prepared for a forward of a network next. Some networks
+    assume only one batch dim, for example, when using ``alf.layers.Reshape()``.
+
+    The reason why we need to do this is because the flow transform can be called
+    with an arbitrary batch shape of ``x`` or ``y``, for example, with computing
```
emailweixu: "with computing" => "when computing"

alf/networks/normalizing_flow_networks.py
```diff
@@ -0,0 +1,538 @@
...
+    a loss with time dimension, or sampling a particular shape from a distribution.
+
+    Args:
+        xy_spec: tensor spec of ``x`` (forward) or ``y`` (inverse)
+        xy:
+        z_spc: tensor spec of ``z`` (conditional input)
```
emailweixu: z_spc => z_spec

alf/networks/normalizing_flow_networks.py
```diff
@@ -0,0 +1,538 @@
...
+
+        if z_outer_rank > 1:
+            z = alf.utils.tensor_utils.BatchSquash(z_outer_rank).flatten(z)
+
+        B = alf.nest.get_nest_batch_size(z)
+        if B < ret.shape[0]:
+            # When the total outer dim of ``z`` is smaller than that of ``xy``,
+            # it means that multiple samples of ``xy`` correspond to one ``z``,
+            # so we need to repeat ``z``'s batch dim.
+            assert ret.shape[0] % B == 0, (
```
emailweixu: Isn't the assert at line 398 enough to ensure this?

alf/networks/normalizing_flow_networks.py
```diff
@@ -0,0 +1,538 @@
...
+        new_y = (y - trans) / self._scale_nonlinear(scale)
+        x = y * self._b + new_y * (~self._b)
+        return x
+
+    def log_abs_det_jacobian(self, x, y):
+        r"""The Jacobian is always a triangular matrix (or can be converted into by
+        row swapping). The diagonal elements are :math:`\mathbb{I}_d` and
+        :math:`\text{diag}(\exp(scale(x_{1:d};z)))`, where the first :math:`d` dims
+        are assumed to be selected by the mask ``self._b``.
+        """
```
emailweixu, line 532: From the name of this function, it calculates log(det (Jacobian)), which should be a scalar quantity (with appropriate batch dimension). But the implementation is not so.

hnyu: > From the name of this function, it calculates log(det (Jacobian)), which should be a scalar quantity (with appropriate batch dimension). But the implementation is not so.

This is just the name convention of PyTorch. For other transforms like StableTanh the returned value of this function can also be a vector. PyTorch will handle this smartly when compute ``log_prob()`` for a transformed distribution.

alf/networks/normalizing_flow_networks.py
```diff
@@ -0,0 +1,538 @@
...
+        ret = (ret, z)
+
+    return ret, bs
+
+
+class _RealNVPTransform(td.Transform):
+    """This class implements each transformation layer of ``RealNVPNetwork``. For
+    details, refer to the docstring of ``RealNVPNetwork``.
+    """
+    domain = td.constraints.real
```
emailweixu: domain and codomain should be td.constraints.real_vector

alf/networks/normalizing_flow_networks.py
emailweixu: It depends on its domain. If the domain is real, it a transform over a single real value and the returned Tensor has same shape as x. But if the domain is real_vector, the ndim of the returned Tensor should 1 less than that of x.

hnyu: > It depends on its domain. If the domain is real, it a transform over a single real value and the returned Tensor has same shape as x. But if the domain is real_vector, the ndim of the returned Tensor should 1 less than that of x.

Right. It does depend on the domain. Currently I have real scalar for RealNVPTransform. But I think there is no difference between the final probs of the two choices?

hnyu: > It depends on its domain. If the domain is real, it a transform over a single real value and the returned Tensor has same shape as x. But if the domain is real_vector, the ndim of the returned Tensor should 1 less than that of x.

The tricky part for RealNVPTransform is that now the domain can be an arbitrary shape. For example, x's spec might be rank-2. So it's difficult to define its domain shape.

emailweixu: Then perhaps domain and codomain can be set at `__init__` as `td.constraints.independent(td.constraints.real, spec.ndim)`

hnyu: > Then perhaps domain and codomain can be set at `__init__` as `td.constraints.independent(td.constraints.real, spec.ndim)`

Sounds good. I've made the change.

hnyu:
After discussions with @Haichao-Zhang and @breakds , I've added a note talking about computing probabilities of transformed distributions.

docs/notes/compute_probs_of_transformed_dist.rst
```diff
@@ -0,0 +1,227 @@
...
+    \end{array}
+
+with :math:`x=g^{-1}(g(x))`, not including :math:`\mu` in the resulting p.d.f.
+Thus PG won't be able to adjust :math:`\mu`.
+
+Transform caches
+^^^^^^^^^^^^^^^^
+
+A cache is usually helpful for a bijective transform where the inverse is computationally
+expensive or numerically instable. The way PyTorch uses a cache is that for an
```
emailweixu: instable => unstable

docs/notes/compute_probs_of_transformed_dist.rst
```diff
@@ -0,0 +1,227 @@
...
+      - :math:`x \leftarrow x`
+      - :math:`x \leftarrow g^{-1}(g(x))` (*)
+
+'*' means potential numerical issues when inverting.
+
+A composition of transforms
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Suppose :math:`g` is now a composition of transforms
+:math:`g=g_N\circ \ldots g_k \ldots \circ g_1`, where :math:`g_k` is the first transform
```
emailweixu: where `g_k` is the *last* transform?

docs/notes/compute_probs_of_transformed_dist.rst
```diff
@@ -0,0 +1,227 @@
...
+that contains trainable parameters. Then for PG, it is completely fine to do either
+of the following things:
+
+
+1. Simply detach :math:`y` and invalidate all caches.
+
+2. Don't detach :math:`y`, turn on caches for transforms :math:`g_N, \ldots, g_{k-1}`,
```
emailweixu: `g_N, \ldots, g_{k+1}`

docs/notes/compute_probs_of_transformed_dist.rst
```diff
@@ -0,0 +1,227 @@
...
+   and detach the input :math:`g^{-1}_{k-1}\circ \ldots \circ g^{-1}_N(y)`
```
emailweixu: `g^{-1}_{k+1} ... `

docs/notes/compute_probs_of_transformed_dist.rst
```diff
@@ -0,0 +1,227 @@
...
+
+In the hindsight, we don't detach the action and exploit the cache for PG to avoid
+inverting. When the transform :math:`g=g_N\circ\ldots\circ g_2` has no trainable parameters
+(e.g., ``StableTanh``), we have parameters only exist in :math:`p_{\theta}(x)`. It
+follows
+
+.. math::
+
+    \begin{array}{rcl}
+        \log p_Y(y) &=& \log p_{\theta}(g^{-1}(y)) - \log \Big|\frac{\partial g(x)}{\partial x}\big|_{x=x}\Big|\\
```
emailweixu: `\Big|\frac{\partial g(x)}{\partial x}\big|_{x=g^{-1}(y)}`

docs/notes/compute_probs_of_transformed_dist.rst
```diff
@@ -0,0 +1,227 @@
...
+
+Action transformations as environment wrappers
+----------------------------------------------
+
+ALF's AC and PPO algorithms always detach the action from ``sample()`` for PG loss,
+without checking if the transforms have trainable parameters or not. This simplicity
+invalidate caches and sometimes causes numerical issues even when all the transforms
+do not have trainable parameters.
+
+In the hindsight, we don't detach the action and exploit the cache for PG to avoid
```
emailweixu: What does "In the hindsight" mean here?

docs/notes/compute_probs_of_transformed_dist.rst
hnyu: > where `g_k` is the _last_ transform?

I added a note (from left to right).

docs/notes/compute_probs_of_transformed_dist.rst
hnyu: > What does "In the hindsight" mean here?

Rephrased for clarity

emailweixu (approved):

PR #1203, opened 2022-02-17
fix bugs in Bayesian linear regression test

runjerry:
Fix bugs in transformed Gaussian covariance in Bayesian linear regression test.


runjerry:
merged into PR 1205

PR #1205, opened 2022-02-17
Pr param bn ln layer

runjerry:
- add 1d and 2d ParamBatchNorm, ParamLayerNorm layers
- make configurations of param layers consistent with alf.nn.layeres
- update param layers and param networks accordingly
- fixed bugs in transformed Gaussian covariance in Bayesian linear regression test.


alf/norm_layers.py
```diff
@@ -0,0 +1,936 @@
+# Copyright (c) 2021 Horizon Robotics and ALF Contributors. All Rights Reserved.
```
emailweixu, line 1: Did you use "git mv batch_norm.py norm_layers.py"? With that, the difference with the original batch_norm.py can be shown and previous change history is still maintained.

alf/layers.py
```diff
@@ -1172,6 +1176,10 @@ def __init__(self,
             self._bn = bn_ctor(out_channels)
         else:
             self._bn = None
+        if use_ln:
+            self._ln = nn.GroupNorm(1, out_channels)
```
emailweixu, line 1180: why not use nn.LayerNorm?

alf/norm_layers.py
```diff
@@ -0,0 +1,936 @@
...
+        """
+        super().__init__()
+        self._num_features = num_features
+        self._n_groups = n_groups
+        self._eps = eps
+        self._momentum = momentum
+        self._track_running_stats = track_running_stats
+        self._set_weight(torch.ones(n_groups, self.weight_length))
+        self._set_bias(torch.zeros(n_groups, self.bias_length))
+        if track_running_stats:
```
emailweixu, line 405: I remember we had a discussion about what is best way of handling BatchNorm for the case of GPVI. For a randomly generated network, the running statistics does not make much sense. And using a single set of BN parameter across all different generated nets might be better?

alf/layers.py
```diff
@@ -1959,7 +2048,10 @@ def forward(self, inputs):
         else:
             res = torch.bmm(inputs, self._weight.transpose(1, 2))
         res = res.transpose(0, 1)  # [B, n, D]
-        res = res.squeeze(1)  # [B, D] if n=1
+        if self._use_norm is not None:
+            res = self._norm(res)
```
emailweixu, line 2052: why do you have keep_group_dim for ParamConv2D but not for ParamFC?

alf/layers.py
```diff
@@ -1959,7 +2048,10 @@ def forward(self, inputs):
...
+        else:
+            res = res.squeeze(1)  # [B, D] if n=1
```
emailweixu, line 2054: Why is this under "else"?

runjerry:
Replaced by the new PR #1211

PR #1206, opened 2022-02-18
Hybrid RL training pipeline

### LINKED ISSUE #1207: Hybrid RL training pipeline

Haichao-Zhang:
This PR implements Hybrid RL training pipeline.

### 1. Usage
The usage is very straightforward, simply providing a valid file path to ``offline_buffer_dir`` in the config on top of existing methods (e.g. SAC).
For example:

`alf.config(
    "TrainerConfig",
    offline_buffer_dir="/data_collection/train/algorithm/ckpt-replay_buffer"
)`

### 2. Implementation
The implementation of new algorithms is also simple, simply implementing two additional functions: ``train_step_offline`` and ``calc_loss_offline`` for customized training and loss calculation based on the offline data provided as input to the functions.
``train_step`` and ``calc_loss`` can also be customized if necessary.


alf/examples/hybrid_rl/hybrid_sac_pendulum.gin
```diff
@@ -0,0 +1,23 @@
...
+
+# we can control when to start the online RL training with
+# rl_train_after_update_steps
+TrainerConfig.rl_train_after_update_steps=0
+
+# we can control the update frequency of the online RL training
+# wrt offline RL by setting rl_train_every_update_steps.
+TrainerConfig.rl_train_every_update_steps=1
+
+# Note that the following file used is just a dymmy replay buffer
```
emailweixu: dymmy => dummy

alf/algorithms/algorithm.py
```diff
@@ -1109,6 +1157,68 @@ def calc_loss(self, info):
...
+                training procedures at different phases.
+        Returns:
+            LossInfo: loss at each time step for each sample in the
+                batch. The shapes of the tensors in loss info should be
+                :math:`(T, B)`.
+        """
+        try:
+            return self.calc_loss(info_offline)
+        except:
+            # the default calc_loss does not compatible with the
```
emailweixu: does not => is not

alf/algorithms/rl_algorithm.py
```diff
@@ -563,20 +582,195 @@ def _train_iter_off_policy(self):
...
 
         with record_time("time/after_train_iter"):
-            train_info = experience.rollout_info
-            experience = experience._replace(rollout_info=())
+            if experience is not None:
+                train_info = experience.rollout_info
+                experience = experience._replace(rollout_info=())
+            else:
+                experience = None
+                train_info = None
```
emailweixu, line 602: need to update the documentation of after_train_iter() in algorithm_interface.py

alf/algorithms/rl_algorithm.py
```diff
@@ -563,20 +582,195 @@ def _train_iter_off_policy(self):
...
             alf.summary.increment_global_counter()
 
-        with torch.set_grad_enabled(config.unroll_with_grad):
-            with record_time("time/unroll"):
-                self.eval()
-                experience = self.unroll(config.unroll_length)
-                self.summarize_rollout(experience)
-                self.summarize_metrics()
+        if self._count == 0 or self._count > self._rl_train_after_update_steps:
+            # 1) RL training
```
emailweixu: why "1)" here?

alf/algorithms/rl_algorithm.py
```diff
@@ -563,20 +582,195 @@ def _train_iter_off_policy(self):
         if not config.update_counter_every_mini_batch:
             alf.summary.increment_global_counter()
 
-        with torch.set_grad_enabled(config.unroll_with_grad):
-            with record_time("time/unroll"):
-                self.eval()
-                experience = self.unroll(config.unroll_length)
-                self.summarize_rollout(experience)
-                self.summarize_metrics()
+        if self._count == 0 or self._count > self._rl_train_after_update_steps:
```
emailweixu: what is the condition "self._count==0" for?
self._count is not checkpointed. So training from checkpoint will have different behavior.

alf/algorithms/algorithm.py
```diff
@@ -1179,35 +1295,86 @@ def train_from_replay_buffer(self, update_global_counter=False):
...
+                        "replay buffer training is not supported")
+
+                    assert config.mini_batch_length is not None, (
+                        "No mini_batch_length is specified for off-policy training"
+                    )
+                    experience, batch_info = self._replay_buffer.get_batch(
+                        batch_size=(mini_batch_size *
+                                    config.num_updates_per_train_iter),
+                        batch_length=config.mini_batch_length)
+                    num_updates = 1
```
emailweixu: Is there any difference between this segment and 1301-1317? Perhaps use a local function _replay() to do both?

alf/algorithms/algorithm.py
```diff
@@ -1533,6 +1714,214 @@ def _update(self, experience, batch_info, weight):
...
+        info = alf.nest.map_structure(
+            lambda x: x.reshape(length, batch_size, *x.shape[1:]), info)
+        # info = dist_utils.params_to_distributions(info, self.train_info_spec)
+        return info
+
+    def _train_hybrid_experience(
+            self, experience, batch_info, offline_experience,
+            offline_batch_info, num_updates, mini_batch_size,
+            mini_batch_length, update_counter_every_mini_batch):
+        """Train using both experience (if available) and aux_experience.
```
emailweixu: aux_experience => offline_experience

alf/algorithms/algorithm.py
```diff
@@ -1533,6 +1714,214 @@ def _update(self, experience, batch_info, weight):
...
+        if experience is not None:
+            experience, batch_info, length, mini_batch_length, batch_size = self._prepare_experience_data(
+                experience, self.experience_spec, batch_info,
+                mini_batch_length, self._replay_buffer)
+
+            if self._processed_experience_spec is None:
+                self._processed_experience_spec = dist_utils.extract_spec(
+                    experience, from_dim=2)
+
+        # TODO: use a different mini_batch_length for aux
```
emailweixu: aux => offline

alf/algorithms/algorithm.py
```diff
@@ -1533,6 +1714,214 @@ def _update(self, experience, batch_info, weight):
...
+                     offline_batch,
+                     offline_binfo,
+                     weight=alf.nest.get_nest_size(offline_batch, 1) /
+                     mini_batch_size)
+                if do_summary:
+                    if exp:
+                        self.summarize_train(exp, train_info, loss_info,
+                                             params)
+                    # TODO: separate summarization
+                    self.summarize_train(offline_exp, offline_train_info,
```
emailweixu: Simply adding `with summary_scope('offline')` should work.

alf/algorithms/algorithm.py
```diff
@@ -1533,6 +1714,214 @@ def _update(self, experience, batch_info, weight):
...
+
+            loss_info = self.calc_loss(train_info)
+        else:
+            train_info = None
+            loss_info = None
+            RL_train = False
+            # some pre-training in the initial phase
+            pre_train = True
+
+        # 2) aux training
```
emailweixu: aux => offline

alf/algorithms/algorithm.py
```diff
@@ -1533,6 +1714,214 @@ def _update(self, experience, batch_info, weight):
...
+            pre_train = False
+            if self._config.temporally_independent_train_step or length == 1:
+                train_info = self._collect_train_info_parallelly(experience)
+            else:
+                train_info = self._collect_train_info_sequentially(experience)
+
+            experience = dist_utils.params_to_distributions(
+                experience, self.processed_experience_spec)
+
+            loss_info = self.calc_loss(train_info)
```
emailweixu: use `_collect_train_info_and_loss_info()`

alf/algorithms/algorithm.py
```diff
@@ -1533,6 +1714,214 @@ def _update(self, experience, batch_info, weight):
...
+                                                batch_info.positions, priority)
+            if self._debug_summaries and alf.summary.should_record_summaries():
+                with alf.summary.scope("PriorityReplay"):
+                    summary_utils.add_mean_hist_summary(
+                        "new_priority", priority)
+                    summary_utils.add_mean_hist_summary(
+                        "old_importance_weight", batch_info.importance_weights)
+        else:
+            assert batch_info is None or batch_info.importance_weights == (), (
+                "Priority replay is enabled. But priority is not calculated.")
```
emailweixu: Share this segment with those in _update()

alf/algorithms/algorithm.py
```diff
@@ -1027,6 +1035,29 @@ def update_with_gradient(self,
...
+        loss_info = self._aggregate_loss(loss_info, valid_masks, batch_info)
+
+        all_params = self._backward_and_gradient_update(loss_info, weight)
+
+        return loss_info, all_params
+
+    def _aggregate_loss(self, loss_info, valid_masks=None, batch_info=None):
+        """Computed aggregated loss.
+
+        Update parameters using the gradient with respect to ``loss_info``.
```
emailweixu: wrong comments

alf/algorithms/algorithm.py
```diff
@@ -1050,11 +1081,28 @@ def update_with_gradient(self,
...
 
+    def _backward_and_gradient_update(
+            self,
+            loss_info,
+            weight=1.0,
+    ):
+        """Do backward and gradient update to all the trainable parameters.
+
+        Args:
+            loss_info (LossInfo): loss with shape :math:`(T, B)` (except for
```
emailweixu: Seems no need to use LossInfo structure. Just a scalar loss should be fine.

alf/algorithms/algorithm.py
```diff
@@ -1050,11 +1081,28 @@ def update_with_gradient(self,
...
+                ``loss_info.scalar_loss``)
+            weight (float): weight for this batch. Loss will be multiplied with
```
emailweixu: This can be omitted if using a scalar for loss because it can be multiplied with the scalar loss by the caller.

alf/algorithms/algorithm.py
```diff
@@ -1179,35 +1295,86 @@ def train_from_replay_buffer(self, update_global_counter=False):
...
+                    mini_batch_size,
+                    config.mini_batch_length,
+                    (config.update_counter_every_mini_batch
+                     and update_global_counter),
+                    whole_replay_buffer_training=config.
+                    whole_replay_buffer_training)
+        else:
+            # hybrid training scheme
+            RL_train = False
+            if self._count == 0 or (
```
emailweixu: why this condition?

alf/algorithms/rl_algorithm.py
```diff
@@ -563,20 +582,195 @@ def _train_iter_off_policy(self):
...
+
+            def _get_full_key(dict, partial_key):
+                full_key = next((key for key in dict if partial_key in key),
+                                None)
+                assert full_key is not None, (
+                    "key containing {} "
+                    "is not found.".format(partial_key))
+                return full_key
+
+            k1 = _get_full_key(buffer_dict, "time_step|reward")
```
emailweixu: what is the use of k1?

alf/algorithms/rl_algorithm.py
```diff
@@ -563,20 +582,195 @@ def _train_iter_off_policy(self):
...
+                buffer_dict[step_type_key], from_dim=2)
+            reward_spec = dist_utils.extract_spec(
+                buffer_dict[reward_key], from_dim=2)
+            discount_spec = dist_utils.extract_spec(
+                buffer_dict[discount_key], from_dim=2)
+
+            env_id_spec = dist_utils.extract_spec(
+                buffer_dict[env_id_key], from_dim=2)
+
+            time_step = TimeStep(
```
emailweixu: time_step=> time_step_spec

alf/algorithms/rl_algorithm.py
```diff
@@ -563,20 +582,195 @@ def _train_iter_off_policy(self):
...
+                # The shape is [env_batch_size, mini_batch_length, ...], where
+                # mini_batch_length denotes the length of the mini_batch
+                for t in range(exp.step_type.shape[1]):
+                    bat = alf.nest.map_structure(lambda x: x[:, t, ...], exp)
+                    self._offline_replay_buffer.add_batch(bat, bat.env_id)
+            else:
+                raise ValueError(
+                    "Unsupported outer rank %s of `exp`" % outer_rank)
+
+        _load_data(exp)
```
emailweixu, line 772: why can't you just use replay_buffer.load_state_dict()?

emailweixu (changes requested):

alf/algorithms/algorithm.py
```diff
@@ -176,6 +179,11 @@ def set_path(self, path):
     def on_policy(self):
         return self._is_on_policy
 
+    @property
+    def has_offline(self):
+        # whether has offline data
+        return self._has_offline
```
hnyu: `self._has_offline` is undefined for non-RL algorithm

alf/algorithms/algorithm.py
```diff
@@ -161,6 +160,10 @@ def __init__(self,
         if optimizer:
             self._optimizers.append(optimizer)
         self._is_on_policy = is_on_policy
+        if config:
+            self._rl_train_every_update_steps = config.rl_train_every_update_steps
+            self._rl_train_after_update_steps = config.rl_train_after_update_steps
+        self._count = 0
```
hnyu: Should be a better name. Iteration? Update? what is its relationship with global counter?

alf/algorithms/algorithm.py
```diff
@@ -1533,6 +1714,214 @@ def _update(self, experience, batch_info, weight):
...
+                        binfo = alf.nest.map_structure(
+                            lambda x: x[batch_indices] if isinstance(
+                                x, torch.Tensor) else x, batch_info)
+                    else:
+                        binfo = None
+                    batch = _make_time_major(batch)
+                    batch = alf.nest.utils.convert_device(batch)
+                else:
+                    batch = None
+                    binfo = None
```
hnyu: This major code block `updates x mini-batches` looks very similar to the one in `_train_experience`. Maybe we can extract the common logic to decrease the maintenance effort in the future.

alf/algorithms/rl_algorithm.py
```diff
@@ -341,6 +357,9 @@ def summarize_rollout(self, experience):
         Args:
             experience (Experience): experience collected from ``rollout_step()``.
         """
+        if not alf.summary.should_record_summaries():
```
hnyu: I think this condition is already checked by the summary wrapper?

alf/data_structures.py
```diff
@@ -481,3 +481,15 @@ def make_experience(time_step: TimeStep, alg_step: AlgStep, state):
...
+
+# some basic structures used by offline replay buffer
+# reconstruction
+Info = namedtuple("Info", [
+    'rl',
+    "rewards",
+    "repr",
+])
+
+RLInfo = namedtuple("RLInfo", [
```
hnyu: These two namedtuple names are too general

hnyu (changes requested):
I feel the pipeline has been bloated by having some duplicate code blocks.

alf/examples/hybrid_rl/hybrid_sac_pendulum.gin
Haichao-Zhang: changed

alf/algorithms/algorithm.py
Haichao-Zhang: changed

alf/algorithms/rl_algorithm.py
Haichao-Zhang: Added documentation

alf/algorithms/rl_algorithm.py
Haichao-Zhang: There used to be a 2) here as well but was lifted to algorithm.py during update. Removed this line of comment.

alf/algorithms/rl_algorithm.py
Haichao-Zhang: Used ``self._count==0`` for constructing the online replay buffer at the very beginning. 
Avoided this by adding a condition checking on the presence of buffer before usage. 
Still thinking about what is a good way to handle training from checkpoint scenario.

alf/algorithms/algorithm.py
Haichao-Zhang: Good point. Extracted to a local function.

alf/algorithms/algorithm.py
Haichao-Zhang: Changed

alf/algorithms/algorithm.py
Haichao-Zhang: Changed.

alf/algorithms/algorithm.py
Haichao-Zhang: Added

alf/algorithms/algorithm.py
Haichao-Zhang: changed to using this function

alf/algorithms/algorithm.py
Haichao-Zhang: Changed to shared with update

alf/algorithms/algorithm.py
Haichao-Zhang: updated

alf/algorithms/algorithm.py
Haichao-Zhang: Correct. Changed to scalar loss input.

alf/algorithms/algorithm.py
Haichao-Zhang: Now the weight is moved out side of the function and multiplied to the loss before calling this function.

alf/algorithms/rl_algorithm.py
Haichao-Zhang: This line was a debugging line. Removed.

alf/algorithms/rl_algorithm.py
Haichao-Zhang: Changed

alf/algorithms/rl_algorithm.py
Haichao-Zhang: I tried this first but was not able to load the buffer correctly with this.
For example, I tried again in commit 3cee63b0a2f5f92fbfe7910459b0ab341b3d0956, the buffer cannot be loaded correctly.

alf/algorithms/algorithm.py
Haichao-Zhang: Changed to return False for non-RL algorithms.

alf/algorithms/algorithm.py
Haichao-Zhang: Changed to update_count. Originally thought about using global counter, but this counter seems to be impacted by a update_global_counter flag (although we set it to be True when calling train_from_replay_buffer).

alf/algorithms/rl_algorithm.py
Haichao-Zhang: Removed

alf/data_structures.py
Haichao-Zhang: This is indeed intended to be a very general and basic structure which is shared among different algorithms.

alf/algorithms/rl_algorithm.py
Haichao-Zhang: Removed ``_count`` and replaced with the existing global_step.

alf/algorithms/algorithm.py
Haichao-Zhang: Removed _count.

alf/algorithms/algorithm.py
Haichao-Zhang: Now more code sharing with a shared function.

Haichao-Zhang:
Updates:
1) more code sharing with common functions
2) example of SAC hybrid training with Murzo buffer (``hybrid_sac_pendulum_muzero_buffer.gin``)
3) note on hybrid/offline RL pipeline (``hybrid_and_offline_rl_training.rst``)

alf/algorithms/algorithm.py
```diff
@@ -1832,69 +1835,53 @@ def _make_time_major(nest):
...
-                                                batch_info.positions, priority)
-            if self._debug_summaries and alf.summary.should_record_summaries():
-                with alf.summary.scope("PriorityReplay"):
-                    summary_utils.add_mean_hist_summary(
-                        "new_priority", priority)
-                    summary_utils.add_mean_hist_summary(
-                        "old_importance_weight", batch_info.importance_weights)
-        else:
-            assert batch_info is None or batch_info.importance_weights == (), (
-                "Priority replay is enabled. But priority is not calculated.")
```
emailweixu: why did you remove this?

alf/algorithms/agent_helpers.py
```diff
@@ -104,7 +104,11 @@ def accumulate_algorithm_rewards(rewards, weights, names, summary_prefix,
             summarize_fn(os.path.join(summary_prefix, "overall"), reward)
         return reward
 
-    def accumulate_loss_info(self, algorithms, train_info):
+    def accumulate_loss_info(self,
+                             algorithms,
+                             train_info,
+                             offline=False,
+                             *offline_args):
```
emailweixu: where is this new argument used? I cannot find it being used in this PR.

alf/algorithms/algorithm.py
Haichao-Zhang: > why did you remove this?

It was extracted into the ``_update_priority()`` function in favor of code sharing with other places and was called here in line 1866. Also added priority update to to offline buffer.

alf/algorithms/agent_helpers.py
Haichao-Zhang: Use for containing additional args (currently only ``pre_train``) for offline RL. Explicitly replaced it with ``pre_train``.

alf/algorithms/algorithm.py
```diff
@@ -1289,13 +1286,16 @@ def train_from_replay_buffer(self, update_global_counter=False):
         """
         config: TrainerConfig = self._config
 
-        if self._replay_buffer.total_size < config.initial_collect_steps:
-            # returns 0 if haven't started training yet; throughput will be 0
+        # returns 0 if haven't started training yet, when ``_replay_buffer`` is
+        # not None and the number of samples in the buffer is less than
+        # ``initial_collect_steps``; throughput will be 0 in this phase.
+        if (self._replay_buffer and
```
hnyu, line 1294: because this is inside ``train_from_replay_buffer``, then the replay buffer should always not be None?

alf/algorithms/algorithm.py
```diff
@@ -1531,6 +1708,237 @@ def _update(self, experience, batch_info, weight):
...
+
+    def _train_hybrid_experience(
+            self, experience, batch_info, offline_experience,
+            offline_batch_info, num_updates, mini_batch_size,
+            mini_batch_length, update_counter_every_mini_batch):
+        """Train using both experience (if available) and offline_experience.
+        We assume that experience can be None.
+        """
+        if experience is not None:
+            experience, processed_exp_spec, batch_info, length, mini_batch_length, batch_size = self._prepare_experience_data(
```
hnyu: This line is too long

alf/algorithms/algorithm.py
```diff
@@ -1531,6 +1708,237 @@ def _update(self, experience, batch_info, weight):
...
+                experience, self.experience_spec, batch_info,
+                mini_batch_length, self._replay_buffer)
+
+            self._processed_experience_spec = processed_exp_spec
+
+        # TODO: use a different mini_batch_length for offline training
+        offline_experience, _, offline_batch_info, length, mini_batch_length, batch_size = self._prepare_experience_data(
```
hnyu: This line is too long

alf/algorithms/algorithm.py
```diff
@@ -1531,6 +1708,237 @@ def _update(self, experience, batch_info, weight):
...
+                                             params)
+                    with alf.summary.scope("offline"):
+                        self.summarize_train(offline_exp, offline_train_info,
+                                             offline_loss_info, None)
+
+        train_steps = 2 * batch_size * mini_batch_length * num_updates
+        return train_steps
+
+    def _hybrid_update(self, experience, batch_info, offline_experience,
+                       offline_batch_info, weight):
```
hnyu, line 1881: Why do you want to use the RL experience and off-experience both at the same time? One alternative way is to treat RL experience and off-line experience as the same arg variable but different values for a unified function call. First accumulate both losses and then call the optimizer step only once.

alf/algorithms/algorithm.py
```diff
@@ -1531,6 +1708,237 @@ def _update(self, experience, batch_info, weight):
...
+            if mini_batch_size < batch_size:
+                # here we use the cpu version of torch.randperm(n) to generate
+                # the permuted indices, as the cuda version of torch.randperm(n)
+                # seems to have a bug when n is a large number, generating
+                # negative or very large values that cause out of bound kernel
+                # error: https://github.com/pytorch/pytorch/issues/59756
+                indices = alf.nest.utils.convert_device(
+                    torch.randperm(batch_size, device='cpu'))
+
+            for b in range(0, batch_size, mini_batch_size):
```
hnyu, line 1840: I am still not sure why we need to separate this training loop from the RL one.  I think the key question (for the general off-line training logic) to answer is: what is the essential difference between replayed RL experience and off-line experience regarding their training? Can we treat RL and off-line experiences conceptually the same (with some flags)?

hnyu (changes requested):

alf/algorithms/algorithm.py
hnyu: > Why do you want to use the RL experience and off-experience both at the same time? One alternative way is to treat RL experience and off-line experience as the same arg variable but different values for a unified function call. First accumulate both losses and then call the optimizer step only once. I think the key question to answer is: what is the essential difference between replayed RL experience and off-line experience regarding their training? Can we treat RL and off-line experiences conceptually the same?

hnyu:
For major pipeline modification/refactor in the future, I suggest we splitting a PR into several smaller ones. Otherwise it's difficult to review. @_@ For new added features/algorithms that are largely separated from existing code, a large PR is OK.

alf/algorithms/algorithm.py
Haichao-Zhang: Fixed.

alf/algorithms/algorithm.py
Haichao-Zhang: Fixed

alf/algorithms/algorithm.py
Haichao-Zhang: > because this is inside `train_from_replay_buffer`, then the replay buffer should always not be None?

In the case that online RL training is not started yet, ``_replay_buffer`` is None.

alf/algorithms/algorithm.py
Haichao-Zhang: The main reason for using this design is that it has no impact on the interfaces of algorithms that does not care about offline training.

alf/algorithms/algorithm.py
Haichao-Zhang: Same as above, the main idea in design is to introduce minimal impacts on existing methods.

alf/algorithms/config.py
```diff
@@ -218,6 +221,23 @@ def __init__(self,
...
+            rl_train_every_update_steps (int): only used in the hybrid training
+                mode. It is used to control the update frequency of the normal
+                (non-offline) part of the RL training  (according to
+                ``global_counter``). Through this flag, we can have a more fine
+                grained control over the update frequencies of online and offline
+                RL training (currently assumes the training frequency of offline
+                RL is always higher or equal to the online RL part).
+                For example, we can set ``rl_train_every_update_steps = 2``
+                to have a train config that executes online RL training at the
+                half frequency of that of the online RL training.
```
emailweixu: online => offline

alf/algorithms/algorithm.py
```diff
@@ -1173,41 +1284,87 @@ def train_from_replay_buffer(self, update_global_counter=False):
...
+                experience, batch_info, num_updates, mini_batch_size = _replay(
                 )
-                experience, batch_info = self._replay_buffer.get_batch(
+            else:
+                experience = None
+                batch_info = None
+                num_updates = 1
+                mini_batch_size = config.mini_batch_size
+
+            with record_time("time/replay"):
```
emailweixu: `"time/offline_replay" to differentiate with "time/replay"

alf/algorithms/rl_algorithm.py
emailweixu: What is the problem? If that is possible, code will be much simpler.

docs/notes/hybrid_and_offline_rl_training.rst
```diff
@@ -0,0 +1,121 @@
...
+we essentially can conduct pure offline RL training.
+
+``rl_train_every_update_steps`` can be used to contol the training frequency
+of the normal (non-offline) part of the RL training w.r.t offline RL training.
+
+
+
+3. Offline Buffer Structure
+-------------------------------------------------
+For generality, we assumes a minimal structure for the offlinereplay buffer,
```
emailweixu: offlinereplay => offline replay

alf/data_structures.py
```diff
@@ -481,3 +481,15 @@ def make_experience(time_step: TimeStep, alg_step: AlgStep, state):
         "priority",
     ],
     default_value=())
+
+# some basic structures used by offline replay buffer
+# reconstruction
+Info = namedtuple("Info", [
+    'rl',
```
emailweixu: need some explanation about the fields. I noticed that "reward" and "repr" are always `{}` when this structure is used. What are the purpose of them?

docs/notes/hybrid_and_offline_rl_training.rst
Haichao-Zhang: Fixed

alf/data_structures.py
Haichao-Zhang: Added more comments. Removed "reward" and "repr" as they are not essential here.

Update: these two are used to be compatible with the interface requirements of ``Agent``. Added comments on this.

alf/algorithms/algorithm.py
Haichao-Zhang: Changed

alf/data_structures.py
```diff
@@ -482,14 +482,18 @@ def make_experience(time_step: TimeStep, alg_step: AlgStep, state):
...
-    "rewards",
-    "repr",
-])
+# Some basic structures used by offline replay buffer for ``rollout_info``.
+# We assume the ``rollout_info`` contains an ``rl`` field, whose structure
+# is defined by ``RLInfo``.
+BasicRolloutInfo = namedtuple(
+    "BasicRolloutInfo",
+    [
+        'rl'  # field containing rl rollout information, as defined in ``RLInfo``
```
emailweixu: as defined in ``BasicRLInfo``

alf/algorithms/rl_algorithm.py
Haichao-Zhang: > What is the problem? If that is possible, code will be much simpler.

When I tried to use ``load_state_dict`` as in 3cee63b, after calling that function with a ``buffer_dict``, the offline_replay buffer is not populated with the data from ``buffer_dict``, and still contains all zeros. Haven't figured out the reason for this yet. Therefore reverted to the previous approach for now.

docs/notes/hybrid_and_offline_rl_training.rst
```diff
@@ -0,0 +1,119 @@
...
+For generality, we assumes a minimal structure for the offline replay buffer,
+which contains the essential information that is common for all RL algorithms.
+Concretely, we assume the offline replay buffer has the structure that is
+described by the following data spec:
+
+.. code-block:: python
+    offline_buffer_data_spec = Experience(
+                                time_step=time_step_spec,
+                                action=self._action_spec,
+                                rollout_info=Info(
```
emailweixu: Info => BasicRolloutInfo
RLInfo => BasicRLInfo

Haichao-Zhang: Sorry I did a rush push earlier... Now fixed these bugs.

alf/data_structures.py
Haichao-Zhang: Updated this and also other places.

Haichao-Zhang:
> For major pipeline modification/refactor in the future, I suggest we splitting a PR into several smaller ones. Otherwise it's difficult to review. @_@ For new added features/algorithms that are largely separated from existing code, a large PR is OK.

Good suggestion. Will do a better PR splitting in the future. Thanks for the efforts on reviewing the PR.

emailweixu (approved):

hnyu (approved):

PR #1209, opened 2022-02-18
Remove the unnecessary print that floods the unit test msg

Haichao-Zhang:
Remove the unnecessary print in layer test that floods the unit test msg, as shown below, which makes finding the useful information from unit test very difficult:

image


hnyu:
I also noticed this, and removed it in this PR #1201

emailweixu (approved):

breakds:
Sorry about this. I think I have accidentally committed this while studying the transformer implementation. Thanks for fixing it!

PR #1210, opened 2022-02-22
Info on encountered red-light in CARLA

Haichao-Zhang:
Added info on encountered red-light, apart from the previously existing info on running red-light.

### DIFF
--- a/alf/environments/suite_carla.py
+++ b/alf/environments/suite_carla.py
@@ -549,6 +549,7 @@ class Player(object):
         self._collision_loc = None  # the location of the car when it starts to have collision
 
         self._prev_violated_red_light_id = None
+        self._prev_encountered_red_light_id = None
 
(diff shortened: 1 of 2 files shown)


alf/environments/suite_carla.py
```diff
@@ -721,7 +722,8 @@ def get_current_time_step(self, current_frame):
         info = OrderedDict(
             success=np.float32(0.0),
             collision=np.float32(0.0),
-            red_light=np.float32(0.0),
+            red_light_run=np.float32(0.0),
+            red_light_met=np.float32(0.0),
```
breakds: Suggest explaining briefly what the two stands for, and units.

Haichao-Zhang: Added comments and also changed their names.

breakds (approved):

PR #1216, opened 2022-02-25
use 'agg' backend for matplotlib and lazy import VideoRecorder

hnyu:
Currently policy_trainer.py has an importing path: ``VideoRecorder`` -> ``render`` -> ``matplotlib``. matplotlib can have some side effects on xserver if a graphic backend is chosen. So we change ``VideoRecorder`` to be a lazy import and use 'agg' backend for matplotlib. As long as ``render.py`` is not mixed with code that relies on matplotlib with interactive UI, using 'agg' should be fine.

### DIFF
--- a/alf/environments/gym_wrappers.py
+++ b/alf/environments/gym_wrappers.py
@@ -129,7 +129,8 @@ class ImageChannelFirst(BaseObservationWrapper):
                 low = observation_space.low
                 high = observation_space.high
                 if np.isscalar(low) and np.isscalar(high):
-                    shape = observation_space.shape[::-1]
+                    shape = (observation_space.shape[-1:] +
+                             observation_space.shape[1:])
                     return gym.spaces.Box(
                         low=low,
                         high=high,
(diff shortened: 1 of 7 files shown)


alf/environments/gym_wrappers.py
```diff
@@ -129,7 +129,8 @@ def transform_space(self, observation_space):
                 low = observation_space.low
                 high = observation_space.high
                 if np.isscalar(low) and np.isscalar(high):
-                    shape = observation_space.shape[::-1]
+                    shape = (observation_space.shape[-1:] +
```
breakds, line 132: Does this suggest the previous implementation is a bug that happen to work because most images are square?

hnyu: > Does this suggest the previous implementation is a bug that happen to work because most images are square?

Right the current one is buggy, only if low and high are scalars and width and height are different.

breakds (approved):

PR #1217, opened 2022-02-26
Make sure alf_config is the one for training

### LINKED ISSUE #1215: alf_config.py in the log directory may not be the one actually used for training
Currently, ALF write alf_config.py after model has been updated. If the conf file has been changed before that, the content in alf_config.py is not same as the one actually used for training.

emailweixu:
Fixes #1215

We load the conf file at the beginning and use it for writing the alf_config later.

### DIFF
--- a/alf/trainers/policy_trainer_test.py
+++ b/alf/trainers/policy_trainer_test.py
@@ -119,4 +119,5 @@ class TrainerTest(alf.test.TestCase):
 
 
 if __name__ == "__main__":
-    alf.test.main()
+    TrainerTest().test_rl_trainer()
+    #alf.test.main()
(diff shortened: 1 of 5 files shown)


alf/trainers/policy_trainer_test.py
```diff
@@ -119,4 +119,5 @@ def test_sl_trainer(self):
 
 
 if __name__ == "__main__":
-    alf.test.main()
+    TrainerTest().test_rl_trainer()
+    #alf.test.main()
```
hnyu: should revert the change here

hnyu (changes requested):

hnyu (approved):

PR #1218, opened 2022-02-26
Enable mini_batch_length > 1 for muzero_algorithm

breakds:
# Motivation

This is part of the effort to make `muzero_algorithm` a generalized predictive representation learning algorithm. This PR focuses on updating the logic in `preprocess_experience` where training targets for the predictive representation learning is computed. Originally the logic only supports `mini_batch_length = 1`. Having `mini_batch_length > 1` is possible but might introduce redundant computation.

# Solution

This PR updates the logic that supports computing the targets when `mini_batch_length = 1`. The improvements are:

1. When `mini_batch_length > 1`, the new logic tries its best to not introduce redundant computation
2. With the help of "unfold" operation, `mini_batch_length = 1` is also just a special case

# Testing

Updated the unit tests so that they include `mini_batch_length > 1` cases as well. Both the new and old unit tests pass.

Actual training result with 2 seeds:

muzero

### DIFF
--- a/alf/algorithms/muzero_algorithm.py
+++ b/alf/algorithms/muzero_algorithm.py
@@ -352,24 +352,6 @@ class MuzeroAlgorithm(OffPolicyAlgorithm):
             beyond_episode_end = positions > episode_end_positions
             positions = torch.min(positions, episode_end_positions)
 
-            # The operation unfold1 (unfold at dimension 1) transform a tensor
-            # of shape [B, T + R, ...] to [B, T, R + 1, ...] by unfolding each
-            # sequence of length T + R into T shorter sequences with indices at
-            # [0:(R+1)], [1:(R+2)], .. until [(T-1):(T+R)].
-
-            # A capped unfolding caps the index for each of such shorter
-            # sequences at the episode boundary if it crosses the episode end.
-
-            capped_unfold1_index = (
-                torch.arange(B)[:, None, None],  # [B, 1, 1]
-                torch.arange(T)[:, None] + torch.min(
-                    steps_to_episode_end.unsqueeze(-1),
-                    torch.arange(R + 1))  # [T, R + 1]
-            )  # [B, T, R + 1]
-
-            def _unfold1_adapting_episode_ends(x):
-                return x[capped_unfold1_index]
-
             if self._reanalyze_ratio > 0:
                 # Here we assume state and info have similar name scheme.
                 mcts_state_field = 'state' + info_path[len('rollout_info'):]
@@ -414,6 +396,24 @@ class MuzeroAlgorithm(OffPolicyAlgorithm):
                         candidate_actions[r] = r_candidate_actions
                     candidate_action_policy[r] = r_candidate_action_policy
                     values[r] = r_values
+
+            # The operation unfold1 (unfold at dimension 1) transform a tensor
+            # of shape [B, T + R, ...] to [B, T, R + 1, ...] by unfolding each
+            # sequence of length T + R into T shorter sequences with indices at
+            # [0:(R+1)], [1:(R+2)], .. until [(T-1):(T+R)].
+
+            # A capped unfolding caps the index for each of such shorter
+            # sequences at the episode boundary if it crosses the episode end.
+
+            capped_unfold1_index = (
+                torch.arange(B)[:, None, None],  # [B, 1, 1]
+                torch.arange(T)[:, None] + torch.min(
+                    steps_to_episode_end.unsqueeze(-1),
+                    torch.arange(R + 1))  # [T, R + 1]
+            )  # [B, T, R + 1]
+
+            def _unfold1_adapting_episode_ends(x):
+                return x[capped_unfold1_index]
 
             # In the logic above, they are computed in folded form to save
             # unnecessary retrieval and computation. They are unfolded here so
--- a/alf/algorithms/muzero_algorithm.py
+++ b/alf/algorithms/muzero_algorithm.py
@@ -295,70 +295,146 @@ class MuzeroAlgorithm(OffPolicyAlgorithm):
     @torch.no_grad()
     def preprocess_experience(self, root_inputs: TimeStep,
                               rollout_info: MCTSInfo, batch_info):
-        """Fill rollout_info with MuzeroInfo
+        """Fill rollout_info with MuzeroInfo.
 
-        Note that the shape of experience is [B, T, ...]
+        Especially, the training targets for representation learning is computed
+        here with reanalyze and/or bootstrapping.
+
+        Note that the shape of experience is [B, T, ...], where B is the batch
+        size T is the mini batch length.
+
         """
         assert batch_info != ()
         replay_buffer: ReplayBuffer = batch_info.replay_buffer
         info_path: str = "rollout_info"
         info_path += "." + self.path if self.path else ""
-        mini_batch_length = root_inputs.step_type.shape[1]
         rollout_value = rollout_info.value
-        assert mini_batch_length == 1, (
-            "Only support TrainerConfig.mini_batch_length=1, got %s" %
-            mini_batch_length)
-
         value_field = info_path + '.value'
         candidate_actions_field = info_path + '.candidate_actions'
         candidate_action_policy_field = (
             info_path + '.candidate_action_policy')
 
+        # Create aliases for mini_batch_size (B), mini_batch_length(T) and
+        # predictive unroll steps (R) to make the implementation below more
+        # succinct.
+        B, T = root_inputs.step_type.shape
+        R = self._num_unroll_steps
+
+        # The implementation below heavily relies on the concept of "unfold".
+        # Take an individual trajectory of size T = 3 as example. Assuming the
+        # steps in this trajectory corresponds to position 11, 12, 13 in the
+        # replay buffer. In order to compute a certain type of targets for this
+        # trajectory with an unroll steps R = 5, a naive implementation will
+        # need to do computation for 15 T * (R + 1) steps at positions
+        #
+        #    11, 12, 13, 14, 15, 16
+        #    12, 13, 14, 15, 16, 17
+        #    13, 14, 15, 16, 17, 18
+        #
+        # respectively. Here we can employ a trick to compute only T + R + 1
+        # times at positions
+        #
+        #    11, 12, 13, 14, 15, 16, 17, 18
+        #
+        # and transform it to the above form. Such transformation that converts
+        # a shape of [T + R] to [T, R + 1] is the "unfold" operations.
+        #
+        # We also name the [T + R] shaped positions as "folded" positions.
+
+        def __unfold_dim1(x: torch.Tensor) -> torch.Tensor:
+            """Perform the aforementioned unfold at dim = 1 of the input tensor.
+
+            """
+            assert x.ndim >= 2
+            assert x.shape[1] == T + R
+            batch_size = x.shape[0]
+            numel = torch.prod(torch.tensor(x.shape[2:],
+                                            dtype=torch.int64)).item()
+            y = x.reshape(batch_size, -1)
+            y = y.unfold(1, numel * (R + 1), numel)
+            return y.reshape(batch_size, T, R + 1, *x.shape[2:])
+
         with alf.device(replay_buffer.device):
-            positions = convert_device(batch_info.positions)  # [B]
-            env_ids = convert_device(batch_info.env_ids)  # [B]
+            # [B]
+            start_positions = convert_device(batch_info.positions)
+            start_env_ids = convert_device(batch_info.env_ids)
+
+            # [B, T + R]
+            folded_positions = start_positions.unsqueeze(-1) + torch.arange(T +
+                                                                            R)
+            # [B, 1]
+            folded_env_ids = start_env_ids.unsqueeze(-1)
 
             if self._reanalyze_ratio > 0:
                 # Here we assume state and info have similar name scheme.
                 mcts_state_field = 'state' + info_path[len('rollout_info'):]
+
+                # Applying the "unfold" trick, where we do reanalyze from the
+                # starting position of B size-T trajectory for an unroll steps
+                # of T + R - 1, and unfold it to [B, T, R + 1]
                 if self._reanalyze_ratio < 1:
-                    batch_size = root_inputs.step_type.shape[0]
-                    r = torch.randperm(
-                        batch_size) < batch_size * self._reanalyze_ratio
+                    r = torch.randperm(B) < B * self._reanalyze_ratio
+                    # [B', T + R, ...], B' = B * reanalyze_ratio
                     r_candidate_actions, r_candidate_action_policy, r_values = self._reanalyze(
-                        replay_buffer, env_ids[r], positions[r],
-                        mcts_state_field)
+                        replay_buffer, start_env_ids[r], start_positions[r],
+                        mcts_state_field, T + R - 1)
+                    # [B', T, R + 1, ...]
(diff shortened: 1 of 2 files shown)


alf/algorithms/muzero_algorithm_test.py
```diff
@@ -204,22 +216,33 @@ def _test_preprocess_experience(self, train_reward_function, td_steps,
...
         processed_experience, processed_rollout_info = muzero.preprocess_experience(
             experience, experience.rollout_info, batch_info)
-        import pprint
-        pprint.pprint(processed_rollout_info)
-        alf.nest.map_structure(lambda x, y: self.assertEqual(x, y),
-                               processed_rollout_info, expected)
 
-    def get_exptected_info(self, sparse_reward):
+        # import pprint
+        # pprint.pprint(processed_rollout_info)
```
emailweixu: removing debugging code.

alf/algorithms/muzero_algorithm.py
```diff
@@ -295,70 +295,146 @@ def _hook(grad, name):
...
+        # times at positions
+        #
+        #    11, 12, 13, 14, 15, 16, 17, 18
+        #
+        # and transform it to the above form. Such transformation that converts
+        # a shape of [T + R] to [T, R + 1] is the "unfold" operations.
+        #
+        # We also name the [T + R] shaped positions as "folded" positions.
+
+        def __unfold_dim1(x: torch.Tensor) -> torch.Tensor:
```
emailweixu: Prefix or suffix "__" is usually reserved for system internals.

emailweixu (changes requested):

alf/algorithms/muzero_algorithm.py
breakds: Leaving only one `_` now.

alf/algorithms/muzero_algorithm_test.py
breakds: Removed.

alf/algorithms/muzero_algorithm.py
```diff
@@ -576,26 +656,31 @@ def _split2(x):
...
+                    replay_buffer: ReplayBuffer,
+                    env_ids,
+                    positions,
+                    mcts_state_field,
+                    horizon: Optional[int] = None):
         """Reanalyze one batch.
 
         This means:
-        1. Re-plan the policy using MCTS for n1 = 1 + num_unroll_steps to get fresh policy
+        1. Re-plan the policy using MCTS for n1 = 1 + horizon to get fresh policy
```
emailweixu, line 671: Somehow my comment on this function was lost in the last review.
The main issue is that new the logic for handling episode end needs to be changed. For example, suppose mini_batch_length is 5 and step 2 is a LAST step. The target for step 0, 1 and 2 should be calculated using the experience from the first episode. While the target for step 3 and 4 should be based on experience from the next episode. For the current code, the target for all the steps are based on the experience from the first episode.

breakds: Thanks a lot for clarifying this. Sorry that I did not think it through clearly - the current implementation is basically cutting it off and keeping only 0, 1, 2. So this is more complicated than "unfold". Let me rethink and revise.

breakds:
PR has been updated with the change we discussed. All unit tests updated and passed. Experiments are still running.

alf/algorithms/muzero_algorithm.py
```diff
@@ -676,28 +802,30 @@ def _reshape(x):
...
+
         # [B, 1]
         env_ids = env_ids.unsqueeze(-1)
+        # [B, n + 1]
+        positions = positions.unsqueeze(-1) + torch.arange(n + 1)
         # [B, 1]
-        episode_end_positions = episode_end_positions.unsqueeze(-1)
-
-        positions = torch.min(positions, episode_end_positions)
+        current_pos = replay_buffer.get_current_position().max().unsqueeze(-1)
```
emailweixu: should not do `max()`. current_position is a vector with same length as the number of environments. 
```python
current_pos = replay_buffer.get_current_positions()[env_ids]
```

alf/algorithms/muzero_algorithm.py
```diff
@@ -295,70 +295,182 @@ def _hook(grad, name):
...
-            positions = convert_device(batch_info.positions)  # [B]
-            env_ids = convert_device(batch_info.env_ids)  # [B]
+            # [B]
+            start_positions = convert_device(batch_info.positions)
+            start_env_ids = convert_device(batch_info.env_ids)
+
+            # [B, T + R], capped at the end of the replay buffer.
+            folded_positions = torch.min(
+                start_positions.unsqueeze(-1) + torch.arange(T + R),
+                replay_buffer.get_current_position().max() - 1)
```
emailweixu: `replay_buffer.get_current_position()[env_ids] - 1`

alf/algorithms/muzero_algorithm.py
```diff
@@ -295,70 +295,182 @@ def _hook(grad, name):
...
+                        value_field)
                 else:
+                    # [B, T + R]
                     values = self._calc_monte_carlo_return(
-                        replay_buffer, env_ids, positions, value_field)
+                        replay_buffer, folded_env_ids, folded_positions,
+                        value_field)
+
+                # [B, T, R + 1]
+                values = _unfold_dim1(values)
```
emailweixu: This `values` should also be applied with `_unfold_dim1_adapting_episode_ends`

alf/algorithms/muzero_algorithm.py
```diff
@@ -295,70 +295,182 @@ def _hook(grad, name):
...
+        #
+        # and transform it to the above form. Such transformation that converts
+        # a shape of [T + R] to [T, R + 1] is the "unfold" operations.
+        #
+        # We also name the [T + R] shaped positions as "folded" positions.
+
+        def _unfold_dim1(x: torch.Tensor) -> torch.Tensor:
+            """Perform the aforementioned unfold at dim = 1 of the input tensor.
+
+            """
```
emailweixu: Not sure how fast unfold is. But if T == 1, we can simply `return x.unsqueeze(1)` to save some time.

alf/algorithms/muzero_algorithm.py
```diff
@@ -295,70 +295,182 @@ def _hook(grad, name):
...
+
+            def _unfold_dim1_adapting_episode_ends(x):
+                # Do the unfolding first.
+                y = _unfold_dim1(x)
+
+                # Fill the beyond episode end locations with the values from the
+                # aforementioned view.
+                y[beyond_episode_end] = y[episode_end_index][
+                    beyond_episode_end]
+                return y
```
emailweixu: Can be simplified as:
```python
    capped_index = (torch.arange(B)[:, None, None],
                    torch.arange(T)[:, None],
                    torch.min(steps_to_episode_end.unsqueeze(-1), torch.arange(R+1)))
    def _unfold_dim1_adapting_episode_ends(x):
        return _unfold_dim1(y)[capped_index]
```
And it's possible to further combine `_unfold_dim1()` and `capped_index` as one single indexing operation `y[some_index]` where `some_index` takes into unfold and capped_index into account.

alf/algorithms/muzero_algorithm.py
breakds: Did not dig into the implementation of `unfold`, but adding two lines to specialize `T == 1` sounds good to me. Added.

alf/algorithms/muzero_algorithm.py
breakds: Updated to `replay_buffer.get_current_position()[start_env_ids, None] - 1`.

alf/algorithms/muzero_algorithm.py
breakds: I had  the wrong impression that all envs in replay buffer are synchronized (but even with that I the suggested approach is much more readable). Thanks!

alf/algorithms/muzero_algorithm.py
breakds: Much nicer! Literally scratched my head trying to find a simpler way to achieve this. Thanks!

alf/algorithms/muzero_algorithm.py
breakds: Fixed.

alf/algorithms/muzero_algorithm.py
```diff
@@ -295,70 +295,177 @@ def _hook(grad, name):
...
+
+            # Slicing ``capped_index`` is used to create a view of shape ([B, T,
+            # R + 1]) but each sequence (b, t) is capped by the index of the
+            # episode end if the sequence crosses the episode boundary.
+            capped_index = (
+                torch.arange(B)[:, None, None],  # [B, 1, 1]
+                torch.arange(T)[:, None],  # [T, 1]
+                torch.min(
+                    steps_to_episode_end.unsqueeze(-1),
+                    torch.arange(R + 1)))  # [B, T, 1]
```
emailweixu: `# [B, T, R + 1]`

alf/algorithms/muzero_algorithm.py
```diff
@@ -295,70 +295,177 @@ def _hook(grad, name):
...
+
+            def _unfold_dim1_adapting_episode_ends(x):
+                # Do the unfolding first.
+                return _unfold_dim1(x)[capped_index]
```
emailweixu: As the comment in the previous review suggested, `_unfold_dim1` and `[capped_index]` can be merged as one single indexing operation.
```python
capped_index = (
        torch.arange(B)[:, None, None],  # [B, 1, 1]
        torch.arange(T)[:, None] + torch.min(
            steps_to_episode_end.unsqueeze(-1),
            torch.arange(R + 1)))  # [B, T, R+1]
def _unfold_dim1_adapting_episode_ends(x):
        return x[capped_index]
```

alf/algorithms/muzero_algorithm.py
breakds: Fixed.

alf/algorithms/muzero_algorithm.py
breakds: Got it now. This also suggest that `unfold_dim1` is implemented in a much more complicated way, so I also updated it there. Thanks for the detailed guidance!

alf/algorithms/muzero_algorithm.py
```diff
@@ -407,11 +495,13 @@ def preprocess_experience(self, root_inputs: TimeStep,
                     batch_info=batch_info,
                     replay_buffer=replay_buffer)
                 exp = self._data_transformer.transform_experience(exp)
-                observation = exp.observation
+                observation = _unfold1(exp.observation)
```
emailweixu: This should also use `_unfold1_adapting_episode_ends`. The previous code does not do capping, which is not right.

alf/algorithms/muzero_algorithm.py
```diff
@@ -295,70 +295,152 @@ def _hook(grad, name):
...
+                        replay_buffer, folded_env_ids, folded_positions,
+                        value_field)
 
+                # [B, T, R + 1]
+                values = _unfold1_adapting_episode_ends(values)
+
+                # [B, T, R + 1]
                 candidate_actions = replay_buffer.get_field(
                     candidate_actions_field, env_ids, positions)
+                # [B, T, R + 1]
```
emailweixu: candidate_actions and candidate_action_policy can be retrieved using folded_env_ids and folded_positions so that we can do a single unfolding for values, candidate_actions and candidate_action_policy at line 452 (unfolding at line 410, 421 and 438 can be removed).

alf/algorithms/muzero_algorithm.py
breakds: You are right. Fixed.

alf/algorithms/muzero_algorithm.py
```diff
@@ -295,70 +295,139 @@ def _hook(grad, name):
...
+            #    unfolded into T sequences of length R. If the original sequence
+            #    cross the boundaries of the episodes, the new unfolded
+            #    sequences may start at different episodes, and may therefore
+            #    have different episode ends.
+            #
+            # With the above in mind, the helper function below achieves
+            # unfolding and adapting on the episode ends individually for each
+            # unfolded sequences.
+
+            # Very similar to ``unfold1_index`` above except for that this also
```
emailweixu: no more unfold1_index

alf/algorithms/muzero_algorithm.py
breakds: Thanks for the suggestion. Updated.
I leave `game_overs`  and `discount` below as they are since the retrieval along will not benefit from folded + unfold and `==` should be trivial enough.

emailweixu: Didn't see the update.

alf/algorithms/muzero_algorithm.py
breakds: This should have been fixed already. Somehow my previous push did not go through. Should be good now.

alf/algorithms/muzero_algorithm.py
```diff
@@ -295,82 +295,138 @@ def _hook(grad, name):
...
+
+            capped_unfold1_index = (
+                torch.arange(B)[:, None, None],  # [B, 1, 1]
+                torch.arange(T)[:, None] + torch.min(
+                    steps_to_episode_end.unsqueeze(-1),
+                    torch.arange(R + 1))  # [T, R + 1]
+            )  # [B, T, R + 1]
+
+            def _unfold1_adapting_episode_ends(x):
+                return x[capped_unfold1_index]
```
emailweixu: can be moved closer to where it is used.

breakds: Sure. Done.

emailweixu (dismissed):

breakds:
Will wait for checking the experiment result before checking-in. Thanks for the patient review @emailweixu !

breakds:
See the experiment result. Unfortunately though they are not perfectly "apple to apple", but it clearly shows that the code  has some problem.

muzero_comparison

|  Color  | Code Version | num_envs | mini_batch_size | mini_batch_length | Training Time |
| ------------- | ------------- | ------------- | ------------- | ------------- | ------------- |
| Blue | `pytorch` branch 2 weeks ago | 4 | 256  | 1 | 14.5 hour |
| Pink | This PR without "eoe" capping | 8 | 64 | 4 | 6.6 hour |
| Grey | This PR | 8 | 256 | 1 | 7.2 hour |
| Green | This PR | 8 | 256 | 2 | 10.5 hour|

More experiments result to come. Meanwhile I'll need to debug more.

breakds:
Updated with a fix on the observation extraction and unfolding order.

alf/algorithms/muzero_algorithm.py
```diff
@@ -455,10 +455,22 @@ def _unfold1_adapting_episode_ends(x):
...
+                # unfold is applied. As a result, we choose to first unfold.
                 observation, step_type = replay_buffer.get_field(
                     ('observation', 'step_type'), folded_env_ids,
                     folded_positions)
+
+                # Unfold (and reshape)
+                observation = _unfold1_adapting_episode_ends(observation)
+                observation = observation.reshape(-1, *observation.shape[2:])
+                step_type = _unfold1_adapting_episode_ends(step_type)
+                step_type = step_type.reshape(-1, *step_type.shape[2:])
```
emailweixu, line 472: with this, batch_info.positions should also be updated before calling transform_experience

breakds: Yep I realized this after running the `mini_batch_length > 1` experiment. Will put up a fix.

breakds: It is fixed.

emailweixu (approved):

PR #1219, opened 2022-03-02
Store transform ctor in DistributionSpec

hnyu:
Currently a DistributionSpec stores transform objects. As discussed, this might be problematic because if this spec is used to build multiple distributions then they all share the same set of transform objects, and setting the params of them in one distribution will unexpectedly change the behaviors of other distributions. So this PR instead stores transform builders in DistributionSpec so that each time a spec builds a distribution, new transform objects are created.

Different distributions might still share some variables across the transforms. These variables are not captured by the `params` of a transform.

### DIFF
--- a/alf/utils/dist_utils.py
+++ b/alf/utils/dist_utils.py
@@ -557,31 +557,10 @@ def _builder_independent(base_builder, reinterpreted_batch_ndims_, **kwargs):
     return td.Independent(base_builder(**kwargs), reinterpreted_batch_ndims_)
 
 
-def _builder_transformed(base_builder, transforms_, transforms_params_,
-                         params_):
-    _set_transform_params(transforms_, transforms_params_)
-    return td.TransformedDistribution(base_builder(**params_), transforms_)
-
-
-def _set_transform_params(transforms, params):
-    """Given a nest of params where each node is a non-composed transform of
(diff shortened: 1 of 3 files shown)


alf/utils/dist_utils.py
```diff
@@ -604,33 +595,57 @@ def _get_independent_builder(obj: td.Independent):
...
+            builders, params = _get_transform_builders_params(transforms.parts)
+            compose_transform_builder = lambda params: td.ComposeTransform(
+                [b(p) for b, p in zip(builders, params)])
+            return compose_transform_builder, params
+        else:
+            builder = _get_transform_builder(transforms)
+            params = _get_transform_params(transforms)
+            new_builder = lambda params: (builder(**params)
+                                          if params != () else builder())
+            return new_builder, params
```
emailweixu: can be simplified as the following if line 617 is changed to `return {}` and `b(p)` at 574  to `b(**p)`:
```python
return builder, params
```

hnyu: > can be simplified as the following if line 617 is changed to `return {}` and `b(p)` at 574 to `b(**p)`:
> 
> ```python
> return builder, params
> ```

Updated

emailweixu (approved):

PR #1220, opened 2022-03-03
Redlight Sensor and light-weight CARLA

Haichao-Zhang:
Training and debugging on CARLA is time consuming.

This PR introduces a lightweight CARLA setting where we can disable camera while can still achieve reasonable performance.
More importantly, under this setting, SAC can show progress much earlier than before, facilitating algorithmic development.

Additionally, a redlight sensor can be added to the observation without too much additional computation and learning cost since it also produces a low-dimensional measurement.

image

This result also suggests that the benefits of learning with image observation could potentially be improved with better techniques such as representation learning.

### DIFF
--- a/alf/environments/carla_sensors.py
+++ b/alf/environments/carla_sensors.py
@@ -995,6 +995,8 @@ class World(object):
 
     # only consider a car for running red light if it is within such distance
     RED_LIGHT_ENFORCE_DISTANCE = 15  # m
+    # set a large value as the default encountered red light distance
+    DEFAULT_ENCOUNTERED_RED_LIGHT_DISTANCE = 1e10  # m
 
     def __init__(self, world: carla.World, route_resolution=1.0):
         """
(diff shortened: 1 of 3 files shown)


alf/environments/carla_sensors.py
```diff
@@ -1307,8 +1308,10 @@ def is_running_red_light(self, actor):
             # red-light id encountered by the actor
             if np.any(same_lane):
                 encountered_red_light_id = self._traffic_light_actors[index].id
+                encountered_red_light_distance = dist[index]
 
-        return violated_red_light_id, encountered_red_light_id
+        return (violated_red_light_id, encountered_red_light_id,
```
breakds, line 1318: Also need to update `Returns` in docstring.

breakds (dismissed):

alf/environments/carla_sensors.py
Haichao-Zhang: Fixed

breakds (approved):

PR #1221, opened 2022-03-09
Munchausen RL

Haichao-Zhang:
Implementation of the Munchausen RL.

Tested on Breakout.   The reward-weight is an import hyper-parameter for tuning.
Currently, it seems that with a proper weight (e.g. 0.5), there are some gain initially, but eventually similar to SAC.
image

### DIFF
--- a/alf/algorithms/sac_algorithm.py
+++ b/alf/algorithms/sac_algorithm.py
@@ -273,7 +273,7 @@ class SacAlgorithm(OffPolicyAlgorithm):
             critic_network_cls, q_network_cls)
 
         self._use_entropy_reward = use_entropy_reward
-        self._munchausen_reward_weight = min(0, munchausen_reward_weight)
+        self._munchausen_reward_weight = max(0, munchausen_reward_weight)
         if munchausen_reward_weight > 0:
             assert not normalize_entropy_reward, (
                 "should not normalize entropy "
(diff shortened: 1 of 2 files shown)


emailweixu:
Are there any experiment results?

alf/algorithms/sac_algorithm.py
```diff
@@ -846,6 +857,30 @@ def _calc_critic_loss(self, info: SacInfo):
...
+                    log_pi_rollout_a = sum(nest.flatten(log_pi_rollout_a))
+
+                munchausen_reward = nest.map_structure(
+                    lambda la, lp: torch.exp(la) * lp, self._log_alpha,
+                    log_pi_rollout_a)
+                munchausen_reward = sum(nest.flatten(munchausen_reward))
+                info = info._replace(
+                    reward=(
+                        info.reward + self._munchausen_reward_weight *
+                        common.expand_dims_as(munchausen_reward, info.reward)))
```
hnyu, line 905: This reward should be added to the next time step? Because you are using the rollout action of the current step and its reward is contained by the next step. The entropy reward and this reward shouldn't be added to the same step.

Haichao-Zhang:
> Are there any experiment results?

When submitting the PR, only tested on pendulum for sanity check. On pendulum, no clear difference compared to SAC.
I'm running more experiments. Will post some here when available.

alf/algorithms/sac_algorithm.py
```diff
@@ -846,6 +857,30 @@ def _calc_critic_loss(self, info: SacInfo):
         When the reward is multi-dim, the entropy reward will be added to *all*
         dims.
         """
```
hnyu, line 856: The comment above has been separated from the entropy reward code.

alf/algorithms/sac_algorithm.py
Haichao-Zhang: This is a bug. Fixed

alf/algorithms/sac_algorithm.py
Haichao-Zhang: Now make the comment embedded to the right place.

alf/algorithms/sac_algorithm.py
```diff
@@ -856,10 +864,46 @@ def _calc_critic_loss(self, info: SacInfo):
...
+                    lambda dist, a: dist.log_prob(a), info.action_distribution,
+                    info.action)
+
+                if self._act_type == ActionType.Mixed:
+                    # For mixed type, add log_pi separately
+                    log_pi_rollout_a = type(self._action_spec)(
+                        (sum(nest.flatten(log_pi_rollout_a[0])),
+                         sum(nest.flatten(log_pi_rollout_a[1]))))
+                else:
+                    log_pi_rollout_a = sum(nest.flatten(log_pi_rollout_a))
```
hnyu, line 886: Why do you want to first add log_pi here?

hnyu, line 886: Why do you want to first add log_pi here? (Unless for each log_alpha you have multiple log_pis)

alf/algorithms/sac_algorithm.py
```diff
@@ -267,6 +273,11 @@ def __init__(self,
             critic_network_cls, q_network_cls)
 
         self._use_entropy_reward = use_entropy_reward
+        self._munchausen_reward_weight = max(0, munchausen_reward_weight)
+        if munchausen_reward_weight > 0:
+            assert not normalize_entropy_reward, (
```
hnyu, line 278: normalize_entropy_reward has no conflict with munchausen_reward_weight > 0?

hnyu, line 278: normalize_entropy_reward has no conflict with munchausen_reward_weight > 0?

hnyu (changes requested):

PR #1223, opened 2022-03-10
Use elastic_namedtuple for BasicRLInfo

Haichao-Zhang:
This PR introduces ``elastic_namedtuple`` and upgrades the ``BasicRLInfo`` structure with it.
With this change, it enables automatic compatibility with future changes of ``RolloutInfo`` made on the algorithm side.

For example,  PR  #1199  uses ``rollout_info.discounted_return`` in SAC ``train_step``, which is  beyond the assumption of the ``BasicRLInfo`` structure:
https://github.com/HorizonRobotics/alf/blob/ca25008bd6ca7491ff19c498f39f4c03381aabe9/alf/algorithms/sac_algorithm.py#L785

With this PR, it provides automatic compatibility with future changes like this.

### DIFF
--- a/alf/data_structures.py
+++ b/alf/data_structures.py
@@ -482,12 +482,32 @@ LossInfo = namedtuple(
     ],
     default_value=())
 
+
+def elastic_namedtuple(name, args):
+    """elastic namedtuple that returns ``()`` for a non-existing attribute,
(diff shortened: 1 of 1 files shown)


le-horizon (approved):
LG.  Thanks for the fix Haichao!

PR #1225, opened 2022-03-10
change ValueNetwork, ActorNetwork, QNetwork, ActorDistributionNetwork to base on Network

hnyu:
Also differentiate between networks that support stateful and stateless input preprocessors. Write test cases for PreprocessorNetwork.

### DIFF
--- a/alf/networks/value_networks.py
+++ b/alf/networks/value_networks.py
@@ -14,6 +14,7 @@
 """ValueNetwork and ValueRNNNetwork."""
 
 import functools
+from typing import Callable
 
 import torch
 import torch.nn as nn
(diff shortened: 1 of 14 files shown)


emailweixu:
The title of this PR "test PreprocessorNetwork with stateful preprocessors" does not seem to match the actual code.

emailweixu:
The title of this PR "test PreprocessorNetwork with stateful preprocessors" does not seem to match the actual code.

hnyu:
> The title of this PR "test PreprocessorNetwork with stateful preprocessors" does not seem to match the actual code.

The main change is in network_test.py where all networks of type `PreprocessorNetwork` is tested against stateful preprocessor and errors are asserted to be raised. The other docstring changes are just updates from old ones.

The point is that with `PreprocessorNetwork` we can't use stateful preprocessors, and thus I made this differentiation in the docstrings.

emailweixu:
In stead of making sure stateful preprocessors cannot be used in some networks, why not change them to accept stateful prepropcessors?

hnyu:
> In stead of making sure stateful preprocessors cannot be used in some networks, why not change them to accept stateful prepropcessors?

I think the concept of `PreprocessorNetwork` is outdated compared to network containers. Networks like `EncodingNetwork` do accept stateful preprocessors. The user can configure a network that preprocesses inputs using `Parallel` and `Sequential`. It is just that we haven't converted some `PreprocessorNetwork` to containers yet. Or we can simply make use of `PreprocessorNetwork` as a simplified network case where stateful preprocessors are not allowed.

emailweixu:
> > In stead of making sure stateful preprocessors cannot be used in some networks, why not change them to accept stateful prepropcessors?
> 
> I think the concept of `PreprocessorNetwork` is outdated compared to network containers. Networks like `EncodingNetwork` do accept stateful preprocessors. The user can configure a network that preprocesses inputs using `Parallel` and `Sequential`. It is just that we haven't converted some `PreprocessorNetwork` to containers yet. Or we can simply make use of `PreprocessorNetwork` as a simplified network case where stateful preprocessors are not allowed.

Right, PreprocessorNetwork is outdated. But it seems to me this PR will smaller if  switching ValueNetwork and ActorNetwork to use EncodingNetwork's input_preprocessors.

hnyu:
> > > In stead of making sure stateful preprocessors cannot be used in some networks, why not change them to accept stateful prepropcessors?
> > 
> > 
> > I think the concept of `PreprocessorNetwork` is outdated compared to network containers. Networks like `EncodingNetwork` do accept stateful preprocessors. The user can configure a network that preprocesses inputs using `Parallel` and `Sequential`. It is just that we haven't converted some `PreprocessorNetwork` to containers yet. Or we can simply make use of `PreprocessorNetwork` as a simplified network case where stateful preprocessors are not allowed.
> 
> Right, PreprocessorNetwork is outdated. But it seems to me this PR will smaller if switching ValueNetwork and ActorNetwork to use EncodingNetwork's input_preprocessors.

I've changed ValueNetwork and ActorNetwork to subclasses of Network. Still keep TransformerNetwork unchanged because it has some special constraints in the inputs and usually doesn't need recurrent in preprocessors.

alf/networks/value_networks.py
```diff
@@ -225,7 +222,9 @@ def __init__(self,
                                     a=-0.03, b=0.03)
 
         self._encoding_net = LSTMEncodingNetwork(
-            input_tensor_spec=self._processed_input_tensor_spec,
+            input_tensor_spec=input_tensor_spec,
```
emailweixu: The base class ValueNetwork also creates self._encoding_net. That does not seem right.

hnyu: > The base class ValueNetwork also creates self._encoding_net. That does not seem right.

It will be replaced by the one in the subclass. This is also the case in the current `ActorDistributionNetwork`. This way we can save some code of ``forward`` and `make_parallel`.

hnyu: Now extracted a base class so that the encoding network won't be made twice.

emailweixu (dismissed):

hnyu:
Forgot to do the same thing for QNetwork. Also noticed that the state_spec of Parallel*Network was not correct because it was default to (). The new commit addresses these two things.

emailweixu (approved):

PR #1228, opened 2022-03-15
Support delayed_update for get_target_updater

emailweixu:
And changed get_target_updater to TargetUpdater to allow more flexible update.

 delayed_update: if True, `target_models` is updated using recent_models
            every `period` steps. If `tau` is 1, the recent_models is `models`
            `period` steps before. If `tau` is not 1, recent_models is
            an exponential moving average of `models` with rate `tau`.
            The use of delayed_update may help to improve the stability of TD
            learning when a small `period` is used.


emailweixu:
> Do we need to add `_recent_models` to `_trainable_attributes_to_ignore`?

No, it is in a list. It will not be found be Module.parameters()

Haichao-Zhang (approved):

PR #1233, opened 2022-03-16
Various improvements to Muzero

emailweixu:
Some of these improvements are from examining EfficientZero implementation.

Enhancements:

1. compute repr target in preprocess_experience to reduce memory usage
2. support data augmenter
3. support use random action instead of repeating action for the steps beyond the episode end.
4. mask out the repr loss for the steps beyond episode end
5. mask out loss for the steps beyond the end of of partial episode due to TimeLimit or ongoing episode
6. support repr_norm_eps to use non-default eps when calculating cosine simarity for representation loss
7. allow different weight for the loss at the initial step of the trajectory
8. support not predicting the reward for the initial step of the trajectory as it is not used by MCTS
9. support reset_reward_sum_period to avoid predicting reward sum of a period longer than what is trained. This is described in EfficientZero paper Appenix A.1

Bug fixes:

1. importance_weights from priority replay is ignored due to the use of scalar_loss


alf/algorithms/muzero_algorithm.py
```diff
@@ -85,10 +85,12 @@ def __init__(self,
                  reanalyze_batch_size=None,
                  full_reanalyze=False,
                  data_transformer_ctor=None,
+                 data_augmenter=None,
```
breakds: Missing docstrings for the 2 new arguments.

breakds (dismissed):
The improvements and bug fix LGTM. Althought there is a test breakage need to be fixed, see comments below.

Also realized my dumbness when not thinking about dividing representation targets into mini batches to save memory. Thanks!

alf/algorithms/mcts_models.py
```diff
@@ -236,17 +269,13 @@ def recurrent_inference(self, state, action):
         Returns:
             ModelOutput: the prediction
         """
-        if not self._handle_bn:
-            dyn_state = self._dynamics_net((state.state, action))[0]
-            return self._predict(state._replace(state=dyn_state))
-        else:
+        current_steps = state.step + 1
```
breakds, line 272: Just found that this breaks the build. Originally `state.step` is only updated for `_handle_bn` case, which means that (I could not directly point out where) `state.step` may be `()`. In that case `() + 1` will throw exception.

emailweixu: Fixed

breakds (dismissed):

breakds (approved):

PR #1237, opened 2022-03-16
[Bug Fix] Correctly compute beyond_episode_end in muzero

breakds:
# Motivation

This PR fixes 3 problems introduced in #1218 

1. Unit test was unintentionally turned off.
2. As a result of 1, the logic of adapting expected value to `mini_batch_length > 1` is left incorrect without being noticed.
3. `beyond_episode_end` is wrong near the replay buffer's end.

Where No.3 is due to

- `beyond_episode_end` is computed with `positions > episode_end_positions`.
- Although `episode_end_positions` is computed correctly, in #1218 `positons` at this moment has already been capped by replay buffer's end.

# Solution

1. Re-enable the unit tests.
2. Update the logic of adapting expected value. It should be much simpler.
3. Fix No.3 by capping after instead of before computing `beyond_episode_end`.


breakds:
After rebasing there are new test failures. Will fix in a bit.

emailweixu:
> After rebasing there are new test failures. Will fix in a bit.

The expected value for beyond_episode_end is not correct.

breakds:
> > After rebasing there are new test failures. Will fix in a bit.
> 
> The expected value for beyond_episode_end is not correct.

Yes, and also `is_partial_trajectory`. This is because the last step of the 2nd environment is the first frame of a new episode. This is fixed now.

emailweixu (approved):

PR #1239, opened 2022-03-17
Use batch_size=2 for Network._test_forward()

emailweixu:
Some layers (e.g. BatchNorm) do not work with batch_size=1. So we use batch_size=2 for _test_forward
Also add a helper funtion is_rnn for Network


hnyu (dismissed):

hnyu:
Needs to update the batch shape in critic_networks_test.py

hnyu (approved):

PR #1242, opened 2022-03-18
some updates on wrappers and transformers

hnyu:
1. Explicitly assert fields is not None in case it is []
2. correct docstring of `get_raw_observation_spec()` which can be called after all gym wrappers are created and no need to wait for data transformers

### DIFF
--- a/alf/environments/gym_wrappers.py
+++ b/alf/environments/gym_wrappers.py
@@ -69,26 +69,39 @@ class BaseObservationWrapper(gym.ObservationWrapper):
         """
         Args:
             env (gym.Env): the gym environment
-            fields (list[str]): fields to be applied transformation, A field str is a multi-level
-                path denoted by "A.B.C". If None, then non-nested observation is transformed
+            fields (list[str]): fields to be applied transformation, A field str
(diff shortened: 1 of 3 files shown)


alf/environments/gym_wrappers.py
```diff
@@ -229,11 +242,16 @@ def transform_space(self, observation_space):
...
-                nested=observation,
-                field=field,
-                func=lambda obs: self.transform_observation(obs, field))
+        if self._fields is not None:
+            for field in self._fields:
+                observation = transform_nest(
+                    nested=observation,
+                    field=field,
+                    func=lambda obs: self.transform_observation(obs, field))
+        else:
```
emailweixu: now this wrapper behaves differently from FrameStacker data transformer. FrameStacker data transformer does not have this "else" branch.

hnyu: > now this wrapper behaves differently from FrameStacker data transformer. FrameStacker data transformer does not have this "else" branch.

Reverted this change

emailweixu (approved):

PR #1246, opened 2022-03-24
let NetworkWrapper support stateful modules

hnyu:


emailweixu:
I am curious what the use scenario is?

hnyu:
> I am curious what the use scenario is?

To put a customized stateful function in a Sequential container. For example, I need to manually accumulate (s_t,a_t) over a trajectory (sum after some projection) and want to make this as a network.

emailweixu (dismissed):

emailweixu:
Please refrain from forced push in the middle of a PR in the future because there is no easy way to know its the difference with the previous version.

emailweixu (approved):

PR #1248, opened 2022-03-24
remove 'action' field from Experience

hnyu:
This field is no longer needed after we moved action to rollout_info. Removed it for consistency of the training logic and less confusion.


hnyu:
It's still needed for summary. Closed for now.

PR #1250, opened 2022-03-26
SummarizeGradient Layer

Haichao-Zhang:
A gradient summarization layer that can be included in a network for monitoring the gradient information.

This is convenient especially for the tensors that we do not have direct access in the algorithm, e.g. raw input tensors or the tensors involved in preprocessing net.

### DIFF
--- a/alf/layers.py
+++ b/alf/layers.py
@@ -3087,30 +3087,30 @@ class SummarizeGradient(ElementwiseLayerBase):
     def __init__(self, name):
         """A layer for summarizing the gradient of the input tensor.
 
-        Summarize the gradient of the input tensor if its gradient
-        calculation is enabled (``requires_grad=True``).
-        Otherwise,  if the input tensor is a leaf node, whose ``requires_grad``
-        is set to False by detault, this layer will set ``requires_grad=True``
-        for the input tensor to enable gradient calculation for summarization.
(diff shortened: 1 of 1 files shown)


alf/layers.py
```diff
@@ -3081,6 +3081,27 @@ def forward(self, input):
...
+            name (str): used to describe the name of the summary, after the
+                tag 'tensor_gradient'.
+        """
+        super().__init__()
+
+        self._name = name
+
+    def forward(self, x):
+        # explicitly turn on gradient calculation in order to summarize gradient
+        x.requires_grad = True
```
emailweixu: This can change the behavior of training. I suggest  do not summarize if x.requires_grad is false.

Haichao-Zhang: Changed to only turn-on ``requires_grad`` for leaf nodes, which is a common use case when we want to summarize the gradient w.r.t. inputs.

Haichao-Zhang: Now changed with a clone of the input as suggested.

emailweixu (dismissed):

Haichao-Zhang:
There is a conflict with a new merge hours ago (#1246). Rebased to pytorch.

emailweixu (approved):

PR #1252, opened 2022-03-28
add is_training()

hnyu:
The current is_replay and is_rollout are not exact indicators of training or not training.

### DIFF
--- a/alf/utils/common.py
+++ b/alf/utils/common.py
@@ -1018,12 +1018,12 @@ def is_eval():
     return _exe_mode == EXE_MODE_EVAL
 
 
-def is_training(alg: alf.algorithms.algorithm.Algorithm):
+def is_training(alg):
     """Return a bool value indicating whether the current code is in a training
     step, for either an on-policy or an off-policy algorithm.
 
     Args:
-        alg: the algorithm to be decided
+        alg (Algorithm): the algorithm to be decided
     """
     return (alg.on_policy and is_rollout()) or is_replay()
 
--- a/alf/utils/common.py
+++ b/alf/utils/common.py
@@ -1000,13 +1000,21 @@ def exe_mode_name():
 def is_replay():
     """Return a bool value indicating whether the current code belongs to
     replaying. Replaying implies off-policy training.
+
+    Any code under ``train_from_replay_buffer()`` of any algorithm is classified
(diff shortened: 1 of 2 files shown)


alf/utils/common.py
```diff
@@ -999,14 +999,14 @@ def exe_mode_name():
 
 def is_replay():
     """Return a bool value indicating whether the current code belongs to
-    unrolling or training.
+    replaying. Replaying implies off-policy training.
```
emailweixu, line 1002: I think we need clearer definition of which part of code belongs to replaying. Similar for is_rollout and is_training.

hnyu: > I think we need clearer definition of which part of code belongs to replaying. Similar for is_rollout and is_training.

Do you mean clearer definition in terms of code or docstring comment?

emailweixu: Describe more clearly what part of code belongs to replay in the doc-string

hnyu:
@Haichao-Zhang I removed `@mark_replay` on `_prepare_experience_data` as this function is always under `_train_from_replay_buffer`

alf/utils/common.py
Haichao-Zhang: Although that the purpose of each function is clear, the definition of ``replay/train/rollout(unroll)`` here is a bit different from other places in ALF, with similar notions, e.g. when summarizing time cost.
This could introduce some confusions. It might be best to keep these terms to have consistent meanings at different places.

emailweixu (approved):

alf/utils/common.py
hnyu: > Although that the purpose of each function is clear, the definition of `replay/train/rollout(unroll)` here is a bit different from other places in ALF, with similar notions, e.g. when summarizing time cost. This could introduce some confusions. It might be best to keep these terms to have consistent meanings at different places.

The purpose of these functions is quite different from other places. They are mainly used as flags for controlling different logics in the same code block (might be very deep in the code hierarchy) to generate different behaviors. If aligning with the terminology of time cost summary, then they can no longer serve this purpose. For example, "time/train" only summarizes ``train_from_unroll()`` for on-policy algorithms but doesn't account for ``rollout_step()`` which is usually what we customize most. Generally the the labels for time cost are more fine-grained.

Haichao-Zhang (approved):

PR #1257, opened 2022-03-29
Fix verify_checkpoint

emailweixu:
There are several problems:
1. Interface change of predict_step
2. Not handling Trainer._progress
3. Better failure message if some configs are set through import. This is because when loading from root_dir, all configs are cleared. But the import command inside the conf file won't run again.

### DIFF
--- a/alf/bin/verify_checkpoint.py
+++ b/alf/bin/verify_checkpoint.py
@@ -81,8 +81,7 @@ def _step(algorithm, time_step, policy_state, trans_state, epsilon_greedy):
         time_step.is_first())
     transformed_time_step, trans_state = algorithm.transform_timestep(
         time_step, trans_state)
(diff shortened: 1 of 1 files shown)


alf/bin/verify_checkpoint.py
```diff
@@ -133,6 +132,20 @@ def _create_algorithm_and_env(root_dir):
     except Exception as e:
         alf.close_env()
         raise e
+    new_configs = dict(alf.get_inoperative_configs())
+    if old_configs is not None:
+        ok = True
+        for k, v in old_configs.items():
+            if k not in new_configs:
+                logging.error("config '%s' is set by the original config file "
+                              "but is noet set by root_dir/alf_config.py" % k)
```
hnyu: noet -> not

hnyu (approved):

PR #1258, opened 2022-04-01
RandomCrop and RelicationPad2d

emailweixu:
pytorch's ReplicationPad2d only works on float data. So we implement one supporting all types dtype.
pytorch's RandomCrop applies same random crop for all images in one batch. This one applies different random crop for each image in a batch.

### DIFF
--- a/alf/layers.py
+++ b/alf/layers.py
@@ -3036,7 +3036,7 @@ class RandomCrop(nn.Module):
     Note that ``torchvision.transforms.RandomCrop`` is different in that it
     applies the same random crop for all the images in the batch.
 
-    Each result image image is a random crop of the padded input image. The padded
+    Each result image is a random crop of the padded input image. The padded
     pixels are from the neareat pixel from the boundary.
 
     Args:
--- a/alf/layers.py
+++ b/alf/layers.py
@@ -20,6 +20,7 @@ import numpy as np
 import torch
 import torch.nn as nn
 import torch.nn.functional as F
+from torch import Tensor
 from typing import Callable, Dict, Iterable, Optional, Tuple, Union
 
 import alf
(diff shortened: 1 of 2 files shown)


alf/layers.py
```diff
@@ -2976,6 +2977,114 @@ def forward(self, input):
...
+        return input[..., yindex, xindex]
+
+
+class RandomCrop(nn.Module):
+    r"""Perform random crop independently for each image in the batch.
+
+    Note that ``torchvision.transforms.RandomCrop`` is different in that it
+    applies the same random crop for all the images in the batch.
+
+    Each result image image is a random crop of the padded input image. The padded
```
le-horizon: remove duplicate "image"

alf/layers.py
```diff
@@ -2976,6 +2977,114 @@ def forward(self, input):
...
+        left, right, top, bottom = self._padding
+        if h != self._h:
+            yindex = torch.arange(-top, h + bottom)
+            yindex[:top] = 0
+            yindex[-bottom:] = h - 1
+            yindex = yindex.unsqueeze(-1)
+            self._yindex = yindex
+            self._h = h
+        else:
+            yindex = self._yindex
```
le-horizon, line 3021: where is self._yindex initialized?

alf/layers.py
```diff
@@ -2976,6 +2977,114 @@ def forward(self, input):
...
+        if w != self._w:
+            xindex = torch.arange(-left, w + right)
+            xindex[:left] = 0
+            xindex[-right:] = w - 1
+            self._xindex = xindex
+            self._w = w
+        else:
+            xindex = self._xindex
```
le-horizon, line 3029: same here?

alf/layers.py
```diff
@@ -2976,6 +2977,114 @@ def forward(self, input):
...
+        starty = torch.randint(-top, H + bottom - h + 1, (B, )).reshape(B, 1)
+        startx = torch.randint(-left, W + right - w + 1, (B, )).reshape(
+            B, 1, 1)
+        # [B, h, 1]
+        y = (starty + torch.arange(h)).clamp_(min=0, max=H - 1).unsqueeze(-1)
+        # [B, 1, w]
+        x = (startx + torch.arange(w)).clamp_(min=0, max=W - 1)
+        # [B, 1, 1]
+        b = torch.arange(B).reshape(B, 1, 1)
+        # The alternative way of input[b, c, y, x] would use a lot more memory
```
le-horizon, line 3084: curious, why would the alternative need more memory?

le-horizon (commented):
only a couple nits.

Thanks,
Le

alf/layers.py
emailweixu: It's ok since if h == self._h, self._yindex is already initialized (self._h is initialized to -1)

alf/layers.py
emailweixu: I am not hundred percent sure. But according to the size of memory allocation request in the error message, it seems that pytorch first make broadcasted indexing array, which costs 8*B*C*H*W bytes.

alf/layers.py
le-horizon: Ah, make sense.  _h here is a marker of whether cache is valid, and _yindex is the cache content.

alf/layers.py
le-horizon: Interesting.

le-horizon (approved):
LGTM

breakds (approved):

PR #1259, opened 2022-04-01
AMPWrapper

emailweixu:
Sometime we want a part of a model to run in a particular AMP context (for example, disabling AMP for a precision sensitive part of modle).  These two wrappers can be used to achieve this.

### DIFF
--- a/alf/layers.py
+++ b/alf/layers.py
@@ -3467,3 +3467,24 @@ def make_parallel_spec(specs, n: int):
                                          spec.minimum, spec.maximum)
 
     return map_structure(_make_spec, specs)
+
+
+class AMPWrapper(nn.Module):
+    """Wrap a layer to run in a given AMP context.
+
+    Args:
+        enabled: whether to enable AMP autocast
+        net: the wrapped network
+    """
+
+    def __init__(self, enabled: bool, net: nn.Module):
+        super().__init__()
+        self._net = net
+        self._enabled = enabled
+
+    def forward(self, input):
+        if torch.is_autocast_enabled() and not self._enabled:
+            input = alf.nest.map_structure(
+                lambda x: x.float() if x.dtype.is_floating_point else x, input)
+        with torch.cuda.amp.autocast(self._enabled):
+            return self._net(input)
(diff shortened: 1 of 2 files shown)


alf/layers.py
```diff
@@ -3467,3 +3467,24 @@ def _make_spec(spec):
...
+
+class AMPWrapper(nn.Module):
+    """Wrap a layer to run in a given AMP context.
+
+    Args:
+        enabled: whether to enable AMP autocast
+        net: the wrapped network
+    """
+
+    def __init__(self, enabled: bool, net: nn.Module):
```
hnyu, line 3480: enabled defaults to True is better?

alf/networks/networks.py
```diff
@@ -284,3 +286,25 @@ def __init__(self, input_tensor_spec, delay=1, name='Delay'):
...
+
+class AMPWrapper(Network):
+    """Wrap a network to run in a given AMP context.
+
+    Args:
+        enabled: whether to enable AMP autocast
+        net: the wrapped network
+    """
+
+    def __init__(self, enabled: bool, net: Network):
```
hnyu, line 299: enabled defaults to True is better?

alf/layers.py
```diff
@@ -3467,3 +3467,24 @@ def _make_spec(spec):
...
+        super().__init__()
+        self._net = net
+        self._enabled = enabled
+
+    def forward(self, input):
+        if torch.is_autocast_enabled() and not self._enabled:
+            input = alf.nest.map_structure(
+                lambda x: x.float() if x.dtype.is_floating_point else x, input)
```
breakds, line 3488: If I understand this correctly, the design is to use `AMPWrapper(false)` to force disabling AMP. If so, it would be easier to read if we provide a separate `DisableAMPWrapper` instead.

alf/layers.py
emailweixu: True and False are useful.

alf/layers.py
emailweixu: We want to able to easily switch between "enable" and "disable" to compare performance. Using a variable for it is easier.

breakds: Make sense. Thanks for the clarification!

breakds (approved):

PR #1265, opened 2022-04-04
NeroPlus optimizer

emailweixu:
This is an enhanced version of the Nero optimizer described in the following  paper:

Yang Liu et. al. "Learning by Turning: Neural Architecture Aware Optimisation" <https://arxiv.org/abs/2102.07227

The enhancement includes:

1. Option for ADAM like update (normalizing_grad_by_norm=False)
2. Upper bound contraint of weight norm (fixed_norm=False)
3. Weight decay

### DIFF
--- a/alf/optimizers/nero_plus.py
+++ b/alf/optimizers/nero_plus.py
@@ -20,28 +20,25 @@ from typing import Tuple, Optional
 from .utils import get_opt_arg
 
 
-def _norm(x, max_norm):
+def _norm(x):
     if x.ndim > 1:
         n = x.reshape(x.shape[0], -1).norm(dim=1)
         return n.reshape(x.shape[0], *([1] * (x.ndim - 1)))
-    elif max_norm is not None:
-        return x.norm()
     else:
-        return x.abs()
+        return x.norm()
 
 
 def _normalize(p,
                max_norm: Optional[float] = 1,
                fixed_norm: bool = True,
                zero_mean: bool = True):
-    if p.ndim > 1:
-        if zero_mean:
+    if zero_mean:
+        if p.ndim > 1:
             p.data -= p.mean(dim=tuple(range(1, p.ndim)), keepdim=True)
-    elif p.ndim == 1:
-        if zero_mean:
+        elif p.ndim == 1:
             p.data -= p.mean()
     if max_norm != math.inf:
-        scale = max_norm / (_norm(p, max_norm) + 1e-30)
+        scale = max_norm / (_norm(p) + 1e-30)
         if not fixed_norm:
             scale.clamp_(max=1)
         p.data *= scale
@@ -77,10 +74,10 @@ class NeroPlus(Optimizer):
     output. For example, ``FC(m, n)`` have two parameters, its weight of shape
     [m, n] and its bias of shape [n]. Its weight have m parameter vectors. Each
     of these m vectors is subject to the norm and mean constraint. For the bias,
-    one element is responsible for one output dimension. So it is not subject the
+    one element is responsible for one output dimension. So it is not subject to the
     norm and zero-mean constraint. Since the range of the output of a model should
     not be constrained, you should set opt_config for the output layers as
-    `dict(fixed_norm=False, max_norm=math.inf)` or use a large finite ``max_norm``
+    `dict(fixed_norm=False, max_norm=math.inf, zero_mean=False)` or use a large finite ``max_norm``
     or ``weight_decay`` to introduce some regularization.
 
     For 2+ D parameter p, its parameter vectors are assumed to be p[0], p[1] ... p[-1].
@@ -137,7 +134,7 @@ class NeroPlus(Optimizer):
                  }],
                  lr: float = 0.01,
                  betas: Tuple[float] = (0.9, 0.999),
-                 eps: float = 1e-30,
+                 eps: float = 1e-7,
                  normalizing_grad_by_norm=False,
                  max_norm: float = 1,
                  weight_decay: float = 0,
@@ -153,10 +150,8 @@ class NeroPlus(Optimizer):
             fixed_norm=fixed_norm,
             zero_mean=zero_mean)
         super().__init__(params, defaults)
-        assert 0 <= betas[0] < betas[0], (
-            "Invalid value for betas[0]=%s" % betas[0])
-        assert 0 < betas[1] < betas[1], (
-            "Invalid value for betas[1]=%s" % betas[1])
+        assert 0 <= betas[0] < 1, ("Invalid value for betas[0]=%s" % betas[0])
+        assert 0 < betas[1] < 1, ("Invalid value for betas[1]=%s" % betas[1])
 
     def add_param_group(self, param_group):
         super().add_param_group(param_group)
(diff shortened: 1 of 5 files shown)


alf/optimizers/nero_plus.py
```diff
@@ -0,0 +1,263 @@
...
+from typing import Tuple, Optional
+from .utils import get_opt_arg
+
+
+def _norm(x, max_norm):
+    if x.ndim > 1:
+        n = x.reshape(x.shape[0], -1).norm(dim=1)
+        return n.reshape(x.shape[0], *([1] * (x.ndim - 1)))
+    elif max_norm is not None:
+        return x.norm()
```
Haichao-Zhang: When ``max_norm`` is not None, based on its meaning, it seems that we should clamp the norm in this case?

alf/optimizers/nero_plus.py
```diff
@@ -0,0 +1,263 @@
...
+    `Yang Liu et. al. Learning by Turning: Neural Architecture Aware Optimisation
+    <https://arxiv.org/abs/2102.07227>`_
+
+    The essence of this optimizer is to keep the norm of each parameter vector
+    fixed and mean at zero during the optimization process. The parameter vector
+    is defined as the part of parameter responsible for one dimension of the
+    output. For example, ``FC(m, n)`` have two parameters, its weight of shape
+    [m, n] and its bias of shape [n]. Its weight have m parameter vectors. Each
+    of these m vectors is subject to the norm and mean constraint. For the bias,
+    one element is responsible for one output dimension. So it is not subject the
```
Haichao-Zhang: subject -> subject to

alf/optimizers/nero_plus.py
```diff
@@ -0,0 +1,263 @@
...
+            pmax_norm, pfixed_norm, pzero_mean = _get_opt_args(
+                p, max_norm, fixed_norm, zero_mean)
+            norm = _norm(p, pmax_norm)
+            if p.ndim > 1:
+                if pzero_mean:
+                    mean = p.mean(dim=tuple(range(1, p.ndim)))
+                    i = mean.abs().argmax()
+                    assert mean[i].abs() < 0.01, (
+                        "Unnormalized parameter: mean()=%s"
+                        "Model should be initialized using nero_initialize()" %
```
Haichao-Zhang: ``nero_initialize()``.  Seems need to update it with ``NeroPlus.initialize()``?

alf/optimizers/nero_plus.py
```diff
@@ -0,0 +1,263 @@
...
+                        mean[i].item())
+                if pmax_norm is not None:
+                    diff = norm / pmax_norm - 1
+                    if pfixed_norm:
+                        diff.abs_()
+                    i = diff.argmax()
+                    assert diff[i] < 0.01, (
+                        "Unnormalized parameter: norm=%s. "
+                        "Model should be initialized using nero_initialize()" %
```
Haichao-Zhang: ``nero_initialize()`` ->  ``NeroPlus.initialize()``?

alf/optimizers/nero_plus.py
runjerry: In current calls of _norm(), is there any case that ``max_norm`` is None?

alf/optimizers/nero_plus.py
```diff
@@ -0,0 +1,263 @@
...
+    The essence of this optimizer is to keep the norm of each parameter vector
+    fixed and mean at zero during the optimization process. The parameter vector
+    is defined as the part of parameter responsible for one dimension of the
+    output. For example, ``FC(m, n)`` have two parameters, its weight of shape
+    [m, n] and its bias of shape [n]. Its weight have m parameter vectors. Each
+    of these m vectors is subject to the norm and mean constraint. For the bias,
+    one element is responsible for one output dimension. So it is not subject the
+    norm and zero-mean constraint. Since the range of the output of a model should
+    not be constrained, you should set opt_config for the output layers as
+    `dict(fixed_norm=False, max_norm=math.inf)` or use a large finite ``max_norm``
```
runjerry: also zero_mean=False?

alf/optimizers/nero_plus.py
```diff
@@ -0,0 +1,263 @@
...
+                    assert mean[i].abs() < 0.01, (
+                        "Unnormalized parameter: mean()=%s"
+                        "Model should be initialized using nero_initialize()" %
+                        mean[i].item())
+                if pmax_norm is not None:
+                    diff = norm / pmax_norm - 1
+                    if pfixed_norm:
+                        diff.abs_()
+                    i = diff.argmax()
+                    assert diff[i] < 0.01, (
```
runjerry, line 180: Is this 0.01 also a hyper parameter in the Nero paper? Should we make it configurable?

alf/optimizers/nero_plus.py
```diff
@@ -0,0 +1,263 @@
...
+                else:
+                    exp_avg = grad
+                if normalizing_grad_by_norm:
+                    sq = _norm(grad, pmax_norm)**2
+                else:
+                    sq = grad**2
+                state['exp_avg_sq'].lerp_(sq, 1 - beta2)
+                denom = state['exp_avg_sq'].sqrt().add_(eps)
+                step_size = lr_scale * lr * math.sqrt(
+                    bias_correction2) / bias_correction1
```
runjerry, line 256: Could you add a brief comment regarding this ``step_size`` computation?

alf/optimizers/nero_plus_test.py
```diff
@@ -0,0 +1,64 @@
...
+            alf.layers.Conv2D(32, 32, 3, strides=2, padding=1),
+            alf.layers.Reshape(-1),
+            alf.layers.FC(
+                4 * 4 * 32,
+                num_classes,
+                weight_opt_args=dict(
+                    fixed_norm=False, weight_decay=1e-3,
+                    max_norm=float('inf'))))
+        NeroPlus.initialize(model)
+        opt = NeroPlus(lr=0.01)
```
runjerry: Do you want to also include test cases beyond the default setting, say original Nero and without running average?

alf/optimizers/nero_plus.py
emailweixu: Good catch

alf/optimizers/nero_plus.py
emailweixu: This check is used to ensure NeroPlus.initialize() is called to normalize the weights before adding them to the normalizer. A slight error is expected due to limited numerical precision. And all the weights will be normalized during the subsequent optimization steps. So it does not matter.

alf/optimizers/nero_plus.py
emailweixu: Added.

alf/optimizers/nero_plus.py
emailweixu: You are right. Removed max_norm from arguments.

alf/optimizers/nero_plus_test.py
emailweixu: added

Haichao-Zhang (approved):

alf/optimizers/nero_plus.py
runjerry: Got it, thanks for the explaination.

runjerry (approved):

PR #1268, opened 2022-04-05
save untransformed time_step in exp

hnyu:
We've done this for unroll and play, but somehow missed for training from experience. 

update: now use a data transformer to set the field of 'untransformed'. Right now it has to be the first data transformer, and it's incompatible with FrameStacker (which could potentially be addressed in fact).

### DIFF
--- a/alf/algorithms/algorithm.py
+++ b/alf/algorithms/algorithm.py
@@ -1492,13 +1492,7 @@ class Algorithm(AlgorithmInterface):
             experience, experience_spec)
         experience = alf.data_structures.add_batch_info(
             experience, batch_info, self._replay_buffer)
-        untransformed_time_step = None
-        if hasattr(experience, 'time_step'):
-            untransformed_time_step = experience.time_step
         experience = self.transform_experience(experience)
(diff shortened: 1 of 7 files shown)


alf/algorithms/algorithm.py
```diff
@@ -1492,7 +1492,13 @@ def _prepare_experience_data(self,
             experience, experience_spec)
         experience = alf.data_structures.add_batch_info(
             experience, batch_info, self._replay_buffer)
+        untransformed_time_step = None
+        if hasattr(experience, 'time_step'):
+            untransformed_time_step = experience.time_step
```
emailweixu: Perhaps it's better to add a new DataTransformer whose sole function is to save the untransformed_time_step. untransformed_time_step is not used in most cases and sometime the time_step use a lot of memory.

hnyu: > Perhaps it's better to add a new DataTransformer whose sole function is to save the untransformed_time_step. untransformed_time_step is not used in most cases and sometime the time_step use a lot of memory.

updated

emailweixu (approved):

PR #1274, opened 2022-04-12
add option of no state resetting for play

hnyu:
Add an option of not resetting states during play. This is mainly used for debugging, for example, using `alf.summary.render` to visualize the prediction statistics. One scenario is to enumerate all skills across episodes played.

### DIFF
--- a/alf/bin/play.py
+++ b/alf/bin/play.py
@@ -82,6 +82,8 @@ def _define_flags():
         'this flag to play a model trained with legacy ALF code.')
     flags.DEFINE_integer('parallel_play', 1,
                          'Play so many simulations simultaneously')
+    flags.DEFINE_bool('state_resetting', True,
+                      "Whether to reset states for StepType.FIRST")
 
 
 FLAGS = flags.FLAGS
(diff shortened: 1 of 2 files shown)


emailweixu:
> 

I still don't understand the use scenario.

hnyu:
> > 
> 
> I still don't understand the use scenario.

For example, if we want to have correlation in the states across episodes when play. One use case is to use state to store an index representing the skill. In this way, all the skills can be examined in order with play. Compared to randomly sampling an index for each episode, this kind of enumeration is better. There might be other scenarios that need this feature for debugging.

emailweixu:
> > > 
> > 
> > 
> > I still don't understand the use scenario.
> 
> For example, if we want to have correlation in the states across episodes when play. One use case is to use state to store an index representing the skill. In this way, all the skills can be examined in order with play. Compared to randomly sampling an index for each episode, this kind of enumeration is better. There might be other scenarios that need this feature for debugging.

It doesn't seems not a generic scenario. In this particular case, perhaps the algorithm can maintain a member variable by itself to handle this?

hnyu:
> > > > 
> > > 
> > > 
> > > I still don't understand the use scenario.
> > 
> > 
> > For example, if we want to have correlation in the states across episodes when play. One use case is to use state to store an index representing the skill. In this way, all the skills can be examined in order with play. Compared to randomly sampling an index for each episode, this kind of enumeration is better. There might be other scenarios that need this feature for debugging.
> 
> It doesn't seems not a generic scenario. In this particular case, perhaps the algorithm can maintain a member variable by itself to handle this?

Sure. I will close this PR.

PR #1275, opened 2022-04-12
Refactor the representation loss for MCTSModel

emailweixu:
Make it an argument of MCTSModel so that we can switch to different losses flexibly.

### DIFF
--- a/alf/algorithms/mcts_models.py
+++ b/alf/algorithms/mcts_models.py
@@ -132,6 +132,7 @@ class MCTSModel(nn.Module, metaclass=abc.ABCMeta):
             initial_alpha=0.0,
             reward_loss: losses.ScalarPredictionLoss = losses.SquareLoss(),
             value_loss: losses.ScalarPredictionLoss = losses.SquareLoss(),
+            repr_loss: Callable = losses.MeanSquaredLoss(batch_dims=2),
             target_entropy=None,
             alpha_adjust_rate=0.001,
             initial_loss_weight: Optional[float] = 1,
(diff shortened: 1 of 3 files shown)


alf/algorithms/mcts_models.py
```diff
@@ -423,24 +430,14 @@ def calc_loss(self, model_output: ModelOutput,
...
-                lambda x: x.reshape(-1, *x.shape[2:]), target.observation)
-            # [B*unroll_steps, ...]
-            repr = _flatten(model_output.state.state)
-            # [B*unroll_steps]
-            repr_loss = self.calc_repr_prediction_loss(repr, target_repr)
-            # [B, unroll_steps]
-            repr_loss = repr_loss.reshape(batch_size, -1)
+            with alf.summary.scope(self._name):
+                # [B, unroll_steps + 1]
+                repr_loss = self._repr_loss(model_output.state.state,
```
breakds, line 435: So we are no longer excluding the first step any more (was in `_flatten`). Before this PR `repr` and `target_repr` will have different shape, and I think this change fixes it.

However, if I understand this correctly, This is trying to bring the following two closer:

1. representation from replay buffer for step `t`, likely computed from a very old model
2. representation generated by the up-to-date model for step `t`

Is this intended?

alf/algorithms/mcts_models.py
```diff
@@ -690,9 +687,6 @@ def __init__(self,
                  train_reward_function=True,
                  train_game_over_function=True,
                  train_repr_prediction=False,
-                 repr_projection_net_ctor=create_repr_projection_net,
```
breakds, line 693: Do we still want to keep `create_repr_projection_net`?

emailweixu: Should be removed.

alf/algorithms/mcts_models.py
emailweixu: Both of them are calculated using the current model.

breakds (approved):

PR #1280, opened 2022-04-13
PPG will save untransformed time step in the replay buffer

breakds:
# Motivation

As @hnyu has pointed out, `PPGAuxAlgorithm`, while maintaining its own replay buffer, saves transformed `TimeStep` instead of untransformed `TimeStep`. We could save untransformed instead, and have them transformed when being sampled, at the cost of slightly more running time.

# Solution

1. Save untransformed `TimeStep` in `observe_for_aux_replay`.
2. `PPGAuxAlgorithm` will inherit the data transformer too.

# Testing

Can still run training without problem. Will run an experiment and post later.

### DIFF
--- a/alf/algorithms/ppg_algorithm.py
+++ b/alf/algorithms/ppg_algorithm.py
@@ -14,6 +14,7 @@
 """Phasic Policy Gradient Algorithm."""
 
 from __future__ import annotations
+from alf.algorithms.data_transformer import SequentialDataTransformer, UntransformedTimeStep
 import torch
 
 from typing import Optional, Tuple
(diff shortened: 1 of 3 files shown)


breakds:
new_ppg

The training performance is the same as before, on procgen bossfight.

hnyu:
I think now in order to access ``.untransformed``, you need to put ``UntransformedTimeStep`` in the ``data_transformer_ctor``.

breakds:
> I think now in order to access `.untransformed`, you need to put `UntransformedTimeStep` in the `data_transformer_ctor`.

This is fixed now. Also added an asert in `PPGAlgorithm` to remind the user about configuring `UntransformedTimeStep` if they forget.

alf/algorithms/ppg_algorithm.py
```diff
@@ -145,6 +149,18 @@ def __init__(self,
...
 
+    @staticmethod
+    def _validate_data_transformer(data_transformer):
+        """Returns True if UntransformedTimeStep is present and is applied
+        before all other data transformers.
+
+        """
+        if type(data_transformer) is UntransformedTimeStep:
+            return True
+        if type(data_transformer) is SequentialDataTransformer:
```
hnyu, line 158: It's always wrapped in SequentialDataTransformer?

breakds: Verified and fixed. Thanks!

hnyu (approved):

PR #1282, opened 2022-04-15
ALF will save a visualization of the algorithm module structure

breakds:
# Motivation

For more complex algorithms, it is useful to be able to visualize the hierarchy of the algorithm, sub algorithms and network structures so that we can make sure the components are correctly wired.

# Solution

Save a visualized graph of algorithm hierarchy when summarizing the training settings.

This is the first implementation and many can be improved from here.

# Testing

An example generated graph for PPG + Transformer + MetaDrive:

algorithm_sturcture

### DIFF
--- a/alf/trainers/policy_trainer.py
+++ b/alf/trainers/policy_trainer.py
@@ -25,6 +25,8 @@ import sys
 import time
 import torch
 import torch.nn as nn
+from PIL import Image
+import numpy as np
 
 import alf
 from alf.algorithms.algorithm import Algorithm, Loss
@@ -105,8 +107,9 @@ def _visualize_alf_tree(module: Algorithm):
     try:
         import graphviz
     except ImportError:
-        raise RuntimeError(
+        logging.warn(
             'Need "graphviz" installed if you want to visualize modules')
+        return None
 
     def _visual_style(node: torch.nn.Module) -> Dict[str, str]:
         if isinstance(node, Loss):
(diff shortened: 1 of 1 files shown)


alf/trainers/policy_trainer.py
```diff
@@ -246,6 +310,11 @@ def _markdownify(paragraph):
             alf.summary.text('diff', _markdownify(git_utils.get_diff()))
             alf.summary.text('seed', str(self._random_seed))
 
+            # Save a rendered directed graph of the algorithm to the root
+            # directory.
+            _visualize_alf_tree(self._algorithm).render(
+                Path(flags.FLAGS.root_dir, 'algorithm_sturcture'))
```
hnyu: should probably use `self._root_dir`

alf/trainers/policy_trainer.py
```diff
@@ -246,6 +310,11 @@ def _markdownify(paragraph):
             alf.summary.text('diff', _markdownify(git_utils.get_diff()))
             alf.summary.text('seed', str(self._random_seed))
 
+            # Save a rendered directed graph of the algorithm to the root
+            # directory.
+            _visualize_alf_tree(self._algorithm).render(
```
hnyu: Do you think it's better to put this image in Tensorboard instead?

breakds: Should be able to do that on top of saving it to the `root_dir`. I could not find a way to directly render graphviz to tensor, so the procedure has to be saving it to the disk, load the image as tensor and summarize it. I'll give it a try tomorrow.

breakds:
Apparently also need to install `graphviz` :) Will make that happen in another PR.

hnyu:
> Apparently also need to install `graphviz` :) Will make that happen in another PR.

I think in order to insteall graphviz, you need to modify the docker file to use apt install.

breakds:
Updated the PR with

1. Also send the graph to tensorboard.
2. If `graphviz` is not present, warn instead of panic.

tensorboard_alg_structure

alf/examples/metadrive/ppg_metadrive_conf.py
```diff
@@ -30,7 +30,7 @@
 
 # Environment Configuration
 alf.config(
-    'create_environment', env_name='BirdEye', num_parallel_environments=36)
+    'create_environment', env_name='BirdEye', num_parallel_environments=3)
```
hnyu: Why having this change to the conf.py?

breakds: Fixed. This is included by accident.

hnyu (approved):

PR #1283, opened 2022-04-19
make Checkpointer.load() configurable

hnyu:
This PR is a follow-up of #1264 . It makes `Checkpointer.load()` configurable, for a scenario of downstream task training after a pre-training stage. Basically we might not want to load the stats for replay buffer, optimizer, or data transformers, because the pretraining and downstream tasks might not share the same data distribution. 

Also reverted play.py to always require strict loading as I don't see the necessity of a non-strict loading in this case.

### DIFF
--- a/alf/algorithms/config.py
+++ b/alf/algorithms/config.py
@@ -35,7 +35,6 @@ class TrainerConfig(object):
                  num_checkpoints=10,
                  confirm_checkpoint_upon_crash=True,
                  no_thread_env_for_conf=False,
-                 load_checkpoint_options={},
                  evaluate=False,
                  num_evals=None,
                  eval_interval=10,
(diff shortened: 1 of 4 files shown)


alf/algorithms/config.py
```diff
@@ -146,12 +146,9 @@ def __init__(self,
                 ``True`` if no evaluation is needed to save resources. The decision
                 of creating an unwrapped env won't affect training; it's used to
                 correctly display inoperative configurations in subprocesses.
-            load_checkpoint_strict (bool): whether to strictly enforce that the keys
-                in ``state_dict`` match the keys returned by module's
-                ``torch.nn.Module.state_dict`` function. If True, will
-                keep lists of missing and unexpected keys and raise error when
-                any of the lists is non-empty; if ``strict=False``, missing/unexpected
-                keys will be omitted and no error will be raised.
+            load_checkpoint_options (dict): optional args for the ``load()`` of
```
emailweixu: This can be achieved simply by making Checkpointer.load() configurable?

emailweixu:
> This PR is a follow-up of #1264 . It adds more checkpoint loading options in TrainerConfig, for a scenario of downstream task training after a pre-training stage. Basically we might not want to load the stats for replay buffer, optimizer, or data transformers, because the pretraining and downstream tasks might not share the same data distribution.
> 
> Also reverted play.py to always require strict loading as I don't see the necessity of a non-strict loading in this case.

Need to update the PR description.

alf/algorithms/config.py
```diff
@@ -35,7 +35,6 @@ def __init__(self,
                  num_checkpoints=10,
                  confirm_checkpoint_upon_crash=True,
                  no_thread_env_for_conf=False,
-                 load_checkpoint_strict=True,
```
Haichao-Zhang, line 38: why remove this option?

hnyu: > why remove this option?

Because now `Checkpointer.load()` can be directly configured.

Haichao-Zhang (approved):

PR #1285, opened 2022-04-19
improve visualize_alf_tree

hnyu:
Several improvements:
1. When visualizing layers, use their `__repr__()` instead of class names. Most layers don't have `__repr__()` defined yet. Need to add `alf.repr_wrapper` for them.
2. Render subgraphs, each for a subalgorithm.
3. use DFS instead of the original BFS so that the graph is better organized. 
4. add a filter function to hide trivial nodes.

A comparison:
before:
Feishu20220418-211820
after (TAACAlgorithm is turned off):
image

### DIFF
--- a/alf/layers.py
+++ b/alf/layers.py
@@ -462,6 +462,18 @@ class FC(nn.Module):
         """
         return ParallelFC(n=n, **self._kwargs)
 
+    def __repr__(self):
+        ret = (f"FC(in={self._input_size},\n"
+               f"   out={self._output_size},\n"
+               f"   activation={self._activation.__name__},\n")
+        if self._use_bias:
+            ret += "   bias=True,\n"
+        if self._use_bn:
+            ret += "   batch_norm=True,\n"
+        if self._use_ln:
+            ret += "   layer_norm=True,\n"
+        return ret + "  )"
+
 
 @alf.configurable
 class FCBatchEnsemble(FC):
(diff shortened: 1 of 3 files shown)


alf/layers.py
```diff
@@ -462,6 +462,18 @@ def make_parallel(self, n: int):
         """
         return ParallelFC(n=n, **self._kwargs)
 
+    def __repr__(self):
```
emailweixu: Perhaps try @alf.repr_wrapper?

hnyu: > Perhaps try @alf.repr_wrapper?

Updated to use repr_wrapper. The only downside is that sometimes the repr str can get really long.

alf/layers.py
```diff
@@ -313,6 +313,7 @@ def weight(self):
 
 
 @alf.configurable
+@alf.config_util.repr_wrapper
```
emailweixu: no need to have config_util: @alf.repr_wrapper

alf/trainers/policy_trainer.py
```diff
@@ -127,31 +138,83 @@ def _visual_style(node: torch.nn.Module) -> Dict[str, str]:
...
-            child_index = reverse_map[child]
+            if _filter_child(field, child):
+                continue
+            if child not in visited:
+                edges += _visit(child, idx, visited)
+            child_idx = visited[child]
             node_records.append(f'<{field}> ({field})')
-            dot.edge(f'{node_index}:{field}', f'{child_index}:caption')
+            edge = (f'{node_index}:{field}', f'{child_idx}:caption')
+            #dot.edge(*edge)
```
breakds: Remove this?

breakds (dismissed):

breakds (approved):

PR #1290, opened 2022-04-27
Algorithm can optionally choose to expose already-handled parameters to its parent

breakds:
## Motivation

The default ALF behavior on assigning parameters to optimizers will

1. Assign sub algorithm's parameters to sub algorithm's optimizer
2. The parent algorithm will not see sub algorithm's parameters if they are already assigned to an optimizer.

This is sometimes needed. For example when combining PPO with Muzero Representation Learner, it is desired that both PPO update and MuZero representation update will be able to train the representation network's parameters.

# Solution

Add an option in `Algorithm` that when set to `True`, it will switch the behavior to expose those already handled parameters to its parent algorithm.

# Testing

Tested together with the MuZero Representation + PPO implementation. The parameters are successfully exposed and training performs nicely.

Since this option is by default set to `False`, existing algorithm **will not** be affected.

### DIFF
--- a/alf/algorithms/algorithm.py
+++ b/alf/algorithms/algorithm.py
@@ -570,12 +570,7 @@ class Algorithm(AlgorithmInterface):
             if isinstance(child, Algorithm):
                 params, child_handled = child._setup_optimizers_(param_to_name)
                 if child.force_params_visible_to_parent:
-                    merged = set(params)
-                    for p in child_handled:
-                        if p in merged:
-                            continue
-                        params.append(p)
-                        merged.add(p)
+                    params += child_handled
                 else:
                     for m in child_handled:
                         assert m not in handled, duplicate_error % param_to_name.get(
--- a/alf/algorithms/algorithm.py
+++ b/alf/algorithms/algorithm.py
@@ -570,7 +570,12 @@ class Algorithm(AlgorithmInterface):
             if isinstance(child, Algorithm):
                 params, child_handled = child._setup_optimizers_(param_to_name)
                 if child.force_params_visible_to_parent:
-                    params = list(set(params + child_handled))
+                    merged = set(params)
+                    for p in child_handled:
+                        if p in merged:
+                            continue
+                        params.append(p)
+                        merged.add(p)
                 else:
                     for m in child_handled:
                         assert m not in handled, duplicate_error % param_to_name.get(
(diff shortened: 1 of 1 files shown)


alf/algorithms/algorithm.py
```diff
@@ -544,10 +569,13 @@ def _add_params_to_optimizer(params, opt):
             assert id(child) != id(self), "Child should not be self"
             if isinstance(child, Algorithm):
                 params, child_handled = child._setup_optimizers_(param_to_name)
-                for m in child_handled:
-                    assert m not in handled, duplicate_error % param_to_name.get(
-                        m)
-                    handled[m] = 1
+                if child.force_params_visible_to_parent:
+                    params = list(set(params + child_handled))
```
emailweixu: I am wondering the use of `set` here can cause issue for checkpoint. See a related PR https://github.com/HorizonRobotics/alf/pull/891

breakds: Make sense. Updated the logic to preserve the order.

alf/algorithms/algorithm.py
```diff
@@ -544,10 +569,18 @@ def _add_params_to_optimizer(params, opt):
...
             if isinstance(child, Algorithm):
                 params, child_handled = child._setup_optimizers_(param_to_name)
-                for m in child_handled:
-                    assert m not in handled, duplicate_error % param_to_name.get(
-                        m)
-                    handled[m] = 1
+                if child.force_params_visible_to_parent:
+                    merged = set(params)
+                    for p in child_handled:
+                        if p in merged:
```
hnyu: When will this happen? By definition, `params` are not handled by the child optimizer, while `child_handled` are those handled. Should these two be disjoint always? 
(And it seems unnecessary to introduce `merged`? Just `if p in params`.

breakds: I think they should be disjoint, but not 100% certain - that is why I added this logic to ensure correctness.

> And it seems unnecessary to introduce merged? Just if p in params.

Agree that `p in params` may not be much slower in most case. I am being conservative here to use a hashtable (i.e. `set`) because the number of parameters can potentially be high (i.e. > 100). `_setup_optimizers_` is called multiple times every iteration for now, and that is why I am a bit sensitive on performance here.

hnyu: > I think they should be disjoint, but not 100% certain - that is why I added this logic to ensure correctness.

I think it's better to assert that they are disjoint instead of silently allowing this, before we know a use case of an overlap in the two returns.

breakds: You are right. After reading the code once more I am convinced they are disjoint. Removing the logic to make it simple. Thanks!

hnyu (approved):

PR #1291, opened 2022-04-27
AsyncEvaluator

### LINKED ISSUE #1271: Improve evaluation during training
There are several problems with the current evaluation during training:

1. It blocks the training. For tasks with long episodes, it can significantly slow down training.
2. It only uses a single environment for evaluation. If we want to evaluate multiple episodes. It can take even more time.

emailweixu:
Fixes #1271 

The AsyncEvaluator creates a different process to evaluate the performance asynchronously so that it will not block the training. It also supports evaluation using parallel environments by setting TrainerConfig.num_eval_environments to number greater than one.

### DIFF
--- a/alf/algorithms/config.py
+++ b/alf/algorithms/config.py
@@ -42,6 +42,7 @@ class TrainerConfig(object):
                  epsilon_greedy=0.,
                  eval_uncertainty=False,
                  num_eval_episodes=10,
+                 num_eval_environments: int = 1,
                  num_summaries=None,
                  summary_interval=50,
                  summarize_first_interval=True,
@@ -162,7 +163,8 @@ class TrainerConfig(object):
                 help prevent a dead loop in some deterministic environment like
                 Breakout. Only used for evaluation.
             eval_uncertainty (bool): whether to evluate uncertainty after training.
-            num_eval_episodes (int) : number of episodes for one evaluation
+            num_eval_episodes (int) : number of episodes for one evaluation.
+            num_eval_environments: the number of environments for evaluation.
             num_summaries (int): how many summary calls are needed throughout the
                 training. If not None, an automatically calculated ``summary_interval``
                 will replace ``config.summary_interval``. Note that this number
@@ -283,6 +285,7 @@ class TrainerConfig(object):
         self.epsilon_greedy = epsilon_greedy
         self.eval_uncertainty = eval_uncertainty
         self.num_eval_episodes = num_eval_episodes
+        self.num_eval_environments = num_eval_environments
         self.num_summaries = num_summaries
         self.summary_interval = summary_interval
         self.summarize_first_interval = summarize_first_interval
--- a/alf/algorithms/config.py
+++ b/alf/algorithms/config.py
@@ -167,7 +167,7 @@ class TrainerConfig(object):
             num_eval_episodes (int) : number of episodes for one evaluation.
             num_eval_environments: the number of environments for evaluation.
             async_eval: whether to do evaluation asynchronously in a different
-                process. Note that this more use more memory.
+                process. Note that this may use more memory.
             num_summaries (int): how many summary calls are needed throughout the
                 training. If not None, an automatically calculated ``summary_interval``
                 will replace ``config.summary_interval``. Note that this number
--- a/alf/algorithms/config.py
+++ b/alf/algorithms/config.py
@@ -43,6 +43,7 @@ class TrainerConfig(object):
                  eval_uncertainty=False,
                  num_eval_episodes=10,
                  num_eval_environments: int = 1,
+                 async_eval: bool = True,
                  num_summaries=None,
                  summary_interval=50,
                  summarize_first_interval=True,
@@ -165,6 +166,8 @@ class TrainerConfig(object):
             eval_uncertainty (bool): whether to evluate uncertainty after training.
             num_eval_episodes (int) : number of episodes for one evaluation.
             num_eval_environments: the number of environments for evaluation.
+            async_eval: whether to do evaluation asynchronously in a different
+                process. Note that this more use more memory.
             num_summaries (int): how many summary calls are needed throughout the
                 training. If not None, an automatically calculated ``summary_interval``
                 will replace ``config.summary_interval``. Note that this number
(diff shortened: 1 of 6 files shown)


alf/metrics/metric.py
```diff
@@ -58,14 +60,18 @@ def result(self):
         raise NotImplementedError(
             'Metrics must define a result() member function')
 
-    def gen_summaries(self, train_step=None, step_metrics=()):
+    def gen_summaries(self,
+                      train_step=None,
+                      step_metrics=(),
+                      other_steps: Dict[str, Number] = dict()):
```
hnyu, line 66: Can this `other_steps` just be merged with `step_metrics`?

alf/metrics/metric.py
```diff
@@ -58,14 +60,18 @@ def result(self):
...
         """Generates summaries against train_step and all step_metrics.
 
         Args:
             train_step: (Optional) Step counter for training iterations. If None, no
                 metric is generated against the global step.
             step_metrics: (Optional) Iterable of step metrics to generate summaries
                 against.
+            other_steps: A dictionary of steps to generate summaries against.
```
hnyu, line 73: It's better to clarify that `other_steps` should directly contain integers instead of metrics, to differentiate from `step_metrics`.

alf/trainers/evaluator.py
```diff
@@ -0,0 +1,253 @@
...
+            eval_dir, flush_secs=config.summaries_flush_secs)
+        alf.summary.enable_summary()
+        logging.info("Evaluator started")
+        with alf.summary.push_summary_writer(summary_writer):
+            while True:
+                job = job_queue.get()
+                if job.type == "eval":
+                    logging.info("Start evaluation")
+                    alf.summary.set_global_counter(job.global_counter)
+                    policy_trainer.Trainer._trainer_progress.update(
```
hnyu: Why do we need to update trainer progress here?

alf/trainers/evaluator.py
```diff
@@ -0,0 +1,253 @@
...
+                        job.global_counter, job.env_steps)
+                    algorithm.load_state_dict(job.state_dict)
+                    done_queue.put(None)
+                    metrics = evaluate(env, algorithm,
+                                       config.num_eval_episodes)
+                    common.log_metrics(metrics)
+                    for metric in metrics:
+                        metric.gen_summaries(
+                            train_step=alf.summary.get_global_counter(),
```
hnyu: Should this just be `job.global_counter`?

alf/trainers/evaluator.py
```diff
@@ -0,0 +1,253 @@
...
+
+        Args:
+            algorithm: the training algorithm
+            current_env_steps: current total number of environment steps.
+        """
+        job = EvalJob(
+            type="eval",
+            env_steps=int(current_env_steps),
+            global_counter=int(alf.summary.get_global_counter()),
+            state_dict=algorithm.state_dict())
```
hnyu: This limited number of EvalJob fields seems to make `gen_summaries` with other metrics not useful. Might have a field that store a list of steps?

hnyu (changes requested):

alf/metrics/metric.py
emailweixu: The type of other_steps indicates so.

alf/trainers/evaluator.py
emailweixu: Changed env_steps to step_metric_values.

alf/trainers/evaluator.py
emailweixu: comment added.

alf/trainers/evaluator.py
emailweixu: They are same because the set_global_counter above.

breakds:
Code change LGTM. The amortized GPU memory consumption should only slightly increase if my understanding is correct.

breakds (dismissed):

alf/trainers/evaluator.py
```diff
@@ -0,0 +1,259 @@
...
+                the evaluation summaries against. Note that it needs to contain
+                "EnvironmentSteps" at least.
+        """
+        job = EvalJob(
+            type="eval",
+            step_metrics=step_metric_values,
+            global_counter=int(alf.summary.get_global_counter()),
+            state_dict=algorithm.state_dict())
+        self._job_queue.put(job)
+        self._done_queue.get()
```
le-horizon: When a previous eval job is still running and eval is called again, the worker is still busy working on the previous job and cannot yet take the job off the queue, so in this case, the evaluator will block on this done_queue.get()?

Maybe let's add a timer here to log to tensorboard how long this done_queue.get() is taking, so that we'll know that the eval process is blocking the training.

le-horizon (commented):
Nice change, Wei!

hnyu (dismissed):

alf/algorithms/config.py
```diff
@@ -165,6 +166,8 @@ def __init__(self,
             eval_uncertainty (bool): whether to evluate uncertainty after training.
             num_eval_episodes (int) : number of episodes for one evaluation.
             num_eval_environments: the number of environments for evaluation.
+            async_eval: whether to do evaluation asynchronously in a different
+                process. Note that this more use more memory.
```
breakds: s/more/may/

breakds (dismissed):

breakds (approved):

alf/trainers/evaluator.py
```diff
@@ -38,29 +40,31 @@
...
+        self._async = config.async_eval
+        num_envs = config.num_eval_environments
+        seed = config.random_seed
+        if self._async:
+            ctx = mp.get_context('spawn')
+            self._job_queue = ctx.Queue()
+            self._done_queue = ctx.Queue()
+            self._worker = ctx.Process(
+                target=_worker,
+                args=(self._job_queue, self._done_queue, conf_file, num_envs,
```
hnyu, line 67: Can we directly pass `config` to `_worker` instead of letting it parse `conf_file`?

emailweixu: The subprocess is created using spawn method. Different from fork, config needs to be parsed again in the spawned subprocess anyway.

hnyu (approved):

alf/trainers/evaluator.py
emailweixu: added

PR #1294, opened 2022-04-28
fix nested observation for TAAC algorithms

hnyu:
A bug similar to #1287 where some places didn't handle a nested observation.

### DIFF
--- a/alf/algorithms/taac_algorithm.py
+++ b/alf/algorithms/taac_algorithm.py
@@ -424,10 +424,10 @@ class TaacAlgorithmBase(OffPolicyAlgorithm):
 
         tau_spec = nest.map_structure(lambda m: action_spec if m else (),
                                       tau_mask)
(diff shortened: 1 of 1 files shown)


alf/algorithms/taac_algorithm.py
```diff
@@ -475,13 +475,14 @@ def _predict_action(self,
                         mode=Mode.rollout):
 
         observation = time_step.observation
+        obs_shape = alf.nest.get_nest_shape(observation)
```
emailweixu: Why can you assume all the fields of observation have the same shape?

hnyu: > Why can you assume all the fields of observation have the same shape?

Here I only need to extract the shape from an arbitrary element, because later I will only use its batch dim.

emailweixu: If it's for batch_size, then you can use `batch_shape = time_step.step_type.shape`

emailweixu (approved):

PR #1299, opened 2022-04-29
fix d_ff in TransformerNetwork

hnyu:
Currently d_ff is not passed to ``TransformerBlock`` which is always using ``4*d_model``.


alf/networks/transformer_networks.py
```diff
@@ -56,7 +56,7 @@ def __init__(self,
                  input_tensor_spec,
                  num_prememory_layers,
                  num_attention_heads,
-                 d_ff,
+                 d_ff=None,
```
emailweixu, line 59: doc-string needs to comment what None means

breakds (approved):

PR #1300, opened 2022-05-01
add nest.transpose and nest.nest_top_level

hnyu:
Add a nest transpose function which could be helpful if we want to swap an outer and inner structure.  See the docstring for example usage.

### DIFF
--- a/alf/nest/nest.py
+++ b/alf/nest/nest.py
@@ -936,9 +936,7 @@ def transpose(nested: Nest, shallow_nest: Nest, new_shallow_nest: Nest = None):
     """
     leaves = flatten_up_to(shallow_nest, nested)
     for leaf in leaves:
-        assert_same_structure(leaves[0], leaf), (
-            "All leaves under the shallow nest should have the same structure!"
-            " Got %s vs. %s" % (leaves[0], leaf))
+        assert_same_structure(leaves[0], leaf)
 
     if new_shallow_nest is None:
         new_shallow_nest = leaves[0]
--- a/alf/nest/nest.py
+++ b/alf/nest/nest.py
@@ -885,11 +885,11 @@ def set_field(nested, field, new_value):
     return transform_nest(nested, field, lambda _: new_value)
 
 
-def transpose(nested, shallow_nest):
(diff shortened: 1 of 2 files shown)


alf/nest/nest.py
```diff
@@ -883,3 +883,36 @@ def set_field(nested, field, new_value):
     """
 
     return transform_nest(nested, field, lambda _: new_value)
+
+
+def transpose(nested, shallow_nest):
+    """Given a nest and its shallow nest structure ``B``, assuming that each leaf
```
breakds: Is "shallow nest structure" of a nest a well-known concept? Not sure what's the best way to phrase this though, but it took me a while to understand that `shallow_nest` here defines "what are considered leaves".

alf/nest/nest.py
```diff
@@ -883,3 +883,36 @@ def set_field(nested, field, new_value):
     """
 
     return transform_nest(nested, field, lambda _: new_value)
+
+
+def transpose(nested, shallow_nest):
```
breakds: I think `inside_out` might better describe the operation, if `transpose` is not a convention.

alf/nest/nest.py
hnyu: > Is "shallow nest structure" of a nest a well-known concept? Not sure what's the best way to phrase this though, but it took me a while to understand that `shallow_nest` here defines "what are considered leaves".

Actually `shallow_nest` is meant to be the top level of a nest. For example, `dict(x=1,y=1)` is a shallow nest of `dict(x=[0,1], y=1)`. This term has been used in `flatten_up_to` and `map_structure_up_to`. But for clarity, I've changed it to "top level".

alf/nest/nest.py
hnyu: > I think `inside_out` might better describe the operation, if `transpose` is not a convention.

I've added a new arg to specify a new shallow nest. In general, it can interchange any two "shallow nests", which is analogous to matrix transpose of two axes.

breakds:
Thanks for the update!

breakds (dismissed):

breakds (approved):

PR #1302, opened 2022-05-02
l2_regularization and weight_decay for NeroPlus

emailweixu:
Differentiate l2 regularization and weight decay as AdamW suggested (https://arxiv.org/abs/1711.05101).

### DIFF
--- a/alf/optimizers/nero_plus.py
+++ b/alf/optimizers/nero_plus.py
@@ -94,7 +94,7 @@ class NeroPlus(Optimizer):
 
     1. Option for ADAM like update (normalizing_grad_by_norm=False)
     2. Upper bound contraint of weight norm (fixed_norm=False)
-    3. Weight decay
+    3. Weight decay and L2 regularization.
 
     To use this optimizer, you should first use ``NeroPlus.initialize()`` to normalize
     the parameter of your model for the given constrains before actually using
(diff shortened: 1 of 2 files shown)


alf/optimizers/nero_plus.py
```diff
@@ -249,6 +258,8 @@ def step(self, closure=None):
                     sq = grad**2
                 state['exp_avg_sq'].lerp_(sq, 1 - beta2)
                 denom = state['exp_avg_sq'].sqrt().add_(eps)
+                if pweight_decay > 0:
+                    p.mul_(1 - lr_scale * lr)
```
Haichao-Zhang: missing ```* pweight_decay```?

emailweixu: good catch. No wonder the unittest coundn't pass when weight decay is used.

Haichao-Zhang (approved):

alf/optimizers/nero_plus.py
```diff
@@ -116,12 +116,16 @@ class NeroPlus(Optimizer):
             the ADAM like behavoir is better.
         max_norm: maximal norm of each parameter vector. A parameter vector is
             part of a parameter responsible for one output dimension.
-        weight_decay: weight decay (L2 penalty).
+        weight_decay: weight decay. This is same as the weight decay of AdamW,
+            which is implemented as substracting `lr * weight_deday * w` from
+            parameter.
+        l2_regularization: L2 penalty. This is same as the weight decay of Adam,
+            which is implemented as adding `weight_decay * w` to gradient.
```
runjerry, line 123: Maybe more intuitive if ``w`` -> ``current_w``? same for line 120

PR #1303, opened 2022-05-02
Several improvements on PPG + Transformer + MetaDrive

breakds:
## Motivation

There are a few changes on the vectorized MetaDrive experiment that are not committed yet. This PR summarizes them.

## Changes

1. Add `history_frame_skip` for other agents' trajectory inputs. Instead of putting all the frames within the history window into the trajectory, it picks only every `history_frame_skip` frames so that the input is significantly smaller, without much loss of information.
2. Observation includes `map_mask` and `agent_mask`, so that the transformer encoder knows what slots need to be ignored.
3. In `layers.TransformerBlock`, correct the meaning of `mask` by reversing it.

## Testing

Ran experiments to make sure the changes aligns with the expectation (i.e. resembles the previous best experiment result). In the graph below:

blue: `num_evs' = 36
orange: `num_envs` = 24

metadrive_ppg

### DIFF
--- a/alf/environments/metadrive/agent_perception.py
+++ b/alf/environments/metadrive/agent_perception.py
@@ -59,6 +59,7 @@ class AgentPerception(object):
     def __init__(self,
                  fov: FieldOfView,
                  history_window_size: int,
+                 history_frame_skip: int = 4,
                  agent_limit: int = 16):
         """Construct an AgentPerception instance.
 
@@ -68,7 +69,9 @@ class AgentPerception(object):
                 generating the agent features, only those agent who is within
                 the FOV result in the feature.
             history_window_size: The feature only tracks this number of
-                historical steps (current step included).
+                historical frames (current frame included).
+            history_frame_skip: Pick every this number of frames to form the historical
+                trajectories for the agents.
             agent_limit: The maximum number of agents shown in the feature. If
                 the number of visible agents exceeds this limit, the farthest
                 ones are filtered out until this limit is satisfied.
@@ -78,10 +81,7 @@ class AgentPerception(object):
         self._fov = fov
         self._agent_limit = agent_limit
         self._unit_feature_size = 7
-        self._spec = TensorSpec(
-            shape=(self._agent_limit, self._history_window_size,
-                   self._unit_feature_size),
-            dtype=torch.float32)
+        self._history_frame_skip = history_frame_skip
 
         self._engine = None
         self._ego = None
@@ -96,6 +96,15 @@ class AgentPerception(object):
         self._visible = None  # 1 = visible, 0 = invisible
         self._history_position = None
         self._history_heading = None
+
+        # Handle Frame Skip
+        H = self._history_window_size
+        self._sampled_index = np.arange(H)[(
+            (H - 1) % self._history_frame_skip)::self._history_frame_skip]
+        self._spec = TensorSpec(
+            shape=(self._agent_limit, self._sampled_index.shape[0],
+                   self._unit_feature_size),
+            dtype=torch.float32)
 
     @property
     def observation_spec(self):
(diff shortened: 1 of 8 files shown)


alf/layers.py
```diff
@@ -2581,7 +2581,7 @@ def calc_output_shape(self, input_shape):
 
 def _masked_softmax(logits, mask, dim=-1):
     if mask is not None:
-        logits.masked_fill_(mask, -float('inf'))
+        logits.masked_fill_(~mask, -float('inf'))
```
emailweixu: The doc-string for mask in TransformerBlock says mask is "A tensor for indicating which slot in ``memory`` will NOT be used"

alf/environments/metadrive/agent_perception.py
```diff
@@ -97,6 +97,15 @@ def __init__(self,
         self._history_position = None
         self._history_heading = None
 
+        # Handle Frame Skip
+        H = self._history_window_size
+        self._sampled_index = np.arange(H)[(
```
emailweixu: `np.arange((H-1)%skip, H, skip)`

alf/environments/metadrive/agent_perception.py
```diff
@@ -190,15 +203,18 @@ def observe(self) -> np.ndarray:
         self._visible[:, -1] = self._fov.within(
             transformed_position.point[:, -1])
         self._visible[~alive, -1] = False
+        sampled_visible = self._visible[:, self._sampled_index]
+        sampled_position = transformed_position.point[:, self._sampled_index]
+        sampled_heading = transformed_heading[:, self._sampled_index]
 
         # Shape is [B,]. Denote whether a car is picked to show in the final
         # feature tensor or not. The criterion is that the car has be visible in
```
emailweixu: has be => has been

emailweixu (changes requested):

alf/environments/metadrive/agent_perception.py
```diff
@@ -153,15 +162,19 @@ def reset(self, engine: BaseEngine, ego: BaseVehicle):
         self._history_heading = np.zeros(
             (self._num_agents, self._history_window_size), dtype=np.float32)
 
-    def observe(self) -> np.ndarray:
+    def observe(self) -> Tuple[np.ndarray, int]:
         """Called upon every observation to produce the feature vectors describing the
         dynamic agents that are visible to the ego car. The vectors are
         transformed so that they are in eog car's body frame.
```
Haichao-Zhang: eog car -> ego car

alf/layers.py
breakds: Sorry that I forgot I have already updated the doc-string. Reverted this change and made changes in the configuration instead.

alf/examples/metadrive/ppg_metadrive_transformer_conf.py
```diff
@@ -61,57 +80,85 @@ class ObservationCombiner(torch.nn.Module):
...
+                    ego_feature=EmbeddingConfig(
+                        d_input=input_tensor_spec['ego'].shape[-1],
+                        hidden=(32, )),
+                    agent_feature=EmbeddingConfig(
+                        d_input=input_tensor_spec['agents'].shape[1] *
+                        input_tensor_spec['agents'].shape[2],
+                        hidden=(32, )),
+                    d_model=d_model),
+                input_tensor_spec=input_tensor_spec), lambda x:
+            (x['map_mask'], x['agent_mask'])), lambda x: (x[0], x[1][0], x[1][
```
emailweixu: This auto-formatting makes it quite hard to read. 
The lambda at line 151 should be indented at the same level ax ObservationCombiner.
The lambda at line 152 should be at a new line and be indented at the same level as MaskedTransformer.
Perhaps use the following to disable the auto-formatting?
```
# yapf: diasble
...
# yapf: enable
```

alf/examples/metadrive/ppg_metadrive_transformer_conf.py
```diff
@@ -61,57 +80,85 @@ class ObservationCombiner(torch.nn.Module):
...
+                    d_model=d_model,
+                    num_heads=num_heads,
+                    memory_size=self._memory_size,
+                    positional_encoding='none'))
+        self._tf_layers = torch.nn.ModuleList(tf_layers)
+
+    def forward(self, inputs):
+        x, map_mask, agent_mask = inputs
+        B = x.shape[0]
+        mask = torch.hstack((torch.ones(B, 1, dtype=bool), map_mask,
```
emailweixu: torch.ones => torch.zeros

emailweixu (changes requested):

alf/examples/metadrive/ppg_metadrive_transformer_conf.py
breakds: Right, thanks for the advice. Updated.

alf/examples/metadrive/ppg_metadrive_transformer_conf.py
breakds: Thanks for pointing out this bad bug. Fixed.
 I'll run the training again to verify before submitting the PR.

breakds:
Verified that the updated PR still works in training (`num_envs` = 24), with matching performance.

metadrive_verify

Haichao-Zhang (approved):

emailweixu (approved):

PR #1304, opened 2022-05-02
Use NeroPlus for muzero_atari_conf

emailweixu:
Previously the combination of Adam and weight decay makes the training quite sensitive to hyperparameters, which can make the prediction of reward and value collapse easily if the l2 regularization is too big. With NeroPlus, we only need to set l2 regularization at the last layer, hence less likely to collapse

Evaluation curve from 4 runs:

muzero_atari_breakout
.


alf/layers.py
```diff
@@ -2413,7 +2416,7 @@ def __init__(self,
 
         conv_fn = _conv_transpose_2d if transpose else nn.Conv2d
         bias = not with_batch_normalization
-        relu = nn.ReLU(inplace=True)
+        relu = activation
```
breakds: It is more readable to just use `activation` without the alias `relu`.

alf/examples/muzero_atari_conf.py
```diff
@@ -427,30 +587,42 @@ def _scale_grad(scale):
     max_unroll_length=5,
     learn_policy_temperature=1.0)
 
+
+def model_ctor(*args, **kwargs):
+    model = SimpleMCTSModel(*args, **kwargs)
+    # Nero optimizer normalizes the parameters. We need to do this in the beginning
+    # so that the target_model is same as the model.
```
breakds, line 599: Question: `target_model` is not optimized, why do we need to run `NeroPlus.initialize()` on it?

alf/examples/muzero_atari_conf.py
```diff
@@ -310,17 +444,28 @@ def _scale_grad(scale):
...
         *_make_trunk(),
         *_scale_grad(1 / num_quantiles if scale_grad_by_num_quantiles else 1),
-        alf.layers.FC(
-            dim,
-            num_quantiles,
-            weight_opt_args=dict(weight_decay=rv_weight_decay),
-            bias_opt_args=dict(weight_decay=rv_bias_decay),
-            bias_initializer=_get_rv_bias_initializer(),
-            kernel_initializer=torch.nn.init.zeros_),
+        alf.layers.AMPWrapper(
```
breakds, line 464: Is this because the sensitivity to precision? Want to confirm this is independent of whether we use `NeroPlus` or not. 
Also suggesting adding comments here in case someone wonders the choice of explicitly disabling AMP when reading this.

alf/examples/muzero_atari_conf.py
```diff
@@ -148,11 +210,52 @@ def define_config(name, default_value):
 alf.config("layers.Conv2D", use_bn=use_bn, bn_ctor=norm_type.BatchNorm2d)
 alf.config("layers.FC", bn_ctor=norm_type.BatchNorm1d)
 
+aug_shift = define_config('aug_shift', 4)
+
+
+def rand_intensity_(x: torch.Tensor) -> torch.Tensor:
+    def _f(x: torch.Tensor) -> torch.Tensor:
+        ratio = 1 + 0.05 * torch.randn(
+            x.shape[0], 1, 1, 1, device=x.device).clamp_(-2, 2)
```
Haichao-Zhang, line 219: clamp_(-2, 2) -> is this an intended feature to have negative pixel values after randomization?

alf/examples/muzero_atari_conf.py
Haichao-Zhang: > Question: `target_model` is not optimized, why do we need to run `NeroPlus.initialize()` on it?

Based on my understanding, `NeroPlus.initialize()` can be regarded as part of model's parameter initialization process.
By doing it when constructing the model, the target model will copy the correct initial parameter values when setting up the target_updater.

breakds: I did not make it clear, sorry! The `model_ctor` is used to construct both the `model` and the `target_model`. It is necessary to run `NeroPlus.initialize()` when constructing the actual `model`. I think running this for `target_model` is not necessary but "it does not hurt". Not sure whether my understanding is correct.

Haichao-Zhang: I think your understanding is correct: ``NeroPlus.initialize() ``  is not necessary for the target model but "it does not hurt".  In this sense, all the param initializations are actually unnecessary for the target model, since they will be wiped away by copying the values from model later.

alf/examples/muzero_atari_conf.py
```diff
@@ -427,30 +587,42 @@ def _scale_grad(scale):
...
     target_update_tau=1.0)
 
 alf.config(
     "MuzeroAlgorithm",
     discount=discount,
     enable_amp=True,
-    representation_learner_ctor=MuzeroRepresentationImpl,
-    mcts_algorithm_ctor=MCTSAlgorithm,
+    mcts_algorithm_ctor=partial(
+        MCTSAlgorithm, num_parallel_sims=8, num_simulations=200),
```
breakds, line 628: Would like to confirm it is intended to use a different set of parameter for rollout (as opposite to reanalyze), so that we search more thoroughly at rollout?

alf/examples/muzero_atari_conf.py
emailweixu: With the clamp, the ratio will be in [0.9, 1.1]

alf/examples/muzero_atari_conf.py
emailweixu: Yes

alf/examples/muzero_atari_conf.py
Haichao-Zhang: You are right. I mis-understood it earlier.

Haichao-Zhang (approved):

breakds (approved):

alf/layers.py
emailweixu: updated.

PR #1307, opened 2022-05-03
Lower bound value target learning

le-horizon:
I included in this PR some changes that help debugging.  Let me know if you don't think they should be merged into pytorch.


alf/algorithms/rl_algorithm.py
```diff
@@ -223,19 +223,21 @@ def __init__(self,
             replay_buffer_length = adjust_replay_buffer_length(
                 config, self._num_earliest_frames_ignored)
 
-            if config.whole_replay_buffer_training and config.clear_replay_buffer:
+            total_replay_size = replay_buffer_length * self._env.batch_size
+            if config.whole_replay_buffer_training:
```
le-horizon, line 227: @breakds do you think it's Ok to remove ```and config.clear_replay_buffer``` here?

emailweixu:
This PR changes 47 files. Many of them are unrelated. I suggest split them into different PRs and describe the reasons of those changes.

PR #695, opened 2020-09-29
draft_retrace

zhuboli:
Change code in file value_ops and td_loss. Default value for train_info is None. If we give the train_info parameter and lambda is not equal to 1 and 0, we will use retrace method. So we do not need to change the code of sac_algorithm or sarsa_algorithm when other people do not want retrace method.

PR #1311, opened 2022-05-04
Lbtq main

le-horizon:
main change for lower bound target q experiments.

still quite large..  Let me know if you'd prefer even smaller pieces.


alf/algorithms/data_transformer.py
```diff
@@ -819,32 +816,96 @@ def transform_experience(self, experience: Experience):
...
+
+                if alf.summary.should_record_summaries():
+                    alf.summary.scalar(
+                        "replayer/" + buffer._name +
+                        ".discount_mean_after_relabel",
+                        torch.mean(discount[:, 1:]))
+
+                result = result._replace(discount=discount)
+                result = result._replace(step_type=step_type)
+                relabeled_rewards = suite_socialbot.transform_reward_tensor(
```
hnyu: My understanding is that `HindsightExperienceTransformer` is a general data transformer independent of the environment. So it feels strange to include `suite_socialbot` here.

alf/algorithms/data_transformer.py
```diff
@@ -922,4 +1022,7 @@ def create_data_transformer(data_transformer_ctor, observation_spec):
     if len(data_transformer_ctor) == 1:
         return data_transformer_ctor[0](observation_spec)
 
+    if HindsightExperienceTransformer in data_transformer_ctor:
+        assert HindsightExperienceTransformer == data_transformer_ctor[0], \
+            "Hindsight relabeling should happen before all other transforms."
```
hnyu: This is no longer true after Break's recent PR on `SequentialDataTransformer`. Basically `UntransformedTimeStep` could appear before any other data transformer. I suggest either justifying why `HindsightExperienceTransformer` has to appear before `UntransformedTimeStep` or removing this assertion here.

alf/algorithms/data_transformer.py
```diff
@@ -874,21 +935,60 @@ def transform_experience(self, experience: Experience):
...
+                           shape[:2])[non_her_or_fst][not_close],
+                       start_pos.unsqueeze(1).expand(
+                           shape[:2])[non_her_or_fst][not_close])
+            logging.warning(msg)
+            # assert False, msg
+            # relabeled_rewards[non_her_or_fst] = goal_rewards[non_her_or_fst]
+
+        final_relabeled_rewards = relabeled_rewards
+        if result.reward.ndim > 2:
+            # multi dimensional env reward, first dim is goal related reward.
```
hnyu: Is this reward assumption too strict and easily violated by a user?

alf/algorithms/data_transformer.py
```diff
@@ -759,6 +750,9 @@ def __init__(self,
                 exp nest.
             desired_goal_field (str): path to the desired_goal field in the
                 exp nest.
+            sparse_reward (bool): Whether to transform reward from -1/0 to 0/1.
```
hnyu: This transform from -1/0 to 0/1 seems specific to certain environments and is not a general thing.

alf/algorithms/data_transformer.py
```diff
@@ -180,6 +182,7 @@ def __init__(self,
                  observation_spec,
                  stack_size=4,
                  stack_axis=0,
+                 convert_only_minibatch_to_device=False,
```
hnyu, line 190: What does this argument mean? docstring should explain its actual meaning.

alf/algorithms/ddpg_algorithm.py
```diff
@@ -237,10 +237,12 @@ def _sample(a, ou):
                                           noisy_action, self._action_spec)
         state = empty_state._replace(
             actor=DdpgActorState(actor=state, critics=()))
+        # action_distribution is not supported for continuous actions for now.
+        # Returns empty action_distribution to fail early.
         return AlgStep(
             output=noisy_action,
             state=state,
-            info=DdpgInfo(action=noisy_action, action_distribution=action))
+            info=DdpgInfo(action=noisy_action, action_distribution=()))
```
hnyu, line 256: Why do you need this change? By default we could think of a deterministic action distribution as an action tensor.

alf/algorithms/ddpg_algorithm.py
```diff
@@ -355,6 +358,11 @@ def calc_loss(self, info: DdpgInfo):
 
         actor_loss = info.actor_loss
 
+        if self._critic_losses[0]._improve_w_nstep_bootstrap:
```
hnyu: This masking seems quite hacky. CriticLoss should not be responsible for actor losses or other losses in general. So this flag really belongs to either an algorithm or some other places.

alf/algorithms/sac_algorithm.py
```diff
@@ -56,7 +57,8 @@
 SacInfo = namedtuple(
     "SacInfo", [
         "reward", "step_type", "discount", "action", "action_distribution",
-        "actor", "critic", "alpha", "log_pi", "discounted_return"
+        "actor", "critic", "alpha", "log_pi", "discounted_return",
+        "future_distance", "her"
```
hnyu: "future_distance" and "her" are not essential to SAC and thus they shouldn't appear here.

alf/algorithms/sac_algorithm.py
```diff
@@ -531,26 +552,52 @@ def _predict_action(self,
...
 
     def predict_step(self, inputs: TimeStep, state: SacState):
-        action_dist, action, _, action_state = self._predict_action(
+        action_dist, action, q_values, action_state = self._predict_action(
             inputs.observation,
             state=state.action,
             epsilon_greedy=self._epsilon_greedy,
             eps_greedy_sampling=True)
+        info = SacInfo(action_distribution=action_dist)
+        if (alf.summary.render.is_rendering_enabled()
```
hnyu: This block of rendering code should be kept as a local change as it's not essential to SAC.

alf/algorithms/sac_algorithm.py
```diff
@@ -274,7 +290,10 @@ def __init__(self,
             )
 
         def _init_log_alpha():
-            return nn.Parameter(torch.tensor(float(initial_log_alpha)))
+            alpha = torch.tensor(float(initial_log_alpha))
+            if alpha_optimizer is None:
+                return alpha
+            return nn.Parameter(alpha)
```
hnyu: If `alpha_optimizer` is None, we can still create it as a parameter?

alf/algorithms/sac_algorithm.py
```diff
@@ -203,6 +208,14 @@ def __init__(self,
...
                 from ``config.epsilon_greedy`` and then
                 ``alf.get_config_value(TrainerConfig.epsilon_greedy)``.
+            rollout_epsilon_greedy (float): epsilon greedy policy for rollout.
+                Together with the following three parameters, the Sac algorithm
+                can be converted to a DQN algorithm.
+            use_epsilon_schedule (float): training schedule for
+                rollout_epsilon_greedy.
+            max_target_action (bool): whether to use the action with the highest
+                target value as the target action for computing bootstrapped value.
+                To mimic the DQN algorithm, set this to True.
```
hnyu: Feels a little strange to include these arguments here because there is usually no good reason to convert SAC to DQN. Maybe create a subclass of SAC for this purpose?

alf/algorithms/td_loss.py
```diff
@@ -31,9 +31,20 @@ class TDLoss(nn.Module):
...
+                 lb_target_q: float = 0.,
+                 default_return: float = -1000.,
+                 improve_w_goal_return: bool = False,
+                 improve_w_nstep_bootstrap: bool = False,
+                 improve_w_nstep_only: bool = False,
+                 lower_bound_constraint: float = 0.,
+                 lb_loss_scale: bool = False,
+                 reward_multiplier: float = 1.,
+                 positive_reward: bool = True,
+                 use_retrace: bool = False,
```
hnyu, line 38: To me, it's not clear which of these arguments are really essential to `TDLoss` in a general sense. Some of them seem only applicable to specific scenarios.

alf/bin/train_play_test.py
```diff
@@ -141,7 +141,10 @@ def _to_conf_params(parameters):
     'TrainerConfig.num_updates_per_train_iter=1',
     'TrainerConfig.mini_batch_length=2',
     'TrainerConfig.mini_batch_size=4',
-    'TrainerConfig.replay_buffer_length=64',
+    # If replay_buffer_length is above 6 (3 iters * 2 unroll length), whole replay buffer training
+    # algorithms (e.g. mbrl_pendulum) will not get enough training data, and will not start training.
+    # This fails during playing, when the config file isn't generated in root_dir.
```
hnyu: Why a longer replay buffer won't give enough training data?

alf/environments/suite_robotics.py
```diff
@@ -43,19 +43,42 @@ def is_available():
...
 class SparseReward(gym.Wrapper):
     """Convert the original :math:`-1/0` rewards to :math:`0/1`.
     """
 
-    def __init__(self, env):
+    def __init__(self,
+                 env,
+                 reward_cap=1.,
+                 positive_reward=True,
+                 append_reward_dim=False):
```
hnyu: Docstrings are needed for the new arguments. The name of `reward_cap` doesn't correctly reflect its usage which is actually a scaling factor. Why do you need `append_reward_dim` (it should be put in another wrapper if needed)?

alf/environments/suite_robotics.py
```diff
@@ -141,14 +209,28 @@ def load(environment_name,
     Returns:
         An AlfEnvironment instance.
     """
-    assert (environment_name.startswith("Fetch")
-            or environment_name.startswith("HandManipulate")), (
-                "This suite only supports OpenAI's Fetch and ShadowHand envs!")
+    assert (
+        environment_name.startswith("Fetch")
+        or environment_name.startswith("HandManipulate")
+        or environment_name.startswith("Ant")
```
hnyu, line 201: Not sure if it's appropriate to include multiworld Ant in this suite? Note that "Fetch" and "HandManipulate" were introduced at the same time by OpenAI with very similar implementations.

alf/utils/math_ops.py
```diff
@@ -457,3 +458,36 @@ def transform(self, x):
 
     def inverse_transform(self, y):
         return y.sign() * ((y / self._alpha).abs().exp() - 1)
+
+
+@alf.configurable
```
hnyu, line 463: These hindsight related functions shouldn't be put in a general file like `math_ops.py`.

hnyu (changes requested):
I only roughly skimmed over this PR. In general, I think it's not clear if it's necessary for most changes to be included in ALF. From my understanding, the best scenario would be to separate LBTQ algorithm from other misc stuff in a clean way with a general API, and only integrate it in ALF. All other stuff can be put in your personal repo which would contain  all experiments details and helper functions, etc. So far I think the generality of some ALF apis has been compromised by the PR.

alf/environments/suite_robotics.py
hnyu: > Not sure if it's appropriate to include multiworld Ant in this suite? Note that "Fetch" and "HandManipulate" were introduced at the same time by OpenAI with very similar implementations.

alf/algorithms/data_transformer.py
le-horizon: Environment rewards can be 0/1 or -1/0 or any form.  But because the Hindsight relabeler needs to know whether goal is achieved to be able to relabel discount and step_type correctly, here, we assume the reward_fn always returns -1/0 reward, and it needs to be transformed into 0/1 form when needed.

Since it relabels discount to 0, this also relabels infinite horizon tasks to episodic.  Updated comment.

alf/algorithms/data_transformer.py
le-horizon: This transform_reward_tensor function is used to translate -1/0 reward into 0/1 reward.  We also depend on the -1/0 reward assumption to determine the relabeled LAST (episode end) steps (also relabel discount 0).

Removed the dependency here, added it in the task gin or alf configs.  This runs the risk of people forgetting to set it in a new task config.

For episodic tasks (sparse_reward==True), the Hindsight transformer does depend on the original reward being -1/0.  In this case, maybe I should move ``transform_reward_tensor'' out of suite_socialbot, into a more general place like reward_utils.py?  Would you prefer this more than the current solution of relying on the task's alf config to setup the transform_reward_tensor function?

alf/algorithms/data_transformer.py
le-horizon: True.  Made a note about the assumption upfront in the class description.
The proper way is maybe to divide the current multi-dim reward into a nested reward field.  Added TODO.

alf/algorithms/data_transformer.py
le-horizon: Good point.  Removed.

alf/algorithms/ddpg_algorithm.py
le-horizon: This is to fail early and clearly.  action is a tensor, not distribution.  Putting action directly there could cause confusion when debugging.  See comment 3 lines above.

alf/algorithms/ddpg_algorithm.py
le-horizon: Removed.

alf/algorithms/sac_algorithm.py
le-horizon: Could an overall optimizer accidentally include alpha parameter and optimize it?

alf/algorithms/sac_algorithm.py
le-horizon: How about I extract this into a separate file for debug rendering?

alf/algorithms/sac_algorithm.py
le-horizon: Yes, does feel hacky.  The code is quite intertwined, so a subclass probably wouldn't do to separate the two classes neatly.  Do you prefer to just remove this part for now?

alf/algorithms/sac_algorithm.py
le-horizon: Where do you think is a good place?  I could group discounted_return, her, future_distance into a batch_info field.  Or should it be somewhere else?

alf/algorithms/td_loss.py
le-horizon: Good point.  Created a subclass to do lower bounding.  Kept a few general functions like clip and retrace in td_loss.

alf/bin/train_play_test.py
le-horizon: because the replay buffer is cleared before it's fully filled.  So training never starts.

alf/environments/suite_robotics.py
le-horizon: Added docstrings.  Renamed reward_cap to reward_weight.  Removed append_reward_dim.

alf/environments/suite_robotics.py
le-horizon: It's probably Ok?  The code here is quite simple.  The if conditions mark clearly where Ant is the env.

alf/utils/math_ops.py
le-horizon: Both data_transformer and suite_robotics use these functions, so util is perhaps a reasonable place?  Should I create a utils/reward_functions.py?

le-horizon (commented):
Thanks for the insightful comments @hnyu .
Tried to address all.  Replies or questions inlined.  PTAL.

Thanks,
Le

alf/algorithms/ddpg_algorithm.py
emailweixu: It is intended to return `action_distribution` here so that some other algorithm can use it (e.g. TracAlgorithm). Do you find it causing problem?

alf/algorithms/sac_algorithm.py
```diff
@@ -152,6 +154,9 @@ def __init__(self,
                  q_network_cls=QNetwork,
                  reward_weights=None,
                  epsilon_greedy=None,
+                 rollout_epsilon_greedy=1.0,
+                 use_epsilon_schedule=0,
```
emailweixu: It's better to change `epsilon_greedy` to accept both a float and a Scheduler. For float, you can use schedulers.as_scheduler to convert it to a scheduler. Similar to `TrainerConfig.priority_replay_alpha`

alf/algorithms/sac_algorithm.py
```diff
@@ -274,7 +290,10 @@ def __init__(self,
             )
 
         def _init_log_alpha():
-            return nn.Parameter(torch.tensor(float(initial_log_alpha)))
+            alpha = torch.tensor(float(initial_log_alpha))
+            if alpha_optimizer is None:
```
emailweixu: I think if you want to fix alpha, you can simply set the learning rate for it to be zero. `???_optimizer==None` ususally means using a default optimizer.

alf/algorithms/sac_algorithm.py
emailweixu: Agree with @hynu. The current situation is that different person want to render different thing.  So it's hard to have an agreed-upon set of figures for rendering. This is different from tensorboard summary where space is not that precious.

alf/algorithms/ddpg_algorithm.py
```diff
@@ -355,6 +358,19 @@ def calc_loss(self, info: DdpgInfo):
...
+        # and pulling additionally a few timesteps from the future to compute
+        # bootstrap values, we here piggyback on n-step TDLoss, but masking
+        # out losses from the 2nd to n-1-th steps.
+        # If this hacky use pattern is to be used frequently in the future,
+        # we should consider refactoring it.
+        if hasattr(self._critic_losses[0],
+                    "_improve_w_nstep_bootstrap") and \
+            self._critic_losses[0]._improve_w_nstep_bootstrap:
+            # Ignore 2nd - nth step actor losses.
+            actor_loss.loss[1:] = 0
```
emailweixu: why do you need to igore the actor loss for those steps? It seems to me it doesn't hurt to keep them.

alf/algorithms/td_loss.py
```diff
@@ -80,10 +82,14 @@ def __init__(self,
...
                 loss. This function takes as input the target and the estimated
                 Q values and returns the loss for each element of the batch.
+            clip: When positive, loss clipping to the range [-clip, clip].
             td_lambda: Lambda parameter for TD-lambda computation.
             normalize_target (bool): whether to normalize target.
                 Note that the effect of this is to change the loss. The critic
                 value itself is not normalized.
+            use_retrace: turn on retrace loss
+                :math:`\mathcal{R} Q(x, a):=Q(x, a)+\mathbb{E}_{\mu}\left[\sum_{t \geq 0} \gamma^{t}\left(\prod_{s=1}^{t} c_{s}\right)\left(r_{t}+\gamma \mathbb{E}_{\pi} Q\left(x_{t+1}, \cdot\right)-Q\left(x_{t}, a_{t}\right)\right)\right]`
+                copied from PR #695.
```
emailweixu: "copied from PR #695" can be removed.

alf/utils/value_ops.py
```diff
@@ -118,6 +118,67 @@ def action_importance_ratio(action_distribution,
     return importance_ratio, importance_ratio_clipped
 
 
+def generalized_advantage_estimation(rewards,
```
emailweixu, line 121: This should be put side-by-side with the original generalized_advantage_estimation for comparison.

alf/utils/value_ops.py
```diff
@@ -213,12 +337,15 @@ def one_step_discounted_return(rewards, values, step_types, discounts):
...
-                                     time_major=True):
+def generalized_advantage_estimation_retrace(rewards,
+                                             values,
+                                             step_types,
+                                             discounts,
+                                             target_value,
+                                             importance_ratio,
+                                             use_retrace=False,
+                                             td_lambda=1.0,
+                                             time_major=True):
```
emailweixu: also should cite retrace paper.

alf/environments/suite_robotics.py
```diff
@@ -43,19 +43,33 @@ def is_available():
...
+@alf.configurable
 class SparseReward(gym.Wrapper):
     """Convert the original :math:`-1/0` rewards to :math:`0/1`.
     """
 
-    def __init__(self, env):
+    def __init__(self, env, reward_weight=1., positive_reward=True):
+        """
+        Args:
+            reward_weight (float): weight of output reward.
```
emailweixu: use type hints at function signature instead
```python
def __init__(self, env, reward_weight: float = 1., positive_reward: float = True)
```

emailweixu (changes requested):
This PR still contains a lot of separable parts. It's better to separate them and describe the motivation for each of them.

alf/algorithms/data_transformer.py
```diff
@@ -180,13 +187,16 @@ def __init__(self,
...
                  fields=None):
         """Create a FrameStacker object.
 
         Args:
             observation_spec (nested TensorSpec): describing the observation in timestep
             stack_size (int): stack so many frames
             stack_axis (int): the dimension to stack the observation.
+            convert_only_minibatch_to_device (bool): whether to convert only the
+                minibatch or the whole batch of data to the default device.
```
hnyu, line 199: Variable name and docstring inconsistent with the code? FrameStacker will always affect the entire batch?

alf/algorithms/data_transformer.py
hnyu: > Environment rewards can be 0/1 or -1/0 or any form. But because the Hindsight relabeler needs to know whether goal is achieved to be able to relabel discount and step_type correctly, here, we assume the reward_fn always returns -1/0 reward, and it needs to be transformed into 0/1 form when needed.
> 
> Since it relabels discount to 0, this also relabels infinite horizon tasks to episodic. Updated comment.

From the docstring, the observation is assumed to have a field "achieved_goal" indicating what goal has been achieved. If this is true, then there is no need to decide goal achieving from the reward, which is not a general setup. (on the other hand, -1/0 is as sparse as 0/1 so the name 'sparse_reward' could be improved).

alf/algorithms/data_transformer.py
```diff
@@ -751,25 +763,39 @@ def __init__(self,
                  her_proportion=0.8,
                  achieved_goal_field="time_step.observation.achieved_goal",
                  desired_goal_field="time_step.observation.desired_goal",
-                 reward_fn=l2_dist_close_reward_fn):
+                 sparse_reward=False,
+                 add_noise_to_goals=False,
+                 threshold=.05,
```
hnyu: Can merge `add_noise_to_goals` and `threshold` into just one `relabeled_goal_noise`, and when it's 0 there is no noise added.

alf/algorithms/data_transformer.py
```diff
@@ -819,32 +845,97 @@ def transform_experience(self, experience: Experience):
...
+            last_step_pos = start_pos + batch_length - 1
+            last_env_ids = env_ids
+            # Get x, y indices of LAST steps for the whole batch, not just the HER part.
             dist = buffer.steps_to_episode_end(last_step_pos, last_env_ids)
             if alf.summary.should_record_summaries():
                 alf.summary.scalar(
                     "replayer/" + buffer._name + ".mean_steps_to_episode_end",
                     torch.mean(dist.type(torch.float32)))
 
+            def _add_noise(t):
```
hnyu: This function `transform_experience()` has already over 200 lines, and it's really difficult to read. I suggest splitting it into several member functions to make it more modular.

alf/algorithms/ddpg_algorithm.py
```diff
@@ -42,7 +42,7 @@
 DdpgInfo = namedtuple(
     "DdpgInfo", [
         "reward", "step_type", "discount", "action", "action_distribution",
-        "actor_loss", "critic", "discounted_return"
+        "actor_loss", "critic", "discounted_return", "future_distance", "her"
```
hnyu: Better not to put DDPG-irrelevant concepts into this Info structure. HER and DDPG are two orthogonal concepts and should be disentangled in the code. Otherwise when we have a third algorithm in the future, this Info structure will be added more fields again.

alf/algorithms/one_step_loss.py
```diff
@@ -16,12 +16,12 @@
...
 
 import alf
-from alf.algorithms.td_loss import TDLoss, TDQRLoss
+from alf.algorithms.td_loss import LowerBoundedTDLoss, TDQRLoss
 from alf.utils import losses
 
 
 @alf.configurable
-class OneStepTDLoss(TDLoss):
+class OneStepTDLoss(LowerBoundedTDLoss):
```
hnyu, line 24: Conceptually OneStepTDLoss is not a child (special case) of LowerBoundedTDLoss. For me, LowerBoundedTDLoss is far more special and should be a completely new class, or a child of OneStepTDLoss.

alf/algorithms/sac_algorithm.py
hnyu: > Where do you think is a good place? I could group discounted_return, her, future_distance into a batch_info field. Or should it be somewhere else?

One ideal treatment is to have a wrapper algorithm maybe called SACHERAlgorithm that wraps an SAC and uses HER. Its info will contain fields `['sac', 'her']`. Actually it might be possible to write a general algorithm wrapper/container that can wrap any off-policy algorithm with HER.

alf/algorithms/sac_algorithm.py
hnyu: > Yes, does feel hacky. The code is quite intertwined, so a subclass probably wouldn't do to separate the two classes neatly. Do you prefer to just remove this part for now?

I think it's definitely possible and clearer to have a subclass? You only need to include these additional arguments and their code in the subclass implementation.

alf/algorithms/sac_algorithm.py
hnyu: > Could an overall optimizer accidentally include alpha parameter and optimize it?

I think the original intent of having `alpha_optimizer=None` is indeed to let the parent algorithm decides which optimizer to use for alpha, instead of not optimizing alpha at all.

alf/algorithms/td_loss.py
```diff
@@ -219,17 +259,272 @@ def _summarize(v, r, td, suffix):
...
         loss = self._td_error_loss_fn(returns.detach(), value)
+        if self._clip > 0:
+            loss = torch.clamp(loss, min=-self._clip, max=self._clip)
 
         if loss.ndim == 3:
             # Multidimensional reward. Average over the critic loss for all dimensions
             loss = loss.mean(dim=2)
 
+        # For the subclass LowerBoundedTDLoss
+        if constraint_loss is not None:
```
hnyu, line 276: The content of a subclass should be implemented in the subclass itself? Unless this code block is shared by multiple subclasses.

alf/bin/train_play_test.py
hnyu: > because the replay buffer is cleared before it's fully filled. So training never starts.

Whole replay buffer training doesn't require the buffer to be full? It just gets all of data in the buffer even though then it's not full.

Just curious: this number of 64 had no issue before? Did you meet this problem for new test cases?

alf/environments/suite_robotics.py
```diff
@@ -105,6 +154,16 @@ def observation(self, observation):
...
 
 
+@alf.configurable
+class RemoveInfoWrapper(gym.Wrapper):
+    """Remove all the info from environment return.
+    """
+
+    def step(self, action):
+        obs, reward, done, info = self.env.step(action)
+        return obs, reward, done, {}
```
hnyu: This wrapper can be put in gym_wrappers.py as a general wrapper.

alf/environments/suite_robotics.py
hnyu: > It's probably Ok? The code here is quite simple. The if conditions mark clearly where Ant is the env.

I mean it's hard for someone else to imagine that metaworld Ant is included in this suite_robotics.py file. The reason  of having "robotics" in the name is because OpenAI Gym originally had "robotics" category which contains Fetch and HandManipulate.

alf/environments/suite_socialbot.py
```diff
@@ -38,11 +40,61 @@ def is_available():
...
+                                  torch.zeros(()))
+    goal_reward *= reward_cap
+    if reward.ndim > 2:
+        reward[:, :, 0] = goal_reward
+    else:
+        reward = goal_reward
+    return reward
+
+
+class SparseReward(gym.Wrapper):
```
hnyu, line 74: You can reuse the one in `suite_robotics.py`?

alf/environments/suite_socialbot.py
```diff
@@ -38,11 +40,61 @@ def is_available():
...
+    goal_reward = goal_reward * reward_cap
+    if isinstance(reward, (np.ndarray, list)):
+        reward[0] = goal_reward
+    else:
+        reward = goal_reward
+    return reward
+
+
+@alf.configurable
+def transform_reward_tensor(reward, reward_cap=1., positive_reward=True):
```
hnyu, line 59: Where is this function used? If it's only in your conf file, then it should be put there instead.

alf/environments/suite_socialbot.py
```diff
@@ -38,11 +40,61 @@ def is_available():
     return social_bot is not None
 
 
+@alf.configurable
+def transform_reward(reward, reward_cap=1., positive_reward=True):
```
hnyu, line 44: Should be a class member of `SparseReward` below, as it is not used by other files / code.

alf/utils/math_ops.py
hnyu: > Both data_transformer and suite_robotics use these functions, so util is perhaps a reasonable place? Should I create a utils/reward_functions.py?

These calculations are very simple (basically one line) and don't have to be factored into functions.

hnyu (changes requested):

alf/algorithms/data_transformer.py
le-horizon: Added comment

alf/algorithms/ddpg_algorithm.py
le-horizon: Yes, when used with e.g. Retrace, a distribution is needed, but action is not a distribution.

alf/algorithms/ddpg_algorithm.py
le-horizon: Good point.  Since we compare with n-step return methods anyways, it may be more fair to keep the other steps' action and alpha losses.

alf/algorithms/sac_algorithm.py
le-horizon: Good point.  For now, added comment that these three fields are optional, added TODO for future work to add the HER algorithm wrapper.

alf/algorithms/sac_algorithm.py
le-horizon: Given it a try, how about this new dqn_algorithm.py?

alf/algorithms/sac_algorithm.py
le-horizon: Nice!

alf/algorithms/sac_algorithm.py
le-horizon: Fair enough, removed.

alf/algorithms/sac_algorithm.py
le-horizon: Oh, I misunderstood the param.  lr=0 sounds good.  Removed.

alf/algorithms/td_loss.py
le-horizon: Done.

alf/algorithms/td_loss.py
le-horizon: Moving this code to the child class requires returning the constraint_loss together with LossInfo, and pollutes the function signature.  Do you prefer that or just keep the code as is?

alf/environments/suite_robotics.py
le-horizon: Done

alf/utils/value_ops.py
le-horizon: Done

alf/utils/value_ops.py
le-horizon: Done

alf/bin/train_play_test.py
le-horizon: Ah, you are right.  Turns out it was due to my earlier change in rl_algorithm which required buffer to be full even when clear_replay_buffer is False.  That change was already reverted.  This change is also reverted.

alf/environments/suite_robotics.py
le-horizon: I see.  How about moving this code to a separate suite_mujoco.py, since all these Ant envs are under: https://github.com/openai/gym/tree/master/gym/envs/mujoco

alf/environments/suite_socialbot.py
le-horizon: Cannot do that.  ``positive_reward`` needs to be consistent with transform_reward_tensor, typically done in config file or command line config.  It needs to be alf.configurable.

alf/environments/suite_socialbot.py
le-horizon: Both suite_robotics and suite_socialbot and their conf files could use this.

alf/environments/suite_socialbot.py
le-horizon: This also assumes first reward dimension being goal reward with multi dim reward.
So perhaps a bit of code duplication is actually better to separate out the env specific reward logic.

alf/utils/math_ops.py
le-horizon: The hindsight data_transformer requires an input function, and the reward function in these two places need to be exactly the same, otherwise, bugs can happen.  Plus it is already not that simple due to multi dim reward. 
 That's why I keep them as alf configurable functions.

alf/algorithms/data_transformer.py
le-horizon: Good point.  Done.

alf/algorithms/data_transformer.py
le-horizon: Good point about goal achieved.  It's still hard to find a general way though.  The reward function is general, and may not return anything about goal achievement, e.g. when it is a dense reward based on distance to goal.  

Calling the variable ``relabel_with_episodic_rewards``, since -1/0 is for continuous and 0/1 is for episodic.

alf/algorithms/data_transformer.py
le-horizon: Good point.  Cleaned up a bit.

alf/algorithms/sac_algorithm.py
le-horizon: Got it.  Removed.

alf/environments/suite_robotics.py
le-horizon: Done.

le-horizon (commented):
Addressed all comments.  Please hold off from reviewing, and let me further split this PR.

Thanks,
Le

alf/algorithms/one_step_loss.py
le-horizon: LowerBoundedTDLoss defaults to TDLoss and can be configured to enable lower bounding, so it's more general than TDLoss.  Maybe I should name it something else?

le-horizon:
Split into https://github.com/HorizonRobotics/alf/pull/1330 and a few others

PR #1314, opened 2022-05-04
Lbtq 4 cudaavail

le-horizon:
Disabling cuda available check with CUDA_VISIBLE_DEVICES= (empty)


hnyu:
This PR contains code that is also present in your other PRs? It's better to separate unrelated code into different PRs.

le-horizon:
Good catch @hnyu .  All 6 PRs are not properly cleaned.  They have a sequential dependency though.  Let's see how that works out..

PR #1317, opened 2022-05-05
CUDA_VISIBLE_DEVICES= (empty) will avoid cuda.is_available check

le-horizon:
disable cuda.is_available check with CUDA_VISIBLE_DEVICES= (empty)
This is useful during hardware issues when the check hangs forever.

### DIFF
--- a/alf/bin/train_play_test.py
+++ b/alf/bin/train_play_test.py
@@ -397,349 +397,349 @@ class TrainPlayTest(alf.test.TestCase):
             eval_dir)
         test_func(episode_returns, episode_lengths)
 
-    def test_ac_breakout(self):
-        self._test(
-            conf_file='ac_breakout_conf.py',
-            skip_checker=self._skip_if_atari_unavailable,
-            extra_train_params=ON_POLICY_TRAIN_PARAMS)
+    # def test_ac_breakout(self):
+    #     self._test(
+    #         conf_file='ac_breakout_conf.py',
+    #         skip_checker=self._skip_if_atari_unavailable,
+    #         extra_train_params=ON_POLICY_TRAIN_PARAMS)
 
-    def test_ac_cart_pole(self):
-        def _test_func(returns, lengths):
-            self.assertGreater(returns[-1], 195)
-            self.assertGreater(lengths[-1], 195)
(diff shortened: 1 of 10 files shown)


alf/utils/common.py
```diff
@@ -48,6 +48,13 @@
 from . import dist_utils, gin_utils
 
 
+def cuda_is_available():
+    # When CUDA_VISIBLE_DEVICES= (empty), simply skip the check.
+    # This is useful during hardware issues when the check hangs forever.
+    cuda_dev = os.getenv("CUDA_VISIBLE_DEVICES", default=None)
+    return (cuda_dev is None or cuda_dev != "") and torch.cuda.is_available()
```
emailweixu, line 55: If CUDA_VISIBLE_DEVICES=="", torch.cuda.is_availble() will return False.

le-horizon: When hardware works properly then yes, but if there is a hardware or driver problem, then this call could hang.  That's the reason for this change.

alf/bin/train_play_test.py
```diff
@@ -397,349 +397,349 @@ def _test_performance(self, root_dir, test_func):
...
+    #     self._test(
+    #         conf_file='oac_halfcheetah_conf.py',
+    #         skip_checker=self._skip_if_mujoco_unavailable,
+    #         extra_train_params=OFF_POLICY_TRAIN_PARAMS)
+
+    # def test_oac_humanoid(self):
+    #     self._test(
+    #         conf_file='oac_humanoid_conf.py',
+    #         skip_checker=self._skip_if_mujoco_unavailable,
+    #         extra_train_params=OFF_POLICY_TRAIN_PARAMS)
```
emailweixu: need to uncomment

emailweixu:
> disable cuda.is_available check with CUDA_VISIBLE_DEVICES= (empty)

The PR title is not correct.

alf/bin/train_play_test.py
le-horizon: Done.

emailweixu:
If this is only useful when the driver or hardware is broken, I'd suggest not to merge it. If people has a GPU machine but the driver or hardware is broken, they should fix those instead.

PR #1318, opened 2022-05-05
Allow flag to turn off torch.use_deterministic_algorithms

le-horizon:
[Disabling use_deterministic_algorithms with flag. Torch reports false alarm during sac_breakout_conf.py running about scatter_add being non-deterministic. In fact, it's still fully deterministic even with use_deterministic_algorithms(False).

### DIFF
--- a/alf/algorithms/config.py
+++ b/alf/algorithms/config.py
@@ -62,7 +62,6 @@ class TrainerConfig(object):
                  mini_batch_length=None,
                  mini_batch_size=None,
                  whole_replay_buffer_training=True,
-                 convert_only_minibatch_to_device=False,
                  replay_buffer_length=1024,
                  priority_replay=False,
                  priority_replay_alpha=0.7,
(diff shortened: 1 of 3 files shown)


alf/algorithms/config.py
```diff
@@ -301,6 +304,7 @@ def __init__(self,
         self.mini_batch_length = mini_batch_length
         self.mini_batch_size = mini_batch_size
         self.whole_replay_buffer_training = whole_replay_buffer_training
+        self.convert_only_minibatch_to_device = convert_only_minibatch_to_device
```
emailweixu: not used anywhere

le-horizon: Removed

le-horizon (commented):
Done.  PTAL.

Thanks,
Le

le-horizon:
closing, changed to https://github.com/HorizonRobotics/alf/pull/1334

PR #1322, opened 2022-05-11
NoisyFC layer

emailweixu:
The Noisy Linear Layer discribed in

Fortunato et. al. `Noisy Networks for Exploration <https://arxiv.org/abs/1706.10295>`_

The goal is to have extended exploration for some tasks.

Some details:

1. The noise for each sample in a batch is different.
2. The noise is maintained as state. It has a probability of `new_noise_prob`
   to change to new noise.
3. Since the initial state is always 0, a new noise will always be generated
   for zero state.

### DIFF
--- a/alf/networks/networks.py
+++ b/alf/networks/networks.py
@@ -320,9 +320,14 @@ class AMPWrapper(Network):
 @alf.configurable
 @alf.repr_wrapper
 class NoisyFC(Network):
-    """The Noisy Linear Layer discribed in
+    r"""The Noisy Linear Layer discribed in
 
     Fortunato et. al. `Noisy Networks for Exploration <https://arxiv.org/abs/1706.10295>`_
+
+    In short, the original weight :math:`w` and bias :math:`b` of FC layer are replaced
+    with :math:`w + w_\sigma \epislon^w` and :math:`b + b_\sigma \epsion^b` where
+    :math:`\epsilon^w` and :math:`\epsilon^b` are noise and :math:`w, w_\sigma, b, b_\sigma`
+    are trainable parameters.
 
     Some details:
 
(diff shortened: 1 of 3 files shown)


alf/networks/networks.py
```diff
@@ -308,3 +315,184 @@ def forward(self, input, state):
...
+            # The initial state is always 0. So we need to generate new noise
+            # for initial state.
+            new_noise = new_noise | ((epsilon_in == 0).all(dim=1) &
+                                     (epsilon_out == 0).all(dim=1))
+            new_noise = new_noise.unsqueeze(-1)
+            epsilon_in = torch.where(new_noise, new_epsilon_in, epsilon_in)
+            epsilon_out = torch.where(new_noise, new_epsilon_out, epsilon_out)
+            noise_in = input * epsilon_in
+            # x = bias_sigma + noise_in @ weight_sigma.t()
+            x = torch.addmm(self._bias_sigma, noise_in, self._weight_sigma.t())
```
Haichao-Zhang, line 496: It might be useful to add a note somewhere that this is the factorized case.

alf/networks/networks.py
```diff
@@ -308,3 +315,184 @@ def forward(self, input, state):
                 lambda x: x.float() if x.dtype.is_floating_point else x, input)
         with torch.cuda.amp.autocast(self._enabled):
             return self._net(input, state)
+
+
+@alf.configurable
+@alf.repr_wrapper
+class NoisyFC(Network):
```
hnyu, line 322: The name hints this class is a layer but in fact it's a network. Maybe changed to a better name? The original paper uses network.

alf/networks/networks.py
```diff
@@ -308,3 +315,186 @@ def forward(self, input, state):
...
+       to change to new noise.
+    3. Since the initial state is always 0, a new noise will always be generated
+       for zero state.
+    4. If it is running in eval mode (i.e., common.is_eval() is True), noise will
+       be disabled (i.e. same as alf.layers.FC).
+    5. The noise is factorized Gaussian noise as described in the paper.
+
+
+    Args:
+        Args:
```
hnyu: duplicate 'Args'

alf/networks/networks.py
```diff
@@ -308,3 +315,186 @@ def forward(self, input, state):
...
+            weight_opt_args: If provided, it will be used as optimizer arguments
+                for weight. And it will be combined with zero_mean=False and
+                fixed_norm=False as optimizer arguments for weight_sigma.
+            bias_opt_args: If provided, it will be used as optimizer arguments
+                for bias. And it will be combined with zero_mean=False as
+                optimizer arguments for bias_sigma.
+    """
+
+    def __init__(self,
+                 input_size: int,
```
hnyu, line 373: Should the standard api for a network be `input_tensor_spec` instead of `input_size`?

alf/networks/networks.py
```diff
@@ -308,3 +315,186 @@ def forward(self, input, state):
...
+        self._bias_sigma.data.fill_(
+            self._std_init / math.sqrt(self._output_size))
+        if self._use_ln:
+            self._ln.reset_parameters()
+        if self._use_bn:
+            self._bn.reset_parameters()
+
+    def _scale_noise(self, size):
+        x = torch.randn(size)
+        return x.sign().mul_(x.abs().sqrt_())
```
hnyu, line 467: This is no longer Gaussian? Why not just `return torch.randn(size)`?

alf/networks/networks.py
```diff
@@ -308,3 +315,186 @@ def forward(self, input, state):
...
+class NoisyFC(Network):
+    """The Noisy Linear Layer discribed in
+
+    Fortunato et. al. `Noisy Networks for Exploration <https://arxiv.org/abs/1706.10295>`_
+
+    Some details:
+
+    1. The noise for each sample in a batch is different.
+    2. The noise is maintained as state. It has a probability of `new_noise_prob`
+       to change to new noise.
```
hnyu, line 336: Without looking at the paper, it is not straightforward to see how the noise is applied to the normal output.

Haichao-Zhang (dismissed):

hnyu (changes requested):

alf/networks/networks.py
emailweixu: fixed.

alf/networks/networks.py
emailweixu: It can be flexible for constructor as many other networks in the same file show.

alf/networks/networks.py
emailweixu: This follows the implementation described in the cited paper.

alf/networks/networks.py
emailweixu: The network in the original paper is really a network consisting of several FC layers.

Haichao-Zhang (approved):

alf/networks/networks.py
```diff
@@ -320,10 +320,15 @@ def forward(self, input, state):
...
 @alf.repr_wrapper
 class NoisyFC(Network):
-    """The Noisy Linear Layer discribed in
+    r"""The Noisy Linear Layer discribed in
 
     Fortunato et. al. `Noisy Networks for Exploration <https://arxiv.org/abs/1706.10295>`_
 
+    In short, the original weight :math:`w` and bias :math:`b` of FC layer are replaced
+    with :math:`w + w_\sigma \epislon^w` and :math:`b + b_\sigma \epsion^b` where
+    :math:`\epsilon^w` and :math:`\epsilon^b` are noise and :math:`w, w_\sigma, b, b_\sigma`
```
hnyu, line 329: This comment of what the noise is applied to is inconsistent with the actual code. The comment says the noise is multiplied to the weight while the code multiplies it with the inputs. For the former, the noise of each sample in a batch is the same.

emailweixu: The noise is different for each sample. And the noise is factorized \epsion_w = \epsion_in \otimes \epsilon_out

hnyu: > The noise is different for each sample. And the noise is factorized \epsion_w = \epsion_in \otimes \epsilon_out

Yeah I know the code has different noises for different samples. But just from the comment here, ":math:`w + w_\sigma \odot \epislon^w`" (the noises are directly applied to the single weight matrix), it's more likely to think that the noise is the same for different samples?

emailweixu: It is equivalent to adding the noise to the weight matrix. And for each sample, the noise is different.

hnyu: > It is equivalent to adding the noise to the weight matrix. And for each sample, the noise is different.

I can get the equivalence. I was just saying that it's easy for someone to misunderstand the noise as the same for different inputs, if not combining the comment here with the details 1 below.

hnyu (approved):

PR #1327, opened 2022-05-13
Suport averaged summary over a summary interval

emailweixu:
In some situations, the single summary value might be too volatile. So we allow averaged value over a summary interval if average_over_interval=True

And use average_over_interval for summarize_reward. This is helpful if the episode is long and reward is sparse so that we don't need to wait until episode end to see if the agent gets any reward.

### DIFF
--- a/alf/summary/summary_ops.py
+++ b/alf/summary/summary_ops.py
@@ -85,6 +85,9 @@ class scope(object):
         _scope_stack.pop()
 
 
+_SUMMARY_DATA_BUFFER = {}
+
+
 def _summary_wrapper(summary_func):
     """Summary wrapper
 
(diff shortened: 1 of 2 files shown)


alf/summary/summary_ops.py
```diff
@@ -85,24 +85,47 @@ def __exit__(self, type, value, traceback):
...
     """Summary wrapper
 
     Wrapper summary function to reduce cost for data computation
     """
 
     @functools.wraps(summary_func)
-    def wrapper(name, data, step=None, **kwargs):
-        if should_record_summaries():
+    def wrapper(name, data, average_over_interval=False, step=None, **kwargs):
+        if average_over_interval:
```
hnyu: `average_over_interval` -> `average_over_summary_interval` to make it clearer about what the interval is?

le-horizon (approved):
LG, good to fix this.

hnyu (approved):

PR #1328, opened 2022-05-13
epsilon_greedy and passing prediction state across steps for MCTSAlgorithm

emailweixu:
Sometime we want to use stateful prediction model for MCTSModel and pass those states   across different steps. One such scenario is the use of stateful noise (NoisyFC) for extended  exploration. So we add the support for this.

Also improve MCTSAlgorithm to support epsilon_greedy so that it can use greedy action for prediction during evaluation.


alf/algorithms/muzero_representation_learner.py
```diff
@@ -268,11 +267,19 @@ def __init__(
                 target_models=[self._target_model],
                 tau=target_update_tau,
                 period=target_update_period)
-            # We assume that the algorithm used for reanalyze will implement the
-            # interface ``set_model`` to indicate that it requires access to the
-            # underlying (target) model.
-            if hasattr(self._reanalyze_algorithm, 'set_model'):
-                self._reanalyze_algorithm.set_model(self._target_model)
+            assert 'model' in inspect.signature(
+                reanalyze_algorithm_ctor).parameters
```
breakds: Since we allow `reanalyze_algorithm_ctor` to not have `model` below, this `assert` is unnecessary?

emailweixu: Oh, this is for debugging to make sure the inspection works as expected. It should be removed.

breakds (dismissed):

breakds (approved):

PR #1331, opened 2022-05-17
add dqn_algorithm, config for atari games, and plot for Qbert

le-horizon:


alf/algorithms/dqn_algorithm.py
```diff
@@ -0,0 +1,128 @@
+# Copyright (c) 2020 Horizon Robotics and ALF Contributors. All Rights Reserved.
```
hnyu: 2020 -> 2022

alf/algorithms/dqn_algorithm.py
```diff
@@ -0,0 +1,128 @@
...
+@alf.configurable
+class DqnAlgorithm(SacAlgorithm):
+    r"""DQN/DDQN algorithm:
+
+    ::
+
+        Mnih et al "Playing Atari with Deep Reinforcement Learning", arXiv:1312.5602
+        Hasselt et al "Deep Reinforcement Learning with Double Q-learning", arXiv:1509.06461
+
+    The difference with DDQN is that a minimum is taken from the two critics,
```
hnyu: DDQN -> DQN

alf/algorithms/dqn_algorithm.py
```diff
@@ -0,0 +1,128 @@
...
+                 rollout_epsilon_greedy=0.1,
+                 num_critic_replicas=2,
+                 env=None,
+                 config: TrainerConfig = None,
+                 critic_loss_ctor=None,
+                 debug_summaries=False,
+                 name="DqnAlgorithm"):
+        """
+        Args:
+            observation_spec (nested TensorSpec): representing the observations.
```
hnyu, line 69: Should better use type hints in the argument list.

alf/algorithms/dqn_algorithm.py
```diff
@@ -0,0 +1,128 @@
...
+                provided to the algorithm which performs ``train_iter()`` by
+                itself.
+            critic_loss_ctor (None|OneStepTDLoss|MultiStepLoss): a critic loss
+                constructor. If ``None``, a default ``OneStepTDLoss`` will be used.
+            debug_summaries (bool): True if debug summaries should be created.
+            name (str): The name of this algorithm.
+        """
+        self._rollout_epsilon_greedy = as_scheduler(rollout_epsilon_greedy)
+        # Disable alpha learning:
+        alpha_optimizer = AdamTF(lr=0)
```
hnyu, line 102: We should also set the initial alpha to be 0 (initial_log_alpha=-inf) to disable the entropy. There is still some issue of basing on SAC because it computes an action distribution for discrete action and if alpha is 0, then this distribution is undefined. 


``` python
     if self._act_type != ActionType.Continuous:
            q_values, critic_state = self._compute_critics(
                self._critic_networks, *critic_network_inputs, state.critic)

            new_state = new_state._replace(critic=critic_state)
            if self._act_type == ActionType.Discrete:
                alpha = torch.exp(self._log_alpha).detach()
            else:
                alpha = torch.exp(self._log_alpha[0]).detach()
            # p(a|s) = exp(Q(s,a)/alpha) / Z;
            logits = q_values / alpha
            discrete_action_dist = td.Categorical(logits=logits)
            if eps_greedy_sampling:
                discrete_action = dist_utils.epsilon_greedy_sample(
                    discrete_action_dist, epsilon_greedy)
            else:
                discrete_action = dist_utils.sample_action_distribution(
                    discrete_action_dist)
```

alf/algorithms/sac_algorithm.py
```diff
@@ -717,7 +723,10 @@ def _critic_train_step(self, inputs: TimeStep, state: SacCriticState,
             probs = common.expand_dims_as(action_distribution.probs,
                                           target_critics)
             # [B, reward_dim]
-            target_critics = torch.sum(probs * target_critics, dim=1)
+            if target_action_picker is not None:
+                target_critics = target_action_picker(target_critics)
```
hnyu: Together with the comment above (about alpha=0), I think we can just use `probs * target_critics` as long as the `probs` peaks at an entry.

hnyu (changes requested):

alf/algorithms/dqn_algorithm.py
```diff
@@ -0,0 +1,128 @@
...
+            # Allow custom optimizer for q_network:
+            critic_optimizer=q_optimizer,
+            alpha_optimizer=alpha_optimizer,
+            debug_summaries=debug_summaries,
+            name=name)
+        assert self._act_type == ActionType.Discrete
+
+    def rollout_step(self, inputs: TimeStep, state: DqnState):
+        return super().rollout_step(
+            inputs, state, eps=self._rollout_epsilon_greedy())
```
hnyu: Don't need to pass this as an argument. You can directly use it at line 554 if `rollout_epsilon_greedy` is defined for `SACAlgorithm`. 

On the other hand, this doesn't seem to be the epsilon greedy of DQN. In DQN, there is an \epsilon chance of uniformly randomly sampling actions and 1-\epsilon chance of getting the greedy action (argmax Q). In SAC, the meaning is different. The \epsilon chance is for sampling from the action distribution instead of uniform sampling. The action distribution might not be uniform depending on the current alpha and Q.

alf/algorithms/dqn_algorithm.py
le-horizon: Done

alf/algorithms/dqn_algorithm.py
le-horizon: Done

alf/algorithms/dqn_algorithm.py
le-horizon: Done

alf/algorithms/dqn_algorithm.py
le-horizon: Here, I'm just not learning alpha, disabling entropy reward, and picking target action greedily.  I think this is enough to sample actions greedily without setting alpha to zero.

alf/algorithms/dqn_algorithm.py
le-horizon: Good point.  I should use uniform to sample the epsilon action.  Let me know how you like this implementation.  (Also updated tb plots.)

alf/algorithms/sac_algorithm.py
le-horizon: I found out that in order to handle multi dim reward properly, some special logic is needed anyways, and it seems most appropriate to put the logic here.  Let me know if you have a better idea.

le-horizon (commented):
Done all, @hnyu Let me know how you like this one.

Thanks,
Le

hnyu:
Does our DQN reproduce the result on QBert ?

le-horizon:
> Does our DQN reproduce the result on QBert ?

It's hard to say as we only ran 1/4 of the frames of reported, but it's in the rough ballpark (DQN is reported to be 1k-4k, DDQN is 10k, while we are at around 2k).
The QBert result is included mostly for record and comparison in the future.

alf/algorithms/dqn_algorithm.py
```diff
@@ -0,0 +1,252 @@
...
+    The difference with DQN is that a minimum is taken from the two critics,
+    similar to TD3, instead of choosing the maximum action using the Q network
+    and evaluating the action value using the target Q network.
+
+    The implementation is based on the SAC algorithm.
+    """
+
+    def __init__(self,
+                 observation_spec,
+                 action_spec,
```
hnyu: can use type hints `alf.tensor_specs.NestedTensorSpec` for these two argments.

alf/algorithms/dqn_algorithm.py
```diff
@@ -0,0 +1,252 @@
...
+        assert self._act_type == ActionType.Discrete
+
+    # Copied and modified from sac_algorithm (discrete actions).
+    def _predict_action(self,
+                        observation,
+                        state: DqnActionState,
+                        epsilon_greedy=None,
+                        eps_greedy_sampling=False,
+                        rollout=False):
+        """The reason why we want to do action sampling inside this function
```
hnyu: This docstring is no longer accurate

alf/algorithms/dqn_algorithm.py
```diff
@@ -0,0 +1,252 @@
...
+        instead of outside is that for the mixed case, once a continuous action
+        is sampled here, we should pair it with the discrete action sampled from
+        the Q value. If we just return two distributions and sample outside, then
+        the actions will not match.
+        """
+        new_state = DqnActionState()
+
+        critic_network_inputs = (observation, None)
+        if self._act_type == ActionType.Mixed:
```
hnyu: No need to have code for ActionType.Mixed

alf/algorithms/dqn_algorithm.py
```diff
@@ -0,0 +1,252 @@
...
+            critic_network_inputs = (observation, (None, continuous_action))
+
+        q_values, critic_state = self._compute_critics(
+            self._critic_networks, *critic_network_inputs, state.critic)
+
+        new_state = new_state._replace(critic=critic_state)
+        if self._act_type == ActionType.Discrete:
+            alpha = torch.exp(self._log_alpha).detach()
```
hnyu: Can only keep this branch.

alf/algorithms/dqn_algorithm.py
```diff
@@ -0,0 +1,252 @@
...
+                discrete_action_dist,
+                epsilon_greedy,
+                # NOTE: This is the only departure from SAC:
+                True)
+        else:
+            discrete_action = dist_utils.sample_action_distribution(
+                discrete_action_dist)
+
+        action_dist = discrete_action_dist
+        action = discrete_action
```
hnyu: Can remove these two lines of only discrete action is considered.

alf/algorithms/dqn_algorithm.py
```diff
@@ -0,0 +1,252 @@
...
+
+        if (self._reproduce_locomotion and rollout
```
hnyu: `reproduce_locomotion` only applies to continuous actions case, and can be removed here.

alf/algorithms/dqn_algorithm.py
```diff
@@ -0,0 +1,252 @@
...
+            # This uniform sampling seems important because for a squashed Gaussian,
+            # even with a large scale, a random policy is not nearly uniform.
+            action = alf.nest.map_structure(
+                lambda spec: spec.sample(outer_dims=observation.shape[:1]),
+                self._action_spec)
+
+        return action_dist, action, q_values, new_state
+
+    # Copied and modified from sac_algorithm (discrete actions).
+    def rollout_step(self, inputs: TimeStep, state: DqnState):
```
hnyu, line 170: You don't have to copy this function? You can directly use rollout epsilon greedy in your defined `_predict_action`

alf/algorithms/dqn_algorithm.py
```diff
@@ -0,0 +1,252 @@
...
+
+        new_state = DqnState(
+            action=action_state, actor=actor_state, critic=critic_state)
+        return AlgStep(
+            output=action,
+            state=new_state,
+            info=DqnInfo(action=action, action_distribution=action_dist))
+
+    # Copied and modified from sac_algorithm (discrete actions).
+    def _critic_train_step(self, inputs: TimeStep, state: DqnCriticState,
```
hnyu: I suggest that you just define a one-hot action distribution (argmax entry has 1) for DQN and directly reuse SAC's `_critic_train_step`. You can do this in your defined `_predict_action` (first return result `action_dist`).

hnyu (changes requested):
It seems to me that you only need to overwrite SAC's `_predict_action` by manually defining a one-hot action distribution in DQN case.

alf/algorithms/dqn_algorithm.py
le-horizon: Done

alf/algorithms/dqn_algorithm.py
le-horizon: Removed

alf/algorithms/dqn_algorithm.py
le-horizon: Done.

alf/algorithms/dqn_algorithm.py
le-horizon: I could do as you said, but I'll have to ignore the input epsilon_greedy to _predict_action(), because sac's rollout uses epsilon_greedy=1.0.  This makes the code harder to maintain (error prone) in the future, so I'd prefer to copy the rollout_step() function over and explicitly pass in the epsilon_greedy parameter.

alf/algorithms/dqn_algorithm.py
le-horizon: Done

alf/algorithms/dqn_algorithm.py
le-horizon: done

alf/algorithms/dqn_algorithm.py
le-horizon: done

alf/algorithms/dqn_algorithm.py
le-horizon: done

le-horizon (commented):
Thanks for the comments @hnyu .

Done all, I'll replace the tb plot once job is done.  PTAL.

Thanks,
Le

alf/algorithms/dqn_algorithm.py
```diff
@@ -0,0 +1,192 @@
...
+        else:
+            # Greedy for train_step to obtain target value from target network.
+            # The greedy action here is the maximizer of q values, and will be used
+            # to obtain target value using the target network.
+            probs = torch.zeros_like(q_values)
+            greedy_act_prob = 1
+        greedy_action = torch.argmax(q_values, dim=-1)
+        probs[(torch.arange(size[0]), greedy_action)] = greedy_act_prob
+        action_dist = td.Categorical(probs=probs)
+        action = dist_utils.epsilon_greedy_sample(action_dist, eps=0.)
```
hnyu: This place should be eps=1? For training, it produces the same result with eps=0 because of one-hot. For rollout and evaluation, it does sampling, requiring eps=1. 

Actually, to minimize confusion here, you can just use `dist_utils.sample_action_distribution(action_dist)`, without introducing eps_greedy.

hnyu (changes requested):
Looks good. Only one last comment.

hnyu:
Also please remember to update the "Algorithms" table in README to include DQN.

alf/algorithms/dqn_algorithm.py
le-horizon: Good point, done.

le-horizon:
> Also please remember to update the "Algorithms" table in README to include DQN.

Done all.  Could you take another look, Haonan?

Thanks,
Le

hnyu (approved):

PR #1332, opened 2022-05-17
convert only one minibatch of data to GPU (to save GPU memory) with t…

le-horizon:
…he overhead of preprocessing (data_transformations) happening in CPU.

### DIFF
--- a/alf/experience_replayers/replay_buffer.py
+++ b/alf/experience_replayers/replay_buffer.py
@@ -71,6 +71,7 @@ class ReplayBuffer(RingBuffer):
                  gamma=.99,
                  reward_clip=None,
                  enable_checkpoint=False,
+                 convert_only_minibatch_to_device=False,
                  name="ReplayBuffer"):
         """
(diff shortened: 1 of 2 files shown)


alf/experience_replayers/replay_buffer.py
```diff
@@ -118,6 +119,9 @@ def __init__(self,
                 Usually consistent with ``TDLoss.gamma``.
             reward_clip (tuple|None): None or (min, max) for reward clipping.
             enable_checkpoint (bool): whether checkpointing this replay buffer.
+            convert_only_minibatch_to_device (bool): when True, only convert a minibatch
```
emailweixu, line 122: It feels weird that this kind of setting is at particular DataTransformer. It should be at the level of Algorithm. And if GPU memory is a concern, now we use fractional unroll_length to reduce it. So this option is not that necessary.

le-horizon:
Good point.  Let's use fractional unroll and close this PR.

PR #1333, opened 2022-05-17
Misc improvements

le-horizon:
adding various asserts, debug messages, and small improvements

### DIFF
--- a/alf/algorithms/merlin_algorithm.py
+++ b/alf/algorithms/merlin_algorithm.py
@@ -543,8 +543,6 @@ class ResnetEncodingNetwork(alf.networks.Network):
             enc_layers.append(res_block)
             in_channels = 64
 
-        if output_activation is None:
-            output_activation = alf.math.identity
         enc_layers.extend([
             nn.Flatten(),
             alf.layers.FC(
--- a/alf/algorithms/merlin_algorithm.py
+++ b/alf/algorithms/merlin_algorithm.py
@@ -543,6 +543,8 @@ class ResnetEncodingNetwork(alf.networks.Network):
             enc_layers.append(res_block)
             in_channels = 64
 
+        if output_activation is None:
+            output_activation = alf.math.identity
         enc_layers.extend([
             nn.Flatten(),
             alf.layers.FC(
--- a/alf/config_util.py
+++ b/alf/config_util.py
@@ -337,7 +337,11 @@ def pre_config(configs):
         try:
             config1(name, value, mutable=False)
             _HANDLED_PRE_CONFIGS.append((name, value))
-        except ValueError:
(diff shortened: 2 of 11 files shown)


alf/algorithms/merlin_algorithm.py
```diff
@@ -543,6 +543,8 @@ def __init__(self,
             enc_layers.append(res_block)
             in_channels = 64
 
+        if output_activation is None:
+            output_activation = alf.math.identity
```
emailweixu: Seems unnecessary. Can provide alf.math.identity as argument.

alf/utils/data_buffer_test.py
```diff
@@ -40,12 +40,20 @@ def get_batch(env_ids, dim, t, x):
...
+    #         device="cpu").unsqueeze(0))
+    if batch_size > 1 and x.ndim > 0 and batch_size == x.shape[0]:
+        a = x
+    else:
+        a = x * torch.ones(batch_size, dtype=torch.float32, device="cpu")
+    if batch_size > 1 and t.ndim > 0 and batch_size == t.shape[0]:
+        pass
+    else:
+        t = t * torch.ones(batch_size, dtype=torch.int32, device="cpu")
+    ox = a.unsqueeze(1).clone().requires_grad_(True)
```
emailweixu: what is the purpose of this change?

alf/config_util.py
```diff
@@ -337,7 +337,11 @@ def pre_config(configs):
         try:
             config1(name, value, mutable=False)
             _HANDLED_PRE_CONFIGS.append((name, value))
-        except ValueError:
+        except ValueError as e:
+            # Most of the times, for command line flags, this warning is a false alarm.
+            # This can be useful in other failures, e.g. when the Config has already been used,
+            # before configuring its value.
+            logging.warning("pre_config potential error: %s", e)
```
emailweixu: This warning is hard to understand. It's better to identify the case of the Config has already been used. Perhaps throw a different type of Exception when config has been used in config1()?

le-horizon: Good point.  Logging error in config1.

alf/algorithms/merlin_algorithm.py
le-horizon: Good point. Removed.

alf/utils/data_buffer_test.py
le-horizon: This is needed because we allow x and t inputs to be scalars, which will be expanded to be consistent with the batch_size.  Made code easier to read, and commented.

le-horizon (commented):
Thanks for the comments.  Done all, PTAL.

Le

PR #1334, opened 2022-05-17
allow command line flag to skip use_torch_deterministic

le-horizon:
This change allows using ``--force_torch_deterministic=False`` to disable use_torch_deterministic.  Sometimes when random seed is set, torch 1.8 complains about _scatter_add kernel does not have a deterministic implementation.  This is a false alarm in our sac atari game runs.  We can safely disable it, but may not be the case for other tasks or algorithms.

### DIFF
--- a/alf/bin/play.py
+++ b/alf/bin/play.py
@@ -53,7 +53,12 @@ def _define_flags():
     flags.DEFINE_integer('random_seed', None, "random seed")
     flags.DEFINE_bool(
         'force_torch_deterministic', True,
-        'torch.use_deterministic_algorithms when random_seed is set')
+        'torch.use_deterministic_algorithms when random_seed is set. '
+        'When it is False, deterministic behavior is not guaranteed, '
+        'but could still be deterministic, e.g. for sac_breakout_conf.py. '
+        'Setting a random seed without setting this to False, training '
+        'could throw this error: _scatter_add kernel does not have a '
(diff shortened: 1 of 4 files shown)


emailweixu:
The training are not repeatable in general without torch.use_deterministic_algorithms. Perhaps in that case we should simply not to set random seed?

le-horizon:
> The training are not repeatable in general without torch.use_deterministic_algorithms. Perhaps in that case we should simply not to set random seed?

It is actually repeatable, and the warning is a false alarm in our sac atari game runs.
It may not be the case for other tasks or algorithms.
Should I still submit it, or keep it in my local branch?

alf/bin/play.py
```diff
@@ -51,6 +51,9 @@ def _define_flags():
         "specify the checkpoint to be loaded. If None, the latest checkpoint under "
         "train_dir will be used.")
     flags.DEFINE_integer('random_seed', None, "random seed")
+    flags.DEFINE_bool(
+        'force_torch_deterministic', True,
+        'torch.use_deterministic_algorithms when random_seed is set')
```
emailweixu: I think we should add more explanation here about the potential issues of using False here.

le-horizon: Good point, done.  PTAL.

emailweixu (approved):

PR #1337, opened 2022-05-21
compress alf snapshot

hnyu:
Store a compressed ALF snapshot as a '.tar' file instead of the raw directory. This will have two benefits:

1. Speed up downloading from our internal cluster. 
2. Avoid creating a file path under the snapshot that is beyond the 255 chars limit when using `wget` for downloading.

### DIFF
--- a/alf/bin/grid_search.py
+++ b/alf/bin/grid_search.py
@@ -229,7 +229,7 @@ class GridSearch(object):
                            id,
                            repeat,
                            token_len=20,
-                           max_len=255):
+                           max_len=50):
         """Generate a run name by writing abbr parameter key-value pairs in it,
         for an easy comparison between different search runs without going
         into Tensorboard 'text' for run details.
@@ -385,7 +385,6 @@ def launch_snapshot_gridsearch():
     it's actually using the right ALF version.
     """
     root_dir = common.abs_path(FLAGS.root_dir)
-    alf_repo = os.path.join(root_dir, "alf")
 
     # write the current conf file as
     # ``<root_dir>/alf_config.py`` or ``<root_dir>/configured.gin``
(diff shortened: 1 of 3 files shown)


alf/bin/grid_search.py
```diff
@@ -229,7 +229,7 @@ def _generate_run_name(self,
                            id,
                            repeat,
                            token_len=20,
-                           max_len=255):
+                           max_len=50):
```
hnyu, line 232: This is also to avoid creating a long url path when downloaded by `wget`.

alf/utils/common.py
```diff
@@ -1440,11 +1440,34 @@ def rsync(src, target, includes):
...
+
+
+def unzip_alf_snapshot(root_dir: str):
+    """Restore an ALF snapshot from a job directory by unzipping the snapshot
+    'tar' file.
+
+    Args:
+        root_dir: the tensorboard job directory
+    """
+    alf_zipped_repo = os.path.join(root_dir, "alf.tar")
```
breakds: Check for the existence of `tar` so that we can prompt clear information when it does not?

alf/utils/common.py
```diff
@@ -1440,11 +1440,34 @@ def rsync(src, target, includes):
...
+    if os.path.isfile(alf_zipped_repo):
+        info("=== Using an ALF snapshot at '%s' ===", alf_zipped_repo)
+    else:
+        info("=== Didn't find a snapshot; using update-to-date ALF ===")
+        return
+
+    os.system("rm -rf %s/alf" % root_dir)
+    os.system("cd %s; tar -xzf alf.tar" % root_dir)
```
breakds: We usually name it `alf.tar.gz` to suggest it is "zipped".

alf/utils/common.py
hnyu: > Check for the existence of `tar` so that we can prompt clear information when it does not?

Already checked below?

alf/utils/common.py
hnyu: > We usually name it `alf.tar.gz` to suggest it is "zipped".

updated the extension

breakds (approved):

PR #1346, opened 2022-06-11
Selective mode play

Haichao-Zhang:
An extension of the current play tool by supporting selectively saving only part of the episodes.
This is useful in several cases, e.g. discovering failure modes.

### DIFF
--- a/alf/trainers/policy_trainer.py
+++ b/alf/trainers/policy_trainer.py
@@ -745,6 +745,35 @@ def _step(algorithm,
           recorder=None,
           sleep_time_per_step=0,
           selective_criteria_func=None):
+    """Perform one step interaction using the outpupt action from ``algorithm``
+    taking ``time_step`` as input. Also record the metrics.
+
+    Note that this function is used both in ``play`` below and ``evaluate`` in
+    ``evaluator.py``.
+
+    Args:
+        algorithm (RLAlgorithm): the algorithm under evaluation
+        env: the environment
+        time_step (TimeStep): current time step
+        policy_state (nested Tensor): state of the policy
+        trans_state (nested Tensor): state of the transformer(s)
+        metrics (StepMetric): a list of metrics that will be updated based on
+            ``time_step``.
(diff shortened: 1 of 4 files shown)


alf/trainers/policy_trainer.py
```diff
@@ -743,16 +743,69 @@ def _step(algorithm,
...
+        metrics (StepMetric): a list of metrics that will be updated based on
+            ``time_step``.
+        render (bool|False): if True, display the frames of ``env`` on a screen.
+        recorder (VideoRecorder|None): recorder the frames of ``env`` and other
+            additional images in prediction step info if present.
+        sleep_time_per_step (int|0): The sleep time between two frames when
+            ``render`` is True.
+        selective_criteria_func (callable|None): a callable for determining
+            whether an episode will be saved to the video file when a valid
+            recorder is provided.
```
emailweixu: should document its arguments

alf/trainers/policy_trainer.py
```diff
@@ -743,16 +743,69 @@ def _step(algorithm,
...
+    if recorder and selective_criteria_func is None:
         recorder.capture_frame(policy_step.info, time_step.is_last())
+
+    elif recorder and selective_criteria_func is not None:
+        env_frame = recorder.capture_env_frame()
+        recorder.cache_frame_and_pred_info(env_frame, policy_step.info)
+
+        if time_step.is_last():
+            # below is an example failure_criteria based on return.
+            # This should be adjusted according to the particular task at hand.
```
emailweixu: Should be removed. They are at 934-935

alf/trainers/policy_trainer.py
Haichao-Zhang: Now added documents on the two arguments.

alf/trainers/policy_trainer.py
Haichao-Zhang: Removed.

emailweixu (approved):

Issue #1347, opened 2022-06-14
documentation not generated


emailweixu:
See for example: https://alf.readthedocs.io/en/latest/api/alf.algorithms.html#alf-algorithms-actor-critic-algorithm

PR #1348, opened 2022-06-14
downgrade protobuf version

hnyu:
Fixed issue #1347 , need to downgrade protobuf version. 

A related RTD issue: @breakds metadrive's doc is not correctly generated. The reason is that if metadrive is not installed, some imports will be skipped, but the typehints still use them.

``` python
try:
    import metadrive
    from metadrive.component.vehicle.base_vehicle import BaseVehicle
    from metadrive.engine.base_engine import BaseEngine
except ImportError:
    from unittest.mock import Mock
    # create 'metadrive' as a mock to not break python argument type hints
    metadrive = Mock()
```

https://alf.readthedocs.io/en/latest/api/alf.environments.metadrive.html

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -21,3 +21,4 @@ torchvision==0.9.1+cpu
 cnest==1.0.4
 gym3==0.3.3
 procgen==0.10.4
+protobuf==3.20.*


breakds:
> Fixed issue #1347 , need to downgrade protobuf version.
> 
> A related RTD issue: @breakds metadrive's doc is not correctly generated. The reason is that if metadrive is not installed, some imports will be skipped, but the typehints still use them.
> 
> ```python
> try:
>     import metadrive
>     from metadrive.component.vehicle.base_vehicle import BaseVehicle
>     from metadrive.engine.base_engine import BaseEngine
> except ImportError:
>     from unittest.mock import Mock
>     # create 'metadrive' as a mock to not break python argument type hints
>     metadrive = Mock()
> ```
> 
> https://alf.readthedocs.io/en/latest/api/alf.environments.metadrive.html

Yep you are right. I am not sure whether it is good to include environment dependencies in the main requirement list though. Is there a good convention in python on including "extra" or "recommended" dependencies?

docs/requirements.txt
```diff
@@ -21,3 +21,4 @@ torchvision==0.9.1+cpu
 cnest==1.0.4
 gym3==0.3.3
 procgen==0.10.4
+protobuf==3.20.*
```
breakds, line 24: Not sure why this is a "downgrade". seems to me like "adding protobuf", no?

hnyu: > Not sure why this is a "downgrade". seems to me like "adding protobuf", no?

protobuf is included in autodoc. Somehow RTD upgrades its version recently which caused the issue.

breakds: Gotcha. Thanks for the clarification.

breakds (approved):

hnyu:
> > Fixed issue #1347 , need to downgrade protobuf version.
> > A related RTD issue: @breakds metadrive's doc is not correctly generated. The reason is that if metadrive is not installed, some imports will be skipped, but the typehints still use them.
> > ```python
> > try:
> >     import metadrive
> >     from metadrive.component.vehicle.base_vehicle import BaseVehicle
> >     from metadrive.engine.base_engine import BaseEngine
> > except ImportError:
> >     from unittest.mock import Mock
> >     # create 'metadrive' as a mock to not break python argument type hints
> >     metadrive = Mock()
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > https://alf.readthedocs.io/en/latest/api/alf.environments.metadrive.html
> 
> Yep you are right. I am not sure whether it is good to include environment dependencies in the main requirement list though. Is there a good convention in python on including "extra" or "recommended" dependencies?

I think you need to make sure to evaluate/import minimal number of metadrive related python files when metadrive is unavailable. For example, also put those imports in the ``try`` block.

PR #1349, opened 2022-06-14
Add time limit for metadrive

breakds:
## Motivation

The training of agent on MetaDrive can often get stuck in a immobilized policy. To avoid getting a lot of very long episodes, we would like to set time limit for it. Fortunately MetaDrive itself provides a parameter called `horizon` to set time limit, and therefore we just make use of it.

## Solution

Added the `time_limit` parameter in `suite_metadrive.load`, with a default of `1200` steps.

## Testing

Has been running experiments with time limit for more than a month now, and can tell that it works correctly via the following figure
time_limit
:

### DIFF
--- a/alf/environments/suite_metadrive.py
+++ b/alf/environments/suite_metadrive.py
@@ -196,7 +196,8 @@ def load(
         map_spec: Union[int, str] = 4,
         crash_penalty: float = 5.0,
         speed_reward_weight: float = 0.1,
-        success_reward: float = 10.0):
+        success_reward: float = 10.0,
+        time_limit: int = 1200):
     """Load the MetaDrive environment and wraps it with AlfMetaDriveWrapper.
     Args:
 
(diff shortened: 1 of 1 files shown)


alf/environments/suite_metadrive.py
```diff
@@ -41,11 +41,10 @@
 def _space_to_spec(space: gym.spaces.box.Box):
     # NOTE: this is meta drive specific conversion function as it
     # assumes low and high are uniform.
-    return BoundedTensorSpec(
-        shape=space.shape,
-        dtype=space.dtype.name,
-        minimum=space.low.flat[0],
-        maximum=space.high.flat[0])
+    return BoundedTensorSpec(shape=space.shape,
```
breakds: Blame the formatter for the extra change :)

Haichao-Zhang (approved):

PR #1350, opened 2022-06-14
Upgrade the nix-based development environment for ALF

breakds:
## Motivation

Point the nix development environment to the up-to-date [nix-alf-devenv](https://github.com/HorizonRobotics/alf-nix-devenv). This mainly includes:

1. Upgrade `pytorch` to 1.11.0
2. Upgrade `metadrive` to its latest version
3. Added `tensor-spline` to the environment

# Solution

N/A

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -6,14 +6,15 @@
         "nixpkgs": [
           "nixpkgs"
         ],
+        "tensor-splines-flake": "tensor-splines-flake",
         "utils": "utils"
       },
       "locked": {
-        "lastModified": 1639518535,
-        "narHash": "sha256-gkVszJhfMXKlv1gWNUr3kEHPmLWXn95FUq0liaA6JxY=",
+        "lastModified": 1654808684,
+        "narHash": "sha256-AVyRuZJSqJFX/O571Yl9JGFKvDZOgqrbTNEujgaNNU8=",
         "owner": "HorizonRobotics",
         "repo": "alf-nix-devenv",
-        "rev": "d7d9d65f3cfca9740fa21b4853b32768e967ba19",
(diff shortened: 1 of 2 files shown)


flake.lock
```diff
@@ -6,14 +6,15 @@
         "nixpkgs": [
           "nixpkgs"
         ],
+        "tensor-splines-flake": "tensor-splines-flake",
```
breakds, line 9: Please note that the update on this file is automatically generated.

flake.nix
```diff
@@ -12,8 +12,8 @@
   };
 
   outputs = { self, nixpkgs, ... }@inputs: inputs.utils.lib.eachSystem [
-    "x86_64-linux" "i686-linux" "aarch64-linux" "x86_64-darwin"
```
le-horizon, line 15: Is this intentional?

flake.nix
```diff
@@ -12,8 +12,8 @@
...
+    "x86_64-linux"
   ] (system: {
-    devShell = inputs.alf-devenv.devShell."${system}";
+    devShell = inputs.alf-devenv.devShells."${system}".default;
```
le-horizon, line 17: will this change affect some of the documentation?

le-horizon (approved):
Not sure about two things.  Otherwise look good.

flake.nix
breakds: Nope this will not.

flake.nix
breakds: Good point. This is intentional as I haven't tested it on MacOS X yet.

le-horizon: Ok then.  It's best to support one frequent use case than to worry about all possible uses.  Feel free to merge.

PR #1351, opened 2022-06-15
compute and summarize gradient noise scale

hnyu:
Add code for computing and summarizing gradient noise scale (GNS). GNS can be used as an indicator  for selecting the mini-batch size (orders of magnitude), and as a rough indicator of the difficulty of a learning task (which might come from the data, the model complexity, or the learning rate).

### DIFF
--- a/alf/optimizers/utils.py
+++ b/alf/optimizers/utils.py
@@ -74,7 +74,7 @@ class GradientNoiseScaleEstimator(nn.Module):
     def __init__(self,
                  batch_size_ratio: float = 0.1,
                  update_rate: float = 0.01,
-                 gradient_norm_clip: float = 10.,
+                 gradient_norm_clip: float = None,
                  name: str = "GNSEstimator"):
         """
         Args:
@@ -88,7 +88,8 @@ class GradientNoiseScaleEstimator(nn.Module):
                 makes the estimated GNS more biased (because quantities at different
                 training steps are averaged) while a larger value (quicker
                 update) makes it have more variances.
-            gradient_norm_clip: a clipping value for gradient norm outliers
+            gradient_norm_clip: a clipping value for gradient norm outliers. If
+                None, no clipping is performed.
         """
         super().__init__()
         self._name = name
@@ -102,7 +103,9 @@ class GradientNoiseScaleEstimator(nn.Module):
     def _calculate_gradient_norm(self, loss: torch.Tensor,
                                  tensors: alf.nest.NestedTensor):
         grads = alf.nest.utils.grad(tensors, loss.mean(), retain_graph=True)
-        norm2 = torch.clamp(global_norm(grads), max=self._grad_norm_clip)
+        norm2 = global_norm(grads)
+        if self._grad_norm_clip is not None:
+            norm2 = torch.clamp(norm2, max=self._grad_norm_clip)
         return norm2**2
 
     def forward(self, loss: torch.Tensor, tensors: alf.nest.NestedTensor):
--- a/alf/optimizers/utils.py
+++ b/alf/optimizers/utils.py
@@ -60,10 +60,10 @@ class GradientNoiseScaleEstimator(nn.Module):
 
     Generally, GNS indicates the noise-to-signal value of SGD. The authors suggest
     that we should choose a batch size close to GNS in order to average out the
-    noise in the gradient.
-
-    We would expect a higher GNS for a difficult learning task, especially when
-    different training samples generate opposite gradient directions.
+    noise in the gradient. In other words, GNS is positively correlated to the
+    current gradient descent difficulty. We would expect a high GNS for a difficult
+    learning task, especially when different training samples generate opposite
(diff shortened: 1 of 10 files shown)


alf/optimizers/utils.py
```diff
@@ -31,3 +36,113 @@ def get_opt_arg(p: nn.Parameter, argname: str, default: Any = None):
...
+    def forward(self, loss: torch.Tensor, tensors: alf.nest.NestedTensor):
+        """Given a loss tensor and a nest of tensors, return the estimated GNS.
+
+        Args:
+            loss: a loss tensor *before* taking the mean. Each entry of the tensor
+                represents an individual loss on a single training sample. Ideally,
+                the samples used for computing these losses should be sampled *with*
+                replacement independently. The loss can have a shape of either
+                ``[T,B]`` or ``[B]``. In the first case, we will only take
+                ``[0,B]`` for computing the gns due to sample correlation along
```
emailweixu: why not simply use all?

alf/optimizers/utils.py
```diff
@@ -31,3 +36,113 @@ def get_opt_arg(p: nn.Parameter, argname: str, default: Any = None):
...
+        shuffled_loss = loss[torch.randperm(B)]
+
+        b = max(1, int(B * self._batch_size_ratio))
+        B -= b
+
+        B_norm2 = self._calculate_gradient_norm(shuffled_loss[b:], tensors)
+        b_norm2 = self._calculate_gradient_norm(shuffled_loss[:b], tensors)
+
+        gradient_norm = (B * B_norm2 - b * b_norm2) / (B - b)
+        var_trace = (b_norm2 - B_norm2) / (1. / b - 1. / B)
```
emailweixu: Cause this is quite different from the original formula, do you have some derivation showing that this is what is needed?

alf/optimizers/utils.py
hnyu: > why not simply use all?

The assumption of using batch statistics to estimate the true trace and gradient norm is that individual samples of a batch are i.i.d. If T>1, then samples along T are correlated.

alf/optimizers/utils.py
hnyu: > Cause this is quite different from the original formula, do you have some derivation showing that this is what is needed?

This is the simplified GNS defined in Appendix A. Compared to the complex one, it doesn't require computing the hessian matrix.

alf/optimizers/utils.py
hnyu: > why not simply use all?

Somehow selecting one `t` out of `T` still results in a more stable GNS estimate than using all `t`.

hnyu:
Currently the estimated result is still not stable if no gradient norm clipping is used. The figure shown for sac_cart_pole_conf.py has a norm_clipping=10. Maybe this also implies that we should consider clipping our gradients when training?

hnyu:
Closed as this metric is still not indicative so far.

alf/optimizers/utils.py
```diff
@@ -116,11 +121,10 @@ def forward(self, loss: torch.Tensor, tensors: alf.nest.NestedTensor):
...
                 the samples used for computing these losses should be sampled *with*
                 replacement independently. The loss can have a shape of either
-                ``[T,B]`` or ``[B]``. In the first case, we will only take
-                ``[0,B]`` for computing the gns due to sample correlation along
-                the temporal axis. On-policy algorithms might have noisy estimates
-                if the ``num_parallel_environments`` is small. Similarly, this
-                happens for a small batch size for off-policy algorithms.
+                ``[T,B]`` or ``[B]``. In the first case, we will randomly select
+                a time step ``t``. The reason is that if the ``T`` steps' gradients
+                are consider together, it will violate the i.i.d. assumption of
```
emailweixu: Treating each sequence as a sample, the samples of a batch are still IID.
And using a single t does not work for RNN.

hnyu: > Treating each sequence as a sample, the samples of a batch are still IID. And using a single t does not work for RNN.

I think as long as the subsequence is randomly sampled with a random start position, using a single t still works (produces meaning statistics for estimating GNS). But so far it turns out neither is very stable. 

Note that each loss is already computed before taking the gradients, so it has considered the temporal dependency. Taking one t won't change this.

hnyu: > Treating each sequence as a sample, the samples of a batch are still IID. And using a single t does not work for RNN.

In fact, there are multiple losses computed along a sequence, each loss for a time step. Even though sequences are iid with each other, the losses are not from iid samples?

hnyu:
Further experiments show that `sac_cart_pole_conf.py` is a very special case. Probably related to its behavior of directly getting action distributions from Q values, which makes the gradient of its critic networks unstable (not sure how to address this issue yet). For other continuous action cases, SAC is fine with this GNS estimator.

image

The last "skills learning" task also uses SAC.

emailweixu (approved):

PR #1356, opened 2022-06-22
add advanced Fetch envs

hnyu:
Improve the existing four Fetch envs by adding an extra 3-dim gripper rotation action, representing the 'xyz' Euler angle rotated based on the previous time step. The rotation action is in [-1,1], and will be multiplied by pi*0.05 before being input to the simulator, where 0.05 is a step scaling factor. 

The 4 new Fetch envs have an action dim of 7. (0-2: dxyz, 3-5: euler_xyz, 6: gripper_open)

To use the new environments, do 

```python
import alf.environments.adv_fetch_envs
import gym
env = gym.make("FetchSlideAdv-0")
```

"FetchSlideAdv-v0"
ezgif com-gif-maker
"FetchPushAdv-v0"
ezgif com-gif-maker
"FetchReachAdv-v0"
ezgif com-gif-maker
"FetchPickAndPlaceAdv-v0"
ezgif com-gif-maker

### DIFF
--- a/alf/environments/adv_fetch_envs/adv_fetch_env.py
+++ b/alf/environments/adv_fetch_envs/adv_fetch_env.py
@@ -1,0 +1,98 @@
+# Copyright (c) 2022 Horizon Robotics and ALF Contributors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import numpy as np
+
+from gym.envs.robotics.fetch_env import FetchEnv
+from gym.envs.robotics import robot_env, rotations, utils
+
+
+class AdvFetchEnv(FetchEnv):
+    def __init__(
+            self,
+            model_path,
+            n_substeps,
+            gripper_extra_height,
+            block_gripper,
+            has_object,
+            target_in_the_air,
+            target_offset,
+            obj_range,
+            target_range,
+            distance_threshold,
+            initial_qpos,
+            reward_type,
+    ):
+        """Initializes a new Fetch environment. Almost the same with ``FetchEnv``.
+        # The only change is from 4 to 7 for ``n_actions`` where the extra 3 dims
+        represent 'xyz' euler angles rotated compared to the previous step.
+
+        Args:
+            model_path (string): path to the environments XML file
+            n_substeps (int): number of substeps the simulation runs on every call to step
+            gripper_extra_height (float): additional height above the table when positioning the gripper
+            block_gripper (boolean): whether or not the gripper is blocked (i.e. not movable) or not
+            has_object (boolean): whether or not the environment has an object
+            target_in_the_air (boolean): whether or not the target should be in the air above the table or on the table surface
+            target_offset (float or array with 3 elements): offset of the target
+            obj_range (float): range of a uniform distribution for sampling initial object positions
+            target_range (float): range of a uniform distribution for sampling a target
+            distance_threshold (float): the threshold after which a goal is considered achieved
+            initial_qpos (dict): a dictionary of joint names and values that define the initial configuration
+            reward_type ('sparse' or 'dense'): the reward type, i.e. sparse or dense
+        """
+        self.gripper_extra_height = gripper_extra_height
+        self.block_gripper = block_gripper
+        self.has_object = has_object
+        self.target_in_the_air = target_in_the_air
+        self.target_offset = target_offset
+        self.obj_range = obj_range
+        self.target_range = target_range
+        self.distance_threshold = distance_threshold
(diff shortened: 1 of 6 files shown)


le-horizon:
Nice additions @hnyu .  A couple high level suggestions:
1) Now the robot seems to be able to bump into itself or damage the table, given the additional freedom.  It would be helpful to return some information such as violations (hitting itself or the table too hard).  Maybe add a TODO?
2) Now the advanced reach goal dimensions could include not just the x-y position, but also the required rotation of the hand, same could be done for push or pickandplace (but would take a bit extra effort for it to make sense for slide).
Without adding these additional target dimensions into the goal state, the task setup seems to be introducing unnecessary complexities: The robot doesn't need x-y-z rotations to finish the tasks.  Maybe add a TODO as well?

hnyu:
> Nice additions @hnyu . A couple high level suggestions:
> 
> 1. Now the robot seems to be able to bump into itself or damage the table, given the additional freedom.  It would be helpful to return some information such as violations (hitting itself or the table too hard).  Maybe add a TODO?
> 2. Now the advanced reach goal dimensions could include not just the x-y position, but also the required rotation of the hand, same could be done for push or pickandplace (but would take a bit extra effort for it to make sense for slide).
>    Without adding these additional target dimensions into the goal state, the task setup seems to be introducing unnecessary complexities: The robot doesn't need x-y-z rotations to finish the tasks.  Maybe add a TODO as well?

Added comments in the docstring. Note currently even without rotations the gripper could still collide with the table and itself. Regarding 2, the goal of this PR is to provide the ability of rotating but not defining new tasks. The new goal definition will largely vary depending on the specific task (not just xyz any more), so should be left to users.

The current xyz goal with advanced gripper can serve as a starting point for testing an RL policy. If with 7-dim actions the policy can't learned to achieve xyz, then it will be impossible to achieve more complex goals.

alf/environments/adv_fetch_envs/adv_fetch_env.py
```diff
@@ -0,0 +1,118 @@
...
+            After adding the rotation angles, the robot gripper might more easily
+            have collisions with the table or itself. Maybe in the future we should
+            have the environment return collision info.
+
+            You can overwrite the function `_sample_goal()` to define new goals
+            given these extra rotation actions, also `_get_obs()` to redefine the
+            achieved goal.
+
+        Args:
+            model_path (string): path to the environments XML file
```
emailweixu: Use type hint instead.

alf/environments/adv_fetch_envs/adv_fetch_env.py
```diff
@@ -0,0 +1,118 @@
...
+            n_substeps (int): number of substeps the simulation runs on every call to step
+            gripper_extra_height (float): additional height above the table when positioning the gripper
+            block_gripper (boolean): whether or not the gripper is blocked (i.e. not movable) or not
+            has_object (boolean): whether or not the environment has an object
+            target_in_the_air (boolean): whether or not the target should be in the air above the table or on the table surface
+            target_offset (float or array with 3 elements): offset of the target
```
emailweixu: Not clear the meaning of this argument. What does a float mean? And what does an array mean?

alf/environments/adv_fetch_envs/adv_fetch_env.py
```diff
@@ -0,0 +1,118 @@
...
+        self.has_object = has_object
+        self.target_in_the_air = target_in_the_air
+        self.target_offset = target_offset
+        self.obj_range = obj_range
+        self.target_range = target_range
+        self.distance_threshold = distance_threshold
+        self.reward_type = reward_type
+        self._quat = None
+
+        robot_env.RobotEnv.__init__(
```
emailweixu, line 81: `super().__init__(...)`

alf/environments/adv_fetch_envs/adv_fetch_env.py
```diff
@@ -0,0 +1,118 @@
...
+            target_offset,
+            obj_range,
+            target_range,
+            distance_threshold,
+            initial_qpos,
+            reward_type,
+    ):
+        """Initializes a new Fetch environment. Almost the same with ``FetchEnv``.
+        # The only change is from 4 to 7 for ``n_actions`` where the extra 3 dims
+        represent 'xyz' euler angles rotated compared to the previous step.
```
emailweixu, line 41: document the meaning of each dimension of the action. It seems that 'xyz' euler angles are not the last 3 dimensions.

alf/environments/adv_fetch_envs/adv_fetch_env.py
hnyu: > Not clear the meaning of this argument. What does a float mean? And what does an array mean?

This class init function is basically copied from the OpenAI FetchEnv. I've added comments.

alf/environments/adv_fetch_envs/adv_fetch_env.py
hnyu: > `super().__init__(...)`

I actually need to call its grandparent's init.

hnyu: > `super().__init__(...)`

I actually need to call its grandparent's init.

alf/environments/adv_fetch_envs/adv_fetch_env.py
hnyu: > document the meaning of each dimension of the action. It seems that 'xyz' euler angles are not the last 3 dimensions.

added

emailweixu (approved):

PR #1358, opened 2022-06-29
Fix fractional unroll_length

emailweixu:
There are two bugs for fractional_unroll:
1. replay_buffer is None for train_from_replay_buffer
2. missing summary for unroll under certain condition

Also support interactionless training at the end. The main purpose of this is to train more on the last environmental steps.


le-horizon (changes requested):
Almost good.

Need to update `common.compute_summary_or_eval_interval` which still relies on the two termination conditions being exclusive.

Thanks,
Le

emailweixu:
> Almost good.
> 
> Need to update `common.compute_summary_or_eval_interval` which still relies on the two termination conditions being exclusive.
> 
> Thanks, Le

Good catch. Fixed.

le-horizon (dismissed):
LG thanks for the fixes, Wei.

alf/trainers/policy_trainer.py
```diff
@@ -606,7 +613,7 @@ def _train(self):
             self._trainer_progress.update(iter_num, total_time_steps)
 
             if ((self._num_iterations and iter_num >= self._num_iterations)
-                    or (self._num_env_steps
+                    or (not self._num_iterations and self._num_env_steps
```
hnyu: no need to `and self._num_env_steps` because it has to be >0 in this case.

hnyu (approved):

PR #1359, opened 2022-06-29
Dynamic Object Sensor

Haichao-Zhang:
Implementation of a dynamic object sensor for Carla, as show in the top-right of the demo below.
This sensor is useful for providing structured perceptions for experiments and visualizations.

https://user-images.githubusercontent.com/21375027/176559697-756393b2-f134-4424-92c5-0bf08215530d.mov

### DIFF
--- a/alf/environments/carla_sensors.py
+++ b/alf/environments/carla_sensors.py
@@ -19,6 +19,11 @@ import cv2
 from enum import IntEnum, auto, Enum
 import math
 import numpy as np
+import matplotlib
+matplotlib.use('Agg')  # 'Agg' no need for xserver!
+import matplotlib.pyplot as plt
+# Style gallery: https://tonysyu.github.io/raw_content/matplotlib-style-gallery/gallery.html
+plt.style.use('seaborn-dark')
 
 import weakref
 import threading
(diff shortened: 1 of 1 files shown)


alf/environments/carla_sensors.py
```diff
@@ -2152,3 +2157,402 @@ def get_current_observation(self, current_frame):
...
+                 history_idx=[-16, -11, -6, -1],
+                 max_history_len=20,
+                 object_filter='vehicle.*',
+                 max_object_number=3,
+                 distance_threshold=100,
+                 with_ego_hist=True,
+                 fill_value=None):
+        """
+        Args:
+            parent_actor (carla.Actor): the parent actor of this sensor
```
breakds, line 2183: Nit: could use type hints instead to document the type, e.g. `parent_actor: carla.Actor` in the argument list.

alf/environments/carla_sensors.py
```diff
@@ -2152,3 +2157,402 @@ def get_current_observation(self, current_frame):
...
+            alf_world (World): the world object keeping all relevant data and
+                some utility functions (e.g., `_get_traffic_light_waypoints`)
+            navigation_sensor (str): the navigation sensor associated with the
+                `parent_actor`
+            history_idx (list[int]): a list of numbers representing the indices
```
breakds, line 2188: Really like this interface design of using `history_idx` list. Question: can `max_history_len` be inferred from this?

alf/environments/carla_sensors.py
```diff
@@ -2152,3 +2157,402 @@ def get_current_observation(self, current_frame):
...
+                of the history information to be rendered for non-ego vehicles.
+                For example, we can set `history_idx=[-1]` for keep only the
+                most recent observation or `history_idx=[-11, -1]` for both the
+                lastest and also the one 10 steps earlier.
+            max_history_len (int): max number of history length preserved
+            object_filter (str): a string representing the type of dynamic
+                objects to be perceived. By default, surrounding vehicles are
+                perceived.
+            max_object_number (int): including ego vehicle
```
breakds: Also explain the behavior when the number of objects exceeds this threshold?

alf/environments/carla_sensors.py
```diff
@@ -2152,3 +2157,402 @@ def get_current_observation(self, current_frame):
...
+                (if with_ego_hist is True); otherwise, all are ado vehicles
+            with_ego_hist (bool): whether inlcude ego history
```
breakds: `hist` can be an ambiguous shorthand, also `max_history_len` did not use the abbreviation. For clarity and consistency, shall we use `with_ego_history` instead?

alf/environments/carla_sensors.py
```diff
@@ -2152,3 +2157,402 @@ def get_current_observation(self, current_frame):
...
+                some utility functions (e.g., `_get_traffic_light_waypoints`)
+            navigation_sensor (str): the navigation sensor associated with the
+                `parent_actor`
+            history_idx (list[int]): a list of numbers representing the indices
+                of the history information to be rendered for non-ego vehicles.
+                For example, we can set `history_idx=[-1]` for keep only the
+                most recent observation or `history_idx=[-11, -1]` for both the
+                lastest and also the one 10 steps earlier.
+            max_history_len (int): max number of history length preserved
+            object_filter (str): a string representing the type of dynamic
```
breakds, line 2193: Maybe give a few examples (or link to carla documentation) about how to set `object_filter`?

alf/environments/carla_sensors.py
```diff
@@ -2152,3 +2157,402 @@ def get_current_observation(self, current_frame):
...
+    """
+
+    def __init__(self,
+                 parent_actor,
+                 alf_world,
+                 history_idx=[-16, -11, -6, -1],
+                 max_history_len=20,
+                 object_filter='vehicle.*',
+                 max_object_number=3,
+                 distance_threshold=100,
```
breakds: Not present in the docstring's `Args`.

alf/environments/carla_sensors.py
```diff
@@ -2152,3 +2157,402 @@ def get_current_observation(self, current_frame):
...
+        self._with_ego_hist = with_ego_hist
+
+        self._object_filter = object_filter
+        self._max_object_numberr = max_object_number
+
+        # dim3 = 1d "presence" + 2d [x, y]
+        self._vehicle_fea_dim = 3
+        self._distance_threshold = distance_threshold
+
+        self._alf_world = alf_world
```
breakds: It appears `self._alf_world` is not used. Maybe just keeping `self._world` is enough (and less confusing)?

alf/environments/carla_sensors.py
```diff
@@ -2152,3 +2157,402 @@ def get_current_observation(self, current_frame):
...
+    def __init__(self,
+                 parent_actor,
+                 alf_world,
+                 history_idx=[-16, -11, -6, -1],
+                 max_history_len=20,
+                 object_filter='vehicle.*',
+                 max_object_number=3,
+                 distance_threshold=100,
+                 with_ego_hist=True,
+                 fill_value=None):
```
breakds: Ditto. Not present in docstring.

alf/environments/carla_sensors.py
```diff
@@ -2152,3 +2157,402 @@ def get_current_observation(self, current_frame):
...
+                "max number of perceivable objects {}. ".format(
+                    len(vehicles), K))
+
+            idx = np.argpartition(distances, K - 1)
+            vehicles = [vehicles[i] for i in idx[:K]]
+            distances = [distances[i] for i in idx[:K]]
+
+        sorted_idx = np.argsort(distances)
+        vehicles = [vehicles[i] for i in sorted_idx]
+        distances = [distances[i] for i in sorted_idx]
```
breakds: Maybe we do not have to duplicate the logic on `sort` and `argpartition`? 
Also in the exceeding threshold case, there will be two sorts performed. Suggest just using `argpartion`, and `len(vehicles) > K` check can be used just for the warning.

alf/environments/carla_sensors.py
```diff
@@ -2152,3 +2157,402 @@ def get_current_observation(self, current_frame):
...
+        full_feature_mat[:, :v_fea_mat.shape[-3], :v_fea_mat.
+                         shape[-2], :v_fea_mat.shape[-1]] = v_fea_mat
+
+        self._current_obs = full_feature_mat.astype(np.float32)
+        return self._current_obs
+
+    @staticmethod
+    def _get_surrounding_actors(bbox_list,
+                                actor_id_list,
+                                criterium,
```
breakds, line 2446: Nit: `predicate`/`predicator` is more often used for this.

hnyu:
What is the additional information that this provides compared to the BEV representation?

Haichao-Zhang:
> What is the additional information that this provides compared to the BEV representation?

This is a structured representation of the dynamic object only.  It is therefore flexible to be combined with other existing sensors in Carla.
The current BEV sensor is a rasterized version of the scene, therefore the range of view field cannot be too large due to the memory limit.

alf/environments/carla_sensors.py
Haichao-Zhang: Good point. Removed ``max_history_len`` since it can be inferred.

alf/environments/carla_sensors.py
Haichao-Zhang: Explanation added.

alf/environments/carla_sensors.py
Haichao-Zhang: Changed to  ``with_ego_history``.

alf/environments/carla_sensors.py
Haichao-Zhang: Added

alf/environments/carla_sensors.py
Haichao-Zhang: Change the argument name to ``view_radius`` and added comments.

alf/environments/carla_sensors.py
Haichao-Zhang: Removed ``self._alf_world``

alf/environments/carla_sensors.py
Haichao-Zhang: Removed ``fill_value`` since it is unnecessary.

alf/environments/carla_sensors.py
Haichao-Zhang: Good point. Now only keeping ``argpartition`` part.

breakds (approved):

PR #1362, opened 2022-08-02
Fix the documentation typo in MCTS exploration policy doc

breakds:
A straight forward typo fix. Also added a pair of brackets for readability.
Will need to regenerate the doc website after check-in.


breakds:
Sigh my local yapf version are too high which causes a lot of unnecessary formatting. Let me fix it.

emailweixu (approved):

PR #1363, opened 2022-08-04
FrameCrop Wrapper

Haichao-Zhang:
Add a FrameCrop wrapper which can be handy when used to only retrieve the pixel contents in an ROI of the original rendered image from the environment, as illustrated in the example below:

image

Similar purpose can possibly be achieved by modifying the rendering viewpoint for some environments; while for others it might be hard to do so.

### DIFF
--- a/alf/environments/gym_wrappers.py
+++ b/alf/environments/gym_wrappers.py
@@ -332,6 +332,53 @@ class FrameResize(BaseObservationWrapper):
 
 
 @alf.configurable
+class FrameCrop(BaseObservationWrapper):
+    def __init__(self, env, sx=0, sy=0, width=84, height=84, fields=None):
+        """Create a FrameCrop instance
+
+        Args:
+             env (gym.Env): the gym environment
+             sx (int): start position along the horizonal direction (x-axis)
+             sy (int): start position along the vertical direction (y-axis)
+             width (int): crop width
+             height (int): crop height
+             fields (list[str]):  fields to be resize, A field str is a multi-level
+                path denoted by "A.B.C". If None, then non-nested observation is resized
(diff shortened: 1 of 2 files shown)


alf/environments/gym_wrappers.py
```diff
@@ -331,6 +331,53 @@ def transform_observation(self, observation):
...
+            "Crop is out of boundary "
+            "along the vertical direction")
+        return gym.spaces.Box(
+            low=0,
+            high=255,
+            shape=[self._width, self._height] + list(obs_shape[2:]),
+            dtype=np.uint8)
+
+    def transform_observation(self, observation):
+        obs = observation[self._sy:self._sy + self._height, self._sx:self._sx +
```
breakds: To make it slightly more flexible, maybe we can add an option to `__init__` to indicate whether channel comes the first or the last? There might be cases when the channel is the first dimension.

alf/environments/gym_wrappers.py
```diff
@@ -331,6 +331,53 @@ def transform_observation(self, observation):
         return obs
 
 
+@alf.configurable
+class FrameCrop(BaseObservationWrapper):
```
breakds, line 335: An unit test can be helpful?

alf/environments/gym_wrappers.py
Haichao-Zhang: Good point. Added.

alf/environments/gym_wrappers.py
Haichao-Zhang: Unit tests added.

breakds (approved):
Thanks!

PR #1365, opened 2022-08-05
Behavior Cloning algorithm

Haichao-Zhang:
Implementation of Behavior Cloning Algorithm together with a complete trainable example.


alf/algorithms/bc_algorithm.py
```diff
@@ -0,0 +1,159 @@
...
+BcState = namedtuple("BcState", ["actor"], default_value=())
+
+BcInfo = namedtuple("BcInfo", ["actor"], default_value=())
+
+BcLossInfo = namedtuple("LossInfo", ["actor"], default_value=())
+
+
+@alf.configurable
+class BcAlgorithm(OffPolicyAlgorithm):
+    r"""Behavior cloning algorithm.
```
emailweixu, line 34: Perhaps give a brief description of the method if it is not very complicated?

alf/algorithms/rl_algorithm.py
```diff
@@ -636,10 +636,6 @@ def _train_iter_off_policy(self):
                         self.summarize_metrics()
                     self._need_to_summarize_rollout = False
 
-        # replay buffer may not have been created yet because unroll is nor performed yet.
```
emailweixu, line 639: For online RL training, this condition can happen when unroll_length<1. It needs to be handled correctly.

alf/algorithms/bc_algorithm.py
```diff
@@ -0,0 +1,159 @@
...
+
+        return actor_info
+
+    def train_step_offline(self,
+                           inputs: TimeStep,
+                           state,
+                           rollout_info,
+                           pre_train=False):
+
+        self._pre_train = pre_train
```
emailweixu: why do you need to assign self._pre_train? It doesn't seem to be used anywhere.

alf/examples/hybrid_rl/bc_pendulum_conf.py
```diff
@@ -0,0 +1,84 @@
...
+actor_network_cls = partial(
+    alf.networks.ActorDistributionNetwork,
+    fc_layer_params=fc_layers_params,
+    activation=activation,
+    continuous_projection_net_ctor=proj_net)
+
+use_action_delta = False
+alf.config('BcAlgorithm', actor_network_cls=actor_network_cls)
+
+num_iterations = 10000000
```
emailweixu: This is 10M iterations. But bc_pendulum_loss.png only shows 100K iterations.

alf/examples/hybrid_rl/bc_pendulum_conf.py
```diff
@@ -0,0 +1,84 @@
...
+
+# training config
+alf.config(
+    "TrainerConfig",
+    initial_collect_steps=0,
+    num_updates_per_train_iter=1,
+    num_iterations=num_iterations,
+    rl_train_after_update_steps=num_iterations + 1000,
```
emailweixu, line 75: does this mean that there is no RL training? Perhaps adding some comment to make this clear.

emailweixu (changes requested):

alf/algorithms/bc_algorithm.py
Haichao-Zhang: Thanks for the comment. Now added description.

alf/algorithms/rl_algorithm.py
Haichao-Zhang: Fixed this issue.

alf/algorithms/bc_algorithm.py
Haichao-Zhang: This line is unused and now removed.

emailweixu (approved):

alf/examples/hybrid_rl/bc_pendulum_conf.py
Haichao-Zhang: The curves were generated by running this conf file. 
bc_pendulum_loss.png and bc_pendulum_return.png are the curves from the first 100K iterations, which is enough in this case.  The full curves are much longer than 100K but it seems the additional iterations are unnecessary.
Now reduced the num_iterations = 100K in conf since 10M is overly large.

alf/examples/hybrid_rl/bc_pendulum_conf.py
Haichao-Zhang: Yes. Comments added.

PR #1367, opened 2022-08-07
add image decoding network with upsampling+conv

hnyu:
Added another implementation of image decoding network which uses upsampling and conv layers. This implementation can avoid checkerboard artifacts as commonly seen in decoders with transposed conv layers. https://distill.pub/2016/deconv-checkerboard

In short, for transposed conv layers, depending on the kernel size and stride, some locations in the next layer might have field overlaps from several locations in the previous layer. This overlap results in a greater default magnitude. Sometimes it's difficult for the network learns to remove this bias.

### DIFF
--- a/alf/networks/encoding_networks.py
+++ b/alf/networks/encoding_networks.py
@@ -14,7 +14,7 @@
 
 import functools
 import numpy as np
-from typing import Callable, Optional, Tuple
+from typing import Callable, Optional, Tuple, Union
 
 import torch
 import torch.nn as nn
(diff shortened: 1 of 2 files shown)


alf/networks/encoding_networks.py
```diff
@@ -232,6 +232,129 @@ def __init__(self,
...
+            start_decoding_channels: the initial number of channels we'd
+                like to have for the feature map. Note that we always first
+                project an input latent vector into a vector of an appropriate
+                length so that it can be reshaped into (``start_decoding_channels``,
+                ``start_decoding_height``, ``start_decoding_width``).
+            upsampling_mode: the argument for choosing an upsampling algorithm
+                for ``torch.nn.Upsample``.
+            same_padding: similar to TF's conv2d ``same`` padding mode. If
+                True, the user provided paddings in ``transconv_layer_params`` will
+                be replaced by automatically calculated ones; if False, it
```
emailweixu, line 297: It's better to simply describe how the padding is calculated for same_padding=True and False

hnyu: > It's better to simply describe how the padding is calculated for same_padding=True and False

``ImageEncodingNetwork`` already has a paragraph explaining the two padding modes. I've added a sentence referring to its docstring.

emailweixu (approved):

PR #1368, opened 2022-08-11
action_sampler and KL exploration policy for MCTSAlgorithm

emailweixu:
Refactor MCTSAlgorithm to allow more flexibilities of sampling actions from the policy found by MCTS. The newly added ActionSampler can maintain internal state to allow temporally correlated sampling, which is helpful for certain tasks.

Also added a different way of calculating policy based on KL divergence.

### DIFF
--- a/alf/algorithms/mcts_algorithm.py
+++ b/alf/algorithms/mcts_algorithm.py
@@ -27,7 +27,7 @@ from alf.algorithms.off_policy_algorithm import OffPolicyAlgorithm
 from alf.utils import dist_utils, summary_utils
 from alf import nest
 from alf.trainers.policy_trainer import Trainer
-from alf.utils import common, tensor_utils
+from alf.utils import common, tensor_utils, action_samplers
 from .mcts_models import MCTSModel, ModelOutput
 
 MAXIMUM_FLOAT_VALUE = float('inf')
@@ -218,7 +218,9 @@ class _MCTSTrees(object):
         return dot
 
 
-MCTSState = namedtuple("MCTSState", ["steps", "pred_state"], default_value=())
+MCTSState = namedtuple(
+    "MCTSState", ["steps", "pred_state", "action_sampler_state"],
+    default_value=())
 MCTSInfo = namedtuple(
     "MCTSInfo", ["candidate_actions", "value", "candidate_action_policy"])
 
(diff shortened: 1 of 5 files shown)


alf/algorithms/mcts_algorithm.py
```diff
@@ -573,28 +596,32 @@ def rollout_step(self, time_step: TimeStep, state: MCTSState):
         else:
             self._build_tree2(trees, to_plays)
 
-        action, info = self._select_action(trees, steps)
+        action_dist, info = self._select_action(trees, steps)
```
breakds: Suggestoins: call it `action_probs` instead of `action_dist` as it seems to me that the convention here is `action_dist` = an actual distribution object, while `action_probs` = a tensor of probabilities.

breakds (dismissed):

alf/algorithms/mcts_algorithm.py
```diff
@@ -1423,6 +1447,53 @@ def calculate_exploration_policy(value, prior, c, tol=1e-6):
...
+
+        p_i = \frac{q_i \exp(v_i/c)}{Z}
+
+    where :math:`Z` is such that :math:`\sum_i p_i = 1`
+
+    Args:
+        value (Tensor): [N, K] Tensor
+        prior (Tensor): [N, K] Tensor
+        c (Tensor): [N, 1] Tensor
+        tol (float): Desired acurracy. The result satisfy :math:`|\sum_i p_i - 1| \le tol`
```
Haichao-Zhang: ```tol``` can be removed?

alf/utils/action_samplers_test.py
```diff
@@ -0,0 +1,56 @@
...
+# limitations under the License.
+
+import torch
+import alf
+from alf.utils.common import zero_tensor_from_nested_spec
+from alf.utils.action_samplers import _CategoricalSeedSamplerBase
+
+
+class ActionSamplersTest(alf.test.TestCase):
+    def test_categorical_seed_sampler(self):
```
Haichao-Zhang, line 22: Need to add this file to ```/.ci-cd/build.sh```?

alf/algorithms/mcts_algorithm.py
emailweixu: removed

alf/algorithms/mcts_algorithm.py
emailweixu: changed.

alf/utils/action_samplers_test.py
emailweixu: added

Haichao-Zhang (approved):

PR #1372, opened 2022-08-12
add discrete VAE

hnyu:
Add a VAE with discrete latent embedding. For gradient backprop, we use the straight-through estimator or straight-through Gumbel-softmax estimator (lower-variance).

It can be difficult for the discrete latent embedding to encode precisions in the inputs, unless the number of categorical variables is large. However, this kind of embedding might be helpful for MI computation between the latent embedding space and some other variable because we can control its granularity. For example, input x and y are encoded into the same set of discrete variables unless they are very different.

### DIFF
--- a/alf/algorithms/vae.py
+++ b/alf/algorithms/vae.py
@@ -239,10 +239,10 @@ class DiscreteVAE(VariationalAutoEncoder):
                  name: str = "DiscreteVAE"):
         """
         Args:
-            z_spec: a tensor spec for the discrete posterior. Regardless of its
-                rank, it will be converted to rank-1, representing a vector of
-                discrete variables. The value bould of each variable must be
-                identical.
+            z_spec: a tensor spec for the discrete posterior. It has to be
+                rank-one, representing a vector of discrete variables.
+                The value bould of each variable must be identical and the lower
+                bound has to be 0.
             input_tensor_spec: the input spec.
             z_network_cls: an encoding network to encode input data into a vector
                 of logits. If ``prior_z_network_cls`` is None, this network must
@@ -263,15 +263,9 @@ class DiscreteVAE(VariationalAutoEncoder):
         """
         Algorithm.__init__(self, name=name)
 
-        assert z_spec.is_discrete
-        self._n_categories = int(z_spec.maximum - z_spec.minimum + 1)
-
-        # convert z_spec to rank-1
-        z_spec = BoundedTensorSpec(
-            shape=(z_spec.numel, ),
-            minimum=0,
-            maximum=self._n_categories - 1,
-            dtype=z_spec.dtype)
+        assert (z_spec.is_discrete and z_spec.ndim == 1
+                and z_spec.minimum == 0)
+        self._n_categories = int(z_spec.maximum + 1)
 
         prior_z_network = None
         if prior_z_network_cls is not None:
@@ -302,7 +296,7 @@ class DiscreteVAE(VariationalAutoEncoder):
 
     @property
     def output_spec(self):
-        """Because the output is a floating one-hot vector, the shape is rank-2.
+        """Because the output is a floating one-hot vector, the shape is rank-two.
         """
         return BoundedTensorSpec(
             shape=self._z_spec.shape + (self._n_categories, ),
@@ -315,9 +309,9 @@ class DiscreteVAE(VariationalAutoEncoder):
             logits2 = torch.zeros_like(logits1)  # assume uniform
         logits1 = torch.nn.functional.log_softmax(logits1, dim=-1)
         logits2 = torch.nn.functional.log_softmax(logits2, dim=-1)
-        kld_loss = torch.nn.KLDivLoss(reduction='none', log_target=True)
         # The expectation is over the target distribution
-        kld = kld_loss(input=logits2, target=logits1)
+        kld = torch.nn.functional.kl_div(
+            input=logits2, target=logits1, reduction='none', log_target=True)
         return kld.sum(dim=(1, 2))  # [B,L,K] -> [B]
 
     def _sampling_forward(self, inputs):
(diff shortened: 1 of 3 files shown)


alf/algorithms/vae.py
```diff
@@ -179,3 +184,143 @@ def train_step(self, inputs, state=()):
...
+class DiscreteVAE(VariationalAutoEncoder):
+    r"""VAE with a discrete posterior distribution. The latent ``z`` might be
+    a single categorical variable or a vector of categorials. Because the
+    re-parameterization trick can no longer be applied to the discrete distribution,
+    we instead use the straight-through gradient estimator to train the encoder.
+
+    ::
+
+        Bengio et al., "Estimating or Propagating Gradients Through Stochastic
+        Neurons for Conditional Computation", 2013.
```
emailweixu, line 200: You may consider Gumbel softmax instead of straight through estimator, which is shown to perform better in https://arxiv.org/pdf/1611.01144.pdf

alf/algorithms/vae.py
```diff
@@ -179,3 +184,143 @@ def train_step(self, inputs, state=()):
...
+            z_logits = prior_z_logits + dist_utils.distributions_to_params(
+                z_dist)['logits']
+            z_dist_spec = dist_utils.extract_spec(z_dist)
+            z_dist = dist_utils.params_to_distributions({
+                'logits': z_logits
+            }, z_dist_spec)
+        else:
+            uniform_prob = 1. / self._n_categories
+            entropy = dist_utils.compute_entropy(z_dist)
+            kl_div_loss = -np.log(uniform_prob) * self._z_spec.numel - entropy
```
emailweixu: should not multiply with `self._z_spec.numel`

alf/algorithms/vae.py
```diff
@@ -179,3 +184,143 @@ def train_step(self, inputs, state=()):
...
+        assert z_spec.is_discrete
+
+        prior_z_network = None
+        if prior_encoder_cls is not None:
+            prior_encoder = prior_encoder_cls(
+                input_tensor_spec=prior_input_tensor_spec)
+            prior_proj_net = OnehotCategoricalProjectionNetwork(
+                input_size=prior_encoder.output_spec.numel, action_spec=z_spec)
+            prior_z_network = alf.nn.Sequential(prior_encoder, prior_proj_net)
+            logits_spec = TensorSpec((np.prod(prior_proj_net._output_shape), ))
```
emailweixu: do you think the code will be cleaner if we simply require that encoder and prior_encoder output the logits directly? And it will be quite easy to support Gumbel softmax.

hnyu: Gumbel softmax seems not what I need. For discrete VAEs we would expect the sampled latent code to be completely discrete (hard labels). Gumbel softmax will still produce soft labels due to using softmax (with a temperature) to replace argmax. 

image

emailweixu: As in the paper, it can also be combined with the straight through trick

hnyu: > As in the paper, it can also be combined with the straight through trick

I'm actually a bit confused about combining Gumbel softmax and straight-through. The Gumbel trick (with argmax) is basically equivalent to Categorical sampling. If eventually we'll use straight-through on the one-hot, why bother using Gumbel trick to replace the categorical sampling in the first place? 

image

hnyu: We could in fact support Gumbel softmax for the latent code, but comment on it to be softly sparse (depending on the temperature) instead of hard onehot. For hard latent code, still use the current onehot straight-through, as Gumbel softmax+straight-through seems unnecessary to me.

hnyu: never mind, I think I understand now that there is still some difference between the two and the paper claims ST gumbel softmax to have lower variance than plain ST.

alf/algorithms/vae.py
hnyu: > should not multiply with `self._z_spec.numel`

self._z_spec.numel is the number of categoricals. The total KLD should be summed over them. `-np.log(uniform_prob)` is the entropy of one categorical.

alf/algorithms/vae.py
```diff
@@ -179,3 +185,179 @@ def train_step(self, inputs, state=()):
...
+                 mode: str = "st",
+                 gumbel_temp_scheduler: Scheduler = ConstantScheduler(1.),
+                 beta: float = 1.,
+                 target_kld_per_categorical: float = None,
+                 beta_optimizer: torch.optim.Optimizer = None,
+                 name: str = "DiscreteVAE"):
+        """
+        Args:
+            z_spec: a tensor spec for the discrete posterior. Regardless of its
+                rank, it will be converted to rank-1, representing a vector of
```
emailweixu: rank-1 is ambiguous. It can be "rank minus one" or "rank one"

alf/algorithms/vae.py
```diff
@@ -179,3 +185,179 @@ def train_step(self, inputs, state=()):
...
+            beta: the weight for KL-divergence
+            target_kld_per_categorical: if not None, then this will be used as the
+                target KLD *per Categorical* to automatically tune beta.
+            beta_optimizer: if not None, will be used to train beta.
+            name (str):
+        """
+        Algorithm.__init__(self, name=name)
+
+        assert z_spec.is_discrete
+        self._n_categories = int(z_spec.maximum - z_spec.minimum + 1)
```
emailweixu: Since the output z is always a vector and bounded in [0, n_gategories), to avoid misuse, we can simply assert z_spec.ndim==1 and z_spec.minimum==0.

alf/algorithms/vae.py
```diff
@@ -179,3 +185,179 @@ def train_step(self, inputs, state=()):
...
+            dtype=torch.float32)
+
+    def _kl_divergence(self, logits1, logits2=None):
+        if logits2 is None:
+            logits2 = torch.zeros_like(logits1)  # assume uniform
+        logits1 = torch.nn.functional.log_softmax(logits1, dim=-1)
+        logits2 = torch.nn.functional.log_softmax(logits2, dim=-1)
+        kld_loss = torch.nn.KLDivLoss(reduction='none', log_target=True)
+        # The expectation is over the target distribution
+        kld = kld_loss(input=logits2, target=logits1)
```
emailweixu: can use nn.functional.kl_div() directly.

emailweixu (approved):

PR #1373, opened 2022-08-12
Cyclical Scheduler

Haichao-Zhang:
A cyclical scheduler that is useful for:
1) making the training better behaved as proposed in https://arxiv.org/pdf/1506.01186.pdf
2) for updating different modules at different paces, e.g. for TD3 and Dreamer.

### DIFF
--- a/alf/utils/schedulers.py
+++ b/alf/utils/schedulers.py
@@ -194,6 +194,93 @@ class ExponentialScheduler(Scheduler):
             self._decay_time)
 
 
+@alf.configurable
+class CyclicalScheduler(Scheduler):
+    """The cyclical scheduler where the value changes cyclically between two bounds.
+    Reference:
+    ::
+        Leslie N. Smith Cyclical Learning Rates for Training Neural Networks, 2017
+        (https://arxiv.org/pdf/1506.01186.pdf)
+
+    This implementation generalizes the original methods in two ways: 1) the
+    initial value can start from either the lower-bound (as in the original method),
+    or upper bound; 2) apart from the linear switching between the bounds, we
+    also support step mode of switching.
+
+    In terms of applications, beyond the standard case of using a cyclical
+    learning rate to improve the learning behavior during NN training, this
+    scheduler is also useful in other cases. One example is in reinforcement
+    learning, sometimes we want to update the parameters of different modules at
+    difference paces. For example, in TD3, we want to update the policy every other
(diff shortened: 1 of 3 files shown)


alf/utils/schedulers.py
```diff
@@ -194,6 +194,93 @@ def __repr__(self):
...
+
+    def __call__(self):
+        progress = self.progress()
+        progress_in_half_cycle = (
+            progress % self._half_cycle_size) / self._half_cycle_size
+        progress_in_cycle = (progress % self._cycle_size) / self._cycle_size
+
+        if self._switch_mode == "step":
+            # step mode changes value at half-cycle point
+            if progress_in_cycle == 0:
```
emailweixu: progress_in_cycle is a float value if progress_type is "percent". Equality comparison will not be satisfied in general.

emailweixu (changes requested):

alf/utils/schedulers.py
Haichao-Zhang: Now changed to a way that is correct for percent mode. Also added unit tests on percent and env_steps modes.

alf/utils/schedulers.py
```diff
@@ -194,6 +194,95 @@ def __repr__(self):
...
+            "step", "linear"
+        }, ("unsupportted switch mode {}".format(switch_mode))
+        self._switch_mode = switch_mode
+        self._current_value = base_lr
+
+    def __call__(self):
+        progress = self.progress()
+
+        # round to one decimal place to facilitate the stage comparison below
+        progress_in_half_cycle = round(
```
emailweixu: It seems that rounding is not necessary.

Haichao-Zhang: ```round``` is used to handle some cases that could occur  due to numerical reasons when ```progress_type``` is ```percent```.

For example, for a case where ```half_cycle_size=0.1```, when the ```progress=0.6```, both ```progress_in_half_cycle``` and ```progress_in_cycle``` should be 0.  However, due to numerical reasons,  the following results are obtained:

image

Rounding is used to avoid this issue.  For other ```progress_type```, it is unnecessary and introduces an approximation.

Now changed to only apply it when ```progress_type``` is ```percent```.  Also changed to using a larger value for ```ndigits``` when rounding to reduce side-effects.

emailweixu (approved):

PR #1375, opened 2022-08-16
Summary for predicted reward during rollout MuZero

emailweixu:
Since the samples from the replay buffer are trained over and over, the summaries generated during replay may not correctly reflect how well the prediction is. The summary during rollout can better reflect the quality of prediction.

Also change print_tree to only show top actions for policy.


breakds (approved):
Thanks, Wei!

PR #1376, opened 2022-08-16
Handle a corner case of summarize_variables

hnyu:
Sometimes we might have inf variables. For example, the log beta is a variable in VAE because sometimes we want to learn it. In other cases we'd like to set it to -inf for a fixed zero beta. So we need to check variable values before summarizing them with a histogram to avoid

```bash
ValueError: autodetected range of [-inf, -inf] is not finite
```

### DIFF
--- a/alf/utils/summary_utils.py
+++ b/alf/utils/summary_utils.py
@@ -115,7 +115,9 @@ def summarize_variables(name_and_params, with_histogram=True):
     """
     for var_name, var in name_and_params:
         var_values = var
-        if with_histogram:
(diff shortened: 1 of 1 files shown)


emailweixu:
inf variable usually indicates training divergence. It doesn't seem right to silently ignore them.

hnyu:
> inf variable usually indicates training divergence. It doesn't seem right to silently ignore them.

If it diverges, it will be reflected in the gradient histogram summary. But we do have some cases where a variable is infinite (and not trained).

emailweixu (approved):

PR #1378, opened 2022-08-17
convert onehot to float in get_mode()

hnyu:
We would expect a one-hot representation is already float in get_mode(), so that it can directly be input to a network.

### DIFF
--- a/alf/utils/dist_utils.py
+++ b/alf/utils/dist_utils.py
@@ -1068,7 +1068,8 @@ def get_mode(dist):
     elif isinstance(dist, td.OneHotCategorical) or \
                 isinstance(dist, td.OneHotCategoricalStraightThrough):
         mode = torch.nn.functional.one_hot(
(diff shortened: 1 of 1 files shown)


alf/utils/dist_utils.py
```diff
@@ -1068,7 +1068,8 @@ def get_mode(dist):
     elif isinstance(dist, td.OneHotCategorical) or \
                 isinstance(dist, td.OneHotCategoricalStraightThrough):
         mode = torch.nn.functional.one_hot(
-            torch.argmax(dist.logits, -1), num_classes=dist.logits.shape[-1])
```
Haichao-Zhang, line 1071: The original code is consistent with the convention that the ``mode`` has the same dtype as ``samples`` generated from ``dist.sample()``.

hnyu: > The original code is consistent with the convention that the `mode` has the same dtype as `samples` generated from `dist.sample()`.

OK, there is no good way of enforcing this. I'll close this PR.

PR #1379, opened 2022-08-17
RewardShifting data transformer

Haichao-Zhang:
A data transformer for reward shifting. 
It is useful in some cases, e.g. for AntMaze task in D4RL, the original reward range is [0, 1], and is typically shifted by -1 before learning in literature:
image
(from "[Offline reinforcement learning with implicit Q-learning"](https://arxiv.org/pdf/2110.06169.pdf))

### DIFF
--- a/alf/algorithms/data_transformer.py
+++ b/alf/algorithms/data_transformer.py
@@ -123,7 +123,7 @@ class SequentialDataTransformer(DataTransformer):
         state_spec = []
         max_stack_size = 1
         for i, ctor in enumerate(data_transformer_ctors):
-            obs_trans = ctor(observation_spec)
+            obs_trans = ctor(observation_spec=observation_spec)
             if isinstance(obs_trans, FrameStacker):
                 max_stack_size = max(max_stack_size, obs_trans.stack_size)
             observation_spec = obs_trans.transformed_observation_spec
@@ -800,7 +800,7 @@ class HindsightExperienceTransformer(DataTransformer):
     """
 
     def __init__(self,
-                 transformed_observation_spec,
+                 observation_spec,
                  her_proportion=0.8,
                  achieved_goal_field="time_step.observation.achieved_goal",
                  desired_goal_field="time_step.observation.desired_goal",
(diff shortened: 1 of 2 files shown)


alf/algorithms/data_transformer.py
```diff
@@ -625,6 +625,27 @@ def forward(self, reward):
...
 
+@alf.configurable
+class RewardShifting(RewardTransformer):
+    """Shift immediate rewards by a displacement of ``bias``.
+
+    Can be used as a reward shaping function passed to an algorithm
+    (e.g. ``ActorCriticAlgorithm``).
+    """
+
+    def __init__(self, observation_spec=(), bias=0.0):
```
breakds: If someone is intended to use it, I would assume they'd always want to provide a `bias`. Might be good to make it a mandatory argument?

alf/algorithms/data_transformer.py
```diff
@@ -625,6 +625,27 @@ def forward(self, reward):
...
+        """
+        Args:
+            observation_spec (nested TensorSpec): describing the observation in timestep
+            bias (float): displacement amount
+        """
+        super().__init__(observation_spec)
+        self._bias = bias
+
+    def forward(self, reward):
+        return reward + self._bias
```
breakds: Though rarely used, what if the reward is multi-dimensional?

Haichao-Zhang: Now added comments on the expected behavior in multi-dim reward case.
Also added ``FunctionalRewardTransformer`` to help users to implement their desired transformations, including the multi-dim case.

alf/algorithms/data_transformer.py
Haichao-Zhang: Good point. Now change to be a mandatory argument. Also did the same change to ``RewardScaling``.

breakds (dismissed):

alf/algorithms/data_transformer.py
Haichao-Zhang: Pushed another commit edf24bacbe16ae7c1fdd02604e34cd8248fa6475 to fix an issue caused by the mandatory argument change.

breakds (approved):

PR #1380, opened 2022-08-19
Implement the PeriodicReset module

breakds:


alf/layers.py
```diff
@@ -3612,3 +3612,57 @@ def forward(self, input):
                 lambda x: x.float() if x.dtype.is_floating_point else x, input)
         with torch.cuda.amp.autocast(self._enabled):
             return self._net(input)
+
+
+class PeriodicReset(torch.nn.Module):
```
hnyu, line 3617: Is it possible to use  `common.Periodically`?

breakds: `common.Periodically` maintains its own counter, which is slightly different than this.

breakds:
This is still a draft. Will update with test result.

PR #1382, opened 2022-08-25
support optional fields for AverageEnvInfoMetric

hnyu:
Sometimes we have lots of stuff in `time_step.env_info` but we don't want to metric everyone of them.

### DIFF
--- a/alf/metrics/metrics.py
+++ b/alf/metrics/metrics.py
@@ -16,6 +16,8 @@
 Converted to PyTorch from the TF version.
 https://github.com/tensorflow/agents/blob/master/tf_agents/metrics/tf_metrics.py
 """
+from typing import List
+
 import torch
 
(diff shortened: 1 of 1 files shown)


alf/metrics/metrics.py
```diff
@@ -553,6 +555,7 @@ def _extract_metric_values(self, time_step):
                            torch.ones_like(time_step.step_type))
 
 
+@alf.configurable(whitelist=['fields'])
```
breakds, line 558: Suggest `whitelist_fields` so that the name indicates this is a whitelist.

breakds (commented):
Thanks Haonan!

breakds (approved):

breakds:
Works as intended and pretty well. I'll merge it.

PR #1384, opened 2022-08-30
Correct handling of DataTransformer for PPG Auxiliary Algorithm

breakds:
# Motvation

Previously `PPGAuxAlgorithm` will inherit the data transformers from the main algorithm `PPGAlgorithm`. This is incorrect because we do not want both algorithms' data transformers to share buffers.

Another problem w.t.r. PPG and data transformers is that usually PPG do not convert the experience to GPU during whole buffer training of auxiliary algorithm. In this case, we would like data transformer(s) to be created on CPU so that they can work with experience on CPU.

# Solution

1. Reconstruct data transformer upon `__init__` of `PPGAuxAlgorithm`.
2. Support `device=` for `create_data_transformer`, so that the user can specify it to be `cpu` when turning off "convert to GPU" of `gather_all()` for PPG (and other algorithm with the same situation).

# Testing

Added data transformer to an existing PPG algorithm config and make sure the training is running correctly.

### DIFF
--- a/alf/algorithms/data_transformer.py
+++ b/alf/algorithms/data_transformer.py
@@ -18,7 +18,7 @@ from functools import partial
 import numpy as np
 import torch
 from torch import nn
-from typing import Iterable
+from typing import Iterable, Optional
 
 import alf
 from alf.data_structures import AlgStep, Experience, namedtuple, StepType, TimeStep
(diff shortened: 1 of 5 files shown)


hnyu (dismissed):
Thanks! Did you run code format checking?

breakds:
> Thanks! Did you run code format checking?

Yes and no. My local `pre-commit`'s version is too high and I haven't fixed it yet. Will update in a bit :)

breakds:
Format is fixed.

hnyu (approved):

PR #1385, opened 2022-08-30
Support injected summarize function from config for summarize_rollout

breakds:
# Motivation

There are times when we want to customize the behavior of summarizing the `env_info` from the environments (in lieu of the default summarization). As discussed with @hnyu, `summarize_rollout` fit the niche pretty well. The only problem here with `summarize_rollout` is that it is part of the algorithm implementation, instead of environment dependent.

# Solution

In this PR I propose to support an "injected_summary" for `summarize_rollout`. Therefore, we can customize the summarization in the *environment's configuration". For example, together with #1382 (thanks to @hnyu), we can let the default env info summarize part of `MetaDrive`'s, and customize the summarization for the rest.

# Example

With the updated MetaDrive base conf, we can now see summary like below automatically as long as `metadrive/base_conf.py` is imported.

inject_summary

### DIFF
--- a/alf/algorithms/rl_algorithm.py
+++ b/alf/algorithms/rl_algorithm.py
@@ -371,16 +371,16 @@ class RLAlgorithm(Algorithm):
                         torch.mean(r),
                         average_over_summary_interval=True)
 
-    @alf.configurable(whitelist=["inject_summary"])
+    @alf.configurable(whitelist=["custom_summary"])
     def summarize_rollout(
             self,
             experience: Experience,
-            inject_summary: Optional[Callable[[Experience], None]] = None):
+            custom_summary: Optional[Callable[[Experience], None]] = None):
         """Generate summaries for rollout.
 
         Args:
             experience: experience collected from ``rollout_step()``.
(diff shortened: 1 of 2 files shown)


breakds:
Still the formatting issue. Will have it fixed in a bit.

breakds:
Format is fixed.

alf/algorithms/rl_algorithm.py
```diff
@@ -371,11 +371,19 @@ def summarize_reward(self, name, rewards):
                         torch.mean(r),
                         average_over_summary_interval=True)
 
-    def summarize_rollout(self, experience):
+    @alf.configurable(whitelist=["inject_summary"])
+    def summarize_rollout(
+            self,
+            experience: Experience,
+            inject_summary: Optional[Callable[[Experience], None]] = None):
```
hnyu: suggest changing name from `inject_summary` to `custom_summary`.

breakds: Sounds good. Fixed.

hnyu (approved):

PR #1386, opened 2022-08-31
Add extra rewards to metadrive

breakds:
# Motivation

To add extra rewards that 

- Encourage keeping in lane and penalize riding the broken line
- Penalize harsh brake and jerky throtle
- Penalize jerky steering

# Solution

In this PR, 3 categories of extra rewards are implemented in `extra_rewards.py`, respectively

1. `EgoKinematicsReward` - penalizes jerky behavior
2. `LaneKeepingReward` - penalizes riding broken line
3. `CrashVehicleReward` - penalizes crashing into other vehicles (on top of the flat crashing cost)

Those extra rewards can be configured in Alf configuration, see `metadrive/base_conf.py`.

Also, while computing those rewards, extra env info are also computed and added. The metadrive base configuration is updated to summarize those extra env info.

# Testing

After tuning, the current set of parameters for the extra costs can produce policies that respect comfortness and knows how to yield. See the video below:

kine_tune9


alf/environments/suite_metadrive.py
```diff
@@ -126,6 +158,10 @@ def _acquire_next_frame(self, action):
         """
         observation, reward, done, info = self._env.step(action)
 
+        # Apply extra rewards
+        extra_reward, extra_env_info = self._compute_extra_reward()
+        reward += extra_reward
+
```
breakds, line 164: To reviewers: the change here can be used as an entry point to read the whole diff. Here extra rewards are computed every time when the underlying environment `step()`.

PR #1387, opened 2022-09-01
Add extra rewards to metadrive

breakds:
# Motivation

To add extra rewards that 

- Encourage keeping in lane and penalize riding the broken line
- Penalize harsh brake and jerky throtle
- Penalize jerky steering

# Solution

In this PR, 3 categories of extra rewards are implemented in `extra_rewards.py`, respectively

1. `EgoKinematicsReward` - penalizes jerky behavior
2. `LaneKeepingReward` - penalizes riding broken line
3. `CrashVehicleReward` - penalizes crashing into other vehicles (on top of the flat crashing cost)

Those extra rewards can be configured in Alf configuration, see `metadrive/base_conf.py`.

Also, while computing those rewards, extra env info are also computed and added. The metadrive base configuration is updated to summarize those extra env info.

# Testing

After tuning, the current set of parameters for the extra costs can produce policies that respect comfortness and knows how to yield. See the video below:

kine_tune9


alf/environments/suite_metadrive.py
```diff
@@ -114,6 +129,23 @@ def render(self, mode):
...
+
+        1. The extra reward (sum of all components)
+        2. The extra environment info generated by computing the extra rewards
+        
+        """
+        result = 0.0
+        extra_env_info = {}
+        for extra_reward in self._all_extra_rewards:
+            rewards, info = extra_reward.evaluate(self._env.engine)
+            extra_env_info = {**extra_env_info, **info}
```
emailweixu: extra_env_info.update(info)

alf/environments/suite_metadrive.py
```diff
@@ -114,6 +129,23 @@ def render(self, mode):
...
+            for _, reward in rewards.items():
```
emailweixu: for reward in rewards.values():

alf/environments/metadrive/extra_rewards.py
```diff
@@ -0,0 +1,348 @@
...
+        lat_acc = _estimate_derivative_2(self._history_heading, is_angle=True)
+        lat_jerk = _estimate_derivative_3(self._history_heading, is_angle=True)
+
+        speed = self._history_speed[-1]
+        lon_acc_cost = self._harsh_brake_cost_func(
+            lon_acc=lon_acc, speed=speed)
+        lon_jerk_cost = self._lon_jerk_cost_func(jerk=lon_jerk, speed=speed)
+        lat_jerk_cost = self._lat_jerk_cost_func(jerk=lat_jerk, speed=speed)
+
+        harsh_brake = 1.0 if speed > 2.0 and lon_acc < -1.2 else 0.0
```
emailweixu: the const value 2.0 and -1.2 is also used as the default argument to harsh_break_cost_func. This may lead to some inconsistency.

alf/environments/metadrive/extra_rewards.py
```diff
@@ -0,0 +1,348 @@
...
+    """
+
+    def __init__(self, cost: float = 20.0):
+        super().__init__()
+        self._cost = cost
+
+    def evaluate(self, engine: BaseEngine):
+        ego: BaseVehicle = engine.managers["agent_manager"].active_agents[
+            "default_agent"]
+        cost = 0.0
```
emailweixu: nitpick: Since this reward is very similar to LaneKeepingReward, I feel the code (line 336-342 and. line 296-303) should also be written in a similar way.

alf/environments/metadrive/extra_rewards.py
breakds: Right this is actually a bit awkward. You are right about different intentions here

1. Harsh brake cost function is the means to encourage mild braking instead of harsh braking
2. Harsh brake boolean is to judge whether a harsh brake happens or not.

I have updated the code logic here to make it clearer.

alf/environments/metadrive/extra_rewards.py
breakds: Very reasonable. Updated to make them consistent.

breakds:
Updated. PTAL.

Haichao-Zhang (approved):

emailweixu (approved):

PR #1389, opened 2022-09-01
Add MuZero configuration for MetaDrive

breakds:
This is the configuration that trains MetaDrive agent with MuZero algorithm. The configuration is partly based on `muzero_atari_conf.py` and partly based on `ppg_metadrive_transformer_conf.py`.


alf/examples/metadrive/muzero_metadrive_transformer_conf.py
```diff
@@ -0,0 +1,412 @@
...
+                        observation_spec['agents'].shape[2],
+                        hidden=(32, ))),
+                input_tensor_spec=observation_spec),
+            lambda x: (~x['map_mask'], ~x['agent_mask'])),
+        lambda x: (x[0], x[1][0], x[1][1]),
+        MaskedTransformer(num_layers=3, num_other_entities=num_other_entities),
+        # Take the corresponding transformer output of the first vector in the
+        # sequence (corresponding to "ego") as the final output of the encoder.
+        lambda x: x[:, 0, :],
+        layers.Reshape(-1),
```
emailweixu: This Reshape seems unnecessary?

alf/examples/metadrive/muzero_metadrive_transformer_conf.py
```diff
@@ -0,0 +1,412 @@
...
+alf.config(
+    "MCTSModel",
+    value_loss=rv_loss,
+    reward_loss=rv_loss,
+    repr_loss=losses.AsymmetricSimSiamLoss(
+        input_size=d_model,
+        proj_hidden_size=512,
+        pred_hidden_size=512,
+        output_size=1024),
+    predict_reward_sum=True,
```
emailweixu, line 326: In muzero_atari_conf, sum_over_reward_prediction or lstm_reward should be True if predict_reward_sum is True. This is because when predict_reward_sum is True, the predicted reward is compared against the reward sum. So the reward prediction should stateful.

alf/examples/metadrive/muzero_metadrive_transformer_conf.py
```diff
@@ -0,0 +1,412 @@
...
+    expand_all_children=False,
+    expand_all_root_children=False,
+    max_unroll_length=8,
+    learn_policy_temperature=1.0)
+
+alf.config(
+    "MuzeroRepresentationImpl",
+    model_ctor=SimpleMCTSModel,
+    num_unroll_steps=5,
+    td_steps=10,
```
emailweixu: since reanalyze_ratio is 1.0, td_steps is not used. What really matter is reanalyze_td_steps

alf/examples/metadrive/muzero_metadrive_transformer_conf.py
breakds: Right, removed.

alf/examples/metadrive/muzero_metadrive_transformer_conf.py
breakds: `td_steps` is a required argument though. Added comment to note that it is not useful here, and also explicitly set `reanalyze_td_steps`.

alf/examples/metadrive/muzero_metadrive_transformer_conf.py
breakds: I had the wrong impression that the "sum over reward" behavior is built-in in the algorithm and will be turned on when `predict_reward_sum` is set to `True`. Let me experiment and fix this. Thanks for pointing out!

breakds: The PR is updated with using LSTM for the reward head and `predict_reward_sum=True`.

predict_reward_sum

Note that both the red curve and blue curve are from the updated `predict_reward_sum=True` + `LSTM` config, suggesting that blue is worse than orange is because of some randomness.

The policy produced by red does can choose to yield or beat at intersection with good judgement (although it is also trained with more iterations). See policy video below:

https://files.breakds.org/wl/?id=WYXl0A05mMNmCcnVU0ajjpK1zbmm0H8O

breakds:
Updated, PTAL.

emailweixu (approved):

PR #1392, opened 2022-09-12
support detach for distributions

hnyu:


alf/utils/common.py
```diff
@@ -1046,15 +1046,19 @@ def _create_ou_process(action_spec):
...
     Returns:
-        detached Tensors with same structure as nests
+        detached Tensors/Distributions with same structure as nests
     """
-    return nest.map_structure(lambda t: t.detach(), nests)
+
+    spec = dist_utils.extract_spec(nests)
+    params = dist_utils.distributions_to_params(nests)
+    params = nest.map_structure(lambda t: t.detach(), params)
+    return dist_utils.params_to_distributions(params, spec)
```
emailweixu: I hate to see the unlikely code path dramatically slow down the common code path. Something like the following (not tested) is better:
```python
def detach_distribution_or_tensor(dist_or_tensor):
  if isinstance(dist_or_tensor, td.Distribution):
    builder, params =  _get_builder(dist)
    params = dict((k, v.detach()) for k, v in params.items())
    return builder(**params)
  else:
    return dist_or_tensor.detach()

def detach(nests: alf.nest.Nest):
  return nest.map_structure(detach_dist_or_tensor, nests)
```

emailweixu (approved):

PR #1395, opened 2022-09-30
BC for CARLA

Haichao-Zhang:
Complete BC example for CARLA.


alf/environments/suite_carla.py
```diff
@@ -385,6 +386,15 @@ def __init__(self,
         else:
             self._bev_sensor = None
 
+        self._with_dynamic_object_sensor = with_dynamic_object_sensor
```
breakds: Seems that `self._with_dyanmic_object_sensor` is not used.

alf/networks/transformer_networks.py
```diff
@@ -222,3 +227,193 @@ def forward(self, inputs, state=()):
...
+        assert self._processed_input_tensor_spec.ndim == 2, (
+            "expect the "
+            "processed spec to have the shape of [entity_num, feature_dim]")
+
+        for size in fc_layer_params:
+            embedding_layers.append(
+                layers.FC(
+                    self._processed_input_tensor_spec.shape[-1],
+                    size,
+                    # activation=activation,
```
breakds: This line is still needed? At least this should be consistent with that in the `kernel_initializer`.

alf/networks/transformer_networks.py
```diff
@@ -222,3 +227,193 @@ def forward(self, inputs, state=()):
...
+            use_bias=False,
+            kernel_initializer=kernel_initializer)
+
+        self._attention_res_branch = layers.FC(
+            fea_dim,
+            fea_dim,
+            use_bias=False,
+            kernel_initializer=kernel_initializer)
+
+    def forward(self, inputs, state=()):
```
breakds, line 341: Reading the code here I think there is no big difference from a standard attention unit? If so, shall we further extract the attention unit out to `layers` and use it here?

alf/examples/carla_conf.py
```diff
@@ -66,3 +86,21 @@ def _make_simple_preproc(spec, use_bias=False):
...
+    # create input mask for flexibily mask out some sensors
+    observation_spec = alf.get_observation_spec()
+    observation_preprocessor_masks = {}
+    for sensor, spec in observation_spec['observation'].items():
+        if sensor in [
+                'gnss',
+                'goal',
+                'radar',
+        ]:  # mask selected seneors out of observation for training
+            print("-----skip {}".format(sensor))
```
breakds: Is this intended or for debugging purpose only?

alf/environments/suite_carla.py
Haichao-Zhang: Removed.

alf/networks/transformer_networks.py
Haichao-Zhang: Uncommented this line. Since the linear activation is used in the paper, now added a linear activation in the conf.

alf/networks/transformer_networks.py
Haichao-Zhang: Moved attention to layers.

alf/examples/carla_conf.py
Haichao-Zhang: This is an intended message on the skipped sensor. Changed to logging instead.

alf/networks/transformer_networks.py
```diff
@@ -222,3 +227,167 @@ def forward(self, inputs, state=()):
...
+
+        if kernel_initializer is None:
+            kernel_initializer = functools.partial(
+                variance_scaling_init,
+                mode='fan_in',
+                distribution='truncated_normal',
+                nonlinearity=activation)
+
+        embedding_layers = nn.ModuleList()
+        print(self._processed_input_tensor_spec)
```
emailweixu: remove

alf/networks/transformer_networks.py
```diff
@@ -222,3 +227,167 @@ def forward(self, inputs, state=()):
...
+        assert self._processed_input_tensor_spec.ndim == 2, (
+            "expect the "
+            "processed spec to have the shape of [entity_num, feature_dim]")
+
+        for size in fc_layer_params:
+            embedding_layers.append(
+                layers.FC(
+                    self._processed_input_tensor_spec.shape[-1],
```
emailweixu: the input size will not correct if there are multiple fc_layer_params

alf/networks/transformer_networks.py
```diff
@@ -222,3 +227,167 @@ def forward(self, inputs, state=()):
...
+
+    def forward(self, inputs, state=()):
+        """
+        Args:
+            inputs (Tensor):  with the shape of [B, N, d], where
+                B denotes batch size, N the number of entities, and d the
+                feature dimension
+            state (nested Tensor): states
+        Returns:
+            - Tensor: shape is [B, k], where k denotes the dimension of output
```
emailweixu: comment what the value of k is

alf/layers.py
```diff
@@ -3612,3 +3612,34 @@ def forward(self, input):
...
+        """Multi-Head attention computation based on the inputs.
+        Args:
+            query (Q): shape [B, head, 1, d]
+            key (K):  shape [B, head, N, d]
+            value (V): shape [B, head, N, d]
+            where B denotes the batch size, head denotes the number of heads,
+            N the number of entities, and d the feature dimension.
+        Return:
+            - the attended results computed as: softmax(QK^T/sqrt(d))V
+            - the attention weight, with the shape [B, head, 1, d]
```
emailweixu: The shape for attention weight is [B, head, 1, N]?

alf/layers.py
```diff
@@ -3612,3 +3612,34 @@ def forward(self, input):
...
+class MultiHeadAttention(nn.Module):
+    """Multi-Head Attention Module."""
+
+    def __init__(self):
+        super().__init__()
+
+    def forward(self, query, key, value):
+        """Multi-Head attention computation based on the inputs.
+        Args:
+            query (Q): shape [B, head, 1, d]
```
emailweixu: Can it be more generally as [B, head, M, d]?

alf/layers.py
```diff
@@ -3612,3 +3612,34 @@ def forward(self, input):
                 lambda x: x.float() if x.dtype.is_floating_point else x, input)
         with torch.cuda.amp.autocast(self._enabled):
             return self._net(input)
+
+
+class MultiHeadAttention(nn.Module):
```
emailweixu: This is not the standard MultiHeadAttention in transformer. Perhaps using a different name (e.g. SimpleAttention)?

emailweixu (changes requested):

alf/layers.py
Haichao-Zhang: Changed name to SimpleAttention to avoid confusion.

alf/layers.py
Haichao-Zhang: Good point. It can handle this more general case. Updated the comments.

alf/layers.py
Haichao-Zhang: Fixed.

alf/networks/transformer_networks.py
Haichao-Zhang: It denotes the dimensionality of the output feature.  Updated the comments.

alf/networks/transformer_networks.py
Haichao-Zhang: You are right. This is a bug. Fixed.

alf/networks/transformer_networks.py
Haichao-Zhang: Removed.

alf/networks/transformer_networks.py
```diff
@@ -222,3 +227,158 @@ def forward(self, inputs, state=()):
...
+    def forward(self, inputs, state=()):
+        """
+        Args:
+            inputs (Tensor):  with the shape of [B, N, d], where
+                B denotes batch size, N the number of entities, and d the
+                feature dimension
+            state (nested Tensor): states
+        Returns:
+            - Tensor: shape is [B, d'], where d' denotes the dimension of output
+                feature (fea_dim)
```
emailweixu: Can be more explicit: It is the  size of the last layer in fc_layer_params (i.e. fc_layer_params[-1])

Haichao-Zhang: Updated the comment according to suggestion.

emailweixu (approved):

PR #1396, opened 2022-10-03
Use torch.div for floor div in replay buffer

breakds:
This is to applease `pytorch` as it is currently warning about `__floordiv__` (i.e. `//`) being deprecated in the near future, and suggesting the correct way to do this is its more cumbersome alternative:

```
torch.div(x, y, rounding_mode="floor")
```

### DIFF
--- a/alf/experience_replayers/segment_tree.py
+++ b/alf/experience_replayers/segment_tree.py
@@ -68,9 +68,9 @@ class SegmentTree(nn.Module):
             """
             Calculate the parent value from its children.
             """
-            # need to use `//` here. Newer versions of pytorch will do automatic
-            # type promtion and will generate float indices if `/` is used.
-            indices = indices // 2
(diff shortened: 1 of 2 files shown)


alf/experience_replayers/segment_tree.py
```diff
@@ -68,9 +68,9 @@ def _step(indices):
             """
             Calculate the parent value from its children.
             """
-            # need to use `//` here. Newer versions of pytorch will do automatic
-            # type promtion and will generate float indices if `/` is used.
-            indices = indices // 2
+            # Here ``torch.div`` with ``rounding_mode="floor"`` will produce
+            # result with integer dtype.
+            indices = torch.div(indices, 2, rounding_mode="floor")
```
emailweixu: "indices >> 1"?

breakds: Right, much more succinct. Updated.

emailweixu (approved):

PR #1397, opened 2022-10-05
Support asynchronous unroll

emailweixu:
Perform unroll and training concurrently. This can be useful in two situations:
1. The environment step is expensive. This can help reducing time spending on unroll
2. Realtime training where the interaction with environment happens in realtime. For this, we can set TrainerConfig.unroll_step_interval to the desired interaction period.

Tested on sac_cart_pole_conf.py and async_unroller_test.py

### DIFF
--- a/alf/algorithms/async_unroller.py
+++ b/alf/algorithms/async_unroller.py
@@ -24,7 +24,6 @@ from typing import Dict, List
 import alf
 from alf.utils import common
 from alf.algorithms.data_transformer import create_data_transformer
-from alf.trainers import policy_trainer
 from collections import namedtuple
 
 UnrollResult = namedtuple("UnrollResult", [
@@ -140,6 +139,8 @@ FLAGS = flags.FLAGS
 
 def _worker(job_queue: mp.Queue, done_queue: mp.Queue, result_queue: mp.Queue,
             conf_file: str, pre_configs: Dict, root_dir: str):
+    from alf.trainers import policy_trainer
+
     def _update_parameter(algorithm, job):
         # Some algorithms use scheduler depending on the global counter
         # or the training progress. So we make sure they are same as
--- a/alf/algorithms/async_unroller.py
+++ b/alf/algorithms/async_unroller.py
@@ -1,0 +1,289 @@
+# Copyright (c) 2022 Horizon Robotics and ALF Contributors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from absl import logging
+from absl import flags
+from queue import Empty, Full
+import torch.multiprocessing as mp
+import sys
+import time
+import torch
+from typing import Dict, List
+
+import alf
+from alf.utils import common
+from alf.algorithms.data_transformer import create_data_transformer
+from alf.trainers import policy_trainer
+from collections import namedtuple
+
+UnrollResult = namedtuple("UnrollResult", [
+    "time_step", "transformed_time_step", "policy_step", "policy_state",
+    "env_step_time", "step_time"
+])
+
+UnrollJob = namedtuple(
+    "UnrollJob", ["type", "step_metrics", "global_counter", "state_dict"],
+    defaults=[None] * 4)
+
+
+class AsyncUnroller(object):
+    """A helper class for unroll asynchronously.
+
+    The following settings in TrainerConfig are related to the functionality of
+    AsyncUnroller: unroll_length, async_unroll, max_unroll_length,
+    unroll_queue_size, unroll_step_interval. See algorithms.config.py for their
+    documentation.
+
+    TODO: redirect the log and summary to the training process. Currently,
+    all the logs are written to a different log file and summary during
+    rollout_step() is not enabled.
+
+    Args:
+        algorithm: the root RL algorithm
+        unroll_queue_size: the size of the queue for transmitting the unroll results
+            to the main process
+        root_dir: directory for saving summary and checkpoints
+        conf_file: config file name
+    """
+
+    def __init__(self, algorithm, unroll_queue_size: int, root_dir: str,
+                 conf_file: str):
+        # The following line is needed for avoiding
+        # "RuntimeError: unable to open shared memory object"
+        # See https://github.com/facebookresearch/maskrcnn-benchmark/issues/103#issuecomment-785815218
+        mp.set_sharing_strategy('file_system')
+        if conf_file.endswith('.gin'):
+            assert not self._async, "async_unroll is not supported for gin_file"
+        ctx = mp.get_context('spawn')
+        self._job_queue = ctx.Queue()
+        self._done_queue = ctx.Queue()
+        self._result_queue = ctx.Queue(unroll_queue_size)
+        pre_configs = dict(alf.get_handled_pre_configs())
+        self._worker = ctx.Process(
+            target=_worker,
+            args=(self._job_queue, self._done_queue, self._result_queue,
+                  conf_file, pre_configs, root_dir))
+        self._worker.start()
+        self.update_parameter(algorithm)
+        self._closed = False
+
+    def gather_unroll_results(self, unroll_length: int,
+                              max_unroll_length: int) -> List[UnrollResult]:
+        """Gather the unroll results:
+
+        Args:
+            unroll_length: the desired unroll length. If is 0, any length up to
+                ``max_unroll_length`` is possible (including zero length) depending
+                on how much data is in the queue.
+            max_unroll_length: maximal length of unroll results. This is only
+                used if ``unroll_length`` is 0.
+        Returns:
+            A list of ``UnrollResult``
+        """
+        unroll_results = []
+        if unroll_length > 0:
+            for i in range(unroll_length):
+                unroll_results.append(self._result_queue.get())
+        else:
+            while not self._result_queue.empty() and len(
+                    unroll_results) < max_unroll_length:
+                unroll_results.append(self._result_queue.get())
+        return unroll_results
+
+    def update_parameter(self, algorithm):
+        """Update the the model parameter for unroll.
+
+        Args:
+            algorithm (RLAlgorithm): the root RL algorithm
+        """
+        step_metrics = algorithm.get_step_metrics()
+        step_metrics = dict((m.name, int(m.result())) for m in step_metrics)
+        job = UnrollJob(
+            type="update_parameter",
+            step_metrics=step_metrics,
+            global_counter=int(alf.summary.get_global_counter()),
+            state_dict=algorithm.state_dict())
+        self._job_queue.put(job)
+        self._done_queue.get()
+
+    def close(self):
+        """Close the unroller and release resources."""
+        if self._closed:
+            return
+        job = UnrollJob(type="stop")
+        self._job_queue.put(job)
+        self._done_queue.get()
+        self._worker.join()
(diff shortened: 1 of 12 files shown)


alf/algorithms/async_unroller.py
```diff
@@ -0,0 +1,289 @@
...
+    The following settings in TrainerConfig are related to the functionality of
+    AsyncUnroller: unroll_length, async_unroll, max_unroll_length,
+    unroll_queue_size, unroll_step_interval. See algorithms.config.py for their
+    documentation.
+
+    TODO: redirect the log and summary to the training process. Currently,
+    all the logs are written to a different log file and summary during
+    rollout_step() is not enabled.
+
+    Args:
```
hnyu, line 58: Move this arg list after \__init\__?

alf/algorithms/config.py
```diff
@@ -126,6 +131,17 @@ def __init__(self,
                 is an on-policy sub-algorithm, we can enable this flag for its
                 training. ``OnPolicyAlgorithm`` always unrolls with grads and this
                 flag doesn't apply to it.
+            async_unroll: whether to unroll asynchronously. If True, unroll will
+                be performed in parallel with training.
+            max_unroll_length: the maximal length of unroll results for each iteration.
+                Only used if async_unroll is True。
+            unroll_queue_size: the size of the queue for transmitting unroll
```
hnyu, line 142: Should better document what happens if the number of unroll results exceeds this queue size.

alf/algorithms/config.py
```diff
@@ -126,6 +131,17 @@ def __init__(self,
                 is an on-policy sub-algorithm, we can enable this flag for its
                 training. ``OnPolicyAlgorithm`` always unrolls with grads and this
                 flag doesn't apply to it.
+            async_unroll: whether to unroll asynchronously. If True, unroll will
+                be performed in parallel with training.
+            max_unroll_length: the maximal length of unroll results for each iteration.
```
hnyu, line 137: Should better document in what conditions this max unroll length might not be reached. Also it's only used if ``unroll_length=0``. aysnc_unroll -> \``async_unroll\``

alf/algorithms/config.py
```diff
@@ -126,6 +131,17 @@ def __init__(self,
...
+                Only used if async_unroll is True。
+            unroll_queue_size: the size of the queue for transmitting unroll
+                results from the unroll process to the main process. Only used
+                if async_unroll is True.
+            unroll_step_interval: if not zero, the time interval in second
+                between each two environment steps. Only used if async_unroll is True。
```
hnyu: Not sure why we need this argument just from the description here. Do we need to sometimes wait idly between two environment steps?

alf/algorithms/async_unroller.py
```diff
@@ -0,0 +1,289 @@
...
+    defaults=[None] * 4)
+
+
+class AsyncUnroller(object):
+    """A helper class for unroll asynchronously.
+
+    The following settings in TrainerConfig are related to the functionality of
+    AsyncUnroller: unroll_length, async_unroll, max_unroll_length,
+    unroll_queue_size, unroll_step_interval. See algorithms.config.py for their
+    documentation.
```
hnyu, line 52: We need a big picture describing how the async unroll schedule works.

alf/algorithms/async_unroller.py
```diff
@@ -0,0 +1,289 @@
...
+            t1 = time.time()
+            env_step_time = t1 - t0
+
+            unroll_result = UnrollResult(
+                time_step=time_step,
+                transformed_time_step=transformed_time_step,
+                policy_step=policy_step,
+                policy_state=policy_state,
+                env_step_time=env_step_time,
+                step_time=step_time)
```
hnyu, line 237: This step_time is actually the time used by the previous step? What is the purpose of this quantity?

alf/algorithms/rl_algorithm.py
```diff
@@ -478,6 +479,74 @@ def _rollout_step(self, time_step: TimeStep, state):
...
+
+            exp_for_training = Experience(
+                time_step=unroll_result.transformed_time_step,
+                action=unroll_result.policy_step.output,
+                rollout_info=dist_utils.distributions_to_params(
+                    policy_step.info),
+                state=policy_state)
+
+            experience_list.append(exp_for_training)
+            original_reward_list.append(time_step.reward)
```
hnyu: Feel like this part of making experiences can be merged with the corresponding part in ``self._unroll`` to reduce code maintenance?

hnyu (changes requested):

alf/algorithms/async_unroller.py
emailweixu: added

alf/algorithms/async_unroller.py
emailweixu: Comments added

alf/algorithms/config.py
emailweixu: updated.

alf/algorithms/config.py
emailweixu: updated.

alf/algorithms/config.py
emailweixu: It is for real time simulation. Comments added.

alf/algorithms/rl_algorithm.py
emailweixu: That's true. But I also feel making more abstraction will lower the readability. There are some code specifically for async_unroll after the loop.

alf/algorithms/async_unroller.py
emailweixu: This is common way for documenting a class. See for example https://pytorch.org/docs/stable/_modules/torch/nn/modules/linear.html#Linear. Some other classes in alf also use this format.

alf/algorithms/async_unroller.py
```diff
@@ -0,0 +1,302 @@
...
+    try:
+        _define_flags()
+        FLAGS(sys.argv, known_only=True)
+        FLAGS.mark_as_parsed()
+        FLAGS.alsologtostderr = True
+        logging.set_verbosity(logging.INFO)
+        logging.get_absl_handler().use_absl_log_file(log_dir=root_dir)
+        logging.use_absl_handler()
+        if torch.cuda.is_available():
+            alf.set_default_device("cuda")
```
breakds, line 173: Do expect the rollout to always be very lite? In case it is not, maybe we can provide an argument to perform the rollout on another GPU?

alf/algorithms/async_unroller.py
```diff
@@ -0,0 +1,302 @@
...
+        remaining = 0
+        step_time = 0
+        t = time.time()
+        while True:
+            policy_state = common.reset_state_if_necessary(
+                policy_state, initial_state, time_step.is_first())
+            transformed_time_step, trans_state = algorithm.transform_timestep(
+                time_step, trans_state)
+            transformed_time_step, trans_state = algorithm.transform_timestep(
+                time_step, trans_state)
```
breakds: Seems like redundant line?

alf/algorithms/async_unroller.py
```diff
@@ -0,0 +1,302 @@
...
+                policy_state=policy_state,
+                env_step_time=env_step_time,
+                step_time=step_time)
+
+            stopped = False
+            # If result_queue is full, result_queue.put() will block, which can
+            # cause deadlock if the main process is trying to update_parameter
+            # or stop at the same time. So we need to periodically check job queue
+            # if the result_queue is full.
+            while result_queue.full():
```
breakds, line 244: It might be useful to add metrics about how full the queue is as a percentage.

alf/algorithms/async_unroller.py
```diff
@@ -0,0 +1,302 @@
...
+                    if job.type == "update_parameter":
+                        _update_parameter(algorithm, job)
+                    elif job.type == "stop":
+                        stopped = True
+                        break
+                    else:
+                        raise KeyError(
+                            'Received message of unknown type {}'.format(
+                                job.type))
+                else:
```
breakds: Nit: Looks like the job consuming logic can be wrapped and re-used. Just in case we need to add a new type of job in the future, it will be easier.

alf/algorithms/rl_algorithm.py
```diff
@@ -478,6 +479,82 @@ def _rollout_step(self, time_step: TimeStep, state):
     @common.mark_rollout
     @data_distributed_when(lambda algorithm: algorithm.on_policy)
     def unroll(self, unroll_length: int):
+        if not self._config.async_unroll:
+            return self._unroll(unroll_length)
+        if self._async_unroller is None:
+            # self._env.close()
```
breakds, line 491: Unintentional comment?

alf/algorithms/rl_algorithm.py
```diff
@@ -478,6 +479,82 @@ def _rollout_step(self, time_step: TimeStep, state):
...
+                              original_reward)
+
+        experience = alf.nest.utils.stack_nests(experience_list)
+        experience = experience._replace(
+            rollout_info=dist_utils.params_to_distributions(
+                experience.rollout_info, self._rollout_info_spec))
+
+        return experience
+
+    def _unroll(self, unroll_length: int):
```
breakds: How about more explicitly `_synchronized_unroll()` ?

alf/algorithms/rl_algorithm.py
breakds: I kind of agree with @hnyu that code duplication here may encourage inconsistent logic during future maintenance. Maybe we can at least add a comment somewhere to reminder the future maintainers that they should try to keep the logic here and below consistent?

alf/algorithms/config.py
```diff
@@ -126,6 +132,25 @@ def __init__(self,
...
+            unroll_queue_size: the size of the queue for transmitting unroll
+                results from the unroll process to the main process. Only used
+                if ``async_unroll`` is True. If the queue is full, the unroll process
+                will wait for the main process to retrieve unroll results from
+                the queue before performing more unrolls.
+            unroll_step_interval: if not zero, the time interval in second
+                between each two environment steps. Only used if ``async_unroll`` is True.
+                This is useful if the interaction with the environment happens
+                in real time (e.g. real world robot or real time simulation) and
+                you want a fixed interaction frequency with the environment.
```
hnyu, line 151: So my understanding is, if an unroll interval is faster than a training iteration, the behavior of the robot would be taking several actions and then pausing and waiting for the training iteration to finish?

alf/algorithms/config.py
```diff
@@ -126,6 +132,25 @@ def __init__(self,
...
+                If the time for one step of training is less than the time for
+                unrolling ``max_unroll_length`` steps, the length of the unroll
+                results will be less than ``max_unroll_length``. Only used if
+                ``async_unroll`` is True and unroll_length==0.
+            unroll_queue_size: the size of the queue for transmitting unroll
+                results from the unroll process to the main process. Only used
+                if ``async_unroll`` is True. If the queue is full, the unroll process
+                will wait for the main process to retrieve unroll results from
+                the queue before performing more unrolls.
+            unroll_step_interval: if not zero, the time interval in second
```
hnyu, line 147: This interval can only lower bound the time interval between two env steps? If the rollout step is really slow, then the actual interval might be longer than this one?

emailweixu: You are right. This is only useful if env step and rollout step together spend less than unroll_step_interval.

alf/algorithms/async_unroller.py
emailweixu: It's generally true. Actually, I suspect that rollout in CPU might be even faster. But I'll leave those for future change.

alf/algorithms/async_unroller.py
emailweixu: fixed

alf/algorithms/async_unroller.py
emailweixu: added

alf/algorithms/async_unroller.py
emailweixu: merged

alf/algorithms/rl_algorithm.py
emailweixu: Uncommented. env is no longer needed for rollout in the main process.

hnyu (approved):

PR #1399, opened 2022-10-10
MoNet algorithm

hnyu:
Implemented MoNet and SpatialBroadcastDecoder. Included an example config file for training MoNet on a simple bouncing squares environment. 

Learning a 3-slot posterior latent:
monet_bouncing_squares


alf/algorithms/monet_algorithm.py
```diff
@@ -0,0 +1,382 @@
...
+            channels = filters[i]
+
+        self._upsampling_path = torch.nn.ModuleList(deconv_blocks)
+
+    def forward(self, inputs: torch.Tensor, state=()):
+        """Do a forward step of the UNet.
+
+        Args:
+            inputs: the input image of shape ``[B,C,H,W]`` where ``C`` can be
+                any value.
```
emailweixu, line 137: comment output

alf/algorithms/monet_algorithm.py
```diff
@@ -0,0 +1,382 @@
...
+        Args:
+            inputs: the input image
+        Returns:
+            AlgStep:
+            - output (VAEOutput): contains the rsampled posterior ``z`` and the
+                mode of the posterior distribution ``z_mode``.
+            - state: empty
+            - info (MoNetInfo):
+                - loss: the overall loss
+                - kld: kl divergence between posterior and prior (before ``bete``)
```
emailweixu: bete -> beta

alf/algorithms/monet_algorithm.py
```diff
@@ -0,0 +1,382 @@
...
+        if self._beta == 0:
+            kld = ()
+        else:
+            # N(0,1) as prior
+            prior_z_dist = self.make_gaussian(torch.zeros_like(z_mean_log_var))
+            kld = td.kl.kl_divergence(z_dist, prior_z_dist)
+
+        rec, mask_rec = self._decoder_step(output.z)
+
+        # [B,G,1,H,W]
```
emailweixu, line 405: Need to comment what G is somewhere

alf/algorithms/monet_algorithm.py
```diff
@@ -0,0 +1,382 @@
...
+            input_tensor_spec: the spec of input images
+            attention_unet_cls: creates the attention UNet that generates masks for
+                the slots. Depending on the value of ``recurrent_attention``, this unet
+                input and output channels might change. The user doesn't need to specify
+                the input and output specs for this UNet, as it is automatically handled
+                by the algorithm. If ``recurrent_attention==True``, this UNet receives
+                4 channels (RGB+attetion_scope) and outputs 2 channels (current attention
+                logits); otherwise it receives 3 channels (RGB) and outputs ``n_slots``
+                channels (all attention logits).
+            encoder_cls: creates the posterior encoder of MoNet. Note that this encoder
```
emailweixu, line 223: Need to comment the input shape and output shape for this network

alf/algorithms/monet_algorithm.py
```diff
@@ -0,0 +1,382 @@
...
+                operates on each individual slot independently, and thus it's invariant
+                to the slot order. The encoder accepts the image and an attention mask
+                for the slot.
+            decoder_cls: creates the decoder of MoNet. The decoder also operates on
```
emailweixu, line 230: comment its input and output shapes. and the reconstructed mask is its logits? (need to apply softmax to get the actual mask)

alf/algorithms/monet_algorithm.py
```diff
@@ -0,0 +1,382 @@
...
+        self._n_slots = n_slots
+        self._beta = beta
+        self._gamma = gamma
+        # Inverse variance of slot reconstruction Gaussians in [1,1.5]
+        inv_var = torch.arange(n_slots) * (0.5 / n_slots) + 1.
+        self._inv_var = inv_var.reshape(1, -1, 1, 1, 1)
+
+    @staticmethod
+    def make_gaussian(z_mean_and_log_var):
+        z_mean = z_mean_and_log_var[..., :z_mean_and_log_var.shape[-1] // 2]
```
emailweixu: dim = z_mean_and_log_var.shape[-1] // 2

alf/algorithms/monet_algorithm.py
```diff
@@ -0,0 +1,382 @@
...
+            self._inv_var.log() - np.log(np.pi * 2) -
+            self._inv_var * alf.math.square(rec - inputs.unsqueeze(1)))
+        rec_loss = _reduce_loss(-torch.logsumexp(rec_log_prob + mask, dim=1))
+
+        mask_rec = torch.nn.functional.log_softmax(mask_rec, dim=1)
+        mask_rec_loss = _reduce_loss(
+            torch.sum(
+                torch.nn.functional.kl_div(
+                    input=mask_rec,
+                    target=mask,
```
emailweixu: According to eq (3) in the MONet paper, the KL divergence is: $D_{KL}(q_\psi(c|x)\|| p_\theta(c|\{z_k\}))$. So it seems that input and target need to be switched.

alf/algorithms/monet_algorithm.py
```diff
@@ -0,0 +1,382 @@
...
+            return l.mean(list(range(1, l.ndim)))
+
+        rec_log_prob = 0.5 * (
+            self._inv_var.log() - np.log(np.pi * 2) -
+            self._inv_var * alf.math.square(rec - inputs.unsqueeze(1)))
+        rec_loss = _reduce_loss(-torch.logsumexp(rec_log_prob + mask, dim=1))
+
+        mask_rec = torch.nn.functional.log_softmax(mask_rec, dim=1)
+        mask_rec_loss = _reduce_loss(
+            torch.sum(
```
emailweixu: Use Tensor.sum() to reduce the indentation

alf/algorithms/monet_algorithm.py
```diff
@@ -0,0 +1,382 @@
...
+        return z_mean_log_var, mask_logprobs
+
+    def _decoder_step(self, z):
+        # z - [B,G,D]
+        # [B,G,C+1,H,W]
+        decoded = self._decoder(z)[0]
+        # First 3 channels are RGB; last is the predicted mask (Alpha)
+        return decoded[:, :, :-1, ...], decoded[:, :, -1:, ...]
+
+    def _rec_loss_step(self, inputs, rec, mask, mask_rec):
```
emailweixu, line 343: it will be easier to read the code if the shapes of the tensors are annotated.

alf/algorithms/monet_algorithm.py
```diff
@@ -0,0 +1,382 @@
...
+            slot_size: the dimension of each slot embedding.
+            input_tensor_spec: the spec of input images
+            attention_unet_cls: creates the attention UNet that generates masks for
+                the slots. Depending on the value of ``recurrent_attention``, this unet
+                input and output channels might change. The user doesn't need to specify
+                the input and output specs for this UNet, as it is automatically handled
+                by the algorithm. If ``recurrent_attention==True``, this UNet receives
+                4 channels (RGB+attetion_scope) and outputs 2 channels (current attention
+                logits); otherwise it receives 3 channels (RGB) and outputs ``n_slots``
+                channels (all attention logits).
```
emailweixu: Comment the input/output shape for each of these two cases.

alf/algorithms/monet_algorithm.py
```diff
@@ -0,0 +1,382 @@
...
+                    3,
+                    strides=1,
+                    padding=1,
+                    use_bias=False,
+                    activation=alf.math.identity),
+                torch.nn.InstanceNorm2d(filters[i], affine=True),
+                torch.nn.ReLU()
+            ]
+            if i > 0:
+                block.append(torch.nn.MaxPool2d(2))
```
emailweixu, line 82: This doesn't seem to match the description in Appendix B.2: "finally downsampled or upsampled by a factor of 2
using nearest neighbour-resizing (no resizing occurs in the last block of each path)."

emailweixu (changes requested):

alf/networks/network.py
```diff
@@ -418,3 +418,28 @@ def wrap_as_network(net, input_tensor_spec):
...
+    def __init__(self,
+                 network: Network,
+                 batch_dims: int = 2,
+                 name: str = "BatchSquashNetwork"):
+        super().__init__(network.input_tensor_spec, network.state_spec, name)
+        assert isinstance(network, Network)
+        self._network = network
+        self._bs = alf.layers.BatchSquash(batch_dims)
+
+        self._output_spec = network.output_spec
```
emailweixu, line 442: the output_spec is not same as the original output_spec

alf/utils/tensor_utils.py
```diff
@@ -406,3 +408,32 @@ def unflatten(self, tensor):
...
+    return torch.cat([im, yy.unsqueeze(1), xx.unsqueeze(1)], dim=1)
+
+
+def spatial_broadcast(z: torch.Tensor, im_shape: Tuple[int]):
+    """Broadcasting an embedding across the image spatial domain. The image shape
+    is assumed to be channel-first.
+
+    Args:
+        z: an embedding to be broadcast spatially
+        im_shape: a tuple of ints where the last two are height and width.
```
emailweixu, line 440: comment return

alf/utils/tensor_utils.py
```diff
@@ -406,3 +408,32 @@ def unflatten(self, tensor):
...
                      tuple(tensor.shape[1:])))
+
+
+def append_coordinate(im: torch.Tensor):
+    """For the image, we append coordinates as two channels. The image is assumed
+    to be channel-first. The coordinates will range from -1 to 1 evenly.
+
+    Args:
+        im: an image of shape ``[B,C,H,W]``.
+    """
```
emailweixu, line 422: comment return

alf/networks/encoding_networks.py
```diff
@@ -252,6 +252,7 @@ def __init__(self,
                  upsample_conv_layer_params: Tuple[Union[int, Tuple[int]]],
                  start_decoding_size: Union[int, Tuple[int]],
                  start_decoding_channels: int,
+                 preprocess_fc_layer_params=None,
```
emailweixu: comment preprocess_fc_layer_params

alf/algorithms/monet_algorithm.py
hnyu: > This doesn't seem to match the description in Appendix B.2: "finally downsampled or upsampled by a factor of 2 using nearest neighbour-resizing (no resizing occurs in the last block of each path)."

I think the description in the paper is not accurate. For downsampling, nearest neighbor is impossible, we need to do maxpool or conv with stride=2. The official code actually uses maxpool.

alf/algorithms/monet_algorithm.py
hnyu: > Need to comment what G is somewhere

Added a comment block about notation convention.

alf/algorithms/monet_algorithm.py
hnyu: > comment its input and output shapes. and the reconstructed mask is its logits? (need to apply softmax to get the actual mask)

Yeah, the mask is in log space. Added comment about non-activated output.

alf/algorithms/monet_algorithm.py
hnyu: > According to eq (3) in the MONet paper, the KL divergence is: DKL(qψ(c|x)||pθ(c|zk)). So it seems that input and target need to be switched.

The current order is correct. pytorch's kl_div target is the first distribution (q), which is the attention mask generated by the attention UNet. 

https://pytorch.org/docs/stable/generated/torch.nn.KLDivLoss.html#torch.nn.KLDivLoss

alf/networks/network.py
hnyu: > the output_spec is not same as the original output_spec

THis is intended to overwrite the automatically calculated output spec which includes all batch dims after the first batch dim. The reason is that BatchSquashNetwork is a special case that accepts multiple batch dims, and we want to keep this multiple batch dims for the output (in case we have more BatchSquashNetworks following). Added a comment about this.

emailweixu (approved):

PR #1401, opened 2022-10-12
Convert MBRL examples to python config

breakds:
# Motivation

Convert the `.gin` configuration to `.py` configuration, partly because of contributing to the deprecation of `.gin`, partly due to my development environment is only set up for running `.py` configured examples.

# Solution

Replaced the two MBRL examples with their `.py` counterpart and reproduced the training result. In addition to that, I also made a few changes to:

1. Ask a few algorithms and modules to accept constructor instead of the instances themselves. This makes it possible to never explicitly access the `observation_spec` and `action_spec` of the environment.
2. Updated some of the `None` handling logic.
3. Added `maximum` keyword argument to `adamw`, as it is required now.

# Testing

Reproduced the result. Orange = using latent representation, Blue = using explicit state prediction.

mbrl_pendulum


breakds:
Fixed the compatibility issue and passed CI.

alf/optimizers/adamw.py
```diff
@@ -162,3 +163,17 @@ def step(self, closure=None):
...
+
+
+# NOTE(breakds): This is to make the torch.optim._functional.adamw() call
+# compatible with both pytorch 1.9- and 1.10+. In 1.10 new required parameters
+# are introduced in this function, and the following wrapper supplies the
+# default value for such parameters.
+if "maximize" in inspect.signature(torch.optim._functional.adamw).parameters:
+
+    def _adamw(*args, **kwargs):
+        return torch.optim._functional.adamw(*args, **kwargs, maximize=False)
```
Haichao-Zhang, line 175: ``maximize`` is not a required argument of adamw in newer version of pytorch?

breakds: It actually becomes a required argument in the newer version.

Haichao-Zhang: It is still optional in pytorch 1.13?
https://pytorch.org/docs/1.13/generated/torch.optim.AdamW.html?highlight=adamw

breakds: `maximize` is an optional argument for `AdamW`. However, here we are directly calling an internal helper function: https://github.com/pytorch/pytorch/blob/master/torch/optim/adamw.py#L186 , where `maximize` is a required argument. Sadly we need to patch it to be compatible.

Haichao-Zhang (approved):

PR #1402, opened 2022-10-14
TCN wrapper

Haichao-Zhang:
A wrapper for creating tasks with temporally correlated noises (TCN) from existing standard tasks.

### DIFF
--- a/alf/environments/alf_wrappers.py
+++ b/alf/environments/alf_wrappers.py
@@ -954,23 +954,20 @@ class TemporallyCorrelatedNoiseWrapper(AlfEnvironmentBaseWrapper):
         Swamy et al. Causal Imitation Learning under Temporally Correlated Noise, arXiv:2202.01312
     """
 
-    def __init__(self, env, sigma=1.0, past_weight=0.5, current_weight=None):
+    def __init__(self, env, sigma=0.5, past_noise_weight=1.0):
         """Create a Temporally Correlated Noise wrapper, which adds temporally
         correlated noise to the action before interacting with the environment:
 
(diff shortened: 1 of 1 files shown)


alf/environments/alf_wrappers.py
```diff
@@ -944,3 +944,49 @@ def _step(self, action):
...
+
+@alf.configurable
+class TemporallyCorrelatedNoiseWrapper(AlfEnvironmentBaseWrapper):
+    """Adding temporally correlated noise to actions.
+    Reference:
+    ::
+        Swamy et al. Causal Imitation Learning under Temporally Correlated Noise, arXiv:2202.01312
+    """
+
+    def __init__(self, env, sigma=1.0, past_weight=0.5, current_weight=0.1):
```
breakds: Seems that there are more than enough arguments :) 
with `sigma`, `past_weight` and `current_weight = 1 - past_weight` it would be simpler and achieve everything that can be achieved by 3 parameters.

Haichao-Zhang: In the paper the authors used two weights that are not sum to one. Updated the arguments to support both cases as suggested.

breakds: Sorry I was just curious why the authors choose to have 3 parameters, when they can just adjust `sigma` assuming weights summing to 1. Thanks!

breakds (dismissed):

alf/environments/alf_wrappers.py
Haichao-Zhang: The previous redundant way maybe a bit convenient when adjusting each factors but is unnecessary. Now reduced the argument list.

breakds (approved):

PR #1406, opened 2022-10-26
check unused parameters

hnyu:
Add a code block for checking unused parameters, to help find potential bugs. This is also related to multi-gpu training

https://github.com/HorizonRobotics/alf/blob/pytorch/alf/utils/distributed.py#L104

### DIFF
--- a/alf/algorithms/algorithm.py
+++ b/alf/algorithms/algorithm.py
@@ -1181,8 +1181,9 @@ class Algorithm(AlgorithmInterface):
 
         all_params = [(self._param_to_name[p], p) for p in all_params]
         unused_parameters = [p[0] for p in all_params if p[1].grad is None]
-        common.warning_once("Find unused parameters, please double check: %s",
-                            unused_parameters)
+        common.warning_once(
+            "Find parameters without gradients, please double check: %s",
+            unused_parameters)
         return all_params, simple_gns
 
     # Subclass may override calc_loss() to allow more sophisticated loss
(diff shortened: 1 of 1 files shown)


alf/algorithms/algorithm.py
```diff
@@ -1179,6 +1179,13 @@ def _backward_and_gradient_update(self, loss):
         if self._grad_scaler is not None:
             self._grad_scaler.update()
 
+        unused_parameters = []
```
breakds: There are cases when this is actually desired. Can we make it optional so that the warning can also state "if this is intended, please add such config to silence it"?

hnyu: > 

The warning will only appear once in the beginning, so I think it's fine?

breakds: Sure. Which also makes the checking after the first warning a bit wasteful?

breakds (dismissed):

alf/algorithms/algorithm.py
hnyu: > Sure. Which also makes the checking after the first warning a bit wasteful?

It's just a for loop which has been everywhere in this function. I'll merge it with the line below.

alf/algorithms/algorithm.py
```diff
@@ -1180,6 +1180,9 @@ def _backward_and_gradient_update(self, loss):
             self._grad_scaler.update()
 
         all_params = [(self._param_to_name[p], p) for p in all_params]
+        unused_parameters = [p[0] for p in all_params if p[1].grad is None]
+        common.warning_once("Find unused parameters, please double check: %s",
```
emailweixu: To be more specific: "Find parameters without gradients"

breakds (approved):

PR #1407, opened 2022-11-01
Support replace() for TensorSpec

breakds:
# Motivation

There are cases when we want to create a new `TensorSpec` or `BoundedTensorSpec` out of an existing one, with `shape`, `dtype` or bounds modified. For example, when adjusting the observation spec of an environment to our needes.

# Solution

Added `replace` (similar to `_replace` in `NamedTuple`) to save the typing in those situations.

# Tests

Added unit tests.

### DIFF
--- a/alf/tensor_specs.py
+++ b/alf/tensor_specs.py
@@ -86,6 +86,26 @@ class TensorSpec(object):
         assert isinstance(array, (np.ndarray, np.number))
         return TensorSpec(array.shape[from_dim:], str(array.dtype))
 
+    def replace(self, shape=None, dtype=None) -> TensorSpec:
+        """Create a new TensorSpec with part of the properties replaced.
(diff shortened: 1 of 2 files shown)


alf/tensor_specs.py
```diff
@@ -86,6 +86,26 @@ def from_array(cls, array, from_dim=0):
         assert isinstance(array, (np.ndarray, np.number))
         return TensorSpec(array.shape[from_dim:], str(array.dtype))
 
+    def replace(self, shape=None, dtype=None) -> TensorSpec:
```
hnyu: Should add argument descriptions and type annotations

breakds: Sure, updated the PR.

hnyu (approved):

PR #1409, opened 2022-11-01
VQ VAE Algorithm

Haichao-Zhang:
Implementation of [VQ-VAE](https://proceedings.neurips.cc/paper/2017/file/7a98af17e63a0ac09ce2e96d03992fbc-Paper.pdf) algorithm.

### DIFF
--- a/alf/algorithms/vq_vae.py
+++ b/alf/algorithms/vq_vae.py
@@ -16,10 +16,12 @@
 import torch
 import torch.nn as nn
 import torch.nn.functional as F
+from typing import Callable
 
 import alf
 from alf.algorithms.algorithm import Algorithm
 from alf.data_structures import AlgStep, LossInfo, namedtuple
+from alf.networks import EncodingNetwork
 
 VqvaeLossInfo = namedtuple(
     "VqvaeLossInfo", ["quantization", "commitment", "reconstruction"],
@@ -44,15 +46,15 @@ class Vqvae(Algorithm):
     """
 
     def __init__(self,
-                 input_tensor_spec,
-                 num_embeddings,
-                 embedding_dim,
-                 encoder_ctor,
-                 decoder_ctor,
-                 optimizer=None,
-                 commitment_loss_weight=1.0,
-                 debug_summaries=False,
-                 name="Vqvae"):
+                 input_tensor_spec: alf.NestedTensorSpec,
+                 num_embeddings: int,
+                 embedding_dim: int,
+                 encoder_ctor: Callable = EncodingNetwork,
+                 decoder_ctor: Callable = EncodingNetwork,
+                 optimizer: torch.optim.Optimizer = None,
+                 commitment_loss_weight: float = 1.0,
+                 debug_summaries: bool = False,
+                 name: str = "Vqvae"):
         """
         Args:
             input_tensor_spec (TensorSpec): the tensor spec of
@@ -74,10 +76,13 @@ class Vqvae(Algorithm):
         self._num_embeddings = num_embeddings
 
         # [n, d]
-        self._embedding = nn.Embedding(self._num_embeddings,
-                                       self._embedding_dim)
-        self._embedding.weight.data.uniform_(-1 / self._num_embeddings,
-                                             1 / self._num_embeddings)
+        self._embedding = torch.nn.Parameter(
+            torch.FloatTensor(self._num_embeddings, self._embedding_dim))
+
+        torch.nn.init.uniform_(
+            self._embedding,
+            a=-1 / self._num_embeddings,
+            b=1 / self._num_embeddings)
 
         self._encoding_net = encoder_ctor(input_tensor_spec)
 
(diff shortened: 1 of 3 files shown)


hnyu:
What is the advantage of VQ VAE over the standard VAE+discrete latent ([DiscreteVAE](https://github.com/HorizonRobotics/alf/blob/pytorch/alf/algorithms/vae.py))?

alf/algorithms/vq_vae.py
```diff
@@ -0,0 +1,164 @@
...
+
+        # [B, n]
+        onehot_encodings = torch.zeros(encoding_indices.shape[0],
+                                       self._num_embeddings)
+
+        onehot_encodings.scatter_(1, encoding_indices, 1)
+
+        # quantization
+        # [B, n] * [n, d] -> [B, d]
+        quantized = torch.matmul(onehot_encodings, self._embedding.weight)
```
hnyu: ``quantized`` can directly be obtained by ``gather`` on ``self._embedding.weight`` using ``encoding_indices``?

alf/algorithms/vq_vae.py
```diff
@@ -0,0 +1,164 @@
...
+        """
+        # [B, d]
+        input_embedding, _ = self._encoding_net(inputs)
+
+        # calculate distances
+        # [B, 1] + [n] + [B, n]
+        distances = (
+            torch.sum(input_embedding**2, dim=1, keepdim=True) + torch.sum(
+                self._embedding.weight**2, dim=1) -
+            2 * torch.matmul(input_embedding, self._embedding.weight.t()))
```
hnyu: I feel like this NN isn't as robust as Gumbel-softmax with hard samples.

alf/algorithms/vq_vae.py
```diff
@@ -0,0 +1,164 @@
...
+    def __init__(self,
+                 input_tensor_spec,
+                 num_embeddings,
+                 embedding_dim,
+                 encoder_ctor,
+                 decoder_ctor,
+                 optimizer=None,
+                 commitment_loss_weight=1.0,
+                 debug_summaries=False,
+                 name="Vqvae"):
```
hnyu: arguments should have type annotations

alf/algorithms/vq_vae.py
```diff
@@ -0,0 +1,164 @@
...
+        super().__init__(debug_summaries=debug_summaries, name=name)
+
+        self._embedding_dim = embedding_dim
+        self._num_embeddings = num_embeddings
+
+        # [n, d]
+        self._embedding = nn.Embedding(self._num_embeddings,
+                                       self._embedding_dim)
+        self._embedding.weight.data.uniform_(-1 / self._num_embeddings,
+                                             1 / self._num_embeddings)
```
hnyu: Can use ``torch.nn.init.uniform_`` on ``self._embedding.weight`` instead

alf/algorithms/vq_vae.py
```diff
@@ -0,0 +1,164 @@
...
+                the parameter of encoder_net, decoder_net and embedding vectors.
+            commitment_loss_weight (float): the weight for commitment loss.
+        """
+        super().__init__(debug_summaries=debug_summaries, name=name)
+
+        self._embedding_dim = embedding_dim
+        self._num_embeddings = num_embeddings
+
+        # [n, d]
+        self._embedding = nn.Embedding(self._num_embeddings,
```
hnyu: If you don't need nn.Embedding as a module but just need its weight, then probably it's better to have ``self._embedding = nn.Parameter(...)``.

hnyu (changes requested):

Haichao-Zhang:
> What is the advantage of VQ VAE over the standard VAE+discrete latent ([DiscreteVAE](https://github.com/HorizonRobotics/alf/blob/pytorch/alf/algorithms/vae.py))?

They are indeed closely related, but have some differences in implementation details, e.g. the generation of the discrete code.  Therefore, they might be suitable to be used in different scenarios. For example, VQ-VAE can be used as an online clustering algorithm over the input dataset.

alf/algorithms/vq_vae.py
Haichao-Zhang: Change to ``nn.Parameter``

alf/algorithms/vq_vae.py
Haichao-Zhang: Switched to use ``torch.nn.init.uniform_``

alf/algorithms/vq_vae.py
Haichao-Zhang: Good point. Simplified this part with indexing operation.

alf/algorithms/vq_vae.py
Haichao-Zhang: Not sure but it could be

alf/algorithms/vq_vae.py
```diff
@@ -0,0 +1,159 @@
...
+
+        # decoding loss
+        rec = self._decoding_net(quantized_st)[0]
+
+        recon_loss = F.mse_loss(rec, inputs, reduction="none").mean(dim=1)
+
+        if self._debug_summaries and alf.summary.should_record_summaries():
+            with alf.summary.scope(self._name):
+                alf.summary.embedding("vq_embedding",
+                                      self._embedding.weight.detach())
```
hnyu: should remove 'weight'

Haichao-Zhang: Fixed

hnyu (approved):

PR #1410, opened 2022-11-01
Add parallelism support for Normal and Beta projection networks.

breakds:
# Motivation

Creating mixture of policy style projection network requires adding an extra `replica` dim to the batch shape of the output distribution, serving as the number of component. This should be done by calling the `make_parallel()` of the corresponding projection network.

# Solution

Added a `parallelism` argument to the affected projection networks for the above purpose. By default the argument is set to `None` so that the behavior of existing code does not change.

The `make_parallel` will now just delegates to this `parallelism` argument. This is slightly better than the naive `make_parallel` as it uses `ParallelFC` under the hood. This is an effortless change.

As some side notes, 

1. `ParallelNormalProjectionNetwork` is now removed in favor of the new argument, which also helps reduce code duplication.
2. Replace a few `as_tensor` with `torch.tensor` to avoid the warnings. Those ocurrences should not affect performance.

# Testing

Added unit tests.

### DIFF
--- a/alf/networks/projection_networks.py
+++ b/alf/networks/projection_networks.py
@@ -292,6 +292,15 @@ class NormalProjectionNetwork(Network):
         stds = self._std_transform(self._std_projection_layer(inputs))
         return self._normal_dist(means, stds), state
 
+    def make_parallel(self, n):
+        parallel_proj_net_args = dict(**self.saved_args)
+        original_parallelism = parallel_proj_net_args.get("parallelism", None)
+        assert original_parallelism is None, (
+            "Calling make_parallel on a network that is already parallelized")
+        parallel_proj_net_args.update(
+            parallelism=n, name="parallel_" + self.name)
+        return type(self)(**parallel_proj_net_args)
+
 
 @alf.configurable
 class StableNormalProjectionNetwork(NormalProjectionNetwork):
--- a/alf/networks/projection_networks.py
+++ b/alf/networks/projection_networks.py
@@ -15,7 +15,7 @@
 from functools import partial
 import math
 import numpy as np
-from typing import Callable
+from typing import Callable, Optional
 
 import torch
 import torch.nn as nn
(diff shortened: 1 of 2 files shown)


hnyu:
Did you try directly calling ``net.make_parallel()``? Usually the naive parallel network is good enough.

breakds:
> Did you try directly calling `net.make_parallel()`? Usually the naive parallel network is good enough.

Thanks for pointing this out! Just check it out and it should work. I think the only difference is `ParallelFC` vs for loop of `FC`s. Given the effort is not that big, I would slightly prefer the current solution.

Haichao-Zhang:
> Did you try directly calling `net.make_parallel()`? Usually the naive parallel network is good enough.

Agree with @hnyu 's comment. We can potentially just use ``proj_net.make_parallel()`` to create network replica, serving as the components in the mixture policy.

Another point is that implementations like ```ParallelNormalProjectionNetwork``` follows ALF's conventional on network parallelism, while the proposed form is not.

breakds:
> > Did you try directly calling `net.make_parallel()`? Usually the naive parallel network is good enough.
> 
> Agree with @hnyu 's comment. We can potentially just use `proj_net.make_parallel()` to create network replica, serving as the components in the mixture policy.
> 
> Another point is that implementations like `ParallelNormalProjectionNetwork` follows ALF's conventional on network parallelism, while the proposed form is not.

Agree with that. I added back `make_parallel` so that it is still the universal interface and follows the convention.

`ParallelNormalProjectionNetwork` itself shares quite some implementation details of complex logic with `NormalProjectionNetwork`. I think it is better to merge it with the `NormalProjectionNetwork`. On the other hand, I can add a `ParallelNormalProjectionNetwork` that delegates to `NormalProjectionNetwork` if needed for compatibility.

emailweixu (approved):

Haichao-Zhang (approved):

PR #1411, opened 2022-11-01
Implement the MixtureProjectionNetwork

breakds:
# Motivation

For experiments that involves mixture of policies, we need a projection network that outputs mixture of same family distributions.

# Solution

Added a composite projection network `MixtureProjectionNetwork` that does so. It requires an argument of base projection network for each of the component. The base projection network will be replicated by calling its `make_parallel`.

Also updated `dist_utils` so that it supports the extraction and construction of `MixtureSameFamily` style distributions.

# Testing

Added unit tests.

### DIFF
--- a/alf/networks/projection_networks.py
+++ b/alf/networks/projection_networks.py
@@ -292,6 +292,15 @@ class NormalProjectionNetwork(Network):
         stds = self._std_transform(self._std_projection_layer(inputs))
         return self._normal_dist(means, stds), state
 
+    def make_parallel(self, n):
+        parallel_proj_net_args = dict(**self.saved_args)
+        original_parallelism = parallel_proj_net_args.get("parallelism", None)
+        assert original_parallelism is None, (
+            "Calling make_parallel on a network that is already parallelized")
+        parallel_proj_net_args.update(
+            parallelism=n, name="parallel_" + self.name)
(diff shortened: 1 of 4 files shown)


emailweixu (dismissed):

alf/networks/projection_networks.py
```diff
@@ -777,3 +777,78 @@ def forward(self, inputs, state=()):
...
+            num_components: int,
+            component_ctor: Callable[[int, TensorSpec], Network],
+            mixture_ctor: Callable[[int, BoundedTensorSpec],
+                                   Network] = CategoricalProjectionNetwork,
+            name: str = "mix_proj_net"):
+        """Constructs an instance of MixtureProjectionNetwork.
+
+        Args
+
+            input_size: the input vector size
```
Haichao-Zhang: Minor format issue: Args -> Args:    and  can remove line 805

Haichao-Zhang (dismissed):

alf/networks/projection_networks.py
breakds: Fixed. Thanks!

breakds:
Need an approval again because the dependent PR is now merged. Thanks!

Haichao-Zhang (approved):

PR #1412, opened 2022-11-03
add RandomCrop gym wrapper

hnyu:
Add a random cropping gym wrapper for data augmentation. After the cropping, the user can choose to append a FrameResize wrapper to normalize the cropped image size.

### DIFF
--- a/alf/environments/gym_wrappers.py
+++ b/alf/environments/gym_wrappers.py
@@ -13,7 +13,9 @@
 # limitations under the License.
 """Wrappers for gym (numpy) environments. """
 
+from functools import partial
 from absl import logging
+from typing import List
 from collections import deque, OrderedDict
 import copy
 import cv2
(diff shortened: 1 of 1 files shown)


alf/environments/gym_wrappers.py
```diff
@@ -331,6 +333,110 @@ def transform_observation(self, observation):
...
+        self._cropping_fraction = cropping_fraction
+        self._channel_order = channel_order
+        self._share_cropping = share_cropping
+        assert channel_order in ['channels_last', 'channels_first']
+        super().__init__(env, fields=fields)
+        self._observation_space = self._dict_space(self.observation_space)
+        self._syx = alf.nest.map_structure(lambda _: None,
+                                           self._observation_space)
+
+    def _dict_space(self, space):
```
breakds: Nit: `self` not used. `@staticmethod` instead?

alf/environments/gym_wrappers.py
```diff
@@ -331,6 +333,110 @@ def transform_observation(self, observation):
         return obs
 
 
+@alf.configurable
+class RandomFrameCrop(BaseObservationWrapper):
```
breakds: The name does not suggest that the `Random` is `Episodic`. Shall we make it more obvious?

breakds (approved):

PR #1421, opened 2023-01-10
alf snapshot only copies alf module

hnyu:
When alf is a dependency installed by another repo, the current alf root obtained by ``common.alf_root()`` will contain many other modules like below:

``` bash
site-packages/alf
site-packages/torch
site-packages/cv2
...
```
(``site-packages`` is alf root here)

Right now when we generate a snapshot, we copy all contents under this alf root, which will take lots of space and time in the above scenario. This PR fixes this by only copying the 'alf' module dir and adjust the snapshot env vars accordingly.

### DIFF
--- a/alf/utils/common.py
+++ b/alf/utils/common.py
@@ -1478,13 +1478,17 @@ def get_alf_snapshot_env_vars(root_dir):
     points to the ALF snapshot under this directory.
     """
     unzip_alf_snapshot(root_dir)
-    # legacy alf repo path for backward compatibility
     legacy_alf_repo = os.path.join(root_dir, "alf")
-    alf_repo = root_dir
+    if os.path.isfile(os.path.join(legacy_alf_repo, "alf")):
+        # legacy alf repo path for backward compatibility
+        # legacy tb dirs: root_dir/alf/alf/__init__.py
+        alf_repo = legacy_alf_repo
+    else:
+        # new tb dirs: root_dir/alf/__init__.py
+        alf_repo = root_dir
     alf_examples = os.path.join(alf_repo, "alf/examples")
     python_path = os.environ.get("PYTHONPATH", "")
-    python_path = ":".join(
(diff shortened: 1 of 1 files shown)


alf/utils/common.py
```diff
@@ -1481,10 +1478,13 @@ def get_alf_snapshot_env_vars(root_dir):
     points to the ALF snapshot under this directory.
     """
     unzip_alf_snapshot(root_dir)
-    alf_repo = os.path.join(root_dir, "alf")
+    # legacy alf repo path for backward compatibility
+    legacy_alf_repo = os.path.join(root_dir, "alf")
+    alf_repo = root_dir
```
emailweixu: this will make alf_examples not working for legacy alf repo

hnyu: > this will make alf_examples not working for legacy alf repo

Right, I've fixed this

alf/utils/common.py
```diff
@@ -1481,7 +1478,14 @@ def get_alf_snapshot_env_vars(root_dir):
     points to the ALF snapshot under this directory.
     """
     unzip_alf_snapshot(root_dir)
-    alf_repo = os.path.join(root_dir, "alf")
+    legacy_alf_repo = os.path.join(root_dir, "alf")
```
breakds, line 1481: Just want to confirm my understanding: legacy mode and non-legacy mode go thought the same procedure in generation of snapshot, and the difference (one extra layer of directory) is handled when getting the snapshot. Is that correct?

hnyu: > Just want to confirm my understanding: legacy mode and non-legacy mode go thought the same procedure in generation of snapshot, and the difference (one extra layer of directory) is handled when getting the snapshot. Is that correct?

Not the same generation. The new mode copies only the 'alf' module, but not its parent dir.

breakds (approved):

PR #1424, opened 2023-02-12
Spare envs for faster reset

le-horizon:
Usage is through alf.environments.utils.create_environments, see example in alf/environments/parallel_environment_test.py

### DIFF
--- a/alf/environments/parallel_environment.py
+++ b/alf/environments/parallel_environment.py
@@ -167,8 +167,8 @@ class ParallelAlfEnvironment(alf_environment.AlfEnvironment):
 
         # handle spare promises
         if not self._blocking:
-            [p() for p in self._spare_promises]
             [p() for p in self._reset_ts if p is not None]
+            [p() for p in self._spare_promises]
             self._reset_ts = [None] * self._num_envs
             self._spare_promises = [
                 env.reset(self._blocking) for env in self._spare_queue
@@ -260,6 +260,10 @@ class ParallelAlfEnvironment(alf_environment.AlfEnvironment):
         if self._closed:
             return
         logging.info('Closing all processes.')
+        [p() for p in self._reset_ts if p is not None]
+        [p() for p in self._spare_promises]
+        self._reset_ts = []
+        self._spare_promises = []
         for env in self._envs:
             env.close()
         for env in self._spare_queue:
--- a/alf/environments/parallel_environment.py
+++ b/alf/environments/parallel_environment.py
@@ -166,13 +166,13 @@ class ParallelAlfEnvironment(alf_environment.AlfEnvironment):
         time_steps = [env.reset(self._blocking) for env in self._envs]
 
         # handle spare promises
-        if self._spare_queue:
+        if not self._blocking:
             [p() for p in self._spare_promises]
+            [p() for p in self._reset_ts if p is not None]
+            self._reset_ts = [None] * self._num_envs
             self._spare_promises = [
                 env.reset(self._blocking) for env in self._spare_queue
             ]
-
-        if not self._blocking:
             time_steps = [promise() for promise in time_steps]
 
         time_steps = self._stack_time_steps(time_steps)
@@ -228,8 +228,6 @@ class ParallelAlfEnvironment(alf_environment.AlfEnvironment):
 
     def _handle_done(self, time_steps):
         for i, reset_ts in enumerate(self._reset_ts):
-            if reset_ts is not None:
-                reset_ts()  # release old unused promises
             env = self._envs[i]
             if time_steps.step_type[i] == alf.data_structures.StepType.LAST:
                 reset_ts = env.reset(blocking=False)
--- a/alf/environments/parallel_environment.py
+++ b/alf/environments/parallel_environment.py
@@ -76,7 +76,7 @@ class ParallelAlfEnvironment(alf_environment.AlfEnvironment):
                     env_constructors[0], env_id=spare_env_id, flatten=flatten)
                 spare_env_id += 1
                 self._spare_queue.append(spare_env)
-            self._last_is_done = []
+            self._last_is_done = None
         self._num_envs = len(env_constructors)
         self._blocking = blocking
         self._start_serially = start_serially
@@ -162,14 +162,23 @@ class ParallelAlfEnvironment(alf_environment.AlfEnvironment):
             Time step with batch dimension.
         """
         time_steps = [env.reset(self._blocking) for env in self._envs]
-        if not self._blocking:
-            time_steps = [promise() for promise in time_steps]
+
+        # handle spare promises
         if self._spare_queue:
-            self._record_done(time_steps)
+            [p() for p in self._spare_promises]
             self._spare_promises = [
                 env.reset(self._blocking) for env in self._spare_queue
             ]
-        return self._stack_time_steps(time_steps)
+
+        if not self._blocking:
+            time_steps = [promise() for promise in time_steps]
+
+        time_steps = self._stack_time_steps(time_steps)
+
+        if self._spare_queue:
+            self._record_done(time_steps)
+
+        return time_steps
 
     def _step(self, actions):
         """Forward a batch of actions to the wrapped environments.
@@ -185,7 +194,7 @@ class ParallelAlfEnvironment(alf_environment.AlfEnvironment):
         """
         if self._spare_queue:
             actions = self._unstack_actions(actions)
-            time_steps = self._handle_last_done(actions)
+            time_steps, reset_calls = self._step_or_handle_last_done(actions)
         else:
             time_steps = [
                 env.step(action, self._blocking) for env, action in zip(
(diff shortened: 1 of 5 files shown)


alf/environments/parallel_environment.py
```diff
@@ -155,22 +183,69 @@ def _step(self, actions):
...
+                ts = env.step(actions[i], self._blocking)
+            time_steps.append(ts)
+        return time_steps
+
+    def _record_done(self, time_steps):
+        if not self._last_is_done:
+            self._last_is_done = [False] * len(time_steps)
+        for i, ts in enumerate(time_steps):
+            if self._flatten:
+                ts = nest.pack_sequence_as(self._time_step_spec, ts)
```
le-horizon: I'm packing the flattened sequence as nest to get the step_type field back.  Is there an easier/more efficient way?

alf/environments/process_environment.py
```diff
@@ -138,6 +138,7 @@ def __init__(self, env_constructor, env_id=None, flatten=False):
         self._reward_spec = None
         self._time_step_spec = None
         self._env_info_spec = None
+        self._conn = None
```
le-horizon, line 141: ProcessEnv returns weird and non-informative errors when _conn isn't initialized correctly, or double initialized.  This change will make such errors more explicit, and easier to debug.

alf/environments/parallel_environment.py
```diff
@@ -155,22 +183,69 @@ def _step(self, actions):
...
+                    actions = list(actions)
+                ts = env.step(actions[i], self._blocking)
+            time_steps.append(ts)
+        return time_steps
+
+    def _record_done(self, time_steps):
+        if not self._last_is_done:
+            self._last_is_done = [False] * len(time_steps)
+        for i, ts in enumerate(time_steps):
+            if self._flatten:
```
emailweixu: You can do record_done after stack_time_step() (line 202) so that you don't need to do pack_sequence_as for every time_step, which is costly.

And it's better to start reset inside record_done() so that the reset can run concurrently with training.

alf/environments/parallel_environment.py
```diff
@@ -141,6 +164,11 @@ def _reset(self):
         time_steps = [env.reset(self._blocking) for env in self._envs]
         if not self._blocking:
             time_steps = [promise() for promise in time_steps]
+        if self._spare_queue:
+            self._record_done(time_steps)
+            self._spare_promises = [
```
emailweixu, line 173: need to run the existing spare_promises before changing to the new promises.

alf/environments/parallel_environment.py
```diff
@@ -141,6 +164,11 @@ def _reset(self):
         time_steps = [env.reset(self._blocking) for env in self._envs]
         if not self._blocking:
             time_steps = [promise() for promise in time_steps]
+        if self._spare_queue:
```
emailweixu: resetting the spare envs can be done immediately after line 164.

alf/environments/parallel_environment.py
```diff
@@ -155,22 +183,69 @@ def _step(self, actions):
         Returns:
             Batch of observations, rewards, and done flags.
         """
-        time_steps = [
-            env.step(action, self._blocking)
-            for env, action in zip(self._envs, self._unstack_actions(actions))
-        ]
+        if self._spare_queue:
+            actions = self._unstack_actions(actions)
+            time_steps = self._handle_last_done(actions)
```
emailweixu: The name "handle_last_done" is confusing. In addition to handle last done, it also performs step.

alf/environments/parallel_environment_test.py
```diff
@@ -100,6 +104,66 @@ def test_step(self):
...
+        time_step0 = env.reset()
+        action_spec = env.action_spec()
+        action = torch.stack([action_spec.sample() for _ in range(num_envs)])
+        time_step1 = env.step(action)
+        self.assertEqual(time_step0.observation.shape,
+                         time_step1.observation.shape)
+        step1_time = time.time()
+        time_step2 = env.step(action)
+        reset_time = time.time()
+        reset_time_diff = reset_time - step1_time
```
emailweixu: why is this called reset_time_diff?

alf/environments/parallel_environment_test.py
```diff
@@ -100,6 +104,66 @@ def test_step(self):
...
+        self.assertEqual(time_step1.observation.shape,
+                         time_step2.observation.shape)
+        self.assertLessEqual(
+            reset_time_diff,
+            sleep_time - 0.5,
+            msg=(f'Reset with spare envs took {reset_time_diff}, too long'))
+        time_step3 = env.step(action)
```
emailweixu, line 168: Also need to test env_id is correct (i.e. from 0 to num_envs - 1)

emailweixu (changes requested):

alf/environments/parallel_environment.py
le-horizon: Good point.

alf/environments/parallel_environment.py
le-horizon: Good catch.

alf/environments/parallel_environment.py
le-horizon: Changed to _step_or_handle_last_done

alf/environments/parallel_environment.py
le-horizon: Good points.  Done both.

alf/environments/parallel_environment_test.py
le-horizon: renamed.  It's the time it took to reset.  The exact moment of after step1 is renamed to step1_t instead.

alf/environments/parallel_environment_test.py
le-horizon: Done.

le-horizon (commented):
Thanks for the great feedback.  Done all.  PTAL.

Thanks,
Le

alf/environments/parallel_environment.py
```diff
@@ -155,14 +192,61 @@ def _step(self, actions):
...
+                # save env_id of spare env, just in case.
+                # Random seeds are handled separately.
+                env._env_id = spare_env_id
+                self._spare_queue.append(env)
+                # Handle promises of the reset()
+                ts = self._spare_promises.pop(0)
+                reset_calls.append(env.reset)
+            else:
+                if self._flatten:
+                    actions = list(actions)
```
emailweixu: Put these two lines at the beginning of this function and make some explanation about why `list(actions)` is necessary.

le-horizon (commented):
Now it allows early reset even when num_spare_envs is 0, as long as blocking = False.  Added test for this as well.

Could you take another look, @emailweixu ?

Cheers,
Le

alf/environments/parallel_environment.py
```diff
@@ -155,14 +194,68 @@ def _step(self, actions):
...
+                # reset was already called, simply use the result from previous stepping
+                time_steps.append(reset_ts)
+                self._reset_ts[i] = None
+            else:
+                time_steps.append(env.step(action, self._blocking))
+        return time_steps
+
+    def _handle_done(self, time_steps):
+        for i, reset_ts in enumerate(self._reset_ts):
+            if reset_ts is not None:
```
emailweixu: How can this happen?

alf/environments/parallel_environment.py
```diff
@@ -139,9 +164,23 @@ def _reset(self):
             Time step with batch dimension.
         """
         time_steps = [env.reset(self._blocking) for env in self._envs]
+
+        # handle spare promises
+        if self._spare_queue:
+            [p() for p in self._spare_promises]
```
emailweixu, line 171: also need to clear self._reset_ts

alf/environments/parallel_environment_test.py
```diff
@@ -100,6 +104,77 @@ def test_step(self):
...
+        assert torch.all(time_step0.step_type == ds.StepType.FIRST)
+        action_spec = env.action_spec()
+        action = torch.stack([action_spec.sample() for _ in range(num_envs)])
+        time_step1 = env.step(action)
+        self.assertEqual(time_step0.observation.shape,
+                         time_step1.observation.shape)
+        step1_t = time.time()
+        time_step2 = env.step(action)
+        reset_t = time.time()
+        reset_time = reset_t - step1_t
```
emailweixu, line 160: I still don't get why this is called reset_time. What I see it is the time for `env.step()`, not reset.

alf/environments/parallel_environment.py
le-horizon: Done

alf/environments/parallel_environment.py
le-horizon: I was handling the _reset() lazily here.  Now it's no longer needed.

alf/environments/parallel_environment_test.py
le-horizon: Oh, this step() internally calls reset.  Added comment.

le-horizon (commented):
Done all, PTAL @emailweixu 

Le

alf/environments/parallel_environment.py
le-horizon: Good point.  I'm using zip instead, no longer need these two lines.

alf/environments/parallel_environment.py
```diff
@@ -171,6 +262,8 @@ def close(self):
         logging.info('Closing all processes.')
         for env in self._envs:
             env.close()
+        for env in self._spare_queue:
```
emailweixu, line 269: clear the promises: _reset_ts, _spare_promises

le-horizon: Done.

emailweixu (approved):

PR #1429, opened 2023-02-15
Update contributing.rst about git pull

emailweixu:
For new version of git,  the default `git pull` will rebase, which is what we want to avoid. So we need to update the instruction.


emailweixu:
ping

breakds (approved):

PR #1430, opened 2023-02-15
In algorithm checkpoint pre-loading

Haichao-Zhang:
This PR implements in-algorithm checkpoint pre-loading, to flexibly support different cases of using an existing checkpoint.

With this PR, we can choose to only pre-load a checkpoint for a sub-algorithm, either from a checkpoint of this sub-algorithm instance or from a composite checkpoint which includes this sub-algorithm.

### DIFF
--- a/alf/algorithms/actor_critic_algorithm.py
+++ b/alf/algorithms/actor_critic_algorithm.py
@@ -139,8 +139,6 @@ class ActorCriticAlgorithm(OnPolicyAlgorithm):
             loss = loss_class(debug_summaries=debug_summaries)
         self._loss = loss
 
-        self._post_init()
-
     def convert_train_state_to_predict_state(self, state):
         return state._replace(value=())
 
--- a/alf/algorithms/actor_critic_algorithm.py
+++ b/alf/algorithms/actor_critic_algorithm.py
@@ -52,8 +52,7 @@ class ActorCriticAlgorithm(OnPolicyAlgorithm):
                  loss=None,
                  loss_class=ActorCriticLoss,
                  optimizer=None,
-                 checkpoint_path=None,
-                 checkpoint_prefix='',
+                 checkpoint=None,
                  debug_summaries=False,
                  name="ActorCriticAlgorithm"):
         """
@@ -94,13 +93,10 @@ class ActorCriticAlgorithm(OnPolicyAlgorithm):
             loss_class (type): the class of the loss. The signature of its
                 constructor: ``loss_class(debug_summaries)``
             optimizer (torch.optim.Optimizer): The optimizer for training
-            checkpoint_path (str): the full path to the checkpoint file saved
-                by ALF, e.g. ``/path_to_experiment/train/algorithm/ckpt-100``.
-            checkpoint_prefix (str): the prefix to the contents in the checkpoint
-                to be loaded. If the checkpoint comes from a previous ALF training
-                session, the standard prefix starts with "alg." (e.g. "alg._sub_alg1").
-                If "", the effects is the same as providing "alg.", which will
-                load the full 'alg' part of the checkpoint.
+            checkpoint (None|str): a string in the format of "prefix@path",
+                where the "prefix" is the multi-step path to the contents in the
+                checkpoint to be loaded. "path" is the full path to the checkpoint
+                file saved by ALF. Refer to ``Algorithm`` for more details.
             debug_summaries (bool): True if debug summaries should be created.
             name (str): Name of this algorithm.
         """
@@ -128,8 +124,7 @@ class ActorCriticAlgorithm(OnPolicyAlgorithm):
             env=env,
             config=config,
             optimizer=optimizer,
-            checkpoint_path=checkpoint_path,
-            checkpoint_prefix=checkpoint_prefix,
+            checkpoint=checkpoint,
             debug_summaries=debug_summaries,
             name=name)
 
--- a/alf/algorithms/actor_critic_algorithm.py
+++ b/alf/algorithms/actor_critic_algorithm.py
@@ -52,6 +52,8 @@ class ActorCriticAlgorithm(OnPolicyAlgorithm):
                  loss=None,
                  loss_class=ActorCriticLoss,
                  optimizer=None,
+                 checkpoint_path=None,
+                 checkpoint_prefix='',
                  debug_summaries=False,
                  name="ActorCriticAlgorithm"):
         """
@@ -92,6 +94,13 @@ class ActorCriticAlgorithm(OnPolicyAlgorithm):
             loss_class (type): the class of the loss. The signature of its
                 constructor: ``loss_class(debug_summaries)``
             optimizer (torch.optim.Optimizer): The optimizer for training
+            checkpoint_path (str): the full path to the checkpoint file saved
+                by ALF, e.g. ``/path_to_experiment/train/algorithm/ckpt-100``.
+            checkpoint_prefix (str): the prefix to the contents in the checkpoint
+                to be loaded. If the checkpoint comes from a previous ALF training
+                session, the standard prefix starts with "alg." (e.g. "alg._sub_alg1").
+                If "", the effects is the same as providing "alg.", which will
+                load the full 'alg' part of the checkpoint.
             debug_summaries (bool): True if debug summaries should be created.
             name (str): Name of this algorithm.
         """
@@ -119,6 +128,8 @@ class ActorCriticAlgorithm(OnPolicyAlgorithm):
             env=env,
             config=config,
             optimizer=optimizer,
+            checkpoint_path=checkpoint_path,
+            checkpoint_prefix=checkpoint_prefix,
             debug_summaries=debug_summaries,
             name=name)
 
--- a/alf/algorithms/actor_critic_algorithm.py
+++ b/alf/algorithms/actor_critic_algorithm.py
@@ -139,6 +139,8 @@ class ActorCriticAlgorithm(OnPolicyAlgorithm):
             loss = loss_class(debug_summaries=debug_summaries)
         self._loss = loss
 
+        self._post_init()
+
     def convert_train_state_to_predict_state(self, state):
         return state._replace(value=())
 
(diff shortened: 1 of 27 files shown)


alf/algorithms/actor_critic_algorithm.py
```diff
@@ -92,6 +94,13 @@ def __init__(self,
             loss_class (type): the class of the loss. The signature of its
                 constructor: ``loss_class(debug_summaries)``
             optimizer (torch.optim.Optimizer): The optimizer for training
+            checkpoint_path (str): the full path to the checkpoint file saved
+                by ALF, e.g. ``/path_to_experiment/train/algorithm/ckpt-100``.
+            checkpoint_prefix (str): the prefix to the contents in the checkpoint
+                to be loaded. If the checkpoint comes from a previous ALF training
+                session, the standard prefix starts with "alg." (e.g. "alg._sub_alg1").
+                If "", the effects is the same as providing "alg.", which will
+                load the full 'alg' part of the checkpoint.
```
hnyu: Is it better to merge these two arguments into one? For example, "alg._sub_alg1@/path/to/checkpoint"?

alf/algorithms/actor_critic_algorithm.py
```diff
@@ -128,6 +139,8 @@ def __init__(self,
             loss = loss_class(debug_summaries=debug_summaries)
         self._loss = loss
 
+        self._post_init()
```
hnyu: Is it possible to automatically call this before load_checkpoint in a recursive way? The current way is really error-prone.

hnyu (changes requested):

alf/algorithms/actor_critic_algorithm.py
Haichao-Zhang: The initial commit implemented auto call: ee8f83e266e452bac3a62872f436c56cc679fed0
Unfortunately, there is a confliction with alf.config (since alf config uses signatures), which is the reason CI failed. 
Also tried another two different approaches for also call but have the same issue. 

Please let me know if you have any suggestions on implementing this auto call.

hnyu: > The initial commit implemented auto call: [ee8f83e](https://github.com/HorizonRobotics/alf/commit/ee8f83e266e452bac3a62872f436c56cc679fed0) Unfortunately, there is a confliction with alf.config (since alf config uses signatures), which is the reason CI failed. Also tried another two different approaches for also call but have the same issue.
> 
> Please let me know if you have any suggestions on implementing this auto call.

Just traverse the entire algorithm tree and call pre-loading for each algorithm at the first line of `PolicyTrainer.train()`? Similar for `PolicyTrainer.play()`.

Haichao-Zhang: > > The initial commit implemented auto call: [ee8f83e](https://github.com/HorizonRobotics/alf/commit/ee8f83e266e452bac3a62872f436c56cc679fed0) Unfortunately, there is a confliction with alf.config (since alf config uses signatures), which is the reason CI failed. Also tried another two different approaches for also call but have the same issue.
> > Please let me know if you have any suggestions on implementing this auto call.
> 
> Just traverse the entire algorithm tree and call pre-loading for each algorithm at the first line of `PolicyTrainer.train()`? Similar for `PolicyTrainer.play()`.

It's doable, but there might be some other use cases not covered by this approach.  E.g. someone simply constructed an algorithm instance with checkpoint arg specified, as in the test case.

hnyu: > > > The initial commit implemented auto call: [ee8f83e](https://github.com/HorizonRobotics/alf/commit/ee8f83e266e452bac3a62872f436c56cc679fed0) Unfortunately, there is a confliction with alf.config (since alf config uses signatures), which is the reason CI failed. Also tried another two different approaches for also call but have the same issue.
> > > Please let me know if you have any suggestions on implementing this auto call.
> > 
> > 
> > Just traverse the entire algorithm tree and call pre-loading for each algorithm at the first line of `PolicyTrainer.train()`? Similar for `PolicyTrainer.play()`.
> 
> It's doable, but there might be some other use cases not covered by this approach. E.g. someone simply constructed an algorithm instance with checkpoint arg specified, as in the test case.

It's just like `compile()` in hobot. The point is that, when needed, we only need to call once on the root algorithm to preload the checkpoints.

alf/algorithms/actor_critic_algorithm.py
Haichao-Zhang: > Is it better to merge these two arguments into one? For example, "alg._sub_alg1@/path/to/checkpoint"?

Done

alf/algorithms/actor_critic_algorithm.py
Haichao-Zhang: Now used auto post init call, which will handle checkpoint loading automatically.

Haichao-Zhang:
Thanks for the suggestion from @emailweixu, now resolved the conflict with alf.config and can automatically call the post init operations, which will load the checkpoint.

alf/algorithms/algorithm.py
```diff
@@ -171,6 +185,96 @@ def __init__(self,
...
+                    # checkpoint file filter
+                    prefix = _remove_prefix(checkpoint_prefix, 'alg.')
+                    checkpoint = {
+                        _remove_prefix(k, prefix + '.'): v
+                        for k, v in checkpoint.items() if k.startswith(prefix)
+                    }
+
+                self.load_state_dict(checkpoint, strict=True)
+                self._checkpoint_pre_loaded = True
+            except:
```
emailweixu: Avoid using except without specifying exception type.

alf/algorithms/algorithm.py
```diff
@@ -171,6 +185,96 @@ def __init__(self,
...
+                raise RuntimeError(
```
emailweixu: This doesn't seem to add any value. In fact, the original exception may be useful for understanding the error.

alf/algorithms/algorithm.py
```diff
@@ -171,6 +185,96 @@ def __init__(self,
...
+                if checkpoint_prefix != '':
+                    # the case when the checkpoint is a subset of the full
+                    # checkpoint file filter
+                    prefix = _remove_prefix(checkpoint_prefix, 'alg.')
+                    checkpoint = {
+                        _remove_prefix(k, prefix + '.'): v
+                        for k, v in checkpoint.items() if k.startswith(prefix)
+                    }
+
+                self.load_state_dict(checkpoint, strict=True)
```
emailweixu: This part can be an independent function. Loading a module by prefix from a checkpoint file may be useful in other context.

alf/algorithms/algorithm.py
Haichao-Zhang: Agree. Removed it.

alf/algorithms/algorithm.py
Haichao-Zhang: Removed

alf/algorithms/algorithm.py
Haichao-Zhang: Good suggestion. Extracted it as an independent function.

alf/algorithms/algorithm.py
```diff
@@ -171,6 +186,69 @@ def __init__(self,
...
+            assert len(prefix_and_path) in [1,
+                                            2], ("invalid checkpoint: "
+                                                 "{}").format(prefix_and_path)
+
+            if len(prefix_and_path) == 1:
+                # only path is provided
+                checkpoint_path = prefix_and_path[0]
+                checkpoint_prefix = 'alg'
+            else:
+                # only path is provided
```
hnyu: comment is incorrect

alf/algorithms/bc_algorithm.py
```diff
@@ -53,6 +53,8 @@ def __init__(self,
                  actor_optimizer=None,
                  env=None,
                  config: TrainerConfig = None,
+                 checkpoint=None,
+                 checkpoint_prefix='',
```
hnyu: should remove checkpoint_prefix

alf/algorithms/dynamics_learning_algorithm.py
```diff
@@ -45,6 +45,8 @@ def __init__(self,
                  hidden_size=256,
                  num_replicas=1,
                  dynamics_network: DynamicsNetwork = None,
+                 checkpoint=None,
+                 checkpoint_prefix='',
```
hnyu: should remove checkpoint_prefix

alf/algorithms/trac_algorithm.py
```diff
@@ -61,13 +61,19 @@ def __init__(self,
                  config=None,
                  ac_algorithm_cls=ActorCriticAlgorithm,
                  action_dist_clip_per_dim=0.01,
+                 checkpoint=None,
+                 checkpoint_prefix='',
```
hnyu: should remove checkpoint_prefix

alf/utils/checkpoint_utils.py
```diff
@@ -47,6 +47,51 @@ def enable_checkpoint(module, flag=True):
     module._alf_checkpoint_enabled = flag
 
 
+def extract_sub_state_dict_from_checkpoint(checkpoint_prefix, checkpoint_path):
+    """Extract a (sub-)state-dictionary from a checkpoint file. The state
+    dictionary is can be a sub-dictionary specified by the
```
hnyu: remove 'is'

hnyu (changes requested):

alf/algorithms/algorithm.py
Haichao-Zhang: Removed

alf/algorithms/bc_algorithm.py
Haichao-Zhang: Removed

alf/algorithms/dynamics_learning_algorithm.py
Haichao-Zhang: Removed

alf/algorithms/trac_algorithm.py
Haichao-Zhang: Removed

alf/utils/checkpoint_utils.py
Haichao-Zhang: Removed

hnyu (approved):

PR #1431, opened 2023-02-16
normalized action alf wrapper; transform_nest supports numeric field name

hnyu:
Added an alf wrapper for normalizing actions. Since our Hobot project no longer has gym envs, I implemented a similar alf wrapper.

### DIFF
--- a/alf/environments/alf_wrappers.py
+++ b/alf/environments/alf_wrappers.py
@@ -1000,7 +1000,8 @@ class NormalizedActionWrapper(AlfEnvironmentBaseWrapper):
     distribution networks can do this, is because we want to set target entropy
     independent of action ranges for algorithms like SAC.
 
-    This wrapper is expected to be used only for individual envs (numpy array).
+    This wrapper can be used only for individual envs (numpy array) or a batched
+    env (tensor).
     """
 
     def __init__(self, env: AlfEnvironment):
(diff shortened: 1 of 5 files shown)


alf/environments/alf_wrappers.py
```diff
@@ -991,3 +991,56 @@ def _step(self, action):
...
+
+
+class NormalizedActionWrapper(AlfEnvironmentBaseWrapper):
+    """Normalize actions into [-1,1].
+
+    The reason why we'd like to normalize the actions, even though our action
+    distribution networks can do this, is because we want to set target entropy
+    independent of action ranges for algorithms like SAC.
+
+    This wrapper is expected to be used only for individual envs (numpy array).
```
emailweixu: It can also be used for the batched env?

alf/environments/alf_wrappers.py
```diff
@@ -991,3 +991,56 @@ def _step(self, action):
...
+    def action_spec(self):
+        return self._action_spec
+
+    def time_step_spec(self):
+        return self._time_step_spec
+
+    def _step(self, action):
+        def _scale_back(a, bound):
+            b0, b1 = bound
+            return (a + 1.) / 2. * (b1 - b0) + b0
```
emailweixu: A slightly faster way is `0.5 * (b1 - b0) * (a+1)  + b0`  since this has only one scalar-Tensor multiplication/division . A even faster way is to cache `b=0.5*(b1-b0)` and `c=b0+b` and conversion will be simply  `b*a+c`. We should try to make our code fast as long as it does not hurt readability much.

emailweixu (approved):

PR #1433, opened 2023-02-22
Persistent displaying of pid

Haichao-Zhang:
PID of the training process is useful for operations such as requesting checkpoint.
However, during the middle of the training, we need to check the PID first, which is a bit inconvenient, especially for the cases of multiple parallel environment or multiple training jobs on the same machine.

This PR displays the pid persistently in the logging message for convenience:

image

### DIFF
--- a/alf/trainers/policy_trainer.py
+++ b/alf/trainers/policy_trainer.py
@@ -285,23 +285,27 @@ class Trainer(object):
         self._random_seed = config.random_seed
         self._rank = ddp_rank
         self._conf_file_content = common.read_conf_file(root_dir)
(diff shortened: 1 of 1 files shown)


hnyu:
In our log file, there is already an instruction in the beginning showing how to save a checkpoint?

Haichao-Zhang:
> In our log file, there is already an instruction in the beginning showing how to save a checkpoint?

Yes, this is to save the trouble to open the log and check, by "persistent displaying of pid", particular useful for the case running on a local machine.

breakds (approved):

PR #1437, opened 2023-03-06
Pr alf upgrade

hnyu:
1. update alf's cicd docker to use ubuntu20.04 (github actions no longer support 18.04)
2. upgrade alf's dependencies to use cuda 11.4
3. provide a docker image for users to easily run ALF jobs

The new setup.py has been tested locally.

The new alf version is updated to 1.1.0

### DIFF
--- a/.ci-cd/Dockerfile.cpu
+++ b/.ci-cd/Dockerfile.cpu
@@ -12,16 +12,20 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM ubuntu:18.04
+FROM ubuntu:20.04
+
+ENV TZ=US/Pacific
+RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
 
 RUN apt update
 
-RUN apt install -y python3.7 python3-pip python3.7-dev python3-setuptools
+# This should install python3.8
+RUN apt install -y python3 python3-pip python3-dev python3-setuptools
 
-RUN python3.7 -m pip install --upgrade pip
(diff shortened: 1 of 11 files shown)


emailweixu (dismissed):

hnyu:
@emailweixu Now for several conf files, when the training finishes normally, there will be a segfault at the end. The training process itself is absolutely fine, but the returncode is no longer zero. This causes train_play_test.py to fail.  For example, 

```python
python -m alf.bin.train --root_dir /tmp/tmp --conf=mbrl_latent_pendulum_conf.py --conf_param="TrainerConfig.num_iterations=2" --nostore_snapshot
```

image


Currently, I don't know a good way to address this. This issue happens only after the software upgrade.

hnyu:
@emailweixu Another issue with unknown reason is that some `self.assertRaises` places cause segfaults even after passing the test. probably has something to do with python3.8 and pybind. I've commented out them. 

https://github.com/HorizonRobotics/alf/pull/1437/commits/853dd65b9ee53eb7aa01a3e6f6af856bd8388b7e#diff-9f47127437483d883eff8a2cf249a139b623ee8a5777cd705cd300fec1410cfbR98

hnyu:
@emailweixu  With the help of @breakds , found a bug in `nest.cpp` which caused the segfaults. Now fixed and upgraded cnest to 1.1.1. https://github.com/HorizonRobotics/cnest/commit/b7a62849ac4531225229cff3a5d5f8fc654bda3f 

The CI test has finally passed on my forked ALF repo here: https://github.com/hnyu/alf/tree/test_cicd_docker

emailweixu (approved):

PR #1438, opened 2023-03-07
Pr save checkpoint structure

Haichao-Zhang:
This PR also saves a json file containing the structure of the model state checkpoint, which facilitates inspecting the structure of the checkpoint without having to load it first. 
This is useful for cases such as extracting a sub-dictionary from the whole (e.g. in in-algorithm checkpoint loading).

### DIFF
--- a/alf/utils/checkpoint_utils.py
+++ b/alf/utils/checkpoint_utils.py
@@ -14,6 +14,7 @@
 
 from absl import logging
 import glob
+import json
 import os
 import torch
 from torch import nn
@@ -95,7 +96,10 @@ class Checkpointer(object):
     """A checkpoint manager for saving and loading checkpoints."""
 
     def __init__(self, ckpt_dir, **kwargs):
-        """A class for making checkpoints.
+        """A class for saving checkpoints. It also saves a json file containing
+        the structure of the model state checkpoint, which facilitates inspecting
+        the structure of the checkpoint without having to load it first. This is
+        useful for cases such as extracting a sub-dictionary from the whole.
 
         Example usage:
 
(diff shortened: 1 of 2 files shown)


alf/utils/checkpoint_utils.py
```diff
@@ -348,5 +352,32 @@ def save(self, global_step):
         torch.save(optimizer_state, f_path + '-optimizer')
         torch.save(replay_buffer_state, f_path + '-replay_buffer')
 
+        if self._global_step == -1:
+
+            def _dummy_value(v):
```
hnyu: map_structure will always reach a leaf node. so this check is unnecessary

alf/utils/checkpoint_utils.py
```diff
@@ -348,5 +352,32 @@ def save(self, global_step):
         torch.save(optimizer_state, f_path + '-optimizer')
         torch.save(replay_buffer_state, f_path + '-replay_buffer')
 
+        if self._global_step == -1:
```
hnyu, line 355: When will this `global_step` be equal to -1? In your test case, it starts with 0.

hnyu (changes requested):

alf/utils/checkpoint_utils.py
Haichao-Zhang: Changed

alf/utils/checkpoint_utils.py
Haichao-Zhang: > When will this `global_step` be equal to -1? In your test case, it starts with 0.

`global_step` is initialized as -1. Therefore, for the first time of checkpoint saving (``if self._global_step == -1``), we will also save the checkpoint structure. This is used to save the structure only once.

hnyu (approved):

PR #1439, opened 2023-03-10
Check if the current thread is main thread before registering signal handler

PinZhang:
Tried to start alf training process in the Prefect flow engine, it failed because `signal.signal` only works in the main thread.

### DIFF
--- a/alf/trainers/policy_trainer.py
+++ b/alf/trainers/policy_trainer.py
@@ -22,6 +22,7 @@ import pprint
 from pathlib import Path
 import re
 import signal
+import threading
 import sys
 import time
 import torch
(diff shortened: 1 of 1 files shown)


PinZhang:
[Python Signals and threads](https://docs.python.org/3/library/signal.html#:~:text=Python%20signal%20handlers%20are%20always,from%20the%20threading%20module%20instead.)

Signals and threads
Python signal handlers are always executed in the main Python thread of the main interpreter, even if the signal was received in another thread. This means that signals can’t be used as a means of inter-thread communication. You can use the synchronization primitives from the [threading](https://docs.python.org/3/library/threading.html#module-threading) module instead.

Besides, only the main thread of the main interpreter is allowed to set a new signal handler.

breakds:
Thanks!

breakds (approved):

PR #1442, opened 2023-03-15
FastParallelEnvironment

emailweixu:
Design:

`FastParallelEnvironment` uses `_penv.ParallelEnvironment` (implemented in C++)    to coordinate step() and reset(). Each ``ProcessEnvironment`` maintains one ``_penv.ProcessEnvironmentCaller`` in the main process and one ``_penv.ProcessEnvironment`` in the client process.

In the client process, ``_penv.ProcessEnvironment.worker()`` runs in a loop to wait for jobs from either ``_penv.ParallelEnvironment`` or ``_penv.ProcessEnvironmentCaller``.

There are 4 types of job:

- step: step the environment. Sent from ``_penv.ParallelEnvironment``. The result is communicated back using shared memory.

- reset: reset the environment. Sent from ``_penv.ParallelEnvironment``. The result is communicated back using shared memory.

- close: close the environment. Sent from ``_penv.ProcessEnvironmentCaller``. This will cause the worker to finish and quit the process.

- call: access other methods of the environment. Sent from ``_penv.ProcessEnvironmentCaller``. This takes advantage of the pipe mechanism used by  the ``ParallelAlfEnvironment``. This is achieved by calling ``call_handler`` to do communication using python pipe. The reason of using the original pipe mechanism for other types of communication is that it is not easy to handle communication of unknow size using shared memory. 

Tested with gym breakout and hobot low level control env, the speed up is about 15%. Not as fast as what I expected.

### DIFF
--- a/alf/environments/parallel_environment.cpp
+++ b/alf/environments/parallel_environment.cpp
@@ -1,0 +1,501 @@
+// Copyright (c) 2023 Horizon Robotics and ALF Contributors. All Rights
+// Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/*
+Boost should be installed first: apt install libboost-all-dev
+Compile with the following command:
+
+PYTHON=python3.8
+g++ -O3 -Wall -shared -std=c++17 -fPIC -fvisibility=hidden -pg \
+ `$PYTHON -m pybind11 --includes` \
+  parallel_environment.cpp \
+  -o _penv`$PYTHON-config --extension-suffix` -lrt
+
+After compilation, a python extension module _penv will be generated, which
+can be imported using "import _penv" from python.
+*/
+
+#include <pybind11/numpy.h>
+#include <pybind11/pybind11.h>
+#include <pybind11/stl.h>
+
+#include <iostream>
+#include <memory>
+#include <string>
+#include <vector>
+
+#include <boost/interprocess/ipc/message_queue.hpp>
+#include <boost/interprocess/shared_memory_object.hpp>
+
+namespace bip = boost::interprocess;
+namespace py = pybind11;
+
+const size_t addr_alignment = 16;
+
+inline size_t align(size_t x) {
+  return ((x + addr_alignment - 1) / addr_alignment) * addr_alignment;
+}
+
+// TODO(emailweixu): directly use cnest c++ function
+std::vector<py::object> Flatten(py::object nest) {
+  py::object flatten = py::module::import("cnest").attr("flatten");
+  auto list = py::list(flatten(nest));
+  std::vector<py::object> ret;
+  ret.reserve(list.size());
+  for (size_t i = 0; i < list.size(); ++i) {
+    ret.push_back(list[i]);
+  }
+  return ret;
+}
+
+// TODO(emailweixu): directly use cnest c++ function
+py::object PackSequenceAs(py::object nest,
+                          const std::vector<py::object>& flat_seq) {
+  py::object pack_sequence_as =
+      py::module::import("cnest").attr("pack_sequence_as");
+  py::list list;
+  for (auto item : flat_seq) {
+    list.append(item);
+  }
+  return pack_sequence_as(nest, flat_seq);
+}
+
+class SharedDataBuffer {
+ public:
+  SharedDataBuffer(py::object data_spec,
+                   size_t num_slices,
+                   const std::string& name);
+  ~SharedDataBuffer();
+  py::object Read(size_t slice_id);
+  void Write(py::object nested_array, size_t slice_id);
+  py::object Read(const std::vector<size_t>& slice_id);
+  void Write(py::object nested_array);
+
+  inline py::object ViewAsNestedArray(size_t slice_id, size_t num_slices);
+  std::vector<py::object> ViewAsFlattenedArray(size_t slice_id,
+                                               size_t num_slices);
+
+  void* GetBufByFieldName(const char* field) {
+    auto array = py::buffer(ViewAsNestedArray(0, num_slices_).attr(field));
+    py::buffer_info info = array.request();
+    return info.ptr;
+  }
+
+ private:
+  std::string name_;
+  size_t num_slices_;
+  py::object data_spec_;
+  std::unique_ptr<bip::shared_memory_object> shm_;
+  std::unique_ptr<bip::mapped_region> region_;
+  char* buf_;
+  std::vector<size_t> offsets_;  // offset of the i-th array
+  std::vector<size_t> sizes_;    // size of each slice of the i-th array
+  std::vector<py::buffer_info> buffer_infos_;
+
+  inline char* GetBuf(int slice_id, int array_id) const {
+    return buf_ + offsets_[array_id] + sizes_[array_id] * slice_id;
+  }
+};
+
+void SharedDataBuffer::Write(py::object nested_array, size_t slice_id) {
+  auto arrays = Flatten(nested_array);
+  for (size_t j = 0; j < arrays.size(); ++j) {
+    auto buffer = py::buffer(arrays[j]);
+    const py::buffer_info& info = buffer.request();
+    if (info.ndim == 0) {
+      memcpy(GetBuf(slice_id, j), info.ptr, info.itemsize);
+    } else {
+      assert(info.strides[0] * info.shape[0] == (signed)sizes_[j]);
+      memcpy(GetBuf(slice_id, j), info.ptr, info.shape[0] * info.strides[0]);
+    }
+  }
+}
+
+void SharedDataBuffer::Write(py::object nested_array) {
+  auto arrays = Flatten(nested_array);
+  for (size_t j = 0; j < arrays.size(); ++j) {
+    auto buffer = py::buffer(arrays[j]);
+    const py::buffer_info& info = buffer.request();
+    assert(info.strides == buffer_infos_[j].strides);
+    assert(info.shape[0] == (signed)num_slices_);
+    memcpy(GetBuf(0, j), info.ptr, info.shape[0] * info.strides[0]);
+  }
+}
+
+SharedDataBuffer::SharedDataBuffer(py::object data_spec,
+                                   size_t num_slices,
+                                   const std::string& name)
+    : name_(name), num_slices_(num_slices), data_spec_(data_spec) {
+  auto outer_dims = py::make_tuple(2);
+
+  auto flattened_data_spec = Flatten(data_spec);
+  size_t offset = 0;
+  for (size_t i = 0; i < flattened_data_spec.size(); ++i) {
+    offsets_.push_back(offset);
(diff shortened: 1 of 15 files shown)


alf/environments/process_environment.py
```diff
@@ -45,7 +47,13 @@ class _MessageType(Enum):
     CLOSE = 6
 
 
-def _worker(conn, env_constructor, env_id=None, flatten=False):
+def _worker(conn,
+            env_constructor,
+            env_id=None,
+            flatten=False,
+            fast=False,
```
breakds, line 54: Need to update the docstring.

alf/environments/process_environment.py
```diff
@@ -110,8 +99,51 @@ def _worker(conn, env_constructor, env_id=None, flatten=False):
         conn.close()
 
 
+def process_call(conn, env, flatten, action_spec):
```
breakds, line 107: As far as I understand, `handle_process_call` sounds like a more accurate name?

alf/environments/process_environment.py
```diff
@@ -110,8 +99,51 @@ def _worker(conn, env_constructor, env_id=None, flatten=False):
...
+
 class ProcessEnvironment(object):
-    def __init__(self, env_constructor, env_id=None, flatten=False):
+    def __init__(self,
+                 env_constructor,
+                 env_id=None,
+                 flatten=False,
+                 fast=False,
+                 num_envs=0,
+                 name=""):
```
breakds, line 151: Update docstring.

alf/environments/parallel_environment.cpp
```diff
@@ -0,0 +1,501 @@
...
+      reinterpret_cast<int32_t*>(timestep_buffer_.GetBufByFieldName("env_id"));
+}
+
+ProcessEnvironment::~ProcessEnvironment() {
+  bip::message_queue::remove(
+      (name_ + ".job_queue." + std::to_string(env_id_)).c_str());
+  bip::message_queue::remove((name_ + ".ready_queue").c_str());
+}
+
+class ProcessEnvironmentCaller {
```
breakds, line 390: I think the name `Caller` is a bit confusing - how about `ProcessEnvironmentProxy` or `ProcessEnvironmentClient` for better readability for the future reader?

alf/environments/parallel_environment.cpp
```diff
@@ -0,0 +1,501 @@
...
+
+// TODO(emailweixu): directly use cnest c++ function
+std::vector<py::object> Flatten(py::object nest) {
+  py::object flatten = py::module::import("cnest").attr("flatten");
+  auto list = py::list(flatten(nest));
+  std::vector<py::object> ret;
+  ret.reserve(list.size());
+  for (size_t i = 0; i < list.size(); ++i) {
+    ret.push_back(list[i]);
+  }
```
breakds: Why is it necessary to convert the `py::list` to `std::vector`?

alf/environments/parallel_environment.cpp
```diff
@@ -0,0 +1,501 @@
...
+    reset_queue_.push_back(num_envs + i);
+  }
+  step_type_buf_ = reinterpret_cast<int32_t*>(
+      timestep_buffer_.GetBufByFieldName("step_type"));
+
+  for (int i = 0; i < num_envs + num_spare_envs; ++i) {
+    job_queues_.emplace_back(std::make_unique<bip::message_queue>(
+        bip::open_or_create,
+        (name + ".job_queue." + std::to_string(i)).c_str(),
+        3,
```
breakds: How is `3` picked for max number of messages?

alf/environments/parallel_environment.cpp
```diff
@@ -0,0 +1,501 @@
...
+    : name_(name),
+      num_envs_(num_envs),
+      action_buffer_(action_spec, num_envs, name + ".action"),
+      timestep_buffer_(timestep_spec, num_envs, name + ".timestep") {}
+
+class ParallelEnvironment : public EnvironmentBase {
+ protected:
+  int num_spare_envs_;
+  std::vector<int> original_env_ids_;  // [num_envs]
+  std::vector<uint8_t> reseted_;       // [num_envs]
```
breakds, line 215: Is there something wrong if we type this as `vector<bool>`?

alf/environments/parallel_environment.cpp
```diff
@@ -0,0 +1,501 @@
...
+    Job job{JobType::step, env_id};
+    job_queues_[original_env_ids_[env_id]]->send(&job, sizeof(job), 0);
+  }
+  for (int i = 0; i < num_envs_; ++i) {
+    bip::message_queue::size_type recvd_size;
+    unsigned int priority;
+    int env_id;
+    ready_queue_.receive(&env_id, sizeof(env_id), recvd_size, priority);
+    assert(recvd_size == sizeof(env_id));
+    if (step_type_buf_[env_id] == (int32_t)StepType::last) {
```
breakds: Nit: `static_cast` over direct cast.

alf/environments/process_environment.py
breakds: Sorry, I actually mistook `process` (as a verb) with `process` (as a noun, the program process). Still, `handle_call` would be better.

alf/environments/parallel_environment.cpp
```diff
@@ -0,0 +1,501 @@
...
+    assert(recvd_size == sizeof(job));
+    if (job.type == JobType::step) {
+      Step(job.env_id);
+    } else if (job.type == JobType::reset) {
+      Reset();
+    } else if (job.type == JobType::close) {
+      env_.attr("close")();
+      break;
+    } else if (job.type == JobType::call) {
+      call_handler_();
```
breakds, line 462: I am still a bit confused by this. It seems that `call_handler_()` is just the original python `ProcessEnvironment`'s handler. Why do we still need it? This has a connection with a follow-up question below.

alf/environments/process_environment.py
```diff
@@ -214,6 +252,8 @@ def __getattr__(self, name):
             Value of the attribute.
         """
         assert self._conn, "Run ProcessEnvironment.start() first"
+        if self._fast:
+            self._penv.call()
```
breakds, line 266: This is a part that is unclear to me. By design, if we delegate all the `main process` / `child process` communication to shared memory buffers in c++, why do we still need to use `Pipe` in fast mode? To put my question in a better way, isn't it possible to make the design such that Pipe is no longer needed in fast mode?

breakds (commented):
Sorry it took quite some time to read through the code and try to find answers to my questions. Here are more comments/questions.

alf/environments/process_environment.py
emailweixu: Updated the description about "call" job in the PR description.

alf/environments/parallel_environment.cpp
emailweixu: The reason for this name is that it only send "call" job to `_penv.ProcessEnvironment`

setup.py
```diff
@@ -15,6 +15,9 @@
 from setuptools import setup, find_packages
 import os
 
+assert os.system(
+    "python3 alf/environments/make_penv.py") == 0, "Fail to build _penv"
```
hnyu: Does this work with pypi package or pip install from ALF repo? A setup file like cnest might be needed https://github.com/HorizonRobotics/cnest/blob/master/setup.py . Feel like a separate penv python repo could be created.

alf/environments/parallel_environment.cpp
emailweixu: changed to py::list.

alf/environments/parallel_environment.cpp
emailweixu: vector<bool> behaves differently from other vector (https://cplusplus.com/reference/vector/vector-bool/). It's more like bit array, which is slower.

alf/environments/parallel_environment.cpp
emailweixu: comment added.

alf/environments/parallel_environment.cpp
emailweixu: updated.

alf/environments/parallel_environment.cpp
emailweixu: Use call_handler to handle other types of communication with unknown size.

alf/environments/process_environment.py
emailweixu: udpated

alf/environments/process_environment.py
emailweixu: fixed.

alf/environments/parallel_environment_test.py
```diff
@@ -173,10 +178,10 @@ def test_non_blocking_reset_with_spare_envs(self):
         time_step4 = env.step(action)
         step4_t = time.time()
         step4_time = step4_t - step3_t
-        self.assertGreaterEqual(
+        self.assertLessEqual(
             step4_time,
-            sleep_time - 0.01,
-            msg=(f'Step without spare envs took {step4_time}, too short'))
+            sleep_time - 0.1,
+            msg=(f'Step witho spare envs took {step4_time}, too short'))
```
hnyu: Why changing to assertLessEqual but the message is still 'short'?

alf/environments/process_environment.py
breakds: Very clear now. Thank you!

setup.py
emailweixu: fixed

alf/environments/parallel_environment_test.py
emailweixu: fixed

breakds (dismissed):

alf/environments/parallel_environment.cpp
```diff
@@ -44,31 +44,25 @@ namespace py = pybind11;
...
+// The size of job queue.
+// Since we always wait for job to finish before sending the next job,
+// we only need a shoft job queue.
+const size_t kJobQueueSize = 3;
+
 inline size_t align(size_t x) {
   return ((x + addr_alignment - 1) / addr_alignment) * addr_alignment;
 }
 
 // TODO(emailweixu): directly use cnest c++ function
```
hnyu, line 56: remove TODO?

alf/environments/parallel_environment.cpp
```diff
@@ -44,31 +44,25 @@ namespace py = pybind11;
...
-  std::vector<py::object> ret;
-  ret.reserve(list.size());
-  for (size_t i = 0; i < list.size(); ++i) {
-    ret.push_back(list[i]);
-  }
-  return ret;
+  return py::list(flatten(nest));
 }
 
 // TODO(emailweixu): directly use cnest c++ function
```
hnyu, line 62: remove TODO?

setup.py
```diff
@@ -57,6 +54,15 @@
...
+            'environments._penv',
+            sources=['alf/environments/parallel_environment.cpp'],
+            extra_compile_args=[
+                '-O3 -Wall -std=c++17', '-fPIC', '-fvisibility=hidden'
+            ])
+    ],
+    cmdclass={'build_ext': build_ext},
     extras_require={
         'metadrive': ['metadrive-simulator==0.2.5.1', ],
     },
```
hnyu, line 68: You need to include the compiled so file in the 'package_data'? Otherwise an installed alf package will miss the so file.

alf/environments/parallel_environment.cpp
breakds: I think Wei means using the one directly from c++, not via python binding.

emailweixu: Yes, that's what I mean.

setup.py
emailweixu: it is handled automatically

hnyu (dismissed):

hnyu (approved):

breakds (approved):

PR #1446, opened 2023-03-16
Render for parallel environment

le-horizon:
An enhancement:
[render first env in parallel env](https://github.com/HorizonRobotics/alf/commit/3aa0fcbfadfc4aadfbc18e5a09ac0f7e92768bc4)

### DIFF
--- a/alf/environments/fast_parallel_environment.py
+++ b/alf/environments/fast_parallel_environment.py
@@ -201,12 +201,19 @@ class FastParallelEnvironment(alf_environment.AlfEnvironment):
     def metadata(self):
         return self._envs[0].metadata
 
-    @property
-    def render_mode(self):
-        if hasattr(self._envs[0], "render_mode"):
-            return self._envs[0].render_mode
-        else:
-            return None
+    # Adding this function causes errors in bipedal_walker etc. envs
+    # when training starts:
+    # AttributeError: 'BipedalWalker' object has no attribute 'render_mode'
+    # and
+    # File "/home/lezhao/alf/alf/environments/fast_parallel_environment.py", line 224, in _step
+    #     return self._to_tensor(self._penv.step(action))
+    # RuntimeError: ProcessEnvironment is interruptted
+    # @property
+    # def render_mode(self):
+    #     if hasattr(self._envs[0], "render_mode"):
+    #         return self._envs[0].render_mode
+    #     else:
+    #         return None
 
     def _to_tensor(self, stacked):
         # we need to do np.copy because the result from _penv.step() or
--- a/alf/environments/fast_parallel_environment.py
+++ b/alf/environments/fast_parallel_environment.py
@@ -194,6 +194,13 @@ class FastParallelEnvironment(alf_environment.AlfEnvironment):
     def time_step_spec(self):
         return self._time_step_spec
 
+    def render(self, mode="rgb_array"):
+        return self._envs[0].render(mode)
+
+    @property
+    def metadata(self):
+        return self._envs[0].metadata
+
     def _to_tensor(self, stacked):
         # we need to do np.copy because the result from _penv.step() or
         # _penv.reset() reuses the same internal buffer.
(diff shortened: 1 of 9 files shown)


alf/environments/parallel_environment.py
```diff
@@ -157,6 +157,9 @@ def reward_spec(self):
     def time_step_spec(self):
         return self._time_step_spec
 
+    def render(self, mode="rgb_array"):
```
breakds, line 156: Why do we need to do parallel  environment render?

le-horizon: Turns out Wei is correct.  We cannot render to rgb using mujoco egl and then start a pygame display at the same time, unless the environment is in a separate process.
alf.environment.create_environment creates parallelenv wrapped processenv.  That's why ParaEnv needs to pass render along from the child process.

emailweixu: Perhaps you can directly use ProcessEnvironment instead of ParallelAlfEnvironment?

le-horizon:
I took the other bug fix out of this PR, for faster reviewing & merge into pytorch.

Discussed with Break, for this change, I will try to skip parallelenv when num_envs == 1, in alf.create_environment.  That should be sufficient.

emailweixu:
> I took the other bug fix out of this PR, for faster reviewing & merge into pytorch.
> 
> Discussed with Break, for this change, I will try to skip parallelenv when num_envs == 1, in alf.create_environment. That should be sufficient.

You cannot do that, create_environment with num_envs ==1 will get a batched environment with batch_size=1. While ProcessEnvironment is not batched.

alf/environments/parallel_environment.py
le-horizon: I tried using ProcessEnvironment directly.

There is a lot of logic inside ParallelEnvironment that needs to be copied: batching, tensor convert from np to torch and from cpu to GPU, (and also custom logic to handle accessing class members like "wrapped_env" and "console_info" etc..)  Uglier than the current change to ParallelEnv, which is Ok and not really a hack, since render and metadata member functions typically are part of an Environment.

I could separate such duplicate logic into a ProcessAdaptorEnvironment, so ParallelEnv will be more dedicated to proc start & stop, and ProcAdaptorEnv can be easily used standalone (wrapping outside ProcessEnvironment).  But that incurs additional compute, several nest loops.  However, there still needs to be some custom logic handling "render" and "metadata", which means this current change seems unavoidable.

Overall, this current change seems the best option so far.  Let me know if you have any better idea, @emailweixu

le-horizon:
Updated the change to include fast parallel env.
Looks like this is the easiest change to allow rendering from a mujoco env with rgb observation into a GUI.
The ProcessEnvironment path would need a lot of duplication.

Let's merge this? @emailweixu @breakds

emailweixu (dismissed):

le-horizon:
Fixed unit test, PTAL @emailweixu  Thanks.

alf/environments/fast_parallel_environment.py
```diff
@@ -194,6 +194,27 @@ def reward_spec(self):
...
+    # and
+    # File "/home/lezhao/alf/alf/environments/fast_parallel_environment.py", line 224, in _step
+    #     return self._to_tensor(self._penv.step(action))
+    # RuntimeError: ProcessEnvironment is interruptted
+    # @property
+    # def render_mode(self):
+    #     if hasattr(self._envs[0], "render_mode"):
+    #         return self._envs[0].render_mode
+    #     else:
+    #         return None
```
emailweixu: remove

alf/environments/parallel_environment.py
```diff
@@ -153,6 +153,27 @@ def reward_spec(self):
...
+    # and
+    # File "/home/lezhao/alf/alf/environments/fast_parallel_environment.py", line 224, in _step
+    #     return self._to_tensor(self._penv.step(action))
+    # RuntimeError: ProcessEnvironment is interruptted
+    # @property
+    # def render_mode(self):
+    #     if hasattr(self._envs[0], "render_mode"):
+    #         return self._envs[0].render_mode
+    #     else:
+    #         return None
```
emailweixu: remove

le-horizon: Done

emailweixu (approved):

PR #1448, opened 2023-03-18
implement bipartite matching loss

hnyu:
This loss is for evaluating the matching between two order-invariant sets. See the DETR paper for details.

### DIFF
--- a/alf/utils/losses.py
+++ b/alf/utils/losses.py
@@ -17,7 +17,8 @@ import torch
 from torch import Tensor
 import torch.nn as nn
 import torch.nn.functional as F
-from typing import Optional, Tuple
+from typing import Optional, Tuple, Callable
+from scipy.optimize import linear_sum_assignment
 
 import alf
 from alf.utils.math_ops import InvertibleTransform, binary_neg_entropy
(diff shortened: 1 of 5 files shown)


hnyu:
Turns out the cicd docker has an old version of torch which breaks the test case. Upgraded.

alf/utils/losses.py
```diff
@@ -682,3 +683,95 @@ def forward(self, pred: Tensor, target: Tensor) -> Tensor:
...
+
+        # mask out any cost with mask values=0
+        if target_mask is not None:
+            target_mask = target_mask.unsqueeze(1)  # [B,1,N]
+            cost_mat = cost_mat * target_mask
+
+        losses = []
+        np_cost_mat = cost_mat.detach().cpu().numpy()
+        # For every sample in the batch, we compute the optimal assignment
+        for i in range(np_cost_mat.shape[0]):
```
hnyu, line 764: Might need to have a pybind module to do this loop.

alf/utils/losses.py
```diff
@@ -682,3 +683,132 @@ def forward(self, pred: Tensor, target: Tensor) -> Tensor:
...
+                loss. For the former two, the loss shape is ``[B]``, while for
+                the 'none', the loss shape is ``[B,N]``.
+        """
+        super().__init__()
+        self._pair_loss_fn = pair_loss_fn
+        self._reduction = reduction
+        assert reduction in ['mean', 'sum', 'none']
+        self._name = name
+
+    def forward(self,
```
Haichao-Zhang, line 734: Why there is a ``forward `` function, ``fast_forward`` and ``__call__``?

emailweixu: same question. What is the difference of forward and fast_forward? Why do we need both of them?

hnyu: > same question. What is the difference of forward and fast_forward? Why do we need both of them?

fast_forward constructs a bug cost matrix across the entire batch. The cost matrix is block diagonal so that objects can only be matched to those in the same batch sample. forward uses a for loop to call linear_sum_assignment multiple times, which is about 2x slower. I've removed the slow version.

Haichao-Zhang (approved):

PR #1450, opened 2023-03-20
pprint namedtuple with py highlight

hnyu:
The default pprint can't handle namedtuple format. Copied a solution from stack overflow to solve this issue. 

image


hnyu:
The ci test depends on PR #1448

emailweixu (approved):

PR #1451, opened 2023-03-20
Update penv's pybind11 module build

breakds:
Currently the name parameter in the `Pybind11Extension` is set to `'environments._penv'`. This causes the compiled `.so` file to be placed in the environments directory, which is parallel to the `alf` package after a proper build.

By changing the name to `'alf.environments._penv'`, the compiled `.so` file should be placed inside the `alf/environments directory`, which is a subdirectory of the `alf` package, as expected.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -56,7 +56,7 @@ setup(
     ],  # And any other dependencies alf needs
     ext_modules=[
         Pybind11Extension(
-            'environments._penv',
+            'alf.environments._penv',
             sources=['alf/environments/parallel_environment.cpp'],
             extra_compile_args=[
                 '-O3', '-Wall', '-std=c++17', '-fPIC', '-fvisibility=hidden'


hnyu:
I've changed it in PR #1448

breakds:
> I've changed it in PR #1448

Sorry I did not notice that. Would be better if we have a dedicate commit for this fix?

hnyu:
> > I've changed it in PR #1448
> 
> Sorry I did not notice that. Would be better if we have a dedicate commit for this fix?

No problem. But I guess you need to first fix the test case due to an upgrade of pytorch state dict in the ci docker.

breakds:
> > > I've changed it in PR #1448
> > 
> > 
> > Sorry I did not notice that. Would be better if we have a dedicate commit for this fix?
> 
> No problem. But I guess you need to first fix the test case due to an upgrade of pytorch state dict in the ci docker.


I see. Will wait for that PR to go in first then. I'll close this PR if #1448 gets merged first. Thanks!

hnyu:
PR #1448  was merged

PR #1452, opened 2023-03-23
[Tiny] Short message on in-algorithm checkpoint loaded

Haichao-Zhang:
Add a short message  on in-algorithm checkpoint loading, to be more informative.

### DIFF
--- a/alf/algorithms/algorithm.py
+++ b/alf/algorithms/algorithm.py
@@ -239,6 +239,8 @@ class Algorithm(AlgorithmInterface):
 
             self.load_state_dict(stat_dict, strict=True)
             self._checkpoint_pre_loaded = True
+            logging.info(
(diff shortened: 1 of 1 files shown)


alf/algorithms/algorithm.py
```diff
@@ -239,6 +239,8 @@ def _preload_checkpoint(self):
 
             self.load_state_dict(stat_dict, strict=True)
             self._checkpoint_pre_loaded = True
+            logging.info(
```
hnyu: use `common.info` will be better highlighted

hnyu (changes requested):

alf/algorithms/algorithm.py
Haichao-Zhang: switched to common.info

hnyu (approved):

PR #1454, opened 2023-03-23
snapshot for extra repos

hnyu:
Support generating a snapshot for extra python repos. 

The difficulty is that only alf has the control of when and where to generate a snapshot, so it's hard to initiate a snapshot generation in any repo other than alf. The current solution is to let other repos call ``add_snapshot_repo_root`` in their ``__init__.py`` file. 

However, the current alf snapshot generation always comes before the conf file is actually parsed, because during parsing a global env is created, which is expected to be in a subprocess (while snapshot is expected in the main process). Without parsing the conf file, any other repo's ``__init__`` won't be evaluated. To make sure that a repo's ``__init__.py`` file is interpreted before alf snapshot generation is called, a solution is that we first informally evaluate the config file and then call ``alf.reset_configs()``. 

With the current solution, now in other repo's ``__init__.py``, all we need to do is something like

```python
from alf.utils.common import add_snapshot_repo_root
add_snapshot_repo_root(HOBOT_ROOT + '/..', 'hobot')
```

### DIFF
--- a/alf/bin/grid_search.py
+++ b/alf/bin/grid_search.py
@@ -327,12 +327,11 @@ class GridSearch(object):
             time.sleep(random.uniform(0, 3))
 
             conf_file = common.get_conf_file()
-            # This is the snapshot stored in grid-search root dir
-            alf_repo = common.abs_path(
-                os.path.join(FLAGS.root_dir, "alf.tar.gz"))
+
             # We still need to keep a snapshot of ALF repo at ``<root_dir>``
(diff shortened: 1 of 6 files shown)


emailweixu:
Can you also add the ability to record the diff and git revision of the extra repos?

hnyu:
> Can you also add the ability to record the diff and git revision of the extra repos?

added and tested

emailweixu (approved):

PR #1455, opened 2023-03-24
rename pprint and improve matching loss

hnyu:
Current `pprint.py` will have a conflict with the built-in repo. So renamed. 
Also improved BipartiteMatchingLoss to directly handle cost matrices.


alf/utils/losses.py
```diff
@@ -777,7 +794,7 @@ def forward(self,
             optimal_loss = optimal_loss.mean(-1)
         elif self._reduction == 'sum':
             optimal_loss = optimal_loss.sum(-1)
-        return optimal_loss
+        return optimal_loss, col_ind.squeeze(-1)
```
emailweixu, line 774: doc-string needs to comment the return values.

alf/utils/losses.py
```diff
@@ -714,26 +714,34 @@ class BipartiteMatchingLoss(object):
...
-                 pair_loss_fn: Callable = torch.cdist,
+                 pair_loss_fn: alf.nest.Nest = torch.cdist,
+                 loss_weights: alf.nest.Nest = 1.,
                  reduction: str = 'mean',
                  name: str = "BipartiteMatchingLoss"):
         """
         Args:
-            pair_loss_fn: the pairwise matching loss function. It should take
-                two sets of length ``N`` and output a cost matrix of shape ``[N,N]``.
+            pair_loss_fn: a nest of pairwise matching loss functions. Each element
```
emailweixu: I don't see the necessity of having nested pair_loss_fn in this module. Why can't you let pair_loss_fn to handle nested input. And perhaps we can even go without pair_loss_fn. Just use cost_mat as input.

hnyu: > I don't see the necessity of having nested pair_loss_fn in this module. Why can't you let pair_loss_fn to handle nested input. And perhaps we can even go without pair_loss_fn. Just use cost_mat as input.

Updated to use a cost mat as the input

alf/utils/losses.py
hnyu: > doc-string needs to comment the return values.

added

alf/utils/losses.py
```diff
@@ -714,54 +714,48 @@ class BipartiteMatchingLoss(object):
...
-            cost_mat = cost_mat * target_mask
+        if cost_mat is None:
+            cost_mat = matching_cost_mat
 
         with torch.no_grad():
+            B, N = matching_cost_mat.shape[:2]
             # [B*N, B*N]
-            max_cost = cost_mat.max() + 1.
-            big_cost_mat = torch.block_diag(*list(cost_mat - max_cost))
+            max_cost = matching_cost_mat.max() + 1.
```
emailweixu, line 755: The comment above should go down?

alf/utils/losses.py
```diff
@@ -714,54 +714,48 @@ class BipartiteMatchingLoss(object):
...
+            big_cost_mat = torch.block_diag(
+                *list(matching_cost_mat - max_cost))
             # fill in all off-diag entries with a max cost
             big_cost_mat = big_cost_mat + max_cost
```
emailweixu: This will also change the diagonal elements?

hnyu: > This will also change the diagonal elements?

The diagonal elements have been subtracted by `max_cost`.

alf/utils/losses.py
hnyu: > The comment above should go down?

moved down

alf/utils/losses.py
emailweixu: The comment is a little bit confusing. 
Adding the same value to all the elements is unnecessary? It will not change the solution.

hnyu: > The comment is a little bit confusing. Adding the same value to all the elements is unnecessary? It will not change the solution.

The max_cost is expected to only affect the off-diagonal entries. Ideally, for each sample b of B, we need to call linear_sum_assignment on the NxN matrix to find a matching. But this is slow. So I created a big matrix of size BNxBN. In order to make sure different NxN matrices won't affect each other, we need to make sure the off-diagonal values are very large, so that each matching is only found within each diagonal block. For example,

```python
[[O1 O1 X X]
 [O1 O1 X X]
 [X X O2 O2]
 [X X O2 O2]]]
```
Each 'O' submatrix is a valid cost matrix. All 'X' values are just paddings to form this big matrix.

emailweixu: I understand your motivation. But I think you don't need to add max_cost. The off-diagonal elements(i,e,zeros) are already guaranteed to be bigger than the diagonal elements.

hnyu: > I understand your motivation. But I think you don't need to add max_cost. The off-diagonal elements(i,e,zeros) are already guaranteed to be bigger than the diagonal elements.

I see. yeah there is no need to add again.

emailweixu (approved):

PR #1463, opened 2023-04-01
Support parallel environments of batched environments

emailweixu:
For the previous version of `FastparallelEnviroment`, each process only holds one environment. So when `num_parallel_environments` is large, we need to create a lot of processes, which may requires a lot of resources.

In the PR, we allow each process to hold more environments through ``BatchedEnvironmentWrapper``.

To use this feature, simply passing batch_size_per_env>1 for ``utils.create_environment``

### DIFF
--- a/alf/environments/alf_wrappers.py
+++ b/alf/environments/alf_wrappers.py
@@ -1055,6 +1055,9 @@ class NormalizedActionWrapper(AlfEnvironmentBaseWrapper):
 class BatchEnvironmentWrapper(AlfEnvironment):
     """Wrapper to make a list of non-batched environment into a batched environment.
 
+    Note the individual environments in ``envs`` are executed sequentially doring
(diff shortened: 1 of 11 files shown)


alf/environments/alf_wrappers.py
```diff
@@ -1049,3 +1050,89 @@ def _scale_back(a, paras):
                                                  self._affine_paras)
         time_step = self._env.step(scaled_action)
         return time_step._replace(prev_action=action)
+
+
+class BatchEnvironmentWrapper(AlfEnvironment):
+    """Wrapper to make a list of non-batched environment into a batched environment.
```
breakds, line 1056: Suggest explaining the "batch" behavior is actually sequentially step the environments here.

breakds (dismissed):

breakds (approved):

PR #1466, opened 2023-04-04
[Small] Updates on make_penv script

breakds:
1. Only install `pybind11` via `pip` if it is not present.
2. Use `sys.executable` to find the path to the actual python binary.

### DIFF
--- a/alf/environments/make_penv.py
+++ b/alf/environments/make_penv.py
@@ -20,9 +20,15 @@ def gen_penv():
     current_path = os.path.abspath(__file__)
     dir_path = os.path.dirname(current_path)
     os.chdir(dir_path)
-    python = f"python{sys.version_info.major}.{sys.version_info.minor}"
-    assert os.system(
-        "pip install pybind11") == 0, "Fail to pip install pybind11"
+
+    try:
+        import pybind11
+    except ImportError:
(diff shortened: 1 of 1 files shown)


breakds:
Just noticed that No.1 is already addressed in #1463

emailweixu (dismissed):

emailweixu (approved):

alf/environments/make_penv.py
```diff
@@ -20,13 +20,14 @@ def gen_penv():
...
     os.chdir(dir_path)
+
     # install pybind11 if it's not installed.
     try:
         import pybind11
     except:
         assert os.system(
             "pip install pybind11") == 0, "Fail to pip install pybind11"
-    python = f"python{sys.version_info.major}.{sys.version_info.minor}"
+    python = sys.executable
```
emailweixu, line 30: What is the problem with the previous version of the code?
The new version does not work on my machine. On my machine, sys.executable get "python", but python-config does not exist.

breakds: I think it is due to https://github.com/pypa/virtualenv/issues/169

Let me revert this change then. I found a way to have `python3.x` in my environment now.

PR #1468, opened 2023-04-04
Store transformed experience in PPG Aux's replay buffer

breakds:
Based on a discussion with Wei, decided to make a modification on how replay buffer is being handled in PPG auxiliary algorithm

## Before the change

1. In `PPGAlgorithm`'s `rollout_step`, it calls aux algorithm's `observe_for_aux_replay` to store an **untransformed** experience in the replay buffer.
2. Upon the creation of `PPGAuxAlgorithm`, the parent `PPGAlgorithm`'s data transformers will be inherited.
3. When `PPGAuxAlgorithm` is performing update, it will load **untransformed** experiences from its replay buffer, and apply data transformers on them to get **transformed** experiences.

## After the change

1. In `PPGAlgorithm`'s `rollout_step`, it calls aux algorithm's `observe_for_aux_replay` to store an **transformed** experience in the replay buffer.
2. Upon the creation of `PPGAuxAlgorithm`, the parent `PPGAlgorithm`'s data transformers will be **ignored**.
3. When `PPGAuxAlgorithm` is performing update, it will load **transformed** experiences from its replay buffer and use them directly.

## Notes

1. I am still keeping `observe_for_aux_replay` because it has a specific replay buffer construction logic.
2. It is more likely to take more memories to store the transformed experience, especially when there are `FrameStacker` with a large `stack_size` on experiences that contains images. In `procgen`'s current configuration though, the memory consumption goes from `8.0 GB` to `9.6 GB`. The advantage is that it saves data transformation when gathering experience from the replay buffer during aux phase. I cannot reliably observe the difference though because it is too small.

## Testing

Running.

### DIFF
--- a/alf/algorithms/ppg/ppg_aux_algorithm.py
+++ b/alf/algorithms/ppg/ppg_aux_algorithm.py
@@ -138,7 +138,7 @@ class PPGAuxAlgorithm(OffPolicyAlgorithm):
 
     def observe_for_aux_replay(self, exp):
         """Save the experience in the replay buffer for auxiliary phase update.
-        
+
         Args:
 
             exp (nested Tensor): Experience to be saved. The shape is [B, ...]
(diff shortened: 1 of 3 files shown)


emailweixu (approved):

breakds:
Tested the `procgen_bossfight` training and the performance is similar to pre-change.

ppg_procgen_bossfight

PR #1471, opened 2023-04-05
Change assert to if for error checking

emailweixu:
It turns out the assert is a noop for release compilation (i.e. when NDEBUG is defined, https://en.cppreference.com/w/cpp/error/assert). So the previous version of the code does not enforce those assertion if it is installed through setup.py.

Also make the error message more informative.

### DIFF
--- a/alf/environments/parallel_environment.cpp
+++ b/alf/environments/parallel_environment.cpp
@@ -95,11 +95,24 @@ class SharedDataBuffer {
   std::vector<size_t> offsets_;  // offset of the i-th array
   std::vector<size_t> sizes_;    // size of each slice of the i-th array
   std::vector<py::buffer_info> buffer_infos_;
+  std::vector<std::vector<ssize_t>> slice_strides_;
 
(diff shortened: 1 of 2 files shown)


alf/environments/parallel_environment.cpp
```diff
@@ -95,12 +95,25 @@ class SharedDataBuffer {
...
 
+void CheckStrides(const py::buffer_info& info,
+                  const std::vector<ssize_t>& strides) {
+  if (info.strides != strides) {
+    throw std::runtime_error(
+        py::str(
+            "strides mismatch. Expected: {}, "
+            "Got: {}, shape: {}. It might be caused by non-continguous array, "
+            "which can be fixed by copying the array.")
+            .format(strides, info.strides, info.shape));
```
hnyu, line 113: Is it possible to also print out more buffer info? Like buffer's name. Just looking at the shape might still not be clear.

emailweixu: buffer_info does not have name

hnyu (approved):

PR #1475, opened 2023-04-10
add a function for checking parameter consistency when using DDP

hnyu:
Add a helper function for checking parameter consistency when using DDP. Usually, DDP will only ensure that parameters requiring gradients will be synced at each iteration. For some parameters that don't require gradient (e.g., target model of SAC), this function provides a way to ensure that they are also consistent across processes. 

This function is also a sanity checker of whether DDP is correctly used in ALF.

### DIFF
--- a/alf/trainers/policy_trainer.py
+++ b/alf/trainers/policy_trainer.py
@@ -45,6 +45,7 @@ from alf.utils import math_ops
 from alf.utils.pretty_print import pformat_pycolor
 from alf.utils.checkpoint_utils import Checkpointer
 import alf.utils.datagen as datagen
+from alf.utils.per_process_context import PerProcessContext
 from alf.utils.summary_utils import record_time
 from .evaluator import Evaluator
 
@@ -559,6 +560,7 @@ class RLTrainer(Trainer):
             self._algorithm.activate_ddp(ddp_rank)
             # Make sure the BN statistics of different processes are synced
             # https://pytorch.org/docs/stable/generated/torch.nn.SyncBatchNorm.html#torch.nn.SyncBatchNorm
+            # This conversion needs to be performed before wrapping modules with DDP.
             self._algorithm = torch.nn.SyncBatchNorm.convert_sync_batchnorm(
                 self._algorithm)
 
@@ -657,6 +659,9 @@ class RLTrainer(Trainer):
                     self._eval()
                 break
 
+            self._check_dpp_paras_consistency(iter_num,
+                                              training_setting_summarized)
+
             if ((self._num_iterations and iter_num >= time_to_checkpoint)
                     or (not self._num_iterations and self._num_env_steps
                         and total_time_steps >= time_to_checkpoint)):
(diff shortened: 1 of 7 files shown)


hnyu:
Using this tool, I was able to detect parameter inconsistency for MultiTaskSAC (which updates the target model based local-batch statistics) but not for SAC.

alf/trainers/policy_trainer.py
```diff
@@ -672,6 +677,77 @@ def _train(self):
...
+        if not training_started:
+            return
+
+        proc_cxt = PerProcessContext()
+        if (proc_cxt.is_distributed
+                and self._config.ddp_paras_check_interval > 0
+                # Assume that DDP will make sure that this modulo check won't
+                # cause a dead lock, i.e., all workers have the same ``iter_num``
+                # at any moment.
+                and iter_num % self._config.ddp_paras_check_interval == 0):
```
emailweixu: nit: return if condition is not satisfied to make the code indent less

alf/trainers/policy_trainer.py
```diff
@@ -672,6 +677,77 @@ def _train(self):
...
+                        for i in range(proc_cxt.num_processes - 1):
+                            their_paras_stat = queue.get()
+                            is_close = np.isclose(
+                                paras_stat, their_paras_stat, atol=1e-6)
+                            if not np.all(is_close):
+                                consistent = False
+                                common.warning(
+                                    "Found inconsistent parameter stats across "
+                                    "DDP processes: %s vs. %s => %s" %
+                                    (paras_stat, their_paras_stat, is_close))
```
emailweixu: should include the name of the parameter. otherwise it will be very hard to know the problem

emailweixu (changes requested):

alf/trainers/policy_trainer.py
hnyu: > should include the name of the parameter. otherwise it will be very hard to know the problem

Currently the stats is computed for the entire parameter space. Are you suggesting that we compute the stats for each named parameter? 

(Because the inconsistency will diffuse very quickly during gradient update, I wonder if providing parameter names helps).

emailweixu: The communication cost of a dict of param name to (mean, std) shouldn't be that much, especially if the interval is large enough.

I think in most situations, the inconsistency will not spread. If a parameter is updated by an optimizer, all the machines will use the same gradient to update it, so it should remain consistent, even if some other parameters become inconsistent. The inconsistency only comes from some special logic, which can be easier to be identified if the name of the parameter is known.

hnyu: > The communication cost of a dict of param name to (mean, std) shouldn't be that much, especially if the interval is large enough.
> 
> I think in most situations, the inconsistency will not spread. If a parameter is updated by an optimizer, all the machines will use the same gradient to update it, so it should remain consistent, even if some other parameters become inconsistent. The inconsistency only comes from some special logic, which can be easier to be identified if the name of the parameter is known.

Right, all parameters being updated will always be the same.

emailweixu (approved):

PR #1476, opened 2023-04-11
alf.import_config()

emailweixu:
There are two motivations for this:

1. config files can be correctly loaed after alf.reset_configs(). Previously, because python only run the imported file once, so after alf.reset_configs(), the config set in the imported file will not be set.

2. save the imported config files. Imported config files are not saved in previous code.

### DIFF
--- a/alf/config_util.py
+++ b/alf/config_util.py
@@ -20,6 +20,8 @@ import inspect
 from inspect import Parameter
 import os
 import pprint
+import runpy
+import shutil
 
 __all__ = [
     'config',
@@ -31,10 +33,13 @@ __all__ = [
     'get_handled_pre_configs',
     'get_inoperative_configs',
     'get_operative_configs',
+    'import_config',
+    'load_config',
     'pre_config',
     'reset_configs',
     'validate_pre_configs',
     'repr_wrapper',
+    'save_config',
 ]
 
 
@@ -217,6 +222,10 @@ _CONF_TREE = {}
 _PRE_CONFIGS = []
 _HANDLED_PRE_CONFIGS = []
 _DEFINED_CONFIGS = []
+_CONF_FILES = {}  # key: file name, value: content
+_CONFIG_MODULES = {}
+_IMPORT_STACK = []
+_ROOT_CONF_FILE = None
 
 
 def reset_configs():
@@ -236,6 +245,10 @@ def reset_configs():
     _DEFINED_CONFIGS.clear()
     _PRE_CONFIGS.clear()
     _HANDLED_PRE_CONFIGS.clear()
+    _CONF_FILES.clear()
+    _CONFIG_MODULES.clear()
+    global _ROOT_CONF_FILE
+    _ROOT_CONF_FILE = None
 
 
 def _remove_config_node(config_name):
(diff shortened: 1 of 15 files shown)


alf/examples/sac_cart_pole_conf.py
```diff
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 import alf
-import alf.examples.sac_conf
+alf.import_config("sac_conf.py")
```
hnyu, line 16: Now how we should provide the path to the conf file? Seems not as straightforward as importing a module.

alf/config_util.py
```diff
@@ -812,3 +825,131 @@ def define_config(name, default_value):
     node.set_default_value(default_value)
     _add_to_conf_tree(['_CONFIG'], '_USER', name, node)
     _DEFINED_CONFIGS.append('_CONFIG._USER.' + name)
+
+
+def _get_conf_file_full_path(conf_file):
```
hnyu, line 830: Comment on the rough logic of how the full path is obtained for the conf file.

alf/config_util.py
```diff
@@ -812,3 +825,131 @@ def define_config(name, default_value):
...
+
+def _add_conf_file(conf_file):
+    if conf_file in _CONF_FILES:
+        return
+    with open(conf_file, "r") as f:
+        _CONF_FILES[conf_file] = f.read()
+
+
+def import_config(conf_file):
+    """Import the config from another file.
```
hnyu, line 862: What will happen if `import_config` and the old `import` are used in a mixed way?

alf/config_util.py
```diff
@@ -812,3 +825,131 @@ def define_config(name, default_value):
...
+    kv = runpy.run_path(conf_file)
+    _IMPORT_STACK.pop()
+    module = ConfigModule()
+    for k, v in kv.items():
+        setattr(module, k, v)
+    _CONFIG_MODULES[conf_file] = module
+    return module
+
+
+def load_config(conf_file):
```
hnyu, line 918: It could be very easy to confuse this function with `import_config`. Should add explanations of their difference.

alf/environments/parallel_environment.cpp
```diff
@@ -114,8 +115,19 @@ void CheckStrides(const py::buffer_info& info,
   }
 }
 
+void SharedDataBuffer::CheckSpec(const py::list& arrays,
+                                 const py::object& nested_array) {
+  if (arrays.size() != sizes_.size()) {
```
hnyu, line 120: This is just checking the sizes but not specs?

alf/utils/common.py
```diff
@@ -993,7 +974,7 @@ def set_random_seed(seed):
         The seed being used if ``seed`` is None.
     """
     if seed is None:
-        seed = int(np.uint32(hash(str(os.getpid()) + '|' + str(time.time()))))
+        seed = hash(str(os.getpid()) + '|' + str(time.time()))
```
hnyu: It is hard to guarantee that other envs (not gym) can accept a negative random seed.

hnyu (changes requested):

alf/config_util.py
emailweixu: comment added to load_config and import_config

alf/config_util.py
emailweixu: That still works most of time. The time it fails is when ``import_config()`` is called from a file A that is loaded by file B using ``import`` and the A and B are not in the same directory.

alf/config_util.py
emailweixu: added

alf/environments/parallel_environment.cpp
emailweixu: updated

alf/utils/common.py
emailweixu: added ``abs()``

alf/examples/sac_cart_pole_conf.py
emailweixu: In this case "sac_conf.py" is in the same directory as "sac_cart_pole_conf.py", so simply "sac_conf.py" works. Added more explanation in ``import_config``

alf/config_util.py
```diff
@@ -832,10 +832,11 @@ def _get_conf_file_full_path(conf_file):
...
-            return candidate
-    dir = os.path.dirname(_IMPORT_STACK[-1])
+        # called from load_config()
+        dir = os.getcwd()
+    else:
+        # callded from import_config()
+        dir = os.path.dirname(_IMPORT_STACK[-1])
     candidate = os.path.join(dir, conf_file)
     if os.path.exists(candidate):
         return candidate
```
hnyu, line 842: Should we by default set ALF_CONFIG_PATH to alf_root? So that any config file under alf root can be found by providing relative paths.

emailweixu: I have the impression that the relative import from the directory obtained from line 834-839 can cover our existing use cases.

hnyu: > I have the impression that the relative import from the directory obtained from line 834-839 can cover our existing use cases.

What if I have two conf files: `~/alf/configs/path1/conf1.py` and `~/alf/configs/path2/conf2.py`, and I want to import `conf1.py` in `conf2.py`? I can do `import_config('../path1/conf1.py')` but if the path is deep sometimes, it's better to directly start with the repo root as `configs/path1/conf1.py`.

emailweixu: I agree with your example. But currently, we don't have that kind of config in our repo.  And if that is needed, one can always set ALF_CONFIG_PATH himself.

hnyu (approved):

PR #1479, opened 2023-04-14
Create eval environment

Haichao-Zhang:
This PR enables the ``create_environment`` function to create a evaluation environment via a configurable ``eval_env_load_fn`` function.

### DIFF
--- a/alf/environments/utils.py
+++ b/alf/environments/utils.py
@@ -74,7 +74,7 @@ def _env_constructor(env_load_fn, env_name, batch_size_per_env, seed, env_id):
 def create_environment(env_name='CartPole-v0',
                        env_load_fn=suite_gym.load,
                        eval_env_load_fn=None,
-                       create_eval_env=False,
+                       for_evaluation=False,
                        num_parallel_environments=30,
                        batch_size_per_env=1,
                        nonparallel=False,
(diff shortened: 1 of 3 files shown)


alf/trainers/evaluator.py
```diff
@@ -170,7 +172,8 @@ def _worker(job_queue: mp.Queue,
             pre_configs: Dict,
             num_parallel_envs: int,
             root_dir: str,
-            seed: Optional[int] = None):
+            seed: Optional[int] = None,
+            create_eval_env: Optional[bool] = False):
```
emailweixu: this argument seems unnecessary. It's always False

Haichao-Zhang: Removed this argument

alf/environments/utils.py
```diff
@@ -73,6 +73,8 @@ def _env_constructor(env_load_fn, env_name, batch_size_per_env, seed, env_id):
 @alf.configurable
 def create_environment(env_name='CartPole-v0',
                        env_load_fn=suite_gym.load,
+                       eval_env_load_fn=None,
+                       create_eval_env=False,
```
hnyu: The argument name is misleading. It can be interpreted as "creating an additional eval env". Can be changed to just "for_evaluation".

hnyu (changes requested):

alf/environments/utils.py
Haichao-Zhang: changed as suggested

hnyu (approved):

emailweixu (approved):

PR #1480, opened 2023-04-14
 Pseudo episode-end pause in recorded video

Haichao-Zhang:
Enables optional pseudo "pause" of the episode end by repeating.
This can be useful when there are some debugging info on the frames.


hnyu (commented):
We can directly configure VideoRecorder

Haichao-Zhang:
> We can directly configure VideoRecorder

Added to configurable list. 
Feel it is convenient to keep the command argument option for this since it is commonly used when we play from command line.

hnyu (approved):

PR #1481, opened 2023-04-15
Fix random seed generation

hnyu:
Current `abs()` sometimes produces overflowed integers that torch cannot handle, reporting errors like below: 
```python     
common.set_random_seed(seed)
  File "/home/haonanyu/venv-hobot/lib/python3.10/site-packages/torch/random.py", line 46, in manual_seed
    return default_generator.manual_seed(seed)
  File "/home/haonanyu/venv-hobot/lib/python3.10/site-packages/torch/random.py", line 46, in manual_seed
    return default_generator.manual_seed(seed)
  File "/home/haonanyu/hobot-alf/alf/alf/utils/common.py", line 989, in set_random_seed
    torch.random.manual_seed(seed)
RuntimeError: Overflow when unpacking long 
```
This PR reverts it back to the old version.

### DIFF
--- a/alf/utils/common.py
+++ b/alf/utils/common.py
@@ -974,7 +974,6 @@ def set_random_seed(seed):
         The seed being used if ``seed`` is None.
     """
     if seed is None:
-        # Has to conver to `int`` otherwise gym will complain
         seed = abs(hash(str(os.getpid()) + '|' + str(time.time())))
     else:
         torch.backends.cudnn.deterministic = True
--- a/alf/utils/common.py
+++ b/alf/utils/common.py
@@ -975,7 +975,7 @@ def set_random_seed(seed):
     """
     if seed is None:
         # Has to conver to `int`` otherwise gym will complain
-        seed = int(np.uint32(hash(str(os.getpid()) + '|' + str(time.time()))))
+        seed = abs(hash(str(os.getpid()) + '|' + str(time.time())))
     else:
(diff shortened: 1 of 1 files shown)


alf/utils/common.py
```diff
@@ -974,7 +974,8 @@ def set_random_seed(seed):
         The seed being used if ``seed`` is None.
     """
     if seed is None:
-        seed = abs(hash(str(os.getpid()) + '|' + str(time.time())))
+        # Has to conver to `int`` otherwise gym will complain
+        seed = int(np.uint32(hash(str(os.getpid()) + '|' + str(time.time()))))
```
hnyu: https://github.com/openai/gym/blob/master/gym/utils/seeding.py#L21

emailweixu: newer version of numpy.uint32 will complain if the number is too big

hnyu: > newer version of numpy.uint32 will complain if the number is too big

Now modulo the seed by 2**32

alf/utils/common.py
```diff
@@ -974,6 +974,7 @@ def set_random_seed(seed):
         The seed being used if ``seed`` is None.
     """
     if seed is None:
+        # Has to conver to `int`` otherwise gym will complain
```
emailweixu: the comment doesn't match code

emailweixu (approved):

PR #1484, opened 2023-04-26
Fix summarize_loss

emailweixu:
Previously, nests with list and unnamed tuple are skipped. Now change to use py_map_structure_with_path to handle them.


emailweixu:
@hnyu

hnyu (approved):

PR #1485, opened 2023-04-26
Add `on_load_checkpoint` hook

breakds:
This PR introduces a new `on_load_checkpoint` hook. It is a simple yet very flexible way to transform the checkpoint right before it is loaded by `Algorithm.load_state_dict`.

By default it does not transform the checkpoint at all. The user can override it to 

1. Remove or add entries to the state dict
2. Implement his/her own way to fuzzy match the entries in the state dict

### DIFF
--- a/alf/algorithms/algorithm.py
+++ b/alf/algorithms/algorithm.py
@@ -23,7 +23,7 @@ import json
 import numpy as np
 import os
 import psutil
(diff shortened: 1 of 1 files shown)


breakds:
Find another way to achieve this without having this patch. Will close.

PR #1487, opened 2023-04-28
Per-frame repeats for low motion video recording

Haichao-Zhang:


alf/bin/play.py
```diff
@@ -64,6 +64,10 @@ def _define_flags():
         'last_step_repeats', 0,
         "If >0, wil repeat such number of times for the last "
         "frame of each episode in the rendered video file.")
+    flags.DEFINE_integer(
+        'per_frame_repeats', 0,
+        "If >0, will repeat such number of times for each frame "
+        "of each episode in the rendered video file.")
```
hnyu, line 70: I still feel that these video recorder related parameters should be removed here and just configured in `VideoRecorder`. When play in the cmd, use `--conf_param='VideoRecorder.per_frame_repeats=2'`.

emailweixu: Why do we need such a feature? If you want to play it slower, you can adjust it in the video player. Or you can simply play it frame by frame.

Haichao-Zhang: > Why do we need such a feature? If you want to play it slower, you can adjust it in the video player. Or you can simply play it frame by frame.

Agree not always needed but sometimes. For example, when the video is uploaded to Feishu doc, unfortunately, there is no way to control the play the of video.

Haichao-Zhang:
Close PR since it is not always necessary.

PR #1488, opened 2023-05-01
More debugging info when strides mismatch in parallel env

breakds:
When strides mismatch happens, it may be caused by the underlying dtype mismatch. This PR

1. Remembers the `dtype` of the specs
2. When stride mismatches, it then goes forward to do one more check on the dtype mismatch and report dtype mismatch instead if it is indeed the cause.

The `dtype` check is done only when stride mismatch is confirmed so that **no actual performance hit** is included.

## Testing

Now with my previous bug, the training crashed with 

```
RuntimeError: dtype mismatch. Expected: float32, Got float64.The offending buffer is [ 0.28346096  0.02934491 -0.2589214 ]
```

Which can still be improved, but would dramatically improve my debugging time.

### DIFF
--- a/alf/environments/parallel_environment.cpp
+++ b/alf/environments/parallel_environment.cpp
@@ -96,6 +96,7 @@ class SharedDataBuffer {
   std::vector<size_t> sizes_;    // size of each slice of the i-th array
   std::vector<py::buffer_info> buffer_infos_;
   std::vector<std::vector<ssize_t>> slice_strides_;
+  std::vector<py::dtype> dtypes_;
 
   inline char* GetBuf(int slice_id, int array_id) const {
     return buf_ + offsets_[array_id] + sizes_[array_id] * slice_id;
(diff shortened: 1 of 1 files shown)


alf/environments/parallel_environment.cpp
```diff
@@ -104,14 +104,17 @@ class SharedDataBuffer {
...
   if (info.strides != strides) {
     throw std::runtime_error(
         py::str(
-            "strides mismatch. Expected: {}, "
+            "Strides mismatch. Expected: {}, "
             "Got: {}, shape: {}. It might be caused by non-continguous array, "
-            "which can be fixed by copying the array.")
-            .format(strides, info.strides, info.shape));
+            "which can be fixed by copying the array. It could also be caused "
+            "by having an array whose dtype is different from its spec. The "
```
emailweixu: Instead of suggesting possible dtype mismatch, we can check whether there is in fact dtype mismatch by comparing info.format

breakds: Good point. Updated. Sorry about having to `rebase` but this PR is small so it should be fine.

emailweixu (approved):

PR #1489, opened 2023-05-01
support partial aggregation and masking for AverageEpisodicAggregationMetric

hnyu:
Sometimes the fields of env info don't always appear in every episode. For example, we randomly sample a different task for each episode and the task env info fields might be different for different episodes.

However, our metric class require a complete env info spec (that contains all the possible fields) during the initialization, and it assumes that each field will be accumulated for every episode. This means that usually we need to fill in dummy metric values for those fields that are actually not active to be compatible. 

This PR makes `AverageEpisodicAggregationMetric` more flexible by allowing the user to skip aggregating certain metric fields by providing "invalid" values. Basically, an invalid value tells the accumulator not to aggregate the current value. For any field, if the values of an entire episode are invalid, then at the episode end, the buffer won't be pushed into a new accumulated value, and the buffer's mean stays the same.

### DIFF
--- a/alf/metrics/metrics.py
+++ b/alf/metrics/metrics.py
@@ -152,6 +152,10 @@ class AverageEpisodicAggregationMetric(metric.StepMetric):
     2. If a field has a postfix "@max", then the aggregated value will be the
        maximum (instead of sum) of step values across the episode.
 
+    This class supports partial aggregation, where if at any step the extracted
+    metric value is not finite (inf or nan), then that step's value will be skipped
+    for aggregation. If a field is skipped for an entire episode, its accumulated
+    value won't be pushed into the metric buffer.
     """
 
     def __init__(self,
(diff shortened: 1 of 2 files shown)


alf/metrics/metrics.py
```diff
@@ -230,40 +238,54 @@ def call(self, time_step):
 
         is_first = time_step.is_first()
 
-        def _update_accumulator_(path, acc, val):
-            """In-place update of the accumulators."""
+        def _update_accumulator_(path, mask, acc, val):
+            """In-place update of the accumulators and mask."""
+            val_valid = torch.isfinite(val)
+            # If at any step the value is valid, then the acc value becomes valid
```
emailweixu, line 253: This behavior should be clearly document in the doc-string of `AverageEpisodicAggregationMetric`

alf/metrics/metrics.py
```diff
@@ -230,40 +238,54 @@ def call(self, time_step):
...
+            mask[:] = torch.where(is_first, torch.zeros_like(mask),
+                                  mask + val_valid)
```
emailweixu: `mask | val_valid`

alf/metrics/metrics.py
```diff
@@ -230,40 +238,54 @@ def call(self, time_step):
...
 
-        def _episode_end_aggregate_(path, buf, acc):
+        def _episode_end_aggregate_(path, mask, buf, acc):
             value = self._extract_and_process_acc_value(
                 acc, last_episode_indices)
             # If the metric's name ends with '@step', the value will
             # be further averaged over episode length so that the
             # result is per-step value.
             if path.endswith('@step'):
                 value = value / self._current_step[last_episode_indices]
```
emailweixu: Should `self._current_step[last_episode_indices]` take into val_valid into account?

emailweixu (changes requested):

alf/metrics/metrics.py
hnyu: > This behavior should be clearly document in the doc-string of `AverageEpisodicAggregationMetric`

Added.

alf/metrics/metrics.py
hnyu: > Should `self._current_step[last_episode_indices]` take into val_valid into account?

Updated to have field-dependent steps. Still keep the original `self._current_step` as it is needed by a subclass.

emailweixu (approved):

PR #1491, opened 2023-05-02
allow outside to set root_dir so load config doesn't depend on FLAGS.…

le-horizon:
…root_dir

(cherry picked from commit b7e413f14f47de9b3e5d701651e2712d8f9aaf06)

### DIFF
--- a/alf/utils/common.py
+++ b/alf/utils/common.py
@@ -574,7 +574,7 @@ def get_gin_file():
 ALF_CONFIG_FILE = 'alf_config.py'
 
 
-def get_conf_file():
+def get_conf_file(root_dir=None):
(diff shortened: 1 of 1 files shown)


alf/utils/common.py
```diff
@@ -574,7 +574,7 @@ def get_gin_file():
 ALF_CONFIG_FILE = 'alf_config.py'
 
 
-def get_conf_file():
+def get_conf_file(root_dir=None):
```
emailweixu, line 577: update doc-string

le-horizon: Done

emailweixu (approved):

PR #1494, opened 2023-05-03
Optionally ignore the value network for deployment purpose in ActorCriticAlgorithm

breakds:
The value network for `ActorCriticAlgorithm` (as well as `PPOAlgorithm` which inherits from it) are only used in training but not for deployment purpose. This PR makes it optional so that an user can load the checkpoint without `value_network` as long as the `value_network_ctor` is set to `None` when constructing the algorithm.

This makes it possible to load a PPO checkpoint with value network that is not designed to be constructed for deployment.

### DIFF
--- a/alf/algorithms/actor_critic_algorithm.py
+++ b/alf/algorithms/actor_critic_algorithm.py
@@ -82,12 +82,14 @@ class ActorCriticAlgorithm(OnPolicyAlgorithm):
                 ``actor_network_ctor`` needs to accept ``input_tensor_spec`` and
                 ``action_spec`` as its arguments and return an actor network.
                 The constructed network will be called with ``forward(observation, state)``.
-            value_network_ctor (Callable): Function to construct the value network.
-                ``value_network_ctor`` needs to accept ``input_tensor_spec`` as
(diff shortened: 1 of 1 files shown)


alf/algorithms/actor_critic_algorithm.py
```diff
@@ -134,6 +140,33 @@ def __init__(self,
...
+        def _deployment_hook(state_dict, prefix: str, unused_loacl_metadata,
+                             unused_strict, unused_missing_keys,
+                             unused_unexpected_keys, unused_error_msgs):
+            """Preprocess the state_dict to delete incompatible keys.
+
+            In pure privileged mode (i.e. not standard encoder) or pure
+            adaptation mode (i.e. no privileged encoder), this hook will exclude
+            the corresponding keys from the state dict when loading the
+            checkpoint, explicitly avoid crashing caused by such
+            incompatibility.
```
emailweixu: This comment (e.g. privileged mode) does not make sense without broader context.

breakds: The comment was move up and didn't remove this. Fixed.

emailweixu (approved):

PR #1495, opened 2023-05-09
Fix ParallelValueNetwork

emailweixu:
Should use allow_non_parallel_input=True for make_parallel

### DIFF
--- a/alf/networks/value_networks.py
+++ b/alf/networks/value_networks.py
@@ -175,7 +175,7 @@ class ParallelValueNetwork(Network):
 
         super().__init__(
             input_tensor_spec=value_network.input_tensor_spec, name=name)
-        self._encoding_net = value_network._encoding_net.make_parallel(n)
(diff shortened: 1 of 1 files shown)


alf/networks/value_networks.py
```diff
@@ -175,7 +175,7 @@ def __init__(self,
 
         super().__init__(
             input_tensor_spec=value_network.input_tensor_spec, name=name)
-        self._encoding_net = value_network._encoding_net.make_parallel(n)
+        self._encoding_net = value_network._encoding_net.make_parallel(n, True)
```
hnyu, line 178: I thought this was intended instead of a bug. There are many other places where make_parallel doesn't allow non-parallel input. What is the criterion here?

emailweixu: Similar to parallel CriticNetwork, parallel ValueNetwork is often called with non-parallel input.

hnyu (approved):

PR #1499, opened 2023-05-17
avoid loading critic networks from state dict during deployment if critic_network_cls is None

le-horizon:


le-horizon:
Addressed Break's concern about the unittest by passing in a config is_eval to RLAlgorithm.
Adjusted code according to the change.
@breakds PTAL.

breakds (approved):

breakds:
I'd suggest @runjerry to also take a look

le-horizon:
> I'd suggest @runjerry to also take a look

Oops, merged early.  @runjerry , could you take a look at this PR as well?  I'll fix as needed.  Thanks!

runjerry:
> > I'd suggest @runjerry to also take a look
> 
> Oops, merged early. @runjerry , could you take a look at this PR as well? I'll fix as needed. Thanks!

Thanks Le for updating this! The PR looks good to me. 

I have only one further question: should we put a customized sac_algorithm under Hobot or change the general alf.sac_algorithm as it is? The reasons for this question are:

1. It seems that this is only needed for our current deployment design rather than general RL settings.
2. Since ``is_eval`` is provided as an option in the base class ``RLAlgorithm``, should we update all other alf rl algorithms that also use critic_networks or value_networks?

PR #1500, opened 2023-05-17
In-algorithm checkpoint loading check

Haichao-Zhang:
Added check on missing-keys and unexpected-keys for in-algorithm checkpoint loading.

### DIFF
--- a/alf/algorithms/algorithm.py
+++ b/alf/algorithms/algorithm.py
@@ -241,7 +241,7 @@ class Algorithm(AlgorithmInterface):
 
             status = self.load_state_dict(stat_dict, strict=True)
             assert not status.missing_keys and not status.unexpected_keys, (
-                "\033[1;31m Checkpoint mis-matches with the model: \033[1;0m \n"
+                "\033[1;31m Checkpoint mismatches with the model: \033[1;0m \n"
                 +
                 "\033[1;31m Missing-keys \033[1;0m (keys in model but not in checkpoint): {}\n"
                 .format(status.missing_keys) +
--- a/alf/algorithms/algorithm.py
+++ b/alf/algorithms/algorithm.py
@@ -239,7 +239,14 @@ class Algorithm(AlgorithmInterface):
             stat_dict = extract_sub_state_dict_from_checkpoint(
                 checkpoint_prefix, checkpoint_path)
 
-            self.load_state_dict(stat_dict, strict=True)
+            status = self.load_state_dict(stat_dict, strict=True)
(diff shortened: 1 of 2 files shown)


alf/algorithms/algorithm.py
```diff
@@ -239,7 +239,14 @@ def _preload_checkpoint(self):
             stat_dict = extract_sub_state_dict_from_checkpoint(
                 checkpoint_prefix, checkpoint_path)
 
-            self.load_state_dict(stat_dict, strict=True)
+            status = self.load_state_dict(stat_dict, strict=True)
```
emailweixu, line 242: does this handle the case of mismatching dimension?

alf/algorithms/algorithm.py
```diff
@@ -239,7 +239,14 @@ def _preload_checkpoint(self):
...
+            assert not status.missing_keys and not status.unexpected_keys, (
+                "\033[1;31m Checkpoint mis-matches with the model: \033[1;0m \n"
```
emailweixu: mis-matches => mismatches

alf/algorithms/algorithm.py
Haichao-Zhang: Yes. The original code can already handle the case of mismatching dimensions. I also add a test case for it now in the second commit.
What happened for the case met yesterday is like what happened in ``3)`` of ``test_mismatch_checkpoint ``, where all the parameters are skipped.

alf/algorithms/algorithm.py
Haichao-Zhang: Fixed

alf/algorithms/algorithm_test.py
```diff
@@ -394,6 +394,62 @@ def test_partial_preloading_and_then_checkpoint_loading(self):
...
+            ckpt_mngr_composed = ckpt_utils.Checkpointer(
+                ckpt_dir_composed, alg=alg_composed)
+            ckpt_mngr_composed.save(0)
+
+            ckpt_path = ckpt_dir_composed + '/ckpt-0'
+
+            # 3）Checkpoint loading with a missing key case by specifying
+            # a wrong prefix (``alg._sub_alg3``). Test Exception will be raised
+            # and missing key message will be generated.
+            with self.assertRaises(Exception) as context:
```
emailweixu: `assertRaises(AssertionError)` otherwise some real error will be ignored.

Haichao-Zhang: Changed

emailweixu (approved):

PR #1502, opened 2023-05-18
Find unused port for DDP

emailweixu:
Sometime the default port is taken and causes failure.


breakds (dismissed):
LGTM.

Finding unused port can be implemented in a simpler way as it is something the operating system should be doing all the time, but since this is an already implemented function it should be ok.

breakds:
There are other places using the original moved function `_get_unused_port`.

breakds (approved):

PR #1505, opened 2023-05-26
Fix the computation of minimum and maximum of frame stacked spec

breakds:
Originally the `minimum` and `maximum` is computed by `np.repeat`. This will actually perform interleaved repeat such that

`[0, 1, 2]` -> `[0, 0, 0, 1, 1, 1, 2, 2, 2]`

which is not what we want. In fact, we should be using `np.tile` style to construct the new `minimum` and `maximum` so that

`[0, 1, 2]` -> `[0, 1, 2, 0, 1, 2, 0, 1, 2]`.

### DIFF
--- a/alf/algorithms/data_transformer.py
+++ b/alf/algorithms/data_transformer.py
@@ -238,18 +238,16 @@ class FrameStacker(DataTransformer):
         else:
             if spec.minimum.shape != ():
                 assert spec.minimum.shape == spec.shape
-                minimum = np.repeat(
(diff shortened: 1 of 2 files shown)


emailweixu (dismissed):
Good catch. Does this affect training?

breakds:
> Good catch. Does this affect training?

I do not think or observe it affects training so probably no gain in training ...

emailweixu (approved):

PR #1507, opened 2023-05-30
Support native batched environment in fast penv

breakds:
`alf` currently support `batch_size_per_env` in a way that when this parameter is specified to be a value > 1, it will create `BatchEnvironmentWrapper` of `batch_size_per_env` environments in a single process.

This PR extends the mechanism to handle a special case when the individual environment is already batched. In this case, each process will just create an environment with the original environment loader while the `batch_size` is set to `batch_size_per_env`.

This is for the preparation of the multi-robot environment.

### DIFF
--- a/alf/environments/utils.py
+++ b/alf/environments/utils.py
@@ -114,11 +114,18 @@ def create_environment(env_name='CartPole-v0',
             will be used for creating the environment if provided. Otherwise,
             ``env_load_fn`` will be used.
         num_parallel_environments (int): num of parallel environments
-        batch_size_per_env (int): if >1, will create ``num_parallel_environments/batch_size_per_env``
+        batch_size_per_env (int): if >1, will create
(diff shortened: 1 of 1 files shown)


alf/environments/utils.py
```diff
@@ -145,7 +154,8 @@ def create_environment(env_name='CartPole-v0',
         env_load_fn = functools.partial(alf_wrappers.MultitaskWrapper.load,
                                         env_load_fn)
 
-    if hasattr(env_load_fn, 'batched') and env_load_fn.batched:
+    if hasattr(env_load_fn,
+               'batched') and env_load_fn.batched and batch_size_per_env == 1:
```
emailweixu, line 166: need to update doc-string for batch_size_per_env

breakds: Updated the doc-string.

emailweixu (approved):

PR #1513, opened 2023-06-27
change _load_from_state_dict from post-order to pre-order

hnyu:
To enable pytorch's `self._register_load_state_dict_pre_hook` correctly in a root alg, the load state dict order should be changed from post-order to pre-order visit, so that the state_dict is processed first in the root alg and its children and grandchildren can have the correct state_dict.

### DIFF
--- a/alf/algorithms/algorithm.py
+++ b/alf/algorithms/algorithm.py
@@ -961,9 +961,7 @@ class Algorithm(AlgorithmInterface):
         if metadata is not None:
             state_dict._metadata = metadata
 
-        def _load(module, prefix='', visited=None):
-            if visited is None:
-                visited = {self}
+        def _load(module, visited, prefix=''):
             if not is_checkpoint_enabled(module):
                 return
             if isinstance(module, Algorithm):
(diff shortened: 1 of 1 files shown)


breakds:
Can you elaborate a bit on why post order traversal makes `_register_load_state_dict_pre_hook` call correct? Is it because we might want the hook of a parent algorithm to alter the state_dict of an offspring algorithm?

hnyu:
> Can you elaborate a bit on why post order traversal makes `_register_load_state_dict_pre_hook` call correct? Is it because we might want the hook of a parent algorithm to alter the state_dict of an offspring algorithm?

The root algorithm might instruct what ckpt paras to ignore for loading. It's convenient to write a hook to pop certain para keys in the root algorithm. So it's important to alter the state dict first in the parent.

breakds:
> > Can you elaborate a bit on why post order traversal makes `_register_load_state_dict_pre_hook` call correct? Is it because we might want the hook of a parent algorithm to alter the state_dict of an offspring algorithm?
> 
> The root algorithm might instruct what ckpt paras to ignore for loading. It's convenient to write a hook to pop certain para keys in the root algorithm. So it's important to alter the state dict first in the parent.

Make sense. Thanks!

breakds (approved):

Haichao-Zhang (approved):

PR #1514, opened 2023-07-17
populate ee cmd to env

le-horizon:
This is needed to populate keybo controller's locomotion command (ee commands) to simulation env.

### DIFF
--- a/alf/environments/fast_parallel_environment.py
+++ b/alf/environments/fast_parallel_environment.py
@@ -188,15 +188,18 @@ class FastParallelEnvironment(alf_environment.AlfEnvironment):
     def task_names(self):
         return self._task_names
 
-    def set_commands(self, commands):
-        """Set commands for each environment.
+    def call_envs_with_same_args(self, name, *args, **kwargs):
+        """Call each environment's named function with the same args.
 
         Args:
-            commands (dict): a dict of locomotion commands for each environment.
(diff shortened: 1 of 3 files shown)


alf/environments/fast_parallel_environment.py
```diff
@@ -188,6 +188,16 @@ def num_tasks(self):
     def task_names(self):
         return self._task_names
 
+    def set_commands(self, commands):
```
emailweixu: I suggest to add a generic call function to call each parallel environment with same arguments:
```
def call_with_same_args(self, func_name, *args, **kwargs);
```

le-horizon: Good point.  Done for both PRs.  PTAL.

alf/environments/process_environment.py
```diff
@@ -377,6 +377,15 @@ def _receive(self):
         raise KeyError(
             'Received message of unexpected type {}'.format(message))
 
+    def set_commands(self, commands):
```
emailweixu: Can be removed

le-horizon: Done.

le-horizon (commented):
Done, PTAL.

Thanks,
Le

emailweixu (approved):

PR #1515, opened 2023-08-01
PretrainedModel and LoRA adapters

hnyu:
For a pretrained model, sometimes we'd like to not only use its frozen weights, but also finetune on top of it. One idea is to add a small adapter layer to each model layer and only tune the adapter weights.

This PR implements a simple wrapper for pretrained models so that certain layers of a model can be adapted. 

A more advanced way (e.g., https://docs.adapterhub.ml/methods.html) would be to redefine APIs for an existing model class to indicate which layers can be adapted in what ways, but this method requires the knowledge of a model structure and it's not usually possible if we only want to simply load a pretrained model (e.g., LIV, R3M) into the memory, without digging into its internal structure.

For now, only torch.nn.Linear, torch.nn.Embedding, and torch.nn.Conv2d can be adapted, using the simple LoRA method.

Example of loading a pretrained model and add an adapter to it:
```python 
from liv.models.model_liv import LIV
from alf.pretrained_models.pretrained_model import PretraindModel
from alf.pretrained_models.model_adapters import LinearAdapter

liv_model = LIV(...)
liv_model = PretrainedModel(liv_model)
liv_model.add_adapter(LinearAdapter)
output = liv_model.encode_text(command)
...
```


hnyu:
Compared to PR https://github.com/HorizonRoboticsInternal/Hobot/pull/517, this one can be treated as new with quite some improvements.

alf/pretrained_models/pretrained_model.py
```diff
@@ -0,0 +1,92 @@
...
+    def reset_adapter(self):
+        """Reset the adapter weights.
+        """
+        for a in self._adapters:
+            a.reset_parameters()
+
+    def forward(self, input):
+        return self._model(input)
+
+    def state_dict(self, *args, **kwargs):
```
emailweixu, line 93: Is `load_state_dict` also needed to be overloaded?

alf/pretrained_models/pretrained_model.py
```diff
@@ -0,0 +1,92 @@
...
+    def __init__(self, model, name: str = 'PretrainedModel'):
+        super().__init__()
+        # Freeze all parameters
+        for para in model.parameters():
+            para.requires_grad = False
+        self._model = model
+        self._adapters = nn.ModuleList()
+        self._name = name
+
+    def add_adapter(self, adapter_cls: Callable = LoRA):
```
emailweixu: Should we allow the possibility of applying both LinearAdapter and Conv2dAdapter to one model?
If not, it should be explicitly forbidden by the code.

alf/pretrained_models/model_adapters.py
```diff
@@ -0,0 +1,252 @@
...
+            kernel_size = (kernel_size, ) * 2
+        nrows = m.out_channels // m.groups * kernel_size[0]
+        ncols = m.in_channels * kernel_size[1]
+        return nrows, ncols
+
+    def forward(self, input):
+        m = self._m[0]
+        return F.conv2d(
+            input,
+            self._adapter_weight(),
```
emailweixu: the third argument of F.conv2d is `bias`

alf/pretrained_models/model_adapters.py
```diff
@@ -0,0 +1,252 @@
...
+    def _adapter_weight(self):
+        w = self._wA
+        if self._wB is not None:
+            w = self._wB @ w
+        w = w.reshape(self._m[0].weight.shape)
+        scaling = self._alpha
+        if self._wB is not None:
+            # If wB is used, wB @ wA will introduce a scale of self._r, so we
+            # need to scale the weight back.
+            scaling /= self._r
```
emailweixu: It seems to me `/sqrt(self._r)` is better. Assuming each dimension of `wA @ x` are independently distributed with same variance, then the variance of each dimension of `wB @ wA @ x` scales linearly with `r`, which means its magnitude scales with `sqrt(r)`

emailweixu (changes requested):

alf/pretrained_models/pretrained_model.py
hnyu: > Should we allow the possibility of applying both LinearAdapter and Conv2dAdapter to one model? If not, it should be explicitly forbidden by the code.

Right now, this is not supported. `LoRA._adapt` already checks '_forward0' for making sure no adaptation exists.

alf/pretrained_models/pretrained_model.py
hnyu: > Is `load_state_dict` also needed to be overloaded?

added

alf/pretrained_models/model_adapters.py
hnyu: > the third argument of F.conv2d is `bias`

Right, but because it's not used so it's skipped here.

alf/pretrained_models/model_adapters.py
hnyu: > It seems to me `/sqrt(self._r)` is better. Assuming each dimension of `wA @ x` are independently distributed with same variance, then the variance of each dimension of `wB @ wA @ x` scales linearly with `r`, which means its magnitude scales with `sqrt(r)`

This calculation seems correct. The LoRA paper uses 1/r simply. https://github.com/microsoft/LoRA/blob/main/loralib/layers.py#L112 I've updated the comment.

alf/pretrained_models/pretrained_model.py
emailweixu: LinearAdapter only adapts Linear layers. And Conv2dAdapter only adapts Conv2d layers. So applying these two will not cause of error, but the result will be unexpected.

hnyu: > LinearAdapter only adapts Linear layers. And Conv2dAdapter only adapts Conv2d layers. So applying these two will not cause of error, but the result will be unexpected.

You're right. The current code actually supports multiple different adapters with a slight modification.

alf/pretrained_models/model_adapters.py
```diff
@@ -189,6 +189,26 @@ def forward(self, input):
...
+    def can_adapt(cls, m: nn.Module) -> bool:
+        return isinstance(m, nn.Embedding)
+
+    @classmethod
+    def _decompose_weight_dims(cls, m: nn.Module):
+        return m.num_embeddings, m.embedding_dim
+
+    def forward(self, input):
+        m = self._m[0]
+        return F.embedding(input, self._adapter_weight(), m.padding_idx,
```
emailweixu: This is quite costly. `F.embedding(input, wB, ...) @ wA` is more efficient when the low rank mode is used.

alf/pretrained_models/model_adapters.py
```diff
@@ -0,0 +1,272 @@
...
+    @classmethod
+    def can_adapt(cls, m: nn.Module) -> bool:
+        return isinstance(m, nn.Linear)
+
+    @classmethod
+    def _decompose_weight_dims(cls, m: nn.Module):
+        return m.out_features, m.in_features
+
+    def forward(self, input):
+        return F.linear(input, self._adapter_weight())
```
emailweixu: `input @ wA.t() @ wB.t()` is more efficient when the low rank mode is used.

alf/pretrained_models/model_adapters.py
hnyu: > This is quite costly. `F.embedding(input, wB, ...) @ wA` is more efficient when the low rank mode is used.

updated

alf/pretrained_models/model_adapters.py
```diff
@@ -0,0 +1,278 @@
...
+        kernel_size = m.kernel_size
+        if not isinstance(kernel_size, tuple):
+            kernel_size = (kernel_size, ) * 2
+        nrows = m.out_channels // m.groups * kernel_size[0]
+        ncols = m.in_channels * kernel_size[1]
+        return nrows, ncols
+
+    def forward(self, input):
+        m = self._m[0]
+        return F.conv2d(
```
emailweixu: The following might be more efficient (at least in terms of FLOPs) (need to figure out `...` part):
```python
a=F.conv2d(input, wA.reshape(r, w.shape[0],  kerneal_size[0], 1), ...)
out=F.conv2d(a, wB.reshape(w.shape[1],  r, 1, kerneal_size[1]), ...)
```

hnyu: > The following might be more efficient (at least in terms of FLOPs) (need to figure out `...` part):
> 
> ```python
> a=F.conv2d(input, wA.reshape(r, w.shape[0],  kerneal_size[0], 1), ...)
> out=F.conv2d(a, wB.reshape(w.shape[0],  r, 1, kerneal_size[1]), ...)
> ```

It's not easy to convolve in this two-stage way in general. For example, if m.padding > 0, then we shouldn't pad twice otherwise it introduces extra padding effects. If we add the padding only to the first/second conv, then the other one should have kernel_size=1 in order to avoid input size decrease. In this case, the low-rank decomposition has a very small benefit.

hnyu: > The following might be more efficient (at least in terms of FLOPs) (need to figure out `...` part):
> 
> ```python
> a=F.conv2d(input, wA.reshape(r, w.shape[0],  kerneal_size[0], 1), ...)
> out=F.conv2d(a, wB.reshape(w.shape[0],  r, 1, kerneal_size[1]), ...)
> ```

I didn't see a general way for this two-stage convolution? For example, if m.groups > 1, then r must be divisible by m.groups to have the correct input-output mapping.

emailweixu: Perhaps we can just do this for the case of groups=1. As convs with groups > 1 are not that expensive compared with groups==1.

alf/pretrained_models/model_adapters.py
```diff
@@ -261,18 +268,51 @@ def can_adapt(cls, m: nn.Module) -> bool:
...
+                self._adapter_weight(),
+                stride=m.stride,
+                padding=m.padding,
+                dilation=m.dilation,
+                groups=m.groups)
+        else:
+            input = F.conv2d(
+                input,
+                self._wA.reshape(self._r, m.in_channels, 1, m.kernel_size[1]),
+                stride=(1, m.stride[1]),
```
hnyu: Not sure if this results in fewer computations as we always have stride=1 for H in the first convolution (e.g., for stride=(2,2))?

alf/pretrained_models/model_adapters.py
```diff
@@ -0,0 +1,318 @@
...
+            tuple: a pair of ints ``(nrows, ncols)``.
+        """
+        raise NotImplementedError()
+
+    def forward(self, input):
+        raise NotImplementedError()
+
+
+@alf.configurable
+class EmbeddingAdapter(LoRA):
```
breakds: What about putting those concrete implementations in a file (module) `lora.py`, so that when used it looks like `lora.EmbeddingAdapter`? This improves readability slightly, in case there are other adapters.

In factor it looks like this `model_adapters.py` looks like it is dedicate for LoRA at this moment.

alf/pretrained_models/pretrained_model.py
```diff
@@ -0,0 +1,96 @@
...
```
breakds: I am a little worried about the side effect introduced by having `add_adapter`. With this, the user can call `add_adapter` flexibly and maybe even after a checkpoint is saved. This also requires the consumer of the lora weights have to call `add_adapter`s exactly the same as when the the lora weights are trained.

An alternative would be to pass the adapter constructors as arguments to the constructor. In this way, it is easily configurable with `alf.config`, `partial` or subclass, so that as long as the consumer and producer get the same configuration, the checkpoints are sure to be compatible.

alf/pretrained_models/model_adapters.py
hnyu: > What about putting those concrete implementations in a file (module) `lora.py`, so that when used it looks like `lora.EmbeddingAdapter`? This improves readability slightly, in case there are other adapters.
> 
> In factor it looks like this `model_adapters.py` looks like it is dedicate for LoRA at this moment.

I've put adapters into model_adapters/lora.py

alf/pretrained_models/pretrained_model.py
hnyu: > I am a little worried about the side effect introduced by having `add_adapter`. With this, the user can call `add_adapter` flexibly and maybe even after a checkpoint is saved. This also requires the consumer of the lora weights have to call `add_adapter`s exactly the same as when the the lora weights are trained.
> 
> An alternative would be to pass the adapter constructors as arguments to the constructor. In this way, it is easily configurable with `alf.config`, `partial` or subclass, so that as long as the consumer and producer get the same configuration, the checkpoints are sure to be compatible.

Moved adapter cls list to the constructor of `PretrainedModel`

breakds (approved):

emailweixu (approved):

PR #1517, opened 2023-08-02
Change DDPG's OU noise to noise process

Haichao-Zhang:
This change makes the DDPG algorithm handles the state of the noise better, in a thread safe manner.


alf/algorithms/ddpg_algorithm.py
```diff
@@ -222,33 +237,37 @@ def predict_step(self, inputs: TimeStep, state):
...
             else:
-                ind_explore = torch.where(
-                    torch.rand(a.shape[:1]) < epsilon_greedy)
-                noisy_a = a + ou()
-                a[ind_explore[0], :] = noisy_a[ind_explore[0], :]
+                choose_noisy_action = torch.where(
+                    torch.rand(a.shape[:1]) < epsilon_greedy)[0]
+                noisy_a = a + noise
+                a[choose_noisy_action[0], :] = noisy_a[
+                    choose_noisy_action[0], :]
```
emailweixu: Can be simplified as:
```pyhon
                choose_noisy_action = torch.rand(a.shape[0]) < epsilon_greedy
                a[choose_noisy_action] += noisy[choose_noisy_action]
```

Haichao-Zhang: Changed as suggested

emailweixu (approved):

PR #1518, opened 2023-08-02
Fix temporally_independent_train_step

emailweixu:
Since TrainerConfig is passed to some sub algorithms, `TrainerConfig.temporally_independent_train_step` can be changed by the sub-algorithm from `None` to False if the sub-algorithm is stateless. However, the parent algorithm may be stateful. But since config.temporally_independent_train_step is not `None` anymore, it will not be changed to True even if the parent algorithm is stateful.

Clearly, we want temporally_independent_train_step to be True if the parent algorithm is stateful.

### DIFF
--- a/alf/algorithms/algorithm.py
+++ b/alf/algorithms/algorithm.py
@@ -163,13 +163,15 @@ class Algorithm(AlgorithmInterface):
         # See the ``force_params_visible_to_parent`` property below for details.
         self._force_params_visible_to_parent = False
         self._grad_scaler = None
+        self._temporally_independent_train_step = None
         if config:
+            self._temporally_independent_train_step = config.temporally_independent_train_step
             self.use_rollout_state = config.use_rollout_state
-            if config.temporally_independent_train_step is None:
-                config.temporally_independent_train_step = (len(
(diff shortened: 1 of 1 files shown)


alf/algorithms/algorithm.py
```diff
@@ -163,13 +163,15 @@ def __init__(self,
         # See the ``force_params_visible_to_parent`` property below for details.
         self._force_params_visible_to_parent = False
         self._grad_scaler = None
+        self._temporally_independent_train_step = None
         if config:
```
hnyu, line 167: config cannot be None because we use `self._config` everywhere in the file?

emailweixu: config is not always passed to the Algorithm constructor.

hnyu: > config is not always passed to the Algorithm constructor.

Then some places will cause errors? Because they don't check if `self._config` is None

emailweixu: Yes, that will cause error. But many algorithm will not go to those places.

hnyu (approved):

PR #1521, opened 2023-08-03
make adapter weights have the same dtype with the base model

hnyu:


alf/pretrained_models/model_adapters/lora.py
```diff
@@ -78,10 +79,10 @@ def __init__(self,
         # adapter size: (nrows + ncols) * r
         # need low-rank adaptation: (nrows + ncols) * r < nrows * ncols
         if rank > ncols * nrows / (nrows + ncols):
-            self._wA = nn.Parameter(torch.Tensor(nrows, ncols))
+            self._wA = nn.Parameter(torch.Tensor(nrows, ncols).to(dtype))
```
emailweixu: Can directly contruct the tensor with the desired dtype.

emailweixu (approved):

PR #1525, opened 2023-08-08
Show the random seed used in play

Haichao-Zhang:
So that it is more convenient to reproduce a previous episode which was generated without specifying a random seed.

### DIFF
--- a/alf/bin/play.py
+++ b/alf/bin/play.py
@@ -121,6 +121,7 @@ def play():
     else:
         alf.config('create_environment', for_evaluation=True, nonparallel=True)
     alf.config('TrainerConfig', mutable=False, random_seed=seed)
+    common.info('Playing with random seed: {}'.format(seed))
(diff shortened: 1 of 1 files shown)


alf/bin/play.py
```diff
@@ -121,6 +121,7 @@ def play():
     else:
         alf.config('create_environment', for_evaluation=True, nonparallel=True)
     alf.config('TrainerConfig', mutable=False, random_seed=seed)
+    common.info('Playing with random seed: {}'.format(seed))
```
emailweixu, line 124: However, if the seed is not explicitly specified when calling common.set_random_seed() at line 112, torch deterministic mode  will not be enabled. So even supplying the same seed later may not reproduce earlier result.

Haichao-Zhang: True. This is not the best way to use random seed for reproducing previous results. Close it for now.

PR #1526, opened 2023-08-09
Fix grid_search

emailweixu:
parse_conf_file() needs to be called before write_config() due to https://github.com/HorizonRobotics/alf/pull/1476

### DIFF
--- a/alf/bin/grid_search.py
+++ b/alf/bin/grid_search.py
@@ -394,10 +394,7 @@ def launch_snapshot_gridsearch():
     # write the current conf file as
     # ``<root_dir>/alf_config.py`` or ``<root_dir>/configured.gin``
     conf_file = common.get_conf_file()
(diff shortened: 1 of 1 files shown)


alf/bin/grid_search.py
```diff
@@ -394,10 +394,7 @@ def launch_snapshot_gridsearch():
     # write the current conf file as
     # ``<root_dir>/alf_config.py`` or ``<root_dir>/configured.gin``
     conf_file = common.get_conf_file()
-    if conf_file.endswith('.gin'):
-        # for gin, we need to parse it first. Otherwise, configured.gin will be
-        # empty
-        common.parse_conf_file(conf_file)
+    common.parse_conf_file(conf_file)
```
hnyu, line 397: Calling parse_conf_file in the main process will create an env, which might be unnecessary or even problematic in certain cases?

PR #1528, opened 2023-08-16
white and black lists for pretrained model

hnyu:
Added whitelist and blacklist for filtering modules to be adapted. This is helpful if we only want to finetune part of a pretrained model.

### DIFF
--- a/alf/pretrained_models/pretrained_model.py
+++ b/alf/pretrained_models/pretrained_model.py
@@ -71,13 +71,13 @@ class PretrainedModel(nn.Module):
                     if re.search(b, name):
                         skip = True
                         break
-            include = (module_whitelist is None)
-            if module_whitelist is not None:
+            elif module_whitelist is not None:
+                skip = True
                 for w in module_whitelist:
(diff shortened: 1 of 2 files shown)


alf/pretrained_models/pretrained_model.py
```diff
@@ -40,27 +43,59 @@ def __init__(self,
...
+                        skip = True
+                        break
+            include = (module_whitelist is None)
+            if module_whitelist is not None:
+                for w in module_whitelist:
+                    if re.search(w, name):
+                        include = True
+                        break
+            if skip or not include:
+                continue
```
emailweixu: I feel the following is a little bit easier to understand:
```python
            skip = False
            if module_blacklist is not None:
                for b in module_blacklist:
                    if re.search(b, name):
                        skip = True
                        break
            elif module_whitelist is not None:
                skip = True
                for w in module_whitelist:
                    if re.search(w, name):
                        skip = False
                        break
            if skip:
                 continue
```

emailweixu (approved):

PR #1530, opened 2023-08-29
add checking root conf file

hnyu:
Sometimes we don't want to evaluate upstream task confs when using `alf.import_config` to avoid their side effects, so we can use `is_root_conf_file(__file__)` in those files.

### DIFF
--- a/alf/config_util.py
+++ b/alf/config_util.py
@@ -945,6 +945,15 @@ def load_config(conf_file):
     return _import_config(conf_file)
 
 
+def is_root_conf_file(conf_file):
+    """Check if a conf file is the root conf file.
+    """
+    global _ROOT_CONF_FILE
+    assert _ROOT_CONF_FILE is not None, "You should first call alf.load_config()"
+    conf_file = _get_conf_file_full_path(conf_file)
+    return _ROOT_CONF_FILE == conf_file
+
+
 def save_config(alf_config_file):
     """Save config files.
 


emailweixu:
> Sometimes we don't want to evaluate upstream task confs when using `alf.import_config` to avoid their side effects, so we can use `is_root_conf_file(__file__)` in those files.

Can you give an example use case of this?

hnyu:
> > Sometimes we don't want to evaluate upstream task confs when using `alf.import_config` to avoid their side effects, so we can use `is_root_conf_file(__file__)` in those files.
> 
> Can you give an example use case of this?

Right now, expert config has some global side effects (e.g., data transformer). I need to import an expert config in an agent config but doesn't want to have those side effects. So I can add a check in the expert config:

```python
if is_root_conf_file(__file__):
    get_expert_constructor_and_config = alf.import_config(
        "expert_constructors.py").get_expert_constructor_and_config

    # prefix@path: the prefix can be inspected from `ckpt-structure.json`. Example:
    # ckpt = "algorithm._rl_algorithm@/tmp/run4/train/algorithm/ckpt-75"
    ckpt = None

    create_expert_algorithm, conf_dict = get_expert_constructor_and_config(
        expert_config, jump_start_ratio=0., ckpt=ckpt)

    alf.config(conf_dict)
```

emailweixu:
> > > Sometimes we don't want to evaluate upstream task confs when using `alf.import_config` to avoid their side effects, so we can use `is_root_conf_file(__file__)` in those files.
> > 
> > 
> > Can you give an example use case of this?
> 
> Right now, expert config has some global side effects (e.g., data transformer). I need to import an expert config in an agent config but doesn't want to have those side effects. So I can add a check in the expert config:
> 
> ```python
> if is_root_conf_file(__file__):
>     get_expert_constructor_and_config = alf.import_config(
>         "expert_constructors.py").get_expert_constructor_and_config
> 
>     # prefix@path: the prefix can be inspected from `ckpt-structure.json`. Example:
>     # ckpt = "algorithm._rl_algorithm@/tmp/run4/train/algorithm/ckpt-75"
>     ckpt = None
> 
>     create_expert_algorithm, conf_dict = get_expert_constructor_and_config(
>         expert_config, jump_start_ratio=0., ckpt=ckpt)
> 
>     alf.config(conf_dict)
> ```

Perhaps the right way is to make sure it doesn't have side effect?

PR #1532, opened 2023-09-01
Avoid out of GPU memory issue when loading checkpoints

Haichao-Zhang:
This PR fixes the GPU out of memory issue when loading checkpoints.

### DIFF
--- a/alf/utils/checkpoint_utils.py
+++ b/alf/utils/checkpoint_utils.py
@@ -61,10 +61,9 @@ def extract_sub_state_dict_from_checkpoint(checkpoint_prefix, checkpoint_path):
             by ALF, e.g. "/path_to_experiment/train/algorithm/ckpt-100".
     """
 
-    map_location = None
-    if not torch.cuda.is_available():
-        map_location = torch.device('cpu')
-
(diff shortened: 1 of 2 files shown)


alf/utils/checkpoint_utils.py
```diff
@@ -249,9 +248,9 @@ def _merge_checkpoint(merged, new):
                 "Checkpoint '%s' does not exist. Train from scratch." % f_path)
             return self._global_step
 
-        map_location = None
-        if not torch.cuda.is_available():
-            map_location = torch.device('cpu')
+        # use ``cpu`` as the map location to avoid GPU RAM surge when loading a
+        # model checkpoint.
+        map_location = torch.device('cpu')
```
hnyu, line 253: Do we need to convert the model weights to gpu later?

Haichao-Zhang: > Do we need to convert the model weights to gpu later?

No need to do that. I have been using this since I encountered the out of GPU memory issue when loading checkpoints.

Haichao-Zhang: Also added a test on this.

emailweixu (approved):

PR #1534, opened 2023-09-13
Fix incorrect throughput print for on-policy algorithm

emailweixu:
On-policy algorithm uses train_from_unroll(). Previouly, it uses `torch.tensor(alf.nest.get_nest_shape(experience)).prod()` to calculate the number samles being trained, which is not correct because get_nest_shape() is not gauranteed to return the correct shape.


hnyu (dismissed):

hnyu:
In general, experience might not contain the field `step_type`, especially for some custom replay buffer cases. For example, a subalgorithm (non-RL) has its own replay buffer and decides on its own how to make an experience. Depending on how the experiences will be used, step_type might be skipped for storing.

hnyu (approved):

PR #1538, opened 2023-09-21
fix monet unet

hnyu:
The current UNet has some issue that the channel numbers of the downsampling path and those of the upsampling path are not always symmetric. This causes an issue for creating skip connections. For example, for a filter list of (32, 64, 64, 128, 128, 128), the result channels are:

```python
DOWN OUTPUT SHAPE:  torch.Size([2, 15, 96, 128])
DOWN OUTPUT SHAPE:  torch.Size([2, 32, 96, 128])
DOWN OUTPUT SHAPE:  torch.Size([2, 64, 48, 64])
DOWN OUTPUT SHAPE:  torch.Size([2, 64, 24, 32])
DOWN OUTPUT SHAPE:  torch.Size([2, 128, 12, 16])
DOWN OUTPUT SHAPE:  torch.Size([2, 128, 6, 8])
UP OUTPUT SHAPE:  torch.Size([2, 128, 3, 4])
UP OUTPUT SHAPE:  torch.Size([2, 128, 6, 8])
UP OUTPUT SHAPE:  torch.Size([2, 128, 12, 16])
UP OUTPUT SHAPE:  torch.Size([2, 128, 24, 32])
```

Then 128 is concat with 64, the network crashes. 

This PR fixes this asymmetry issue.

### DIFF
--- a/alf/algorithms/monet_algorithm.py
+++ b/alf/algorithms/monet_algorithm.py
@@ -58,7 +58,10 @@ class MoNetUNet(alf.networks.Network):
             filters: a tuple of output channels along the downsampling path, each for
                 a conv layer. The upsampling path uses a reversed tuple.
             nonskip_fc_layers: a tuple of fc layer sizes for the bottleneck connection
-                (nonskip) of the UNet.
+                (nonskip MLP in the illustration) of the UNet. Note that there
+                will be implicitly an additional FC layer after this MLP to project
+                its output to a proper dimension (the output size of the downsampling
+                path) before being reshaped for the upsampling path.
             output_channels: final output channels. The output features are non-activated.
         """
         super().__init__(input_tensor_spec=input_tensor_spec, name=name)
--- a/alf/algorithms/monet_algorithm.py
+++ b/alf/algorithms/monet_algorithm.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import Callable, Tuple
+from typing import Callable, Tuple, List
 import numpy as np
 
 import torch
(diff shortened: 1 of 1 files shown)


alf/algorithms/monet_algorithm.py
```diff
@@ -90,42 +90,43 @@ def __init__(self,
...
-            fc_layer_params=nonskip_fc_layers,
-            last_layer_size=last_skip_tensor_spec.numel,
-            last_activation=torch.relu_)
+            input_tensor_spec=alf.TensorSpec((last_skip_tensor_spec.numel, )),
+            fc_layer_params=nonskip_fc_layers)
 
-        self._reshape = alf.layers.Reshape(*last_skip_tensor_spec.shape)
+        self._reshape = torch.nn.Sequential(
+            alf.layers.FC(nonskip_fc_layers[-1], last_skip_tensor_spec.numel,
+                          torch.relu_),
```
emailweixu, line 101: This can be part of nonskip_mlp. And should update document of nonskip_fc_layers about the additional FC layer.

hnyu: > This can be part of nonskip_mlp. And should update document of nonskip_fc_layers about the additional FC layer.

The reason why I chose to put this additional FC layer in `self._reshape` instead of `self._nonskip_mlp` is that sometimes we want to perform extra operations on the output of `self._nonskip_mlp` (as in the vonet case). This output can be obtained by calling `self.encode()`. If the additional FC is in `self._nonskip_mlp` then the encoding dim might be large depending on `last_skip_tensor_spec.numel` and is not straightforwardly controllable.

emailweixu: Should still comment about this additional fc in doc-string of the class.

hnyu: > Should still comment about this additional fc in doc-string of the class.

added

emailweixu (approved):

PR #1548, opened 2023-10-14
Add mask_out_loss_for_last_step for TrainerConfig

emailweixu:
By default, ALF mask out loss for LAST step. However, we may want to have some flexibility in some situations. So we add this flag.

### DIFF
--- a/alf/algorithms/config.py
+++ b/alf/algorithms/config.py
@@ -38,6 +38,7 @@ class TrainerConfig(object):
                  unroll_parameter_update_period: int = 10,
                  use_rollout_state=False,
                  temporally_independent_train_step=None,
+                 mask_out_loss_for_last_step=True,
                  num_checkpoints=10,
                  confirm_checkpoint_upon_crash=True,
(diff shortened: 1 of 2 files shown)


Haichao-Zhang (approved):

alf/algorithms/config.py
```diff
@@ -159,6 +160,11 @@ def __init__(self,
                 states will be taken from the replay buffer; otherwise they will
                 be set to 0. In the case of True, the ``train_state_spec`` of an
                 algorithm should always be a subset of the ``rollout_state_spec``.
+            mask_out_loss_for_last_step (bool): If True, the loss for the last
+                step of each sequence will be masked out. For RL training,
```
Haichao-Zhang, line 164: ``mask_out_loss_for_last_step`` is True does not always do loss masking. The current logic also requires  ``is_rl()`` is True for ``mask_out_loss_for_last_step`` to take effect.

emailweixu: If not is_rl(), then there is no last step either.

PR #1550, opened 2023-10-17
[Tiny] Fix deprecated seaborn style

breakds:
`seaborn-dark` has been deprecated in `matplotlib` long time ago, and at the latest version it will report error instead of warning, and suggesting using `seaborn-v0_8-dark` if we insist. This PR follows that suggestion. I think this is better compared to the alternative of pinning the version of `matplotlib`.

### DIFF
--- a/alf/summary/render.py
+++ b/alf/summary/render.py
@@ -19,7 +19,10 @@ import matplotlib
 matplotlib.use('Agg')  # 'Agg' no need for xserver!
 import matplotlib.pyplot as plt
 # Style gallery: https://tonysyu.github.io/raw_content/matplotlib-style-gallery/gallery.html
-plt.style.use('seaborn-dark')
+# The seaborn styles shipped by Matplotlib are deprecated since 3.6,
+# as they no longer correspond to the styles shipped by seaborn.
+# However, they will remain available as 'seaborn-v0_8-<style>'.
+plt.style.use('seaborn-v0_8-dark')
 try:
     import rpack
 except ImportError:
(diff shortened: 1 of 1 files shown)


alf/summary/render.py
```diff
@@ -19,7 +19,10 @@
 matplotlib.use('Agg')  # 'Agg' no need for xserver!
 import matplotlib.pyplot as plt
 # Style gallery: https://tonysyu.github.io/raw_content/matplotlib-style-gallery/gallery.html
-plt.style.use('seaborn-dark')
+# The seaborn styles shipped by Matplotlib are deprecated since 3.6,
+# as they no longer correspond to the styles shipped by seaborn.
+# However, they will remain available as 'seaborn-v0_8-<style>'.
+plt.style.use('seaborn-v0_8-dark')
```
hnyu: Will matplotlib with a lower version be compatible with this option?

breakds: As long as it is not too low. I am not sure how low is too low though, probably below 3.0.

breakds: The problem is that if I put a `< 3.6` in our CICD docker image, it will be too restrictive. But otherwise my current PR will be blocked.

hnyu: > As long as it is not too low. I am not sure how low is too low though, probably below 3.0.

Tested it's not available in our server docker. Need to make this compatible.

hnyu: You can try and catch

breakds: Thanks. Updated with try/catch.

hnyu (approved):

PR #1553, opened 2023-10-18
Fix SummarizeGradient

emailweixu:
Sometimes (perhaps caused by torch 2.1?) the cloned tensor is not a leaf. And torch does not allow setting requires_grad for non-leaf tensor.

So changed to summarizing the gradient only if requires_grad is True (summarize_tensor_gradients() does this).

### DIFF
--- a/alf/layers.py
+++ b/alf/layers.py
@@ -3212,30 +3212,23 @@ class SummarizeGradient(ElementwiseLayerBase):
     def __init__(self, name):
         """A layer for summarizing the gradient of the input tensor.
 
-        Summarize the gradient of the input tensor. Always first cloning the
-        input tensor and then setting ``requires_grad=True`` for the cloned
-        tensor to enable gradient calculation for summarization.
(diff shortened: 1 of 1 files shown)


Haichao-Zhang (dismissed):

alf/layers.py
```diff
@@ -3212,30 +3212,23 @@ class SummarizeGradient(ElementwiseLayerBase):
     def __init__(self, name):
         """A layer for summarizing the gradient of the input tensor.
 
-        Summarize the gradient of the input tensor. Always first cloning the
-        input tensor and then setting ``requires_grad=True`` for the cloned
-        tensor to enable gradient calculation for summarization.
+        Summarize the gradient of the input tensor. Note that if the input tensor
+        does not requires gradient (i.e. x.requires_grad is False), gradient
+        will not be summarized. (This is because )
```
hnyu: Unfinished sentence?

Haichao-Zhang (approved):

PR #1557, opened 2023-11-05
Fix sac alpha_loss bug

Haichao-Zhang:
There is a bug in the current sac's ``alpha_loss``, which will cause problems when the ``action_spec`` contains multiple specs, e.g. multi-head case (apart from the Mixed Action case which is specially handled)
- the target_entropy is calculated as the aggregated entropy across all action specs: 
``target_entropy = np.sum(list(map(target_entropy, flat_action_spec)))``
- in the alpha loss, the aggregated ``target_entropy`` as calculated above is used to calculate the alpha loss for each individual spec in ``action_spec``, which is incorrect when the number of specs (denoted as ``N``) in ``action_spec`` is larger than one. In this case, the target entropy is effectively scaled by a factor of ``N``.

### DIFF
--- a/alf/algorithms/sac_algorithm.py
+++ b/alf/algorithms/sac_algorithm.py
@@ -784,9 +784,15 @@ class SacAlgorithm(OffPolicyAlgorithm):
         return state, info
 
     def _alpha_train_step(self, log_pi):
-        neg_entropy = sum(nest.flatten(log_pi))
-        alpha_loss = self._log_alpha * (
-            -neg_entropy - self._target_entropy).detach()
+        if self._act_type == ActionType.Mixed:
+            alpha_loss = nest.map_structure(
+                lambda la, lp, t: la * (-lp - t).detach(), self._log_alpha,
+                log_pi, self._target_entropy)
+            alpha_loss = sum(nest.flatten(alpha_loss))
+        else:
+            neg_entropy = sum(nest.flatten(log_pi))
+            alpha_loss = self._log_alpha * (
+                -neg_entropy - self._target_entropy).detach()
         return alpha_loss
 
     def train_step(self, inputs: TimeStep, state: SacState,
--- a/alf/algorithms/sac_algorithm.py
+++ b/alf/algorithms/sac_algorithm.py
@@ -789,9 +789,9 @@ class SacAlgorithm(OffPolicyAlgorithm):
             # in the mixed action case, both log_pi and _target_entropy are tuples of two
             # elements, corresponding to the discrete action and continuous action respectively
             for i in range(2):
-                neg_entropy = sum(nest.flatten(log_pi[i]))
-                alpha_loss = self._log_alpha * (
-                    -neg_entropy - self._target_entropy[i]).detach()
+                neg_entropy_i = sum(nest.flatten(log_pi[i]))
+                alpha_loss += self._log_alpha[i] * (
+                    -neg_entropy_i - self._target_entropy[i]).detach()
         else:
             neg_entropy = sum(nest.flatten(log_pi))
             alpha_loss = self._log_alpha * (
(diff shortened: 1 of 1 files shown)


alf/algorithms/sac_algorithm.py
```diff
@@ -784,10 +784,16 @@ def _critic_train_step(self, inputs: TimeStep, state: SacCriticState,
         return state, info
 
     def _alpha_train_step(self, log_pi):
-        alpha_loss = nest.map_structure(
-            lambda la, lp, t: la * (-lp - t).detach(), self._log_alpha, log_pi,
-            self._target_entropy)
-        return sum(nest.flatten(alpha_loss))
+        if self._act_type == ActionType.Mixed:
+            alpha_loss = nest.map_structure(
+                lambda la, lp, t: la * (-lp - t).detach(), self._log_alpha,
```
emailweixu: this still have problem if the continuous action is nested

Haichao-Zhang: Now also fixed.

alf/algorithms/sac_algorithm.py
```diff
@@ -784,10 +784,19 @@ def _critic_train_step(self, inputs: TimeStep, state: SacCriticState,
...
-            lambda la, lp, t: la * (-lp - t).detach(), self._log_alpha, log_pi,
-            self._target_entropy)
-        return sum(nest.flatten(alpha_loss))
+        if self._act_type == ActionType.Mixed:
+            alpha_loss = 0
+            # in the mixed action case, both log_pi and _target_entropy are tuples of two
+            # elements, corresponding to the discrete action and continuous action respectively
+            for i in range(2):
+                neg_entropy = sum(nest.flatten(log_pi[i]))
+                alpha_loss = self._log_alpha * (
```
emailweixu: +=

Haichao-Zhang: Sorry the previous fix was not done correctly. Fixed the loss aggregation and also added the missing index to ``self._log_alpha``.

emailweixu (approved):

hnyu:
> There is a bug in the current sac's `alpha_loss`, which will cause problems when the `action_spec` contains multiple specs, e.g. multi-head case (apart from the Mixed Action case which is specially handled)
> 
> * the target_entropy is calculated as the aggregated entropy across all action specs:
>   `target_entropy = np.sum(list(map(target_entropy, flat_action_spec)))`
> * in the alpha loss, the aggregated `target_entropy` as calculated above is used to calculate the alpha loss for each individual spec in `action_spec`, which is incorrect when the number of specs (denoted as `N`) in `action_spec` is larger than one. In this case, the target entropy is effectively scaled by a factor of `N`.

I think you are not using the original alpha_train_step correctly? For a typical non-mixed action case, the log_pi, log_alpha, and target_entropy should be *scalars*, no matter how complex the action nest is. In other words, the log_pi passed to `alpha_train_step()` should be the sum of the nested action log_probs before calling this function. In this case, the calculation and use of target entropy are reasonable. In fact, `train_step()` has done this as below:

```python
if self._act_type == ActionType.Mixed:
            # For mixed type, add log_pi separately
            log_pi = type(self._action_spec)((sum(nest.flatten(log_pi[0])),
                                              sum(nest.flatten(log_pi[1]))))
 else:
            log_pi = sum(nest.flatten(log_pi))
```

For mixed-action case, log_pi, log_alpha, and target_entropy are all two-elements tuples. That means discrete action alpha and continuous action alpha have separate target entropies. The continuous action still has one overall log_pi, no matter what the action nest looks like. 

It's weird because SAC's  `train_step()` has done the summation of log_prob for you before calling `alpha_train_step`. I think your custom SAC algorithm somehow calculates the target entropy and log_pi in a different way.

PR #1560, opened 2023-11-15
Improve TrainerConfig.num_evals

emailweixu:
When TrainerConfig.num_evals is set, there should be so many evaluations performed uniformed spread across the training.

However, previous way of determining the evaluation interval using compute_summary_or_eval_interval() can be too small because the number of iterations caluated using `config.num_env_steps / (num_envs * config.unroll_length)` is too small due to short episodes (note that the last step of an episode is not counted towards num_env_steps).

So change to use progress() to determine whether to do the evaluation.

### DIFF
--- a/alf/trainers/policy_trainer.py
+++ b/alf/trainers/policy_trainer.py
@@ -629,6 +629,13 @@ class RLTrainer(Trainer):
         if self._evaluate and iter_num == 0:
             self._eval()
 
+        if self._evaluate:
+            if self._config.num_evals is None:
+                self._num_evals_performed = iter_num // self._eval_interval
+            else:
+                self._num_evals_performed = int(
+                    self.progress() * self._config.num_evals)
+
         while True:
             t0 = time.time()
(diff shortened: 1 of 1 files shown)


hnyu:
Does not quite get the point. If the number of iterations calculated using `config.num_env_steps / (num_envs * config.unroll_length)` is not right, shouldn't we fix this instead? Because this value will impact many other statistics.

emailweixu:
> Does not quite get the point. If the number of iterations calculated using `config.num_env_steps / (num_envs * config.unroll_length)` is not right, shouldn't we fix this instead? Because this value will impact many other statistics.

It is hard to fix it for the following reason. In ALF, the first step of an episode is not counted toward EnvironmentSteps metric. But the first step is counted toward unroll_length. So there is not a simple formula that can calculate the eval interval from `config.num_env_steps` and `config.unroll_length`

hnyu (approved):

PR #1566, opened 2023-11-21
Support schedulers for TargetUpdater

emailweixu:


alf/utils/common.py
```diff
@@ -186,7 +187,7 @@ def __init__(self,
                  models,
                  target_models,
                  tau=1.0,
```
Haichao-Zhang: also also add data type annotation ``Union[int, Scheduler]`` to tau?

Haichao-Zhang (dismissed):

Haichao-Zhang (approved):

PR #1568, opened 2023-11-28
`FastParallelEnvironment` can optionally use "spawn" as the start method

breakds:
## Motivation

The subprocesses of `ProcessEnvironment` is currently created via the `"fork"` start method. The benefit is that `"fork"` is fast and it inherits the alf configurations automatically.

However, there can be cases when inheriting everything via `"fork"` is not desired. We ran into such case with `MuJoCo`'s EGL context for rendering. See https://github.com/HorizonRoboticsInternal/Hobot/issues/756 for details.

## What is changed?

1. Support `start_method="spawn"` for `FastParallelEnvironment`. This is done by using "spawn" context to start the process for `_worker`.
2. When `start_method="spawn"`, a special initialization is done to make sure that the configuration is properly loaded.
3. To avoid infinite recursion (i.e. parsing the configuration ⇨ load parallel environment ⇨ subprocess `_worker` ⇨ parsing configuration ⇨ load parallel environment ...)
    - During the above said initialization, there is `SpawnedProcessContext` being created at the very beginning of `_worker()`. Where the (single) environment constructor is passed in.
    - The above context, when set, will alter the behavior of `get_env()`. Instead of creating the top-level `ParallelEnvironment`, it will just create that single environment for the spawned process. Infinite recursion is avoided, while also `get_action_spec()` and `get_observation_spec()` which depends on `get_env()` continue to work as expected.

There is also a very small fix for newer version of gym, which (very annoyingly) add `TimeLimit.truncated` sometimes to the env info.

## How is this tested?

1. Using `ppo_cart_pole_conf.py` to test the before/after. See graph below:

spawn_process_env_comparison_ppo_cart_pole

2. Tested this in Hobot with "play". Previously I cannot "play" with both camera rendering and the passive mujoco viewer. Now it works perfectly without exception.

### DIFF
--- a/alf/config_helpers.py
+++ b/alf/config_helpers.py
@@ -246,7 +246,7 @@ def get_env():
         if isinstance(ctx, SpawnedProcessContext):
             _env = ctx.create_env()
             return _env
-        
+
         if _is_parsing:
             random_seed = get_config_value('TrainerConfig.random_seed')
         else:
--- a/alf/config_helpers.py
+++ b/alf/config_helpers.py
@@ -26,6 +26,7 @@ from alf.config_util import config1, get_config_value, load_config, pre_config, 
 from alf.environments.utils import create_environment
 from alf.utils.common import set_random_seed
 from alf.utils.per_process_context import PerProcessContext
+from alf.utils.spawned_process_utils import SpawnedProcessContext, get_spawned_process_context
 
 __all__ = [
     'close_env', 'get_raw_observation_spec', 'get_observation_spec',
@@ -190,21 +191,13 @@ def adjust_config_by_multi_process_divider(ddp_rank: int,
         config1('TrainerConfig.evaluate', False, raise_if_used=False)
 
 
-def parse_config(conf_file, conf_params, create_env: bool = True):
+def parse_config(conf_file, conf_params):
     """Parse config file and config parameters
-
-    Note: by default a global environment will be created (which can be obtained
-    by alf.get_env()) and random seed will be initialized by this function using
-    common.set_random_seed(). Such behavior can be turned off by setting
-    ``create_env = False`` (one use case is when creating ``ProcessEnvironment``
-    wiht "spawn" instead of "fork", where each subprocess will need to parse the
-    config again but not creating the root environment by themselves).
 
     Args:
         conf_file (str): The full path of the config file.
         conf_params (list[str]): the list of config parameters. Each one has a
             format of CONFIG_NAME=VALUE.
-        create_env: Whether to create the environment at the end of parsing.
 
     """
     global _is_parsing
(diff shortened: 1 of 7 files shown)


alf/config_helpers.py
```diff
@@ -223,7 +228,8 @@ def parse_config(conf_file, conf_params):
         _is_parsing = False
 
     # Create the global environment and initialize random seed
-    get_env()
+    if create_env:
```
emailweixu: This won't solve the problem for all configuration. Some will call alf.get_observation_spec() or alf.get_action_spec() inside the configuration. Both function will call get_env()

breakds: I see what you mean. The tricky part is that at this moment when environment is being created, even the parent process cannot do `get_observation_spec` yet because of that `env = get_env()` call. Let me think of a solution. 

The "simple fix" might be not fixing it, i.e. just throw exception for the cases when `get_env()` gets called during config parse in the subprocesses ... which at least stops the infinite recursion.

breakds: A viable approach might be hacking the `create_environment` function with pre configs. Let me do some experiment.

breakds: I thought about another solution, which is to create the environment at main process, get the action/observation spec and send them to the subprocess.

However, this defeats the purpose that we do not want to construct the environment at the main process (e.g. mujoco at this moment will create the camera and mess up EGL context ...)

breakds:
Okay I think I find a good middle ground to solve the `get_env()` issue. Updated the PR. See "What is changed" in the description, which is also updated.

alf/config_helpers.py
```diff
@@ -193,14 +194,11 @@ def adjust_config_by_multi_process_divider(ddp_rank: int,
 def parse_config(conf_file, conf_params):
     """Parse config file and config parameters
 
-    Note: a global environment will be created (which can be obtained by
-    alf.get_env()) and random seed will be initialized by this function using
-    common.set_random_seed().
-
```
emailweixu: This comment is still valid.

alf/utils/common.py
```diff
@@ -621,12 +621,9 @@ def parse_conf_file(conf_file):
 
     It also looks for FLAGS.gin_param and FLAGS.conf_param for extra configs.
 
-    Note: a global environment will be created (which can be obtained by
-    alf.get_env()) and random seed will be initialized by this function using
-    common.set_random_seed().
-
```
emailweixu: restore the comment

alf/config_helpers.py
breakds: The comment is added back. Mistakenly removed when clearing some appended docstring.

alf/utils/common.py
breakds: Fixed.

emailweixu (approved):

PR #1570, opened 2023-12-04
Avoid unnecessary import (gin, gym)

emailweixu:
They are not always available.

### DIFF
--- a/alf/environments/utils.py
+++ b/alf/environments/utils.py
@@ -18,7 +18,6 @@ import numpy as np
 import random
 
 import alf
-from alf.environments import suite_gym
 from alf.environments import thread_environment, parallel_environment, fast_parallel_environment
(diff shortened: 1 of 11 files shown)


alf/environments/utils.py
```diff
@@ -172,6 +171,12 @@ def create_environment(env_name='CartPole-v0',
 
     """
 
+    if env_name is None:
+        env_name = 'CartPole-v0'
```
Haichao-Zhang, line 176: This seems to be the same as using ``CartPole-v0`` as the default value for the argument ``env_name``? Not sure if there is any particular reason for this change.

Haichao-Zhang (dismissed):

breakds (dismissed):

Haichao-Zhang (dismissed):

PR #1572, opened 2023-12-07
S5: Simplified State Space Layers for Sequence Modeling

emailweixu:
https://openreview.net/forum?id=Ai8Hw3AXqks

Tested on sMNIST got similar result as running the original S5 code.

Results from this PR
Train Loss: 0.00022 -- Val Loss: 0.03176 --Test Loss: 0.02213 -- Val Accuracy: 0.9955 Test Accuracy: 0.9954

Results from the original code
Train Loss: 0.00119 -- Val Loss: 0.02077 --Test Loss: 0.01888 -- Val Accuracy: 0.9950 Test Accuracy: 0.9954


alf/networks/s5/utils.py
```diff
@@ -0,0 +1,311 @@
...
+
+    """
+    # Make -HiPPO
+    hippo = make_HiPPO(N)
+
+    # Add in a rank 1 term. Makes it Normal.
+    P = np.sqrt(np.arange(N) + 0.5)
+
+    # HiPPO also specifies the B matrix
+    B = np.sqrt(2 * np.arange(N) + 1.0)
```
breakds, line 51: Seems `B = np.sqrt(2) * P`

alf/networks/s5/s5.py
```diff
@@ -0,0 +1,421 @@
...
+
+    Note: this implementation corresponds to s5.ssm.S5SSM with the following settings:
+
+        C_init='lecun_normal'
+        bidirectional=False
+        conj_sym=True
+        clip_eigs=True
+        discretization='zoh'
+
+    Args:
```
breakds, line 140: missing `num_blocks` in `Args:`.

alf/networks/s5/s5.py
```diff
@@ -0,0 +1,421 @@
...
+                 dt_max: float = 0.1,
+                 step_rescale: float = 1.0,
+                 name="S5"):
+        super().__init__(
+            input_tensor_spec=alf.TensorSpec((data_dim, )),
+            state_spec=alf.TensorSpec((state_dim, )),
+            name=name)
+        assert state_dim % 2 == 0
+        assert state_dim // 2 % num_blocks == 0
+        cstate_dim = state_dim // 2
```
breakds, line 167: Would be helpful to explain that `c` here means complex.

alf/networks/s5/s5.py
```diff
@@ -0,0 +1,421 @@
...
+    Corresponds to s5.layers.SequenceLayer
+
+    Note: for sequence input, the input is assumed to be [T, B, C]
+
+    """
+
+    def __init__(self,
+                 ssm_ctor: Callable[..., nn.Module],
+                 dropout: float,
+                 activation: str = "half_glu2",
```
breakds, line 337: Need to explain the activation here, e.g. what are the possible options and their effects.

alf/networks/s5/s5.py
```diff
@@ -0,0 +1,421 @@
...
+    """
+    Args:
+        x: real tensor with shape [..., N]
+    Returns:
+        complex tensor with shape [..., N // 2]
+    """
+    return torch.view_as_complex(x.reshape(*x.shape[:-1], x.shape[-1] // 2, 2))
+
+
+def _view_as_real(x):
```
breakds, line 58: Why do we want to view complex numbers as real, if all the operators can handle complex numbers (or maybe they cannot)?

alf/networks/s5/utils.py
```diff
@@ -0,0 +1,311 @@
...
+    # Store the final gradients for s and Lambda
+    tl.store(grad_s_ptr + col_offsets * 2, grad_s_real, mask=mask)
+    tl.store(grad_s_ptr + col_offsets * 2 + 1, -grad_s_imag, mask=mask)
+    tl.store(grad_lambda_ptr + col_offsets * 2, grad_lambda_real, mask=mask)
+    tl.store(
+        grad_lambda_ptr + col_offsets * 2 + 1, -grad_lambda_imag, mask=mask)
+
+
+class _ssm_forward(torch.autograd.Function):
+    BLOCK_SIZE = 128
```
breakds: How is the block size picked?

alf/networks/s5/utils.py
```diff
@@ -0,0 +1,311 @@
...
+        grid = lambda meta: (triton.cdiv(n, meta['BLOCK_SIZE']), )
+
+        if Lambda.dtype == torch.complex64:
+            diag_ssm_forward_kernel_complex[grid](torch.view_as_real(s),
+                                                  torch.view_as_real(x),
+                                                  torch.view_as_real(y),
+                                                  torch.view_as_real(Lambda),
+                                                  length, batch_size, dim,
+                                                  _ssm_forward.BLOCK_SIZE)
+        else:
```
breakds: Slightly safer to have `elif Lambda.dtype.is_floating_point`, to avoid sending a `torch.complex128` here. And then with another `else` to report the error.

alf/networks/s5/utils.py
```diff
@@ -0,0 +1,311 @@
...
+
+    P = V.conj().T @ P
+    B_orig = B
+    B = V.conj().T @ B
+    return Lambda_real + 1j * Lambda_imag, P, B, V, B_orig
+
+
+@triton.jit
+def diag_ssm_forward_kernel(s_ptr, x_ptr, lambda_ptr, y_ptr, length,
+                            batch_size, dim, BLOCK_SIZE: tl.constexpr):
```
breakds: Suggesting adding some documentation here about the "shape" of the `_ptr`s, using `length`, `batch` and `dim`

alf/networks/s5/s5.py
```diff
@@ -0,0 +1,421 @@
...
+    Note: this implementation corresponds to s5.ssm.S5SSM with the following settings:
+
+        C_init='lecun_normal'
+        bidirectional=False
+        conj_sym=True
+        clip_eigs=True
+        discretization='zoh'
+
+    Args:
+        data_dim: dimension of input (H)
```
breakds: "dimension of input and output  (H)" to provide more information?

alf/networks/s5/s5.py
```diff
@@ -0,0 +1,421 @@
...
+        ds/dt = As + Bu
+        A = V \Lambda V^{-1}
+        ds/dt = V \Lambda V^{-1}s + Bu
+        dV^{-1}s/dt = \Lambda V^{-1}s + V^{-1}Bu
+        y = C x + D u
+        y = CV V^{-1}s + Du
+
+    where :math:`s` is the state, :math:`u` is the input, :math:`y` is the output,
+    :math:`A` is the state transition matrix, :math:`B` is the input matrix,
+    :math:`C` is the output matrix, :math:`D` is the feedforward matrix,
```
breakds, line 119: Maybe mention `D` her is also assumed to be diagonal?

alf/networks/s5/utils.py
```diff
@@ -0,0 +1,311 @@
...
+
+def diag_ssm_forward(s, x, Lambda):
+    r"""Diagonal SSM forward pass
+
+    Calculate :math:`y_t = Lambda * y_{t-1} + x_t` for t > 0
+    and :math:`y_0 = Lambda * s + x_0`
+
+    Args:
+        s (torch.Tensor): shape is [batch_size, state_dim]
+        x (torch.Tensor): shape is [length, batch_size, data_dim]
```
breakds: Here `x` comes from Bu, whose last dimension seems to be `state_dim`?

alf/networks/s5/utils.py
```diff
@@ -0,0 +1,311 @@
...
+
+    @staticmethod
+    def backward(ctx, grad_y):
+        s, y, Lambda = ctx.saved_tensors
+        length, batch_size, dim = y.shape
+        grad_y = grad_y.contiguous()
+        n = batch_size * dim
+        grad_s = torch.empty_like(s)
+        grad_x = torch.empty_like(grad_y)
+        grad_lambda = torch.empty_like(s)
```
breakds: Nice trick. I am not sure whether this is very commonly used - but I was assuming `grad_lambda`'s shape is `[dim]` instead of `[B, dim]` until I saw the `grad_lambda.sum` below and fully understood the trick.

Suggest documenting the intention of making this temporary `grad_lambda` the same shape as `s` here.

alf/networks/s5/utils.py
```diff
@@ -0,0 +1,311 @@
...
+
+
+@triton.jit
+def diag_ssm_backward_kernel_complex(
+        s_ptr, lambda_ptr, y_ptr, grad_s_ptr, grad_x_ptr, grad_lambda_ptr,
+        grad_y_ptr, length, batch_size, dim, BLOCK_SIZE: tl.constexpr):
+
+    # autograd for complex numbers calculates \partial f / \partial z^*
+    # so we need to take conjugate during the calculation.
+    # https://pytorch.org/docs/stable/notes/autograd.html#autograd-for-complex-numbers
```
breakds: Suggesting moving the comments here to right above line 225 for better readability-locality.

alf/networks/s5/utils.py
```diff
@@ -0,0 +1,311 @@
...
+    Lambda = tl.load(lambda_ptr + col_offsets % dim, mask=mask, other=0)
+    for t in range(length):
+        offsets = t * batch_size * dim + col_offsets
+        x = tl.load(x_ptr + offsets, mask=mask, other=0)
+        s = s * Lambda + x
+        tl.store(y_ptr + offsets, s, mask=mask)
+
+
+@triton.jit
+def diag_ssm_backward_kernel(s_ptr, lambda_ptr, y_ptr, grad_s_ptr, grad_x_ptr,
```
breakds: Mathematically working out the backward rules:

triton_diag_ssm_backward_deduction

which is consistent with the implementation here, assuming we are looking at a portion of vectors with `offsets` as the indices.

breakds (commented):
Finished first round of review and LGTM. A few readability comments.
Learned a lot from reading the code here.

alf/networks/s5/utils.py
emailweixu: You are right. The code is from the original S5 repo, which is from https://github.com/srush/annotated-s4/blob/main/s4/s4.py

alf/networks/s5/utils.py
emailweixu: Not only the line after 225, all the load/store of the imaginary part of a gradient are related to the comment.

alf/networks/s5/utils.py
emailweixu: It is chosen from a few different values: 64,128,256

alf/networks/s5/utils.py
emailweixu: done

alf/networks/s5/utils.py
emailweixu: comment added.

alf/networks/s5/utils.py
emailweixu: fixed

alf/networks/s5/s5.py
emailweixu: Sometimes, it's more efficient to treat them as real. For example, if we only want real the part of `A * x` where A and x are complex.

alf/networks/s5/s5.py
emailweixu: added

alf/networks/s5/s5.py
emailweixu: added

alf/networks/s5/s5.py
emailweixu: updated

alf/networks/s5/s5.py
emailweixu: added

alf/networks/s5/s5.py
emailweixu: added.

alf/networks/s5/utils.py
```diff
@@ -83,248 +81,263 @@ def make_DPLR_HiPPO(N):
...
+            return grad_s, grad_x, grad_lambda.sum(dim=0)
+
+    diag_ssm_forward_triton = _ssm_forward.apply
+
+except ImportError:
+    from alf.utils.common import warning_once
+
+    def diag_ssm_forward_triton(s, x, Lambda):
+        warning_once("Triton is not installed. Using slow diag_ssm_forward.")
+        return diag_ssm_forward_slow(s, x, Lambda)
```
le-horizon, line 340: This has the disadvantage of falsely pass unit test when slow and triton paths actually differ.
Maybe try something like this:
https://github.com/HorizonRobotics/alf/commit/2ac807c39a1ca88bf36a51386b9630619966d1fe

emailweixu: If triton is installed, it is always tested. If triton is not installed, then there is no way to test.

le-horizon: If say our CI docker image accidentally removes triton, then the unit tests will silently pass, instead of raise an error.

emailweixu: Our CI docker only runs in CPU.

le-horizon (dismissed):
LGTM.  Thanks for the changes!

breakds (approved):

PR #1574, opened 2023-12-07
fix a potential bug for batch_size_per_env setting of evaluator

runjerry:
Encountered this bug in hobot when configuring ``batch_size_per_env>1`` and turning on evluate.

### DIFF
--- a/alf/trainers/evaluator.py
+++ b/alf/trainers/evaluator.py
@@ -60,6 +60,7 @@ class Evaluator(object):
         if conf_file.endswith('.gin'):
             assert not self._async, "async_eval is not supported for gin_file"
         num_envs = config.num_eval_environments
+        batch_size_per_env = num_envs
         seed = config.random_seed
         if self._async:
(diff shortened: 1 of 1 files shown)


emailweixu:
> Encountered this bug in hobot when configuring `batch_size_per_env>1` and turning on evluate.

Can you describe the bug and how this fixes the bug?

Perhaps you can use create_envirionment.eval_batch_size_per_env without changing the code.

runjerry:
> > Encountered this bug in hobot when configuring `batch_size_per_env>1` and turning on evluate.
> 
> Can you describe the bug and how this fixes the bug?
> 
> Perhaps you can use create_envirionment.eval_batch_size_per_env without changing the code.

Oh yes, configuring ``eval_batch_size_per_env`` solves the problem. Will update common_conf in hobot. Thanks!

runjerry:
Can be solved by configuring ``create_environment.eval_batch_size_per_env``.

PR #1576, opened 2023-12-08
handle target entropy for SAC mixed action type

hnyu:
This PR further addresses the issue mentioned in #1557 , making sure that the target entropy is either a scalar or a pair of scalars, no matter how complex the action spec is. When action type is mixed, we need to compute the target entropy as a pair of values.

To enforce a fixed alpha=0, we can set init_log_alpha to a very negative value and its learning rate to 0.

### DIFF
--- a/alf/algorithms/sac_algorithm.py
+++ b/alf/algorithms/sac_algorithm.py
@@ -17,6 +17,7 @@ from absl import logging
 import numpy as np
 import functools
 from enum import Enum
+import math
 
 import torch
 import torch.nn as nn
@@ -281,8 +282,16 @@ class SacAlgorithm(OffPolicyAlgorithm):
                 "Only continuous/discrete action is supported for multidimensional reward"
             )
 
+        finite_alpha = math.isfinite(initial_log_alpha)
+
         def _init_log_alpha():
-            return nn.Parameter(torch.tensor(float(initial_log_alpha)))
+            if finite_alpha:
+                # Only when the initial log alpha is finite, we can train it with
+                # a finite loss. If we create a parameter for an infinite alpha and
+                # don't add its loss to the overall loss, then DDP might complain.
+                return nn.Parameter(torch.tensor(float(initial_log_alpha)))
+            else:
+                return torch.tensor(float(initial_log_alpha))
 
         if self._act_type == ActionType.Mixed:
             # separate alphas for discrete and continuous actions
@@ -327,13 +336,15 @@ class SacAlgorithm(OffPolicyAlgorithm):
             self.add_optimizer(actor_optimizer, [actor_network])
         if critic_optimizer is not None and critic_networks is not None:
             self.add_optimizer(critic_optimizer, [critic_networks])
-        if alpha_optimizer is not None:
-            self.add_optimizer(alpha_optimizer, nest.flatten(log_alpha))
+
+        if finite_alpha:
+            if alpha_optimizer is not None:
+                self.add_optimizer(alpha_optimizer, nest.flatten(log_alpha))
+            if self._act_type == ActionType.Mixed:
+                self._log_alpha_paralist = nn.ParameterList(
+                    nest.flatten(log_alpha))
 
         self._log_alpha = log_alpha
-        if self._act_type == ActionType.Mixed:
-            self._log_alpha_paralist = nn.ParameterList(
-                nest.flatten(log_alpha))
 
         if max_log_alpha is not None:
             self._max_log_alpha = torch.tensor(float(max_log_alpha))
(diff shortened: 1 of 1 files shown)


alf/algorithms/sac_algorithm.py
```diff
@@ -784,19 +798,13 @@ def _critic_train_step(self, inputs: TimeStep, state: SacCriticState,
...
-            alpha_loss = self._log_alpha * (
-                -neg_entropy - self._target_entropy).detach()
-        return alpha_loss
+        # ``log_pi`` should either be a scalar or a pair (mixed action case),
+        # so is ``self._target_entropy``
+        alpha_loss = nest.map_structure(
+            lambda la, lp, t: (la * (-lp - t).detach() if isinstance(
+                la, nn.Parameter) else torch.zeros_like(lp)), self._log_alpha,
+            log_pi, self._target_entropy)
+        return sum(nest.flatten(alpha_loss))
```
emailweixu: It seems to me this revert the bug fix in https://github.com/HorizonRobotics/alf/pull/1557/files

alf/algorithms/sac_algorithm.py
```diff
@@ -281,8 +282,16 @@ def __init__(self,
                 "Only continuous/discrete action is supported for multidimensional reward"
             )
 
+        finite_alpha = math.isfinite(initial_log_alpha)
```
Haichao-Zhang: Seems currently alpha has to be infinite to be non-trainable? But will there be cases that alpha is a plain finite number but we want to fix it unchanged?

alf/algorithms/sac_algorithm.py
hnyu: > It seems to me this revert the bug fix in https://github.com/HorizonRobotics/alf/pull/1557/files

Yeah this is intentional. I think that bug fix is unnecessary because log_pi is guaranteed to be either a scalar (non mixed action case) or a pair of scalars (mixed action case), according to our current code (no matter how complex the action spec is). The `map_structure()` in `alpha_train_step()` is only effective in the Mixed action case. We just need to make sure the target entropy is correctly computed in the Mixed action case, which is what this PR does.

See my comment https://github.com/HorizonRobotics/alf/pull/1557#issuecomment-1800203849

alf/algorithms/sac_algorithm.py
hnyu: > Seems currently alpha has to be infinite to be non-trainable? But will there be cases that alpha is a plain finite number but we want to fix it unchanged?

If alpha is a finite number and we don't want it to be trained, then according to my past experience we can set its learning rate as 0. This PR addresses the issue of an infinite loss computed when alpha is 0. If alpha is infinite, according to the math, it simply can't be trained.

emailweixu: How about simply giving a very negative initial_log_alpha such as -1000?

hnyu: > How about simply giving a very negative initial_log_alpha such as -1000?

But it can still be trained unless we set its learning rate to 0? I think allowing an exact of 0 will be more principled.

Haichao-Zhang: ``infinite_alpha`` is a bit mis-leading, it is actually negtative_infinite_log_alpha?  
Feels that this newly introduced logics makes the sac algorithm a bit more complicated while only handle one additional particular case.

hnyu: > `infinite_alpha` is a bit mis-leading, it is actually negtative_infinite_log_alpha? Feels that this newly introduced logics makes the sac algorithm a bit more complicated while only handle one additional particular case.

I don't have an `infinite_alpha` variable? The `finite_alpha` is just a boolean flag. 

I didn't mean to introduce any new logic. It is just the case that our old code didn't support the case of `alpha=0`. Because in this case, the loss will become -inf and it will pollute the entire training loss. A special handling has to be needed for that. Unless we agree that we don't want the case of `alpha=0`, then this special handling is necessary.  

The logic is, I introduce the change not because that I "intend" to fix the alpha, but because we have to discard the training loss and parameter when `init_alpha=0`. (no loss but keeping the parameter will cause the "unused para" warning).

emailweixu: > > How about simply giving a very negative initial_log_alpha such as -1000?
> 
> But it can still be trained unless we set its learning rate to 0? I think allowing an exact of 0 will be more principled.

assert exp(-1000) == 0

hnyu: > > > How about simply giving a very negative initial_log_alpha such as -1000?
> > 
> > 
> > But it can still be trained unless we set its learning rate to 0? I think allowing an exact of 0 will be more principled.
> 
> assert exp(-1000) == 0

Reverted the alpha parameter change.

emailweixu (approved):

PR #1577, opened 2023-12-09
representation learner inside SAC

emailweixu:
With this, a target model of the representation can be maintained and be used to calculate the target critics. This is the strategy used by many other algorithms involving TD learning, (e.g. CURL, MuZero, etc).

### DIFF
--- a/alf/algorithms/sac_algorithm.py
+++ b/alf/algorithms/sac_algorithm.py
@@ -815,7 +815,7 @@ class SacAlgorithm(OffPolicyAlgorithm):
                                *self._reward_spec.shape).long()
         return q_values.gather(2, action).squeeze(2)
 
-    def _critic_train_step(self, observation, target_observation, reward,
+    def _critic_train_step(self, observation, target_observation,
                            state: SacCriticState, rollout_info: SacInfo,
                            action, action_distribution):
         critics, critics_state = self._compute_critics(
@@ -847,7 +847,8 @@ class SacAlgorithm(OffPolicyAlgorithm):
             target_critics = torch.sum(
                 discrete_act_dist.probs * target_critics, dim=-1)
 
-        target_critic = target_critics.reshape(reward.shape)
+        target_critic = target_critics.reshape(target_critics.shape[0],
+                                               *self._reward_spec.shape)
 
         target_critic = target_critic.detach()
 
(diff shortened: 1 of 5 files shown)


Haichao-Zhang (dismissed):
Thanks for the change! The proposed change properly handles the target version of the representation part, which is good in the presence of a representation module. 

I'm thinking of whether it is possible to achieve this goal within the ``Agent`` and/or ``ReprLearningAlgorithm`` class instead of ``SacAlgorithm``, since in that way, we can still reuse our previous modular design in terms of the functionality of each module in Agent. This could also make it easier to apply this Repr learning algorithm to other RL algorithm beyond SAC.

alf/algorithms/sac_algorithm.py
```diff
@@ -270,6 +281,23 @@ def __init__(self,
...
+                observation_spec=observation_spec,
+                action_spec=action_spec,
+                reward_spec=reward_spec,
+                config=config)
+            target_repr_alg = repr_alg_ctor(
+                observation_spec=observation_spec,
+                action_spec=action_spec,
+                reward_spec=reward_spec,
+                config=config)
+            observation_spec = repr_alg.output_spec
```
hnyu, line 298: A repr alg doesn't necessarily contain a property called "output_spec"? Use assertion first will be better.

alf/algorithms/sac_algorithm.py
```diff
@@ -773,7 +847,8 @@ def _critic_train_step(self, inputs: TimeStep, state: SacCriticState,
             target_critics = torch.sum(
                 discrete_act_dist.probs * target_critics, dim=-1)
 
-        target_critic = target_critics.reshape(inputs.reward.shape)
+        target_critic = target_critics.reshape(target_critics.shape[0],
+                                               *self._reward_spec.shape)
```
hnyu, line 853: Why is this change necessary?

alf/algorithms/sac_algorithm.py
emailweixu: added

alf/algorithms/sac_algorithm.py
emailweixu: This is because inputs.reward is no longer available

hnyu (approved):

PR #1578, opened 2023-12-11
Support different learning rate scheduler for param group

emailweixu:
This is used for S5 training.

### DIFF
--- a/alf/optimizers/optimizers.py
+++ b/alf/optimizers/optimizers.py
@@ -168,6 +168,7 @@ def wrap_optimizer(cls):
             if isinstance(lr, Callable):
                 self._lr_scheduler = lr
                 kwargs["lr"] = float(lr())
+        self._lr_schedulers = []
 
         super(NewCls, self).__init__([{'params': []}], **kwargs)
         if gradient_clipping is not None:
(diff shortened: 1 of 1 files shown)


alf/optimizers/optimizers.py
```diff
@@ -198,8 +199,14 @@ def step(self, closure=None):
         """
         if self._lr_scheduler is not None:
             lr = float(self._lr_scheduler())
-            for param_group in self.param_groups:
-                param_group['lr'] = lr
+            for i, (lr_scheduler, param_group) in enumerate(
+                    zip(self._lr_schedulers, self.param_groups)):
+                if lr_scheduler is not None:
+                    param_group['lr'] = lr_scheduler()
```
breakds, line 205: If my understanding is correct, in this particular case `param_group['lr']` has already been assigned to this scheduler below at line 340 when the group is added?

emailweixu: 340 is only executed once when the param_group is added.
Here we need to update 'lr' according to lr_schedule for every call to step().

breakds: Ah I see. This is calling the `scheduler` to acquire the current learning rate. Sorry about the misunderstanding.

breakds (approved):

PR #1579, opened 2023-12-12
render_heatmap_fast

emailweixu:
Different from the existing render_heatmap(), this function does not use matplotlib and is much faster and does not affect the video recording speed.

The following is an example of rendered multiple heatmap (ignore those read ellipses and black dots.)

image

### DIFF
--- a/alf/summary/render.py
+++ b/alf/summary/render.py
@@ -34,6 +34,7 @@ except ImportError:
 
 import cv2
 
+import torch
 import torch.distributions as td
 
 import alf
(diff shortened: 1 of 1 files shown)


hnyu:
Not quite sure about the strip on the left. Is that the value bar?

emailweixu:
> Not quite sure about the strip on the left. Is that the value bar?

It's the color bar representing the color of different value (i.e. different position in the bar).

hnyu:
> > Not quite sure about the strip on the left. Is that the value bar?
> 
> It's the color bar representing the color of different value (i.e. different position in the bar).

The color bar should has a gradually changing value range? Somehow the bars in the figure have a sudden change from white to black, which is confusing.

emailweixu:
> > > Not quite sure about the strip on the left. Is that the value bar?
> > 
> > 
> > It's the color bar representing the color of different value (i.e. different position in the bar).
> 
> The color bar should has a gradually changing value range? Somehow the bars in the figure have a sudden change from white to black, which is confusing.

If you look carefully, you can notice the change is not abrupt, though it is quite fast. This is because typically, the range of the value in the heatmap at one instant is quite small compared to the predefined range.

hnyu (approved):

PR #1580, opened 2023-12-13
fix a bug of target repr in SAC

hnyu:


alf/algorithms/sac_algorithm.py
```diff
@@ -884,7 +886,7 @@ def train_step(self, inputs: TimeStep, state: SacState,
             # usage can be reduced because its computation graph will not be kept.
             with torch.no_grad():
                 tgt_repr_step = self._target_repr_alg.predict_step(
-                    inputs, rollout_info.repr)
+                    inputs, state.target_repr)
```
emailweixu, line 889: I suspect "rollout_info.repr" is from copilot.

emailweixu (approved):

PR #1582, opened 2023-12-17
Better backward compatibility for SAC

Haichao-Zhang:
Better backward compatibility for SAC. 
For example, when using ``compare_training.py`` in Hobot to compare with a commit before introducing repr changes in SAC (with latest alf). 

Also this makes sub-class of SAC algorithm does not need to explicitly handle the repr field in SACLossInfo, if repr is not relevant to them.


le-horizon (approved):
LG, thanks for the fix.

PR #1590, opened 2024-01-09
Fix a bug in sac initial uniform sampling 

runjerry:
e.g. ``observation`` has shape ``[32, 5]`` and ``observation _spec`` has shape ``[5, ]``

then ``get_outer_rank(observation, observation_spec)`` will return the extra rank, i.e. the number of extra dims, which is ``1``, rather than the size of the extra dims, which should be ``[32]``.

### DIFF
--- a/alf/algorithms/oac_algorithm.py
+++ b/alf/algorithms/oac_algorithm.py
@@ -152,14 +152,15 @@ class OacAlgorithm(SacAlgorithm):
             return mu + shift
 
         if (rollout and not self._training_started):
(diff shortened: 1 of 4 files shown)


Haichao-Zhang:
Could you provide more details on when there will be an issue and provide unit test on these cases?

runjerry:
> Could you provide more details on when there will be an issue and provide unit test on these cases?

Just updated the PR description and revised SAC and OAC unit tests to reflect the previous bug. Thanks!

Haichao-Zhang (approved):

PR #1591, opened 2024-01-10
Addressing primacy bias via periodic model and parameter resetting

Haichao-Zhang:
Implementation of the method described in:  "[The Primacy Bias in Deep Reinforcement Learning](https://arxiv.org/abs/2205.07802)".


alf/algorithms/sac_algorithm.py
```diff
@@ -447,6 +450,15 @@ def _filter(x):
             tau=target_update_tau,
             period=target_update_period)
 
+        self._periodic_reset = common.PeriodicReset(
+            models=_filter([
+                self._actor_network, self._critic_networks,
+                self._target_critic_networks, repr_alg, target_repr_alg,
+                self._log_alpha
+            ]),
+            post_processings=[self._update_target],
```
runjerry: Maybe we can skip the ``TargetUpdater`` here given that ``self._target_critic_networks`` are also reset?

alf/utils/common.py
```diff
@@ -325,6 +325,76 @@ def forward(self):
...
+            for i, m in enumerate(self._models):
+                if isinstance(m, alf.networks.Network):
+                    self._copy_model_or_parameter(m.copy(), m)
+                elif isinstance(m, torch.nn.Parameter):
+                    self._copy_model_or_parameter(
+                        self._init_param_values[id(m)], m)
+            for c in self._post_processings:
+                c()
+
+        if self._counter >= period:
```
runjerry: seems that this line can be removed?

Haichao-Zhang: > seems that this line can be removed?

Not quite sure about your question. This line is checking if the condition for resetting is met.

alf/algorithms/sac_algorithm.py
Haichao-Zhang: > Maybe we can skip the `TargetUpdater` here given that `self._target_critic_networks` are also reset?

We need to keep the post_preocessing since we need to re-copy the param values from critic network to target whenever a reset is done.

alf/utils/common.py
runjerry: Seems that the same condition is checked at line 384?

alf/algorithms/sac_algorithm.py
runjerry: Yes, I think the updated version is faithful to the original paper.

alf/utils/common.py
Haichao-Zhang: Oh, I see. Removed. Thanks!

Haichao-Zhang:
Update: now added support to self-defined network reset behavior according to suggestions from @emailweixu (https://github.com/HorizonRobotics/alf/pull/1591/commits/60d80819ef5e74717012d8c944cdd196f746f11b) 
Now it should be compatible with @runjerry 's network reset code and more flexible :)

runjerry (approved):
Looks good to me, thanks!

alf/utils/common.py
```diff
@@ -325,6 +325,85 @@ def forward(self):
...
+        else:
+            for ws, wt in zip(s.parameters(), t.parameters()):
+                wt.data.copy_(ws)
+            if self._reset_buffers:
+                for ws, wt in zip(s.buffers(), t.buffers()):
+                    wt.copy_(ws)
+
+    def forward(self):
+        self._counter += 1
+        period = self._period()
```
hnyu, line 388: The scheduler uses global training iter as the counter, so `period` here is scheduled according to how many training iterations. But `self._counter` is maintained according to num of updates. Is this an expected behavior?

Haichao-Zhang: > The scheduler uses global training iter as the counter, so `period` here is scheduled according to how many training iterations. But `self._counter` is maintained according to num of updates. If one iter contains multiple updates, then the comparison below `self._counter >= period` is not correct.

This is a known behavior. Already discussed this with @runjerry weeks ago when this PR was submitted. The decision was to keep it the same as target update.

Haichao-Zhang: > The scheduler uses global training iter as the counter, so `period` here is scheduled according to how many training iterations. But `self._counter` is maintained according to num of updates. Is this an expected behavior?

Adjusted and added comments: https://github.com/HorizonRobotics/alf/pull/1603

PR #1593, opened 2024-01-11
DSAC algorithm

runjerry:
Add the Distribution SAC algorithm and corresponding config. Tested on cheetah-run and hopper-hop of Deepmind Control.


runjerry:
This PR is ready for review. I am sorry that it ends up a long one.

alf/algorithms/dsac_algorithm.py
```diff
@@ -0,0 +1,462 @@
...
+            self._critic_networks,
+            inputs.observation[:-bs],
+            rollout_info.action[:-bs],
+            tau_info.tau_hat[:-bs],
+            state.critics,
+            replica_min=False)
+        # Extend with zeros so that the tensors can be correctly reshape to
+        # [T, B, ...]
+        zeros = torch.zeros_like(critics[:bs])
+        critics = torch.cat((critics, zeros), dim=0)
```
Haichao-Zhang, line 287: can use ``tensor_utils.tensor_extend_zero``?

alf/algorithms/dsac_algorithm.py
```diff
@@ -0,0 +1,462 @@
...
+
+        with torch.no_grad():
+            target_critics, target_critics_state, _ = self._compute_critics(
+                self._target_critic_networks, inputs.observation[bs:],
+                action[bs:], tau_info.next_tau_hat[bs:], state.target_critics,
+                tau_info.next_delta_tau[bs:])
+            # Prepend with zeros so that the tensors can be correctly reshape to
+            # [T, B, ...]
+            zeros = torch.zeros_like(target_critics[:bs])
+            target_critics = torch.cat((zeros, target_critics), dim=0)
```
Haichao-Zhang, line 301: can use ``tensor_utils.tensor_extend_zero``?

alf/algorithms/dsac_algorithm.py
```diff
@@ -0,0 +1,462 @@
...
+
+        if self._debug_summaries and alf.summary.should_record_summaries():
+            with torch.no_grad():
+                critics_eval = self._compute_critics(
+                    self._critic_networks,
+                    inputs.observation[:-bs],
+                    rollout_info.action[:-bs],
+                    self._eval_tau,
+                    state.critics,
+                    replica_min=False)[0]
```
Haichao-Zhang, line 317: why do we need to re-eval the critic? Is it possible to re-use the ``critics`` computed above?

alf/algorithms/dsac_algorithm.py
```diff
@@ -0,0 +1,462 @@
...
+        assert self._act_type == ActionType.Continuous, (
+            "Only continuous action space is supported for dsac algorithm.")
+        if self.has_multidim_reward():
+            assert not min_critic_by_critic_mean, (
+                "min_critic_by_critic_mean is not supported for multi-dim rewards."
+            )
+
+        self._min_critic_by_critic_mean = min_critic_by_critic_mean
+        if alpha_optimizer is None:
+            self._epistemic_alpha = True
```
Haichao-Zhang, line 163: Suggest to add comments on the logic of ``alpha``, e.g. when ``epistemic_alpha`` will be used and what does it represent

alf/networks/encoding_networks.py
```diff
@@ -721,9 +721,17 @@ def __init__(self,
...
+        #     "The input shape {} should be like (N,)!".format(spec.shape)
+        # input_size = spec.shape[0]
+        if spec.ndim == 1:
+            input_size = spec.shape[0]
+        elif spec.ndim == 2:
+            input_size = spec.shape[-1]
+        else:
+            raise ValueError(
+                f"The input shape {spec.shape} should be like (N, )"
+                "or (N, D, ).")
```
Haichao-Zhang: We might need to document this part on the expected behaviors after the change, including the ``spec.ndim == 2`` case

Haichao-Zhang:
Another suggestion is to include return curves for some benchmark tasks for reference.

alf/algorithms/dsac_algorithm.py
runjerry: It seems that ``tensor_utils.tensor_extend_zero`` only extend one slice of zeros along ``dim``? I to extend a slice of size [bs] here.

alf/algorithms/dsac_algorithm.py
runjerry: Same concern as above.

alf/algorithms/dsac_algorithm.py
runjerry: The idea is to use a "fixed" delta_tau for evaluation rather than the randomly sampled one as used to compute ``critics`` for more debugging purpose. This does introduce extra computational cost, but only when ``debug_summaries=True``. I can summarize the computed critics instead if that makes better sense.

alf/algorithms/dsac_algorithm.py
runjerry: Added ,thanks.

alf/networks/encoding_networks.py
runjerry: Added.

runjerry:
> Another suggestion is to include return curves for some benchmark tasks for reference.

Good point, added.

Haichao-Zhang (dismissed):

runjerry:
Sorry for the further update. Just realized that the previous commits cannot work with nested observations and actions, and as a result, cannot be applied to hobot training.

Haichao-Zhang (approved):

PR #1594, opened 2024-01-12
Fix some sync_progress issues

Haichao-Zhang:
Fixed some sync_progress issues and updated unit tests.

### DIFF
--- a/alf/environments/alf_environment.py
+++ b/alf/environments/alf_environment.py
@@ -222,8 +222,11 @@ class AlfEnvironment(object):
         """A placeholder interface to make it consistent with some derived classes
         such as ``ParallelAlfEnvironment`` and ``FastParallelAlfEnvironment``.
         """
-        pass
-    
+        raise RuntimeError(
+            'This is a not a parallel environment. '
+            'Change to use parallel environment or set config.sync_progress_to_envs=False'
+        )
+
     def close(self):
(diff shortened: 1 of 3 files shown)


alf/environments/alf_environment.py
```diff
@@ -218,6 +218,12 @@ def step(self, action):
         self._current_time_step = self._step(action)
         return self._current_time_step
 
+    def sync_progress(self):
```
Haichao-Zhang, line 221: Adding this since in the call to ``env.sync_progress()`` in ``policy_trainer.py``, ``env`` could be a non-parallel env. Adding this interface to unify the interface without additional codes to do conditional call based on the typr of env.

alf/environments/alf_environment.py
```diff
@@ -218,6 +218,12 @@ def step(self, action):
...
+        """A placeholder interface to make it consistent with some derived classes
+        such as ``ParallelAlfEnvironment`` and ``FastParallelAlfEnvironment``.
+        """
+        pass
```
emailweixu: suggest to raise error so that user knows something is wrong.

Haichao-Zhang: > suggest to raise error so that user knows something is wrong.

Good point. Done.

emailweixu (approved):

PR #1596, opened 2024-01-17
allow specifying interpolation type when resizing images

hnyu:
Sometimes for resizing an image of integer type, we need to use nearest interpolation.


alf/utils/tensor_utils.py
```diff
@@ -423,7 +423,7 @@ def append_coordinate(im: torch.Tensor):
     assert len(im.shape) == 4, "Image must have a shape of [B,C,H,W]!"
     y = torch.arange(-1., 1., step=2. / im.shape[-2])
     x = torch.arange(-1., 1., step=2. / im.shape[-1])
-    yy, xx = torch.meshgrid(y, x)
+    yy, xx = torch.meshgrid(y, x, indexing='ij')
```
hnyu, line 426: This is for compatibility with higher versions of pytorch, where the default indexing is no longer 'ij'.

emailweixu (approved):

PR #1597, opened 2024-01-17
revert get_all_parameters() to handle exception

runjerry:
When ThreadEnvironment is used, some parameters from hobot environment could raise exception, say ``quaternion.base.data``.

### DIFF
--- a/alf/utils/common.py
+++ b/alf/utils/common.py
@@ -1449,7 +1449,14 @@ def get_all_parameters(obj):
             # when called in a wrong context (e.g. Algorithm.experience_spec)
             if isinstance(getattr(type(obj), name, None), property):
                 continue
-            attr = getattr(obj, name, None)
+            attr = None
+            try:
+                attr = getattr(obj, name)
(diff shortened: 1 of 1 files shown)


alf/utils/common.py
```diff
@@ -1449,7 +1449,14 @@ def get_all_parameters(obj):
             # when called in a wrong context (e.g. Algorithm.experience_spec)
             if isinstance(getattr(type(obj), name, None), property):
                 continue
-            attr = getattr(obj, name, None)
+            attr = None
+            try:
+                attr = getattr(obj, name)
+            except:
+                # some attrbutes are property function, which may raise exception
+                # when called in a wrong context (e.g. Algorithm.experience_spec)
```
emailweixu: Property function  is handled by line 1450.
This is for handling the case that the attribute cannot be accessed.

runjerry: Oh yes, updated.

emailweixu (approved):

PR #1598, opened 2024-01-17
relax procgen version

hnyu:
python 3.10 can only find procgen>=10.0.6


hnyu:
@breakds  ping. Does higher version of procgen have an issue?

breakds (approved):

PR #1601, opened 2024-01-19
DOAC algorithm and DSAC update

runjerry:
- Updated oac algorithm to support nested actions
- Add doac algorithm and unit tests
- Update dsac algorithm to work with temporally dependent train_step


runjerry:
By running more random seeds, it seems to me that the performance of DSAC in this PR is on par with that of the main branch version. So this PR is ready for review.

alf/algorithms/doac_algorithm.py
```diff
@@ -0,0 +1,263 @@
...
+        # assert isinstance(action_dist, td.TransformedDistribution), (
+        #     "Squashed distribution is expected from actor_network.")
+        # assert isinstance(
+        #     action_dist.base_dist, dist_utils.DiagMultivariateNormal
+        # ), ("the base distribution should be diagonal multivariate normal.")
+        # normal_dist = action_dist.base_dist
+        # unsquashed_mean = normal_dist.mean
+        # unsquashed_std = normal_dist.stddev
+        # unsquashed_var = normal_dist.variance
+        # new_state = new_state._replace(actor_network=actor_network_state)
```
Haichao-Zhang: remove these lines?

alf/algorithms/doac_algorithm.py
```diff
@@ -0,0 +1,263 @@
...
+        action_dist_type = nest.map_structure(
+            lambda dist: isinstance(dist, td.TransformedDistribution),
+            action_dist)
+        action_base_dist_type = nest.map_structure(
+            lambda dist: isinstance(dist.base_dist, dist_utils.
+                                    DiagMultivariateNormal), action_dist)
+        action_dist_type = nest.flatten(action_dist_type)
+        action_base_dist_type = nest.flatten(action_base_dist_type)
+        assert all(nest.flatten(action_dist_type)), (
+            "Squashed distribution is expected from actor_network.")
```
Haichao-Zhang: If line 181-182 is to ensure the distributions are squashed, the current code block does not seem to be able to ensure this? Currently it only checks if the base_dist is DiagMultiNormal, that does not have any guarantee on the outer distribution containing the base_dist?

alf/algorithms/doac_algorithm.py
```diff
@@ -0,0 +1,263 @@
...
+                return transformed_action
+
+            def mean_shift_fn(mu, dqda, sigma):
+                if self._dqda_clipping:
+                    dqda = torch.clamp(dqda, -self._dqda_clipping,
+                                       self._dqda_clipping)
+                norm = torch.sqrt(torch.sum(torch.mul(dqda * dqda,
+                                                      sigma))) + 1e-6
+                shift = self._explore_delta * torch.mul(sigma, dqda) / norm
+                return mu + shift
```
Haichao-Zhang, line 201: duplicated with line 186-192

alf/algorithms/oac_algorithm.py
```diff
@@ -163,12 +159,40 @@ def mean_shift_fn(mu, dqda, sigma):
...
+            unsquashed_mean = nest.map_structure(lambda dist: dist.mean,
+                                                 normal_dist)
+            unsquashed_std = nest.map_structure(lambda dist: dist.stddev,
+                                                normal_dist)
+            unsquashed_var = nest.map_structure(lambda dist: dist.variance,
+                                                normal_dist)
+
+            def _prepare_critic_action(dist_mean):
+                critic_action = dist_mean.detach().clone()
+                critic_action.requires_grad = True
```
Haichao-Zhang: Curious about what are the cases the ``critic_action.requires_grad`` is False and we have to manually turn it on?

alf/algorithms/doac_algorithm.py
```diff
@@ -0,0 +1,263 @@
...
+            norm = torch.sqrt(torch.sum(torch.mul(dqda * dqda, sigma))) + 1e-6
+            shift = self._explore_delta * torch.mul(sigma, dqda) / norm
+            return mu + shift
+
+        if (rollout and not self._training_started):
+            # get batch size with ``get_outer_rank`` and ``get_nest_shape``
+            # since the observation can be a nest in the general case
+            outer_rank = nest_utils.get_outer_rank(observation,
+                                                   self._observation_spec)
+            outer_dims = alf.nest.get_nest_shape(observation)[:outer_rank]
```
Haichao-Zhang, line 165: remembered that these appeared in other algorithms as well. maybe we can extract them as a common utility function so that it will be easier to be used in different places?

alf/algorithms/dsac_algorithm.py
```diff
@@ -266,39 +263,22 @@ def _compute_critics(self,
...
-                                           rollout_info.action)
+        # [B * T, n_quantiles] or [B, n_quantiles]
         critics, critics_state, _ = self._compute_critics(
             self._critic_networks,
-            critic_observation,
-            critic_action,
-            tau_info.tau_hat[:-bs],
+            inputs.observation,
+            rollout_info.action,
+            tau_info.tau_hat,
```
Haichao-Zhang, line 273: what's the reason for this change?

alf/algorithms/dsac_algorithm.py
```diff
@@ -266,39 +263,22 @@ def _compute_critics(self,
...
-                self._target_critic_networks, target_critic_observation,
-                target_critic_action, tau_info.next_tau_hat[bs:],
-                state.target_critics, tau_info.next_delta_tau[bs:])
-            # Prepend with zeros so that the tensors can be correctly reshape to
-            # [T, B, ...]
-            zeros = torch.zeros_like(target_critics[:bs])
-            target_critics = torch.cat((zeros, target_critics), dim=0)
+                self._target_critic_networks, inputs.observation, action,
+                tau_info.next_tau_hat, state.target_critics,
+                tau_info.next_delta_tau)
```
Haichao-Zhang, line 281: what's the reason for changes?

alf/algorithms/doac_algorithm.py
runjerry: My bad, removed.

alf/algorithms/doac_algorithm.py
runjerry: The ``action_dist_type`` extracted from line 173-175 was checked for ``TransformedDistribution``?

alf/algorithms/doac_algorithm.py
runjerry: Yes, similar codes are also used in ``oac_algorithm``. The mean_shift_fn depends on some member variables, so I leave it there. I find two other functions that could be extracted for both oac and doac.

alf/algorithms/doac_algorithm.py
runjerry: Good catch, remvoed.

alf/algorithms/dsac_algorithm.py
runjerry: Previously, both ``inputs.observation`` and ``rollout_info.action`` are chunked to extract the first [B * (T-1), ...] tensors to forward the critics, instead of forwarding the full tensors of shape [B * T, ...] as done by SAC and other alf algoirthms, to save some computation time, since they will be chunked either way during TD loss computation. However, this will not work with the case when ``temporally_independent_train_step=False``. So I just changed back to the default case.

alf/algorithms/dsac_algorithm.py
runjerry: Similar as above, previously, both ``inputs.observation`` and ``action`` are chunked to extract the last [B * (T-1), ...] tensors to forward the critics, instead of forwarding the full tensors of shape [B * T, ...]. Now change it back.

alf/algorithms/oac_algorithm.py
runjerry: There might be a better way to make a clone for such purpose? So that I will not have to ``detach`` first and then turn ``requires_grad`` on after the ``clone``.

Haichao-Zhang: This seems to be similar to the case in sac for dqda? If yes, then no need to detach then turn on require_grad? Just directly call transformed_action directly on ``critic_action``?

alf/algorithms/doac_algorithm.py
Haichao-Zhang: You are right.

alf/algorithms/oac_algorithm.py
```diff
@@ -163,12 +159,40 @@ def mean_shift_fn(mu, dqda, sigma):
...
+            def mean_shift_fn(mu, dqda, sigma):
+                if self._dqda_clipping:
+                    dqda = torch.clamp(dqda, -self._dqda_clipping,
+                                       self._dqda_clipping)
+                norm = torch.sqrt(torch.sum(torch.mul(dqda * dqda,
+                                                      sigma))) + 1e-6
+                shift = self._explore_delta * torch.mul(sigma, dqda) / norm
+                return mu + shift
+
+            critic_action = nest.map_structure(_prepare_critic_action,
```
Haichao-Zhang: Is it possible to implement a ``MeanShiftDistributionTransformation``, which will include ``mean_shift_fn`` inside? In that way, the code can be simplified since we don't need to manually do the process of mean_shift -> distribution transform?

runjerry: Updated according to our discussion. Many thanks for the constructive feedback!

Haichao-Zhang (approved):
LGTM

PR #1602, opened 2024-01-20
add assert to avoid silently converting num_para_envs to 1 when nonparallel is true

le-horizon:
When nonparallel and num_envs > 1, the two parameters conflict with each other.
Silently converting can cause unexpected behavior.  It's better to throw error and notify the user.

The caveat is that all callers have to specify num_parallel_env=1 whenever nonparallel=True is set.

### DIFF
--- a/alf/environments/utils.py
+++ b/alf/environments/utils.py
@@ -185,7 +185,7 @@ def create_environment(env_name='CartPole-v0',
         _get_wrapped_fn(env_load_fn), 'no_thread_env', False)
 
     if nonparallel:
-        num_parallel_environments = 1
+        assert num_parallel_environments == 1, "nonparallel is True"
         batch_size_per_env = 1
 
     if batch_size_per_env is None:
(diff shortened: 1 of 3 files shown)


emailweixu (approved):

alf/environments/utils.py
```diff
@@ -185,7 +185,7 @@ def create_environment(env_name='CartPole-v0',
         _get_wrapped_fn(env_load_fn), 'no_thread_env', False)
 
     if nonparallel:
-        num_parallel_environments = 1
+        assert num_parallel_environments == 1, "nonparallel is True"
```
emailweixu, line 189: update the doc-string

emailweixu (changes requested):

alf/environments/utils.py
le-horizon: Done, and added extra assert on batch_size_per_env.  PTAL.

emailweixu: Where is the extra assert on batch_size_per_env?

le-horizon: Ah, pushed to the wrong remote..  Now it's Ok.

emailweixu (approved):

PR #1604, opened 2024-01-23
Support capacity scheduling with maskout value specification

Haichao-Zhang:
This PR adds supports to maskout value specification in capacity scheduling.

### DIFF
--- a/alf/optimizers/optimizers.py
+++ b/alf/optimizers/optimizers.py
@@ -250,7 +250,7 @@ def wrap_optimizer(cls):
                         state['rng_state'] = common.get_torch_rng_state()
                     else:
                         rng_state = state['rng_state']
-                        common.set_torch_rng_state(rng_state)
+                        common.set_torch_rng_state(*rng_state)
 
                     # generate capacity mask using the same random number generator state
                     n = p.numel()
@@ -267,7 +267,7 @@ def wrap_optimizer(cls):
                         p.data[mask] = self._masked_out_value
 
             # recover the original random number generator state
-            common.set_torch_rng_state(org_rng_state)
+            common.set_torch_rng_state(*org_rng_state)
 
     @common.add_method(NewCls)
     def step(self, closure=None):
--- a/alf/optimizers/optimizers.py
+++ b/alf/optimizers/optimizers.py
@@ -15,7 +15,7 @@
 import copy
 import numpy as np
 import torch
-from typing import Callable, Union
+from typing import Callable, Dict, Union
 
 import alf
 from alf.utils import common
@@ -106,6 +106,7 @@ def wrap_optimizer(cls):
                  repulsive_weight=1.,
                  capacity_ratio: Union[float, Scheduler] = 1.0,
                  min_capacity: int = 8192,
+                 masked_out_value: Union[float, None] = None,
                  name=None,
                  **kwargs):
         """
@@ -150,9 +151,11 @@ def wrap_optimizer(cls):
                 elements are turned on for training. The remaining elements
                 are frozen. ``capacity_ratio`` can be a scheduler to control
                 the capacity over the training process.
-            min_capacity (int): For each parameter, at least so many elements
+            min_capacity: For each parameter, at least so many elements
                 are turned on for training.
-
+            masked_out_value: the value to be set for the masked out parameters, i.e.,
+                parameters whose mask value is True. If None, no operation will be applied.
+                Otherwise, set the parameter values as the specified value.
             name (str): the name displayed when summarizing the gradient norm. If
                 None, then a global name in the format of "class_name_i" will be
                 created, where "i" is the global optimizer id.
@@ -170,6 +173,11 @@ def wrap_optimizer(cls):
                 kwargs["lr"] = float(lr())
         self._lr_schedulers = []
 
+        capacity_ratio_scheduler = alf.utils.schedulers.as_scheduler(
+            capacity_ratio)
+
+        self._capacity_ratio = capacity_ratio_scheduler()
+
         super(NewCls, self).__init__([{'params': []}], **kwargs)
         if gradient_clipping is not None:
             self.defaults['gradient_clipping'] = gradient_clipping
@@ -177,9 +185,10 @@ def wrap_optimizer(cls):
         self._gradient_clipping = gradient_clipping
         self._clip_by_global_norm = clip_by_global_norm
         self._parvi = parvi
-        self._capacity_ratio = alf.utils.schedulers.as_scheduler(
-            capacity_ratio)
+        self._capacity_ratio_scheduler = capacity_ratio_scheduler
+
         self._min_capacity = min_capacity
+        self._masked_out_value = masked_out_value
         self._norms = {}  # norm of each parameter
         if parvi is not None:
             assert parvi in ['svgd', 'gfsf'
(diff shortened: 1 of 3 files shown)


alf/optimizers/optimizers.py
```diff
@@ -380,6 +428,10 @@ def add_param_group(self, param_group):
         else:
             super(NewCls, self).add_param_group(param_group)
 
+        capacity_ratio = self._capacity_ratio
```
emailweixu: capacity_ratio should be from scheduler.

alf/optimizers/optimizers.py
```diff
@@ -380,6 +428,10 @@ def add_param_group(self, param_group):
...
+        states = self._generate_capacity_mask(capacity_ratio)
+        self._apply_capacity_mask(capacity_ratio, states)
```
emailweixu: why are these additional self._apply_capacity_mask needed for parvi_step?

alf/optimizers/optimizers.py
```diff
@@ -192,6 +201,74 @@ def __init__(self,
...
+                    if 'rng_state' not in state:
+                        rng_state = torch.get_rng_state()
+                        s['rng_state'] = rng_state
+                    else:
+                        rng_state = state['rng_state']
+                        torch.set_rng_state(rng_state)
+                    n = p.numel()
+                    ratio = max(self._min_capacity / n, capacity_ratio)
+                    mask = torch.rand_like(p) >= ratio
+                    s['mask'] = mask
```
emailweixu: It seems that mask generation can be put into apply_mask, so that we don't need to save all the masks in state.

alf/optimizers/optimizers.py
```diff
@@ -170,16 +173,22 @@ def __init__(self,
                 kwargs["lr"] = float(lr())
         self._lr_schedulers = []
 
+        capacity_ratio_scheduler = alf.utils.schedulers.as_scheduler(
+            capacity_ratio)
+
+        self._capacity_ratio = capacity_ratio_scheduler()
```
emailweixu: why not use scheduler as the old code?

alf/optimizers/optimizers.py
Haichao-Zhang: Changed the name to initial_capacity_ratio. The reason is that we need to record initial capacity ratio in order to be used in ``add_param_group``, since  the capacity scheduler (``self._capacity_ratio``) is inaccessible in ``add_param_group``.

alf/optimizers/optimizers.py
Haichao-Zhang: Not sure if I understand your question correctly. This is in ``add_param_group``, not in ``parvi_step``. 
The reason to ``_apply_capacity_mask`` in ``add_param_group`` is to correctly handle the parameter values before any forward step.

alf/optimizers/optimizers.py
Haichao-Zhang: Changed as suggested

alf/optimizers/optimizers.py
Haichao-Zhang: Change to the original way of naming

alf/optimizers/optimizers.py
```diff
@@ -170,16 +173,23 @@ def __init__(self,
                 kwargs["lr"] = float(lr())
         self._lr_schedulers = []
 
+        capacity_ratio = alf.utils.schedulers.as_scheduler(capacity_ratio)
+
+        # record initial capacity ratio in order to be used in ``add_param_group``,
+        # within which the capacity scheduler (``self._capacity_ratio``) is inaccessible
+        self._initial_capacity_ratio = capacity_ratio()
```
emailweixu: why not assign self._capacity_ratio?

alf/optimizers/optimizers.py
```diff
@@ -192,6 +202,77 @@ def __init__(self,
...
+
+            for param_group in self.param_groups:
+                for p in param_group['params']:
+                    state = self.state[p]
+                    s = states[p]
+
+                    # get, save and set random number generator state
+                    if 'rng_state' not in state:
+                        rng_state = torch.get_rng_state()
+                        s['rng_state'] = rng_state
```
emailweixu: `state['rng_state'] = rng_state` and line 270-271 can be removed.

alf/optimizers/optimizers.py
```diff
@@ -192,6 +202,77 @@ def __init__(self,
...
+        states = {}
+        if capacity_ratio < 1:
+            for param_group in self.param_groups:
+                for p in param_group['params']:
+                    s = {}
+                    if self._masked_out_value is None:
+                        # only save previous param value if masked_out_value is unspecified
+                        s['old_param'] = p.data.clone()
+                    states[p] = s
+        return states
```
emailweixu: Given the current implementation, it's better to change this to `_clone_params()` and return a dict where the value is the cloned parameter.

alf/optimizers/optimizers.py
Haichao-Zhang: Changed as suggested

alf/optimizers/optimizers.py
Haichao-Zhang: Changed function name and other relevant places accordlingly.

alf/optimizers/optimizers.py
Haichao-Zhang: Changed. I actually tried this initially when implementing this PR this yesterday, but was unsuccessful. Probably due to other issues at that time.  And that's why I recorded the initial capacity ratio value in the original PR.
Bow this seems unnecessary.

alf/optimizers/optimizers.py
```diff
@@ -192,6 +198,79 @@ def __init__(self,
...
+        Returns:
+            an empty dictionary if ``self._masked_out_value`` is None; otherwise, return
+            a dictionary with the parameter as the key and the current parameter value as
+            the value of the dictionary.
+        """
+        param_values = {}
+        if capacity_ratio < 1:
+            for param_group in self.param_groups:
+                for p in param_group['params']:
+                    if self._masked_out_value is None:
```
emailweixu: combine with line 215

Haichao-Zhang: Yes, it should be combined. Done

Haichao-Zhang:
Update: now behavior is normal, when using masked_out_values=0: 
image


The abnormal behavior (very bad performance) I mentioned days ago leads to the discovery of the bug fixed in 4c4ac73f80365f3b868e2c83c16d13ba2e0b8ad3. 
The issue was that previously we only save and recover the torch cpu rng_state, therefore when running on GPU, the mask for the same parameter in each optimizer step is **different**.  Therefore, in the current main branch, the capacity scheduling actually leads to an effects of stopping gradient on a subset of randomly selected parameter elements per gradient step. 
In the case of using masked_out_value, the randomly selected parameter elements with be set to zero value. Because the selected subset is random each time even if the capacity ratio is the same, this will leads to a bigger change to the effective function represented by the network, resulting bad and strange performance.

Haichao-Zhang:
Update II: addressed comments in 14b29809a2973f0ab93d866b44a8433727689bb7.
1. paired the rng_state with device obtained from ``alf.get_default_device``, and also use the corresponding device when setting rng_state. By doing this, we can avoid introducing additional member attributes to optimizer.
2. added unit test. Tested locally on both cpu and gpu the test cases passed.

Haichao-Zhang:
Now saved a random number generator as member attribute 79f24f6f2fb98558d9eb06fe84215df0d18f0286

alf/optimizers/optimizers.py
```diff
@@ -192,6 +201,80 @@ def __init__(self,
...
+
+            for param_group in self.param_groups:
+                for p in param_group['params']:
+                    state = self.state[p]
+                    old_param_val = old_param_values.get(p, None)
+                    # get, save and set random number generator state
+                    if 'rng_state' not in state:
+                        # record random number generator state in ``self.state``
+                        state['rng_state'] = common.get_torch_rng_state(
+                            self._random_number_generator.device)
```
emailweixu: `self._random_number_generator.get_state()`

alf/optimizers/optimizers.py
```diff
@@ -192,6 +201,80 @@ def __init__(self,
...
+                    else:
+                        rng_state = state['rng_state']
+                        common.set_torch_rng_state(*rng_state)
+
```
emailweixu, line 258: `self._random_number_generator.set_state()`

I wonder how the unittest can pass.

Haichao-Zhang: > `self._random_number_generator.set_state()`
> 
> I wonder how the unittest can pass.

The reason was that the unittests are designed for the previous version before using rnd as member attributes. 
Updated the unittests and also in optimizers.

This issue only present in https://github.com/HorizonRobotics/alf/commit/79f24f6f2fb98558d9eb06fe84215df0d18f0286, previous results (e.g. the return curves were obtained before introducing this commit, and does not have this issue)

alf/optimizers/optimizers.py
Haichao-Zhang: Fixed

alf/optimizers/optimizers_test.py
```diff
@@ -252,6 +253,64 @@ def _test(i, sampled_predictive=False):
...
+
+                        # use str type as the key so that later map_structure can be used
+                        # for nested checking
+                        rand_matrix_for_capacity_mask[str(id(p))] = torch.rand(
+                            p.shape, generator=local_rng)
+            return rand_matrix_for_capacity_mask
+
+        rand_matrix0 = _generate_rand_matrix_for_capacity_mask(opt)
+
+        opt.zero_grad()
```
emailweixu, line 291: Instead of comparing with a mask generated outside of opt. It's better to infer the mask from the parameter before and after the backward() call. And the mask should not change with multiple backward() calls.

Haichao-Zhang: Good suggestion. Done

alf/optimizers/optimizers_test.py
```diff
@@ -21,7 +22,8 @@
 from alf.algorithms.particle_vi_algorithm import ParVIAlgorithm
 from alf.optimizers import Adam, AdamTF
 from alf.tensor_specs import TensorSpec
-from alf.utils import tensor_utils
+from alf.utils import common, tensor_utils
+from alf.utils.schedulers import LinearScheduler
```
emailweixu: not used

alf/optimizers/optimizers_test.py
```diff
@@ -252,6 +254,64 @@ def _test(i, sampled_predictive=False):
         self.assertLess(mean_err, 0.5)
         self.assertLess(cov_err, 0.5)
 
+    @parameterized.parameters(
+        dict(capacity_ratio=0.5, masked_out_value=None, opt_steps=3),
+        dict(capacity_ratio=0.5, masked_out_value=0, opt_steps=5))
+    def test_capacity_scheduling(self, capacity_ratio, masked_out_value,
+                                 opt_steps):
+        layer = torch.nn.Linear(5, 3)
```
emailweixu: use a large parameter and check the mask is approximately close to the desired capacity.

Haichao-Zhang: Added. A constant capacity is used to keep the test not too complicated.

emailweixu (approved):

PR #1607, opened 2024-02-05
Handle capacity scheduling for optimizers with lazy state init

Haichao-Zhang:
Torch has refactored some optimizers.
After refactorization, some torch optimizers (e.g. Adam) implementation in torch switched to lazy state initialization during step(), and the way to check if lazy init is done or not is based on checking the len(state):
https://github.com/pytorch/pytorch/blob/b2e0f8d82d721f6598113f64a408b0017bb90cb2/torch/optim/adam.py#L102-L103

In our case, since we added ``rng_state`` to state in add_param_group, len(state)  becomes 1 before entering step(), and the current implementation in torch will regard that the lazy state init has already done, breaking the subsequent access ops on some states that are expected to be exist:
https://github.com/pytorch/pytorch/blob/b2e0f8d82d721f6598113f64a408b0017bb90cb2/torch/optim/adam.py#L120-L121

This PR fixes this issue and supports capacity scheduling to all kinds of optimizers regardless of their internal way of handling states.

### DIFF
--- a/alf/optimizers/optimizers.py
+++ b/alf/optimizers/optimizers.py
@@ -363,7 +363,7 @@ def wrap_optimizer(cls):
 
         super(NewCls, self).step(closure=closure)
 
-        if len(customzied_states) > 1:
+        if len(customzied_states):
             self._append_customized_states(customzied_states)
 
         if not isinstance(self, NeroPlus):
--- a/alf/optimizers/optimizers.py
+++ b/alf/optimizers/optimizers.py
@@ -15,7 +15,7 @@
 import copy
 import numpy as np
 import torch
-from typing import Callable, Dict, Union
+from typing import Callable, Dict, List, Union
 
 import alf
 from alf.utils import common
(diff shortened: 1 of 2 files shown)


alf/optimizers/optimizers.py
```diff
@@ -318,8 +353,19 @@ def step(self, closure=None):
...
+            # if no stepping done yet, remove customized states to avoid any impacts on
+            # some specific optimizers' internal procedures, e.g. lazy state initialization
+            # https://github.com/pytorch/pytorch/blob/b2e0f8d82d721f6598113f64a408b0017bb90cb2/torch/optim/adam.py#L102-L103
+            customzied_states = self._remove_customized_states()
+        else:
+            customzied_states = {}
+
         super(NewCls, self).step(closure=closure)
 
+        if len(customzied_states) > 1:
```
emailweixu: `if len(customized_states) > 0` or simply `if customized_states`

Haichao-Zhang: Fixed

emailweixu (approved):

PR #1609, opened 2024-02-07
Bug fix for pretty_print.py

emailweixu:
PrettyPrinter._format() in alf/utils/pretty_print.py changed content of class member _dispatch. _dispatch is also used by the original PrettyPrinter. This can cause problem when the original pprint is used somewhere, because the original PrettyPrinter does not have format_namedtuple_items().

This PR fixed this problem by explicitly using PrettyPrinter.format_namedtuple_items()

### DIFF
--- a/alf/nest/nest.py
+++ b/alf/nest/nest.py
@@ -40,7 +40,11 @@ NestedTensor = Union[
 
 
 def flatten(nest):
-    """(C++) Returns a flat list from a given nested structure."""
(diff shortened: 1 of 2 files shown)


emailweixu:
ping @hnyu

hnyu (approved):

hnyu:
Sorry my github notifcations of the ALF repo has some issues.

PR #1613, opened 2024-02-15
Set DDP PerProcessContext in spawned process

emailweixu:
Also use get_env() instead of env_constructor to get the environment to avoid creating env multiple times.


emailweixu:
ping @breakds

breakds (approved):

PR #1616, opened 2024-02-21
Support python 3.11 and torch2.2

emailweixu:
1. Update dependencies to newer versions.
2. Fix some errors caused by deprecated np types.
3. Fix some unittests
4. Updated the CI docker file
5. Removed SocialRobot from the docker image because it depends on an older version of gazebo, which is not available for ubuntu 22.04
6. Removed procgen from the docker image since it does not support python3.11
7. Also updated the docker image for GPU training: horizonrobotics/cuda:11.8.0-py3.11-torch2.2-ubuntu22.04

The updated code can still work with older version of python (3.8, 3.9, 3.10).


hnyu:
Need to change README about the installation? Currently it's still python 3.7+

hnyu (approved):

PR #1617, opened 2024-02-23
fix rng_state

Haichao-Zhang:
When loading from saved ckpt, the rnd_state should be byte type.

### DIFF
--- a/alf/optimizers/optimizers.py
+++ b/alf/optimizers/optimizers.py
@@ -252,8 +252,8 @@ def wrap_optimizer(cls):
                             'rng_state'] = self._random_number_generator.get_state(
                             )
                     else:
-                        rng_state = state['rng_state']
-                        self._random_number_generator.set_state(rng_state)
+                        self._random_number_generator.set_state(
+                            state['rng_state'])
 
                     # generate capacity mask using the same random number generator state
                     n = p.numel()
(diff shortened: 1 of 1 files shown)


alf/optimizers/optimizers.py
```diff
@@ -252,7 +252,8 @@ def _adjust_capacity(self, capacity_ratio: float, old_param_values: Dict):
                             'rng_state'] = self._random_number_generator.get_state(
                             )
                     else:
-                        rng_state = state['rng_state']
+                        # the state['rng_state'] could be float tensor if loaded from ckpt
+                        rng_state = state['rng_state'].cpu().byte()
```
emailweixu: This should be done in load_state_dict.

Haichao-Zhang: Done

alf/optimizers/optimizers.py
```diff
@@ -482,6 +482,22 @@ def add_param_group(self, param_group):
...
+        This function first call the parent's ``load_state_dict()`` function, and
+        then make sure the ``rng_state`` is in correct data type.
+        """
+        super(NewCls, self).load_state_dict(state_dict)
+        if len(self.state) > 0:
+            for param_group in self.param_groups:
+                for p in param_group['params']:
+                    state = self.state[p]
+                    if 'rng_state' in state:
+                        state['rng_state'] = state['rng_state'].cpu().byte()
```
emailweixu: Is rng_state always in cpu?

Haichao-Zhang: Handled it in a more general way by using ``_rng_state_device``

emailweixu (approved):

PR #1619, opened 2024-02-27
check unsync'ed scheduler progress; sync progress during play; make sure progress is updated when loading from ckpt

le-horizon:
total 3 changes.

(cherry picked from commit b7854b933ba855c3d15efc8143ad15d217121553)

### DIFF
--- a/alf/utils/schedulers.py
+++ b/alf/utils/schedulers.py
@@ -65,6 +65,12 @@ class Scheduler(object):
             progress_type (str): one of "percent", "iterations", "env_steps"
         """
         assert progress_type in _progress
+        if not alf.get_config_value('create_environment.nonparallel'):
+            assert alf.get_config_value(
+                'TrainerConfig.sync_progress_to_envs'), (
+                    "When using parallel environments, the progress should be "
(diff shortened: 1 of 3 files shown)


alf/utils/schedulers.py
```diff
@@ -65,6 +65,12 @@ def __init__(self, progress_type):
             progress_type (str): one of "percent", "iterations", "env_steps"
         """
         assert progress_type in _progress
+        if not alf.get_config_value('create_environment.nonparallel'):
+            assert alf.get_config_value(
+                'TrainerConfig.sync_progress_to_envs'), (
+                    "When using parallel environments, the progress should be "
+                    "synchronized to the environments. Please set "
+                    "TrainerConfig.sync_progress_to_envs to True.")
```
emailweixu, line 73: This effectively requires that sync_progress_to_envs=True whenever scheduler is used, even if it is not used in environment. A better way is to do this check when the value of the scheduler is requested in the process environment.

Besides this, scheduler progress is not available to evaluation environment now.

Let me fix all these.

PR #1620, opened 2024-02-28
Improve the handling of scheduler progress in process environment

emailweixu:
If a scheduler is used in the process environment, an error will be raised if sync_progress_to_envs is not set to True.

Also added sync_progress() to evaluator.py

### DIFF
--- a/alf/trainers/evaluator.py
+++ b/alf/trainers/evaluator.py
@@ -267,6 +267,7 @@ def evaluate(env: AlfEnvironment, algorithm: RLAlgorithm,
     """
     batch_size = env.batch_size
     env.reset()
+    env.sync_progress()
     time_step = common.get_initial_time_step(env)
     algorithm.eval()
     policy_state = algorithm.get_initial_predict_state(env.batch_size)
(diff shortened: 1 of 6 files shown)


alf/trainers/evaluator.py
```diff
@@ -267,6 +267,7 @@ def evaluate(env: AlfEnvironment, algorithm: RLAlgorithm,
     """
     batch_size = env.batch_size
     env.reset()
+    env.sync_progress()
```
le-horizon, line 270: This has to be qualified so that only parallel_envs are allowed, otherwise sync_progress() from a thread env will throw RuntimeError?

le-horizon (commented):
Thanks for the fixes, Wei!

Two points to fix
a) We still need the four lines of changes in policy_trainer from this PR: https://github.com/HorizonRobotics/alf/pull/1619/files
Otherwise, 1) progress from ckpts won't be updated, and 2) progress during play won't be correct.

b) inlined

Thanks,
Le

alf/trainers/evaluator.py
emailweixu: Now the default implementation of sync_progress() does not fail.

le-horizon (approved):
LG, thanks for the fixes!

PR #1621, opened 2024-03-02
Add example capacity scheuling conf and train_play_test

Haichao-Zhang:
Add an example conf to illustrate its usage.

image

### DIFF
--- a/alf/bin/train_play_test.py
+++ b/alf/bin/train_play_test.py
@@ -460,6 +460,11 @@ class TrainPlayTest(alf.test.TestCase):
             conf_file='dyna_actrepeat_sac_bipedalwalker_conf.py',
             extra_train_params=OFF_POLICY_TRAIN_PARAMS)
 
+    def test_bipedal_walker_w_capacity_schedule(self):
+        self._test(
+            conf_file='sac_bipedal_walker_capacity_schedule_conf.py',
+            extra_train_params=OFF_POLICY_TRAIN_PARAMS)
+
     def test_dyna_actrepeat_sac_pickplace(self):
         self._test(
             conf_file="dyna_actrepeat_sac_pickplace.gin",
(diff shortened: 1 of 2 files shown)


Haichao-Zhang:
@hnyu you can use this conf for testing, and you can set a smaller value for eval_interval. I have tested it with eval_interval=1 without any problem.

hnyu:
Confirmed that if I replace AdamTF with Adam, then the error occurs. Is capacity scheduling only compatible with AdamTF at this moment?

Haichao-Zhang:
> Confirmed that if I replace AdamTF with Adam, then the error occurs. Is capacity scheduling only compatible with AdamTF at this moment?

Should be compatible with all optimizers.
I also tested with Adam yesterday. No problem as well.

Run again and screenshot:

image

image

hnyu:
Seems that for torch>2, Adam has some changes that make this crash happen. I thought I was using the docker image with torch1.11, but in fact it was torch2.1. So if I use docker with torch1.11, the error is gone. 

You can try loading horizonrobotics/cuda:11.8-torch2.1-py3.8-ubuntu20.04 and see the error happening.

Haichao-Zhang:
Good that you can run without problem under the standard setting (as in hobot).

It is possible that new changes (especially API related ones) from torch team on the optimizers may cause problems.
This is a general issue if the API changes from torch team is not backward compatible.

hnyu (approved):
For anyone with torch>2, if using capacity scheduling, make sure to use AdamTF instead of Adam at this moment.

Haichao-Zhang:
> For anyone with torch>2, if using capacity scheduling, make sure to use AdamTF instead of Adam at this moment.

Yes. The reason is that AdamTF is implemented by ourselves and is not affected by torch's changes on optimizers.

PR #1622, opened 2024-03-02
Include batch dimension for std/scale even if state_dependent_(std/sc…

emailweixu:
…ale) is False

Sometimes the distribution needs to be saved in replay buffer, which requires all the tensor have a batch dimension. Also if the distribution generated in train_step is returned as one member of info, it also needs to have batch dimension.

### DIFF
--- a/alf/networks/projection_networks.py
+++ b/alf/networks/projection_networks.py
@@ -27,6 +27,7 @@ from alf.tensor_specs import TensorSpec, BoundedTensorSpec
 from alf.networks.network import Network, wrap_as_network
 from alf.utils import dist_utils
 import alf.utils.math_ops as math_ops
+from alf.utils.tensor_utils import expand_batch_dim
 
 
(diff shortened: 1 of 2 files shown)


alf/networks/projection_networks.py
```diff
@@ -663,7 +665,8 @@ def __init__(self,
             self._scale = nn.Parameter(
                 action_spec.constant(scale_bias_initializer_value),
                 requires_grad=True)
-            self._scale_projection_layer = lambda _: self._scale
+            self._scale_projection_layer = lambda x: expand_batch_dim(
```
hnyu: Why not just use `tensor_extend_new_dim(self._scale, 0, x.shape[0])`?

emailweixu: changed to use tensor_extend_new_dim

hnyu (approved):

PR #1623, opened 2024-03-02
check grad when summary

hnyu:
A tensor's grad could be None if the tensor requires grad but it's never involved in loss calculation (e.g., when a flag ignores certain losses).

### DIFF
--- a/alf/utils/summary_utils.py
+++ b/alf/utils/summary_utils.py
@@ -469,9 +469,10 @@ def summarize_tensor_gradients(name, tensor, batch_dims=1, clone=False):
     """
 
     def _hook(grad, name):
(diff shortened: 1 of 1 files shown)


emailweixu:
> A tensor's grad could be None if the tensor requires grad but it's never involved in loss calculation (e.g., when a flag ignores certain losses).

Why do you need to summarize its gradient in this case?

hnyu:
> > A tensor's grad could be None if the tensor requires grad but it's never involved in loss calculation (e.g., when a flag ignores certain losses).
> 
> Why do you need to summarize its gradient in this case?

I use the same code for different flags. So just don't want to modify the source in the case when the loss is excluded.

emailweixu (approved):

PR #1626, opened 2024-03-16
raise exception when cxt manager exits

hnyu:
Need to handle exceptions happening inside the context manager, otherwise no error message will be reported, and code will continue execution jumping out of the context, until something weird happens. For example,

```python
with alf.utils.common.eval_context():
            low_level_step = self._low_level_alg.predict_step(
                low_time_step, state.low)
```
If low_level_alg has any error, the current code won't report it.

### DIFF
--- a/alf/utils/common.py
+++ b/alf/utils/common.py
@@ -336,15 +336,15 @@ class PeriodicReset(nn.Module):
 
         m <-- re-initialize.
 
-    Note: 
+    Note:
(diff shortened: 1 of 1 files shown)


Haichao-Zhang (approved):

breakds (approved):
Silencing the exception does not sound like a good default behavior. Weird that context manager was designed this way by the Python guys.

PR #1627, opened 2024-03-21
fix batch_ensemble layers initialization

runjerry:
Currently, ``nn.Module.__init__()`` is called twice, by both the BatchEnsemble layer and its parent, as a result,  the extra ``nn.Parameter``'s defined by the BatchEnsemble layer end up unregistered in the ``nn.Module``.

### DIFF
--- a/alf/layers.py
+++ b/alf/layers.py
@@ -575,6 +575,9 @@ class FCBatchEnsemble(FC):
     def reset_parameters(self):
         """Reinitialize parameters."""
         super().reset_parameters()
+        # We need to the check the existence of ``_s`` and ``_r`` since
+        # ``reset_parameters()`` is also called by the init function of the parent
+        # class when both ``_s`` and ``_r`` are not initialized yet.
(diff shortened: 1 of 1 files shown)


alf/layers.py
```diff
@@ -560,33 +569,27 @@ def __init__(self,
...
-        self._r.data.mul_(2)
-        self._r.data.sub_(1)
-        self._s.data.mul_(2)
-        self._s.data.sub_(1)
-        if self._use_ensemble_bias:
-            nn.init.uniform_(
-                self._ensemble_bias.data,
-                a=-self._bias_init_range,
-                b=self._bias_init_range)
+        if hasattr(self, '_r') and hasattr(self, '_s'):
```
emailweixu, line 581: Add a comment about why need to check for the existence of _r and _s

runjerry: Added.

emailweixu (approved):

PR #1628, opened 2024-03-22
Encoding network batch ensemble

runjerry:
Make encoding network work with batch_ensemble layers. 

Handel inputs with ``emsemble_ids`` by wrapping ``input_preprocessors`` and ``preprocessing_combiner`` with ``Parallel``.

### DIFF
--- a/alf/networks/encoding_networks.py
+++ b/alf/networks/encoding_networks.py
@@ -125,9 +125,9 @@ class ImageEncodingNetwork(_Sequential):
         if flatten_output:
             if use_batch_ensemble and output_ensemble_ids:
                 nets.append(
-                    NetworkWrapper(
-                        lambda inputs: (alf.layers.Reshape((-1, ))(inputs[0]),
-                                        alf.layers.Identity()(inputs[1])), ()))
+                    Parallel((alf.layers.Reshape(
+                        (-1, )), alf.layers.Identity()),
+                             ((), TensorSpec((), dtype=torch.int64))))
             else:
                 nets.append(alf.layers.Reshape((-1, )))
 
@@ -823,11 +823,11 @@ class EncodingNetwork(_Sequential):
                 assert output_tensor_spec.numel % input_size == 0
             if use_batch_ensemble and output_ensemble_ids:
                 nets.append(
-                    NetworkWrapper(
-                        lambda inputs: (alf.layers.Reshape(output_tensor_spec.
-                                                           shape)(inputs[0]),
-                                        alf.layers.Identity()(inputs[1])),
-                        (output_tensor_spec, TensorSpec(()))))
+                    Parallel(
+                        (alf.layers.Reshape(output_tensor_spec.shape),
+                         alf.layers.Identity()),
+                        (output_tensor_spec, TensorSpec(
+                            (), dtype=torch.int64))))
             else:
                 nets.append(alf.layers.Reshape(output_tensor_spec.shape))
 
--- a/alf/networks/encoding_networks.py
+++ b/alf/networks/encoding_networks.py
@@ -18,7 +18,7 @@ from typing import Callable, Optional, Tuple, Union
 
 import torch
 import torch.nn as nn
-from .containers import _Sequential, Parallel
+from .containers import Sequential, _Sequential, Parallel
 from .network import Network, NetworkWrapper
 import alf
 import alf.layers as layers
@@ -40,7 +40,6 @@ class ImageEncodingNetwork(_Sequential):
                  conv_layer_params,
                  use_batch_ensemble=False,
                  ensemble_size=10,
-                 output_ensemble_ids=True,
                  same_padding=False,
                  activation=torch.relu_,
                  kernel_initializer=None,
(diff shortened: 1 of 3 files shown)


alf/networks/encoding_networks.py
```diff
@@ -700,6 +737,14 @@ def __init__(self,
             assert isinstance(spec, TensorSpec), \
                 "The spec must be an instance of TensorSpec!"
 
+        if input_with_ensemble_ids:
+            nets = [
+                Parallel(
+                    (_Sequential(nets, input_tensor_spec=input_tensor_spec),
```
emailweixu: why `_Sequential` instead of `Sequential`?

alf/networks/encoding_networks.py
```diff
@@ -709,11 +754,16 @@ def __init__(self,
             net = ImageEncodingNetwork(
                 input_channels, (height, width),
                 conv_layer_params,
+                use_batch_ensemble=use_batch_ensemble,
+                ensemble_size=ensemble_size,
+                output_ensemble_ids=output_ensemble_ids,
```
emailweixu: output_ensemble_ids=True

alf/networks/encoding_networks.py
```diff
@@ -709,11 +754,16 @@ def __init__(self,
...
                 activation=activation,
                 kernel_initializer=kernel_initializer,
                 flatten_output=True)
             spec = net.output_spec
             nets.append(net)
+        if isinstance(spec, tuple):
```
emailweixu: `if use_batch_ensemble:`

alf/networks/encoding_networks.py
```diff
@@ -733,9 +783,18 @@ def __init__(self,
             assert isinstance(fc_layer_params, tuple)
             fc_layer_params = list(fc_layer_params)
 
+        if use_batch_ensemble:
+            assert ensemble_size > 1
+            fc_layer_ctor = functools.partial(
+                layers.FCBatchEnsemble,
+                ensemble_size=ensemble_size,
+                output_ensemble_ids=output_ensemble_ids)
```
emailweixu: output_ensemble_ids=True

alf/networks/encoding_networks.py
```diff
@@ -755,7 +814,7 @@ def __init__(self,
                 last_kernel_initializer = kernel_initializer
 
             nets.append(
-                layers.FC(
+                fc_layer_ctor(
```
emailweixu, line 820: for batch ensemble: output_ensemble_ids=output_ensemble_ids

emailweixu (changes requested):

alf/networks/encoding_networks.py
runjerry: Updated.

alf/networks/encoding_networks.py
runjerry: Updated.

alf/networks/encoding_networks.py
runjerry: Updated.

alf/networks/encoding_networks.py
runjerry: Updated. Thanks for catching this issue.

alf/networks/encoding_networks.py
runjerry: The interface of ``Sequential`` wrapper will wrap the input modules into an extra layer of tuple, which fails the ``isinstance(module, Callable)`` check for some of the test cases, say when ``nets = [NestSum()]``.

emailweixu: I don't get it. I think `Sequential(*nets, input_tensor_spec=input_tensor_spec)` is equivalent to `_Sequential(nets, input_tensor_spec=input_tensor_spec)`

alf/networks/encoding_networks.py
emailweixu: This means that the unittests do not properly test different scenarios?

runjerry: That is a good point. I then realized that handling the ``output_ensemble_ids`` option for ``EncodingNetwork`` with and without ``last_layer_size`` becomes a bit tedious, so I just removed this option and always create BatchEnsemble layers with ``output_ensemble_ids=True``.

alf/networks/encoding_networks.py
runjerry: Oops, forgot the star sign, updated.

alf/networks/encoding_networks.py
```diff
@@ -69,6 +71,9 @@ def __init__(self,
             conv_layer_params (tuppe[tuple]): a non-empty tuple of
                 tuple (num_filters, kernel_size, strides, padding), where
                 padding is optional
+            use_batch_ensemble (bool): whether use BatchEnsemble Conv layers.
```
emailweixu: should comment the input and the output in this case.

alf/networks/encoding_networks.py
```diff
@@ -659,6 +682,13 @@ def __init__(self,
                 used.
             use_fc_bn (bool): whether use Batch Normalization for fc layers.
             use_fc_ln (bool): whether use Layer Normalization for fc layers.
+            use_batch_ensemble (bool): whether use BatchEnsemble FC and Conv
```
emailweixu: should comment the output of the network, i.e., it is tuple with ensemble_ids

alf/networks/encoding_networks.py
runjerry: Updated.

alf/networks/encoding_networks.py
runjerry: Updated.

emailweixu (approved):

PR #1631, opened 2024-03-25
fix normalizing flow test

hnyu:
The variable `bs` was incorrectly overwritten by `z`'s batch. Now use a different name for that.


runjerry (approved):
Thanks for the fix!

PR #1633, opened 2024-03-28
Batch ensemble ddpg

runjerry:
Update ``actor_network``, ``critic_network``, and ``ddpg_algorithm`` to work with batch_ensemble layers.

### DIFF
--- a/alf/algorithms/ddpg_algorithm.py
+++ b/alf/algorithms/ddpg_algorithm.py
@@ -74,13 +74,18 @@ class DdpgAlgorithm(OffPolicyAlgorithm):
                  config: TrainerConfig = None,
                  ou_stddev=0.2,
                  ou_damping=0.15,
+                 noise_clipping=None,
                  critic_loss_ctor=None,
                  num_critic_replicas=1,
                  target_update_tau=0.05,
                  target_update_period=1,
+                 actor_update_period=1,
                  rollout_random_action=0.,
                  dqda_clipping=None,
                  action_l2=0,
+                 use_batch_ensemble=False,
+                 ensemble_size=10,
+                 input_with_ensemble_ids=False,
                  actor_optimizer=None,
                  critic_optimizer=None,
                  checkpoint=None,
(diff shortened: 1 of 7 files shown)


alf/algorithms/ddpg_algorithm.py
```diff
@@ -139,6 +148,17 @@ def __init__(self,
                 gradient dqda element-wise between ``[-dqda_clipping, dqda_clipping]``.
                 Does not perform clipping if ``dqda_clipping == 0``.
             action_l2 (float): weight of squared action l2-norm on actor loss.
+            use_batch_ensemble (bool): whether to use BatchEnsemble FC and Conv2D
```
Haichao-Zhang, line 158: Ideally, we might should make these batch ensemble related parameters transparent to the ddpg_algorithm? Basically, the ddpg_algorithm should not use batch_ensemble related parameters in the ideal case.

runjerry: That's a good point. Currently ddpg needs the ``use_batch_ensemble`` to do some post processing when forwarding critic networks during training. Let me think it over if there might be some alternative methods to work around.

alf/algorithms/ddpg_algorithm.py
```diff
@@ -281,14 +318,39 @@ def _update_random_action(spec, noisy_action):
         if self._rollout_random_action > 0:
             nest.map_structure(_update_random_action, self._action_spec,
                                pred_step.output)
-        return pred_step
+
+        if self.need_full_rollout_state():
```
emailweixu, line 348: We want the algorithm use the same ensemble_id during an entire episode. This means that it should store ensembled_id in state and use the same ensemble_id to call actor_network

runjerry: Oh yes, good point, I think that is the reason why I had to tweak the ``ddpg_algorithm_test`` to pass the toy unittest. Updated.

PR #1634, opened 2024-03-29
Add summarize_play to rl_algorithm

runjerry:
Similar as ``summarize_rollout``, for injecting custom summary functions during play and eval.

### DIFF
--- a/alf/algorithms/rl_algorithm.py
+++ b/alf/algorithms/rl_algorithm.py
@@ -447,6 +447,14 @@ class RLAlgorithm(Algorithm):
             if len(field) == 1:
                 summary_utils.summarize_distribution("action_dist", field[0])
 
+    @alf.configurable(whitelist=["custom_summary"])
+    def summarize_play(
+            self,
+            time_step: TimeStep,
+            custom_summary: Optional[Callable[[TimeStep], None]] = None):
+        if custom_summary is not None:
+            custom_summary(time_step)
+
     def summarize_metrics(self):
         """Generate summaries for metrics ``AverageEpisodeLength``,
         ``AverageReturn``, etc.
--- a/alf/algorithms/rl_algorithm.py
+++ b/alf/algorithms/rl_algorithm.py
@@ -455,7 +455,10 @@ class RLAlgorithm(Algorithm):
         """Generate summaries for play or evaluate.
 
         Args:
-            experience: experience collected from ``env.step`` of play.
+            experience: experience of one step rollout in the environment during 
+            play or evaluation, this is in contrast to the experience input to 
+            ``summarize_rollout`` where a sequence of multiple rollout steps
+            are collected.
             custom_summary: when specified it is a function that will be called every
                time when this ``summarize_play`` hook is called. This provides
                a convenient way for the user to customize ``summarize_play`` from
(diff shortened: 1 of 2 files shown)


alf/algorithms/rl_algorithm.py
```diff
@@ -447,6 +447,14 @@ def summarize_train(self, experience, train_info, loss_info, params):
             if len(field) == 1:
                 summary_utils.summarize_distribution("action_dist", field[0])
 
+    @alf.configurable(whitelist=["custom_summary"])
+    def summarize_play(
+            self,
+            time_step: TimeStep,
+            custom_summary: Optional[Callable[[TimeStep], None]] = None):
+        if custom_summary is not None:
+            custom_summary(time_step)
```
emailweixu: I think for play or aync eval, ALF doesn't write anything for tensorboard.

runjerry: It doesn't. I just want to inject summarize function from config for eval/play, as analogous as PR #1385, maybe can do it elsewhere?

emailweixu: My bad. ALF does write tensorboard summary for eval.

emailweixu: I feel it makes sense to make summarize_play have same signature as summarize_rollout, i.e., accepting Experience instead of TimeStep. Also please add doc-string.

runjerry: Updated.

alf/trainers/policy_trainer.py
```diff
@@ -973,6 +973,7 @@ def _step(algorithm,
         time.sleep(sleep_time_per_step)
 
     next_time_step = env.step(policy_step.output)
+    algorithm.summarize_play(next_time_step.cpu())
```
emailweixu: Need to make it an Experience

runjerry: Corrected.

alf/trainers/policy_trainer.py
```diff
@@ -973,6 +973,12 @@ def _step(algorithm,
         time.sleep(sleep_time_per_step)
 
     next_time_step = env.step(policy_step.output)
+    experience = Experience(
```
emailweixu: you can use `make_experience`: `make_experience(time_step, policy_step, policy_state)`
Note that time_step should be used instead of next_time_step

alf/algorithms/rl_algorithm.py
```diff
@@ -447,6 +447,23 @@ def summarize_train(self, experience, train_info, loss_info, params):
...
 
+    @alf.configurable(whitelist=["custom_summary"])
+    def summarize_play(
+            self,
+            experience: Experience,
+            custom_summary: Optional[Callable[[Experience], None]] = None):
+        """Generate summaries for play or evaluate.
+
+        Args:
+            experience: experience collected from ``env.step`` of play.
```
emailweixu: It is better to explain that the experience here is different from the one for summarize_rollout in that the experience here   is just for one step, where the one for summarize_rollout is a sequence.

alf/trainers/policy_trainer.py
runjerry: Updated accordingly.

alf/algorithms/rl_algorithm.py
runjerry: Added further docstring.

emailweixu (approved):

PR #1638, opened 2024-04-20
Also print out the expected shape upon stride mismatch

breakds:
## Motivation

This is to slightly improve the visibility of the error "Strides mismatch", we can sometimes happen when the code has bug in specs. Specifically, this PR updated the error message so that instead of only showing the expected strides, it also shows the **expected shape**, which usually makes it easy for debugging.

## Testing

In the recent effort of creating an offline dataset, I got a dataset in which the image shape is different from the expected one. Alf reports an error of:

```console
RuntimeError: Strides mismatch. Expected: [14400, 160, 1], Got: [480, 3, 1], shape: [90, 160, 3]). It might be caused by non-continguous array, which can be fixed by copying the array. The offending buffer is ...
```

With this change, the error message will become:

```console
RuntimeError: Strides mismatch. Expected: [14400, 160, 1] (non-batch shape is [3, 90, 160]), Got: [480, 3, 1] (shape is [90, 160, 3]). It might be caused by non-continguous array, which can be fixed by copying the array. The offending buffer is ...
```

Which makes it slight easier to spot the problem as shape is just more intuitive.

## Note

Note that `CheckStrides` remain only checking the strides. The shape is just for visibility purpose.

### DIFF
--- a/alf/environments/parallel_environment.cpp
+++ b/alf/environments/parallel_environment.cpp
@@ -217,7 +217,7 @@ SharedDataBuffer::SharedDataBuffer(py::object data_spec,
     buffer_infos_.emplace_back(array.request());
 
     // Stores the strides and shapes (ignoring the first dimension which is a
-    // dummpy batch dimension) of each slice.
+    // dummy batch dimension) of each slice.
     auto& strides = buffer_infos_.back().strides;
     slice_strides_.emplace_back(strides.begin() + 1, strides.end());
     const std::vector<ssize_t>& shape = buffer_infos_.back().shape;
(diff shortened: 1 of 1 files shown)


alf/environments/parallel_environment.cpp
```diff
@@ -208,8 +215,14 @@ SharedDataBuffer::SharedDataBuffer(py::object data_spec,
     py::buffer array = py::reinterpret_borrow<py::buffer>(
         spec.attr("numpy_zeros")(outer_dims));
     buffer_infos_.emplace_back(array.request());
+
+    // Stores the strides and shapes (ignoring the first dimension which is a
+    // dummpy batch dimension) of each slice.
```
emailweixu: dummpy -> dummy

breakds: Fixed.

emailweixu (approved):

PR #1640, opened 2024-05-01
Improve the support for MixtureProjectionNetwork

emailweixu:
When MixtureProjectionNetwork is used for action distribution, we need to correctly handle get_mode() and summarize_distribution() for it.

### DIFF
--- a/alf/utils/dist_utils.py
+++ b/alf/utils/dist_utils.py
@@ -1128,10 +1128,19 @@ def get_mode(dist):
         # approach to compute the mode, by using the mode of the component
         # distribution that has the highest component probability.
         # [B]
+        batch_shape = dist.batch_shape
         ind = get_mode(dist.mixture_distribution)
         # [B, num_component, d]
         component_mode = get_mode(dist.component_distribution)
-        mode = component_mode[torch.arange(component_mode.shape[0]), ind]
+        if len(batch_shape) == 1:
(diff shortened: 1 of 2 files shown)


alf/utils/dist_utils.py
```diff
@@ -1128,10 +1128,19 @@ def get_mode(dist):
...
         # distribution that has the highest component probability.
         # [B]
+        batch_shape = dist.batch_shape
         ind = get_mode(dist.mixture_distribution)
         # [B, num_component, d]
         component_mode = get_mode(dist.component_distribution)
-        mode = component_mode[torch.arange(component_mode.shape[0]), ind]
+        if len(batch_shape) == 1:
+            mode = component_mode[torch.arange(batch_shape[0]), ind]
+        elif ind.ndim == 2:
```
breakds: How about `len(batch_shape) == 2` to be consistent with the above `if`, and also `d0, d1 = batch_shape` below?

alf/utils/summary_utils.py
```diff
@@ -327,6 +347,19 @@ def _summarize_one(path, dist):
             else:
                 return
 
+            if ind is not None:
```
breakds, line 350: Would be great to add a one line comment saying this is for handling the mixture distribution, picking the component with highest weight.

breakds (approved):

PR #1642, opened 2024-05-07
Improve normalized advantage calculation

emailweixu:
1. Use LazyBatchNorm for advantage normalization.The motivation is that for DDP training, the normalization statistics will combine the statistics from all GPUs.

2. Calculate normalized advantage in PPOAlgorithm.preprocess so that the normalization is based on a much larger batch.

### DIFF
--- a/alf/algorithms/actor_critic_loss.py
+++ b/alf/algorithms/actor_critic_loss.py
@@ -28,21 +28,6 @@ ActorCriticLossInfo = namedtuple("ActorCriticLossInfo",
                                  ["pg_loss", "td_loss", "neg_entropy"])
 
 
-def _normalize_advantages(advantages, variance_epsilon=1e-8):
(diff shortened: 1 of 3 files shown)


hnyu:
> 2. Calculate normalized advantage in PPOAlgorithm.preprocess so that the normalization is based on a much larger batch.

Not sure why computing normalized advantage in preprocess is based on a larger batch? Is it due to mini_batch_length smaller than full batch length?

emailweixu:
> > 2. Calculate normalized advantage in PPOAlgorithm.preprocess so that the normalization is based on a much larger batch.
> 
> Not sure why computing normalized advantage in preprocess is based on a larger batch? Is it due to mini_batch_length smaller than full batch length?

In preprocess, all the samples from the rollout are processed. In loss, only mini_batch_size samples are processed.

hnyu (approved):

PR #1644, opened 2024-05-10
ignore inf in loss tb summary, and report valid_rate

le-horizon:
This change is to report statistics like precision / recall of segmentation maps, where often cases the target segment does not appear in the image.  In those cases, recall will be invalid, and we supply inf as placeholder value for the statistic.

This change ignores these invalid values, and also reports the rate of how often the metric is valid.

(cherry picked from commit 72cf8fa4516322d09690c3e12eae75569846a747)

### DIFF
--- a/alf/algorithms/algorithm.py
+++ b/alf/algorithms/algorithm.py
@@ -1207,7 +1207,25 @@ class Algorithm(AlgorithmInterface):
             loss_info.loss * weight)
 
         loss_info = loss_info._replace(gns=gns)
-        loss_info = alf.nest.map_structure(torch.mean, loss_info)
+
+        new_fields_values = {}
+
+        def _mean_ignore_inf(path, loss):
+            inf_mask = torch.isinf(loss)
+            if inf_mask.any():
+                # populate the valid_rate field only when value can be inf.
+                new_fields_values[
+                    path + '-valid_rate'] = 1.0 - inf_mask.float().mean()
+            if inf_mask.all():
+                return torch.tensor(np.float32(np.inf), device=loss.device)
+            else:
+                return loss[~inf_mask].mean()
+
+        loss_info = alf.nest.py_map_structure_with_path(
+            _mean_ignore_inf, loss_info)
+
+        for k, v in new_fields_values.items():
+            loss_info = alf.nest.set_field(loss_info, k, v)
 
         return loss_info, all_params
 


alf/algorithms/algorithm.py
```diff
@@ -1207,7 +1207,25 @@ def update_with_gradient(self,
...
+            if inf_mask.all():
+                return torch.tensor(np.float32(np.inf), device=loss.device)
+            else:
+                return loss[~inf_mask].mean()
+
+        loss_info = alf.nest.py_map_structure_with_path(
+            _mean_ignore_inf, loss_info)
+
+        for k, v in new_fields_values.items():
+            loss_info = alf.nest.set_field(loss_info, k, v)
```
hnyu, line 1228: A better way is to do this in the specific algorithm. You can generate the mask and take valid losses. The loss shape might not be [B] anymore, but you can assign loss.mean() to ``scalar_loss` in LossInfo.

In a general case, an inf loss can be quite problematic (the objective is wrong, the gradient has exploded, etc). So we don't want to ignore it quietly. 

Also, if the quantity to be summarized is not a 'loss', then probably it should be put in LossInfo. You can create a customized summary in the alg's `calc_loss` function.

le-horizon: Loss_info's extra field requires all tensors to be of the same shape, and this function computes the mean of all those values, and summarizes them.

If I put anything inside loss_info extra, then it's automatically summarized.  The automatic summary will not be correct due to missing (inf) values.  That's why I needed this change in alf, and cannot be done just in the algorithm.  (I have the corresponding change in maps_pred_algo.train_step to populate the fields for summary.)

Should I use this change or maybe create a new field in loss_info for custom tb summary?

le-horizon: Btw, infs are not ignored.  They are summarized into "valid_rate" plots, when valid rate is not 100%.

hnyu: Why do we have to put the precision / recall in loss_info since they are not losses? Can just return any new fields in the train step info and customize the summary logic in the alg's calc_loss. You can refer to MapsPredAlgorithmV2's calc_loss, where the info is not LossInfo but a general MapsPredInfo.  In fact, all those images and videos generated in train_step are handled in this way.

le-horizon: Good point.  Done.

PR #1645, opened 2024-05-13
Fix DDP logging

emailweixu:
With this fix, the logging work as desired when multigpu training is used

```
I0513 12:43:54.905132 126015779825472 policy_trainer.py:666] [rank 01]  [pid: 2128791] nuplan_extent/planning/script/config/rl/hybrid_ppo_conf.py -> ddp: 1 time=17.778 throughput=230.39
```


le-horizon (approved):
LG, thanks for the fix!

PR #1646, opened 2024-05-15
Log sending evaluation job

emailweixu:
If the evaluation process dies without giving any message, the main process can stuck at sending the evaluation job without any output. The logging for sending evaluation job can help identify such situation.


le-horizon (approved):
LG.

PR #1647, opened 2024-05-17
Support on-demand debug for multigpu training

emailweixu:
For multigpu training, all the training workers are spawned subprocesses. But pdb session cannot start in a subprocess because the lack of stdin. So we change to lauch of ddp processes so that the main process will be the rank-0 worker so that we can start a pdb session by using `kill -10`


le-horizon (approved):
LG, thanks for the improvements!

PR #1648, opened 2024-05-29
Forward max width and height args to rpack in pack_image_nest

QuantuMope:
Simple PR that allows `max_width` and `max_height` args to be forwarded to `rpack.pack`. Needed to control the dimensions of the packed image in cases where multiple minimal area solutions exist.

Also, fixed some minor spelling errors concerning `interploation --> interpolation`.

### DIFF
--- a/alf/summary/render.py
+++ b/alf/summary/render.py
@@ -19,6 +19,7 @@ import matplotlib
 matplotlib.use('Agg')  # 'Agg' no need for xserver!
 import matplotlib.pyplot as plt
 from typing import Optional
+from absl import logging
 # Style gallery: https://tonysyu.github.io/raw_content/matplotlib-style-gallery/gallery.html
 # The seaborn styles shipped by Matplotlib are deprecated since 3.6,
 # as they no longer correspond to the styles shipped by seaborn.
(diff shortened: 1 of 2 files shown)


alf/summary/render.py
```diff
@@ -175,7 +181,8 @@ def pack_image_nest(cls, imgs):
         # first get all images' sizes (w,h)
         sizes = [(i.shape[1], i.shape[0]) for i in imgs]
         # call rpack for an approximate solution: [(x,y),...] positions
-        positions = rpack.pack(sizes)
+        positions = rpack.pack(
+            sizes, max_width=max_width, max_height=max_height)
```
hnyu: What if there is no such a solution given the max_width and max_height? Will that crash the function?

QuantuMope: In that case, a `rpack.PackingImpossibleError` will be thrown. Do you prefer having the user handle the error or catching this and rerunning without limits perhaps?

hnyu: > In that case, a `rpack.PackingImpossibleError` will be thrown. Do you prefer having the user handle the error or catching this and rerunning without limits perhaps?

I think the second option is better.

QuantuMope: Changed the behavior to catch the error and rerun with constraints dropped. A warning is also printed when this occurs. PTAL.

hnyu (approved):

breakds (approved):

PR #1650, opened 2024-05-29
Use GPU before loading environment

emailweixu:
For some environment, it takes very long to load the environment. This PR change the code to use GPU before loading environments so `cluster_train` idle will not show those GPUs.

### DIFF
--- a/alf/environments/utils.py
+++ b/alf/environments/utils.py
@@ -16,6 +16,7 @@ import functools
 import inspect
 import numpy as np
 import random
+import torch
 
 import alf
 from alf.environments import suite_gym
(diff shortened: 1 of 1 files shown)


alf/environments/utils.py
```diff
@@ -173,6 +174,10 @@ def create_environment(env_name='CartPole-v0',
 
     """
 
+    # Use GPU before loading environments so cluster_train idle will not show
+    # the GPU
+    tmp = torch.zeros((32, ))
```
hnyu, line 179: I understand the motivation of this. But the comment seems unrelated to ALF?

emailweixu: Updated the comment

hnyu (approved):

PR #1651, opened 2024-05-29
Option for disabling AMP for NormalProjectionNetwork

emailweixu:
When AMP is enabled for training, fp16 may not be enough for NormalProjectionNetwork. Hence
introduce the option to disable AMP for it.

### DIFF
--- a/alf/layers.py
+++ b/alf/layers.py
@@ -21,6 +21,7 @@ import numpy as np
 import torch
 import torch.nn as nn
 import torch.nn.functional as F
+import torch.distributions as td
 from torch import Tensor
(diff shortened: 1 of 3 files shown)


hnyu:
Can you explain why fp16 is not enough for NormalProjectionNetwork?

emailweixu:
> Can you explain why fp16 is not enough for NormalProjectionNetwork?

when calculating the logprob, it needs to compute (x-mean)^2/sigma^2.  The result can be not accurate enough if sigma is small compare to mean.

hnyu (approved):

PR #1652, opened 2024-05-29
Fix DDP for on-policy training

emailweixu:
All the steps for computing loss needs to be wrapped in  @data_distributed_when.
Previously, only unroll() is wrapped. Now merge unroll() and loss computation to one function which can be wrapped by  @data_distributed_when

### DIFF
--- a/alf/algorithms/rl_algorithm.py
+++ b/alf/algorithms/rl_algorithm.py
@@ -27,7 +27,8 @@ from alf.algorithms.algorithm import Algorithm
 from alf.algorithms.async_unroller import AsyncUnroller
 from alf.experience_replayers.replay_buffer import ReplayBuffer
 from alf.data_structures import (AlgStep, Experience, make_experience,
-                                 TimeStep, BasicRolloutInfo, BasicRLInfo)
+                                 StepType, TimeStep, BasicRolloutInfo,
+                                 BasicRLInfo)
 from alf.utils import common, dist_utils, summary_utils
 from alf.utils.summary_utils import record_time
 from alf.utils.distributed import data_distributed_when
@@ -509,7 +510,6 @@ class RLAlgorithm(Algorithm):
         return policy_step
 
     @common.mark_rollout
-    @data_distributed_when(lambda algorithm: algorithm.on_policy)
     def unroll(self, unroll_length: int):
         if self._config.async_unroll:
             return self._async_unroll(unroll_length)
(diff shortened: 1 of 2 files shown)


hnyu:
Surprised that we didn't use DDP for on-policy before?

alf/algorithms/rl_algorithm.py
```diff
@@ -715,20 +715,35 @@ def train_iter(self):
...
+        """User may override this for their own training procedure."""
+        alf.summary.increment_global_counter()
+
+        train_info, loss_info, experience = self._compute_train_info_and_loss_info_on_policy(
+            self._config.unroll_length)
+
         with record_time("time/train"):
-            train_info = experience.rollout_info
-            experience = experience._replace(rollout_info=())
-            steps = self.train_from_unroll(experience, train_info)
```
breakds, line 730: This seems to be the only place that `train_from_unroll` is used. Since now we are breaking it up, is there still anything preventing us from deprecating it?

alf/algorithms/rl_algorithm.py
```diff
@@ -715,20 +715,35 @@ def train_iter(self):
...
+    @data_distributed_when(lambda algorithm: algorithm.on_policy)
+    def _compute_train_info_and_loss_info_on_policy(self, unroll_length):
         with record_time("time/unroll"):
             with torch.cuda.amp.autocast(self._config.enable_amp):
                 experience = self.unroll(self._config.unroll_length)
             self.summarize_metrics()
 
+        train_info = experience.rollout_info
+        experience = experience._replace(rollout_info=())
+        loss_info = self.calc_loss(train_info)
```
breakds, line 727: From profiling point of view, this used to be under `time/train`. Probably negligible but may deserve letting everyone know about this change.

breakds (approved):

breakds:
> Surprised that we didn't use DDP for on-policy before?

I also had this question. My understanding is that this is not a problem if all learnable parameters are only used by `unroll()`. The backward hook is placed on the output of `unroll`'s output, and during backward of the loss when that output is being processed, synchronization is triggered.

breakds:
Thanks for catching this and sorry about the overlook!

alf/algorithms/rl_algorithm.py
emailweixu: Yes, calc_loss is typically negligible. And we usually don't use on-policy.

PR #1653, opened 2024-05-29
Unroll with AMP for off policy training

emailweixu:
Also
1. Fixed action dtype when amp is enabled. Environment needs float32 instead of float16
2. Fix dtype mismatch for storing batch to replay buffer when AMP is enabled.

### DIFF
--- a/alf/layers.py
+++ b/alf/layers.py
@@ -3691,7 +3691,7 @@ def to_float32(nested):
     Args:
         nested (nested Tensor): a nest of tensors
     Returns:
-        nested Tensor: a nest of tensors with dtype torch.float32
+        nested Tensor: a nest of tensors/distributions with dtype torch.float32
     """
 
     def _to_float32(x):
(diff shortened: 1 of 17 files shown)


Haichao-Zhang (dismissed):

alf/layers.py
```diff
@@ -3683,6 +3685,31 @@ def _make_spec(spec):
     return map_structure(_make_spec, specs)
 
 
+def to_float32(nested):
+    """Change the dtype of all the tensors in ``nested`` to torch.float32.
+
+    Args:
+        nested (nested Tensor): a nest of tensors
+    Returns:
+        nested Tensor: a nest of tensors with dtype torch.float32
```
Haichao-Zhang: ``Returns:
        nested Tensor``: when ``x`` is an instance of ``td.Distribution``, the return type is ``td.Distribution``?

Haichao-Zhang (dismissed):

Haichao-Zhang (approved):

hnyu:
This `with (torch.set_grad_enabled(config.unroll_with_grad),
                  torch.cuda.amp.autocast(self._config.enable_amp))` with syntax is only available in higher version torch.

PR #1655, opened 2024-05-30
Use running statistics to normalize advantages for PPO and AC

emailweixu:
For some tasks, there are rare but importance events which may not appear in one batch. In these situations, the advantage normalization based on one single batch may not correctly normalize the advantage using the desired statistics. This PR changes the normalization based on the EMA statistics of the past batches.

Also add support for normalizing the weighted sum of the advantage instead of normalization for individual reward dimensions.

### DIFF
--- a/alf/algorithms/actor_critic_loss.py
+++ b/alf/algorithms/actor_critic_loss.py
@@ -104,8 +104,6 @@ class ActorCriticLoss(Loss):
         self._use_gae = use_gae
         self._lambda = td_lambda
         self._use_td_lambda_return = use_td_lambda_return
-        assert not use_td_lambda_return or use_gae, (
-            "use_td_lambda_return requires use_gae=True!")
         if normalize_scalar_advantages:
             self._adv_norm = torch.nn.BatchNorm1d(
                 num_features=1,
--- a/alf/algorithms/actor_critic_loss.py
+++ b/alf/algorithms/actor_critic_loss.py
@@ -32,7 +32,7 @@ def normalize(batch_norm, x):
     batch_norm.train()
     momentum = batch_norm.momentum
     if batch_norm.num_batches_tracked * momentum < 1.0:
-        # First the first few batches, we do cumulative moving average
+        # For the first few batches, we do cumulative moving average
         batch_norm.momentum = None
     batch_norm(x)
     batch_norm.momentum = momentum
--- a/alf/algorithms/actor_critic_loss.py
+++ b/alf/algorithms/actor_critic_loss.py
@@ -28,15 +28,33 @@ ActorCriticLossInfo = namedtuple("ActorCriticLossInfo",
                                  ["pg_loss", "td_loss", "neg_entropy"])
 
 
+def normalize(batch_norm, x):
+    batch_norm.train()
+    momentum = batch_norm.momentum
+    if batch_norm.num_batches_tracked * momentum < 1.0:
+        # First the first few batches, we do cumulative moving average
+        batch_norm.momentum = None
+    batch_norm(x)
+    batch_norm.momentum = momentum
+    # We use the running mean and variance of the advantages to normalize
+    # since the batch may not be large enough to properly normalize within
+    # the batch.
+    batch_norm.eval()
+    return batch_norm(x)
+
+
 @alf.configurable
 class ActorCriticLoss(Loss):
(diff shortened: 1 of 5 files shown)


alf/algorithms/actor_critic_loss.py
```diff
@@ -28,15 +28,33 @@
...
+    # since the batch may not be large enough to properly normalize within
+    # the batch.
+    batch_norm.eval()
+    return batch_norm(x)
+
+
 @alf.configurable
 class ActorCriticLoss(Loss):
     def __init__(self,
+                 reward_dim=1,
```
breakds, line 49: Need to update the docstring for `reward_dim`.

alf/algorithms/actor_critic_loss.py
```diff
@@ -67,6 +85,11 @@ def __init__(self,
             normalize_advantages (bool): If True, normalize advantage to zero
                 mean and unit variance within batch for caculating policy
                 gradient. This is commonly used for PPO.
+            normalize_scalar_advantages (bool): If False, the normalization is
```
breakds, line 89: I think it is not very clear to the reader about the relationship between `normalize_advantages` and `normalize_scalar_advantages`, especially when both of them are `True`. 

In `__init__` it seems that `normalize_scalar_advantages` supersedes `normalize_advantages`, but in `forward` it seems to suggest the other way around.

alf/algorithms/actor_critic_loss.py
emailweixu: updated.

alf/algorithms/actor_critic_loss.py
emailweixu: updated dod-string

alf/algorithms/actor_critic_loss.py
```diff
@@ -81,13 +106,23 @@ def __init__(self,
         self._use_gae = use_gae
         self._lambda = td_lambda
         self._use_td_lambda_return = use_td_lambda_return
+        if normalize_scalar_advantages:
```
hnyu, line 109: For me, I feel it's more natural that `normalize_advantages=True` is a pre-condition for using `normalize_scalar_advantages` which basically determines how we are going to normalize the advantage: either for individual reward dims or for the combined scalar.

alf/algorithms/actor_critic_loss.py
```diff
@@ -28,15 +28,33 @@
                                  ["pg_loss", "td_loss", "neg_entropy"])
 
 
+def normalize(batch_norm, x):
+    batch_norm.train()
+    momentum = batch_norm.momentum
+    if batch_norm.num_batches_tracked * momentum < 1.0:
+        # First the first few batches, we do cumulative moving average
```
hnyu: First -> For

alf/algorithms/actor_critic_loss.py
emailweixu: I feel it's cumbersome to need to specify both normalize_advantages=True normalize_scalar_advantages=True

alf/algorithms/actor_critic_loss.py
emailweixu: fixed

hnyu (approved):

alf/algorithms/actor_critic_loss.py
```diff
@@ -81,13 +106,23 @@ def __init__(self,
...
+                affine=False,
+                track_running_stats=True)
+            normalize_advantages = False
+        elif normalize_advantages:
+            self._adv_norm = torch.nn.BatchNorm1d(
+                num_features=reward_dim,
+                eps=1e-8,
+                momentum=advantage_norm_momentum,
+                affine=False,
+                track_running_stats=True)
```
runjerry, line 123: I found that changing this to True could lead to a mismatch between the layer and the ckpt which does not have keys ``running_mean`` and ``running_var``. I wonder if we might want to keep it False and config in conf files instead?

emailweixu: You can try the following:
alf.config('Checkpointer.load', strict=False)

runjerry: I gave it a try in both the agent and expert conf files, it does not work out-of-the-box, will dig further to figure out why.

PR #1656, opened 2024-05-31
Add config_only_args arg to alf.configurable to prevent local config changes

QuantuMope:
This PR adds the `config_only_args` arg to the `alf.configurable` decorator as well as a unit test.

The functionality can be described as:
1. Any arg specified in this list can only be changed via `alf.config` and will otherwise raise a `ValueError`. 
2. Any entry in `config_only_args` cannot also be found in `blacklist`, otherwise a `ValueError` is raised.

This will allow users to eliminate possible mismatches in parameters that may arise from local config changes using `partial()`.

### DIFF
--- a/alf/config_util.py
+++ b/alf/config_util.py
@@ -528,9 +528,10 @@ def _make_wrapper(fn, configs, signature, has_self, config_only_args):
             signature does not contains parameter for self. This should be True
             if fn is __init__() function of a class.
         config_only_args (list[str]): list of args that should be guarded. In other
-            words, their values can only be set / changed globally via alf.config().
-            This protects against the case where local changes via partial() can
-            cause unintended side effects.
+            words, their values can only be set / changed globally via ``alf.config()``.
+            This protects against local untracked changes as a result of 1) using
+            ``partial()`` or 2) setting an argument when calling the function, which
+            can cause unintended side effects.
     Returns:
         The wrapped function
     """
@@ -563,8 +564,7 @@ def _make_wrapper(fn, configs, signature, has_self, config_only_args):
                 config.set_used()
 
         for config_only_arg in config_only_args:
-            if config_only_arg in set_positional_args or config_only_arg in kwargs.keys(
-            ):
+            if config_only_arg in set_positional_args or config_only_arg in kwargs:
                 raise ValueError(
                     f"The arg '{config_only_arg}' of {fn.__qualname__} is guarded but has been modified. "
                     f"Most likely partial() was used to change this value, which is not allowed."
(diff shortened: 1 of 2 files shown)


alf/config_util.py
```diff
@@ -557,13 +561,21 @@ def _wrapper(*args, **kwargs):
...
+                    )
+
         return fn(*args, *unspecified_positional_args, **kwargs,
                   **unspecified_kw_args)
 
     return _wrapper
 
 
-def _decorate(fn_or_cls, name, whitelist, blacklist):
+def _decorate(fn_or_cls, name, whitelist, blacklist, guard_kwargs):
```
hnyu: It's not actually 'kwargs'? It's better that we also have this guard for positional arg?

alf/config_util.py
```diff
@@ -557,13 +561,21 @@ def _wrapper(*args, **kwargs):
                     unspecified_kw_args[name] = config.get_value()
                 config.set_used()
 
+        for key in guard_kwargs:
+            if key in kwargs and key in configs:
```
hnyu: We can just directly assert that any key in guard_kwargs shouldn't appear in kwargs or args? (i.e., it has to be in `configs`)

hnyu (changes requested):

alf/config_util.py
```diff
@@ -576,6 +588,10 @@ def _decorate(fn_or_cls, name, whitelist, blacklist):
         blacklist (list[str]): A blacklisted set of kwargs that should not be
             configurable. All other kwargs will be configurable. Only one of
             ``whitelist` or ``blacklist`` should be specified.
+        guard_kwargs (list[str]): list of kwargs that should be guarded. In other
```
emailweixu: The name `guard_kwargs` does really convey the meaning. How about  `config_only_args`?

alf/config_util.py
emailweixu: +1

QuantuMope: Done.

alf/config_util.py
QuantuMope: I've expanded the feature for positional args as well.

alf/config_util.py
QuantuMope: Good idea. Changed.

QuantuMope (commented):
Addressed comments. PTAL.

alf/config_util.py
```diff
@@ -527,6 +527,10 @@ def _make_wrapper(fn, configs, signature, has_self):
         has_self (bool): whether the first argument is expected to be self but
             signature does not contains parameter for self. This should be True
             if fn is __init__() function of a class.
+        config_only_args (list[str]): list of args that should be guarded. In other
+            words, their values can only be set / changed globally via alf.config().
+            This protects against the case where local changes via partial() can
+            cause unintended side effects.
```
hnyu: This also protects the case where a value is specified by the code when calling the function?

alf/config_util.py
```diff
@@ -557,13 +562,21 @@ def _wrapper(*args, **kwargs):
                     unspecified_kw_args[name] = config.get_value()
                 config.set_used()
 
+        for config_only_arg in config_only_args:
+            if config_only_arg in set_positional_args or config_only_arg in kwargs.keys(
```
hnyu: Can be directly `config_only_arg in kwargs`

hnyu (commented):
Only some minors comments

alf/config_util.py
QuantuMope: Right. Updated comments to mention this case as well.

alf/config_util.py
QuantuMope: Changed.

QuantuMope (commented):
Addressed comments. PTAL.

hnyu (approved):

PR #1657, opened 2024-06-04
Properly handle the required 'step' in optimizer states

Haichao-Zhang:
Properly handle the required 'step' in optimizer states in higher version torch.

### DIFF
--- a/alf/optimizers/optimizers.py
+++ b/alf/optimizers/optimizers.py
@@ -509,9 +509,6 @@ def wrap_optimizer(cls):
 
     @common.add_method(NewCls)
     def __setstate__(self, state):
-        def _move_key_as_the_first(d, key):
-            return {key: d.pop(key)}
-
         for param_group in self.param_groups:
             for p in param_group['params']:
                 p_state = self.state[p]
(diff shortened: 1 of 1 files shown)


alf/optimizers/optimizers.py
```diff
@@ -501,6 +507,28 @@ def load_state_dict(self, state_dict):
...
+
+        for param_group in self.param_groups:
+            for p in param_group['params']:
+                p_state = self.state[p]
+                if 'step' not in p_state:
+                    # optimizers in higher version torch requires 1) the presence of 'step'
+                    # in state, and 2) it appears as the first key of the state dictionary.
+                    # Therefore we explicitly create it if it does not exist and make it
+                    # the key that comes first
+                    p_state.update({'step': 0})
```
emailweixu: Is this needed by all optimizers?

Haichao-Zhang: Some of the optimizers have this requirement, such Adam, AdamW, ASGD. Here we handle it in a general way without differentiating different variations while without too much additional overhead. Updated comment

alf/optimizers/optimizers.py
```diff
@@ -501,6 +507,28 @@ def load_state_dict(self, state_dict):
...
+                    # Some of the optimizers such as Adam/AdamW/ASGD etc in higher version
+                    # torch require 1) the presence of 'step' in state, and 2) it appears
+                    # as the first key of the state dictionary. Therefore we explicitly
+                    # create it if it does not exist and make it the key that comes first
+                    p_state.update({'step': 0})
+                    p_state = {
+                        **_move_key_as_the_first(p_state, 'step'),
+                        **p_state
+                    }
+                    self.state[p] = p_state
```
emailweixu: can be simplified as: `self.state[p] = {'step': 0, **p_state}`

Haichao-Zhang: Simplified. Thanks for the suggestion.

emailweixu (approved):

PR #1658, opened 2024-06-04
fix random seed

hnyu:
This PR addresses two issues.
1. Currently if we don't specify a random seed for training, the tensorboard will record the random seed as None. This prevents us reproducing the training job using the actual random seed. So we need to explicitly generate a random seed for ddp_rank=0 or -1. 
2. For spawned env, we need to explicitly pass TrainerConfig.random_seed in case we need it inside the spawned env, to control the envs group behavior based on the global training random seed.

### DIFF
--- a/alf/environments/process_environment.py
+++ b/alf/environments/process_environment.py
@@ -300,12 +300,21 @@ class ProcessEnvironment(object):
         ddp_num_procs = PerProcessContext().num_processes
         ddp_rank = PerProcessContext().ddp_rank
 
+        pre_configs = alf.get_handled_pre_configs()
+        if self._start_method == 'spawn':
+            # If we spawn an env, TrainerConfig.random_seed of the main processs
+            # won't be inherited. In case we need this info later in the env, we
+            # need to pass it explicitly, so that later we can do
+            # ``alf.get_config_value("TrainerConfig.random_seed")``.
+            pre_configs.append(
+                ('TrainerConfig.random_seed',
+                 alf.get_config_value('TrainerConfig.random_seed')))
         self._process = mp_ctx.Process(
             target=_worker,
-            args=(conn, self._env_constructor, self._start_method,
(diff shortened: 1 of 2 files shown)


emailweixu:
Why not just set the random seed in config if reproducing is needed?

alf/environments/process_environment.py
```diff
@@ -300,12 +300,21 @@ def start(self, wait_to_start=True):
...
         self._process = mp_ctx.Process(
             target=_worker,
-            args=(conn, self._env_constructor, self._start_method,
-                  alf.get_handled_pre_configs(), self._env_id, self._flatten,
-                  self._fast, self._num_envs, self._torch_num_threads,
-                  ddp_num_procs, ddp_rank, self._name),
+            args=(conn, self._env_constructor, self._start_method, pre_configs,
+                  self._env_id, self._flatten, self._fast, self._num_envs,
+                  self._torch_num_threads, ddp_num_procs, ddp_rank,
+                  self._name),
```
emailweixu, line 317: the seed of the spawned environment is already handled by _env_constructor() in utils.py

hnyu:
> Why not just set the random seed in config if reproducing is needed?

Sometimes when we trained a job we forgot to set the seed but later we want to reproduce?

alf/environments/process_environment.py
hnyu: > the seed of the spawned environment is already handled by _env_constructor() in utils.py

Yeah I know that each env will get its own seed set in create_environment. But here it's setting the global seed for each DDP rank, which cannot be easily obtained by each env later.

breakds: I think the idea here is just to set the random seed in the configuration of the spawned environment process, so that within the spawned environment process `TrainerConfig.random_seed` can be accessed to get the "per rank seed".

breakds (approved):

PR #1661, opened 2024-06-21
Add sole_config function for enforced one-time configuring

QuantuMope:
This PR adds a new `sole_init` flag in the original `alf.config` along with a simple wrapper function, `alf.sole_config`.

If the flag is set to true, the config value can only be set once, with any previous or future calls raising a `ValueError`. This is helpful in enforcing a singular point of initialization, thus eliminating any potential side effects from possible prior or future overrides. This is especially a problem when working deeply nested `alf.config` calls.

Before, a somewhat similar behavior could be achieved through the `mutable` flag, but this flag would only raise a warning, which can easily get drowned out in logs.

### DIFF
--- a/alf/config_util.py
+++ b/alf/config_util.py
@@ -37,6 +37,7 @@ __all__ = [
     'load_config',
     'pre_config',
     'reset_configs',
+    'sole_config',
     'validate_pre_configs',
     'repr_wrapper',
     'save_config',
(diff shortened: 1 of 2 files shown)


alf/config_util.py
```diff
@@ -306,6 +344,11 @@ def config1(config_name, value, mutable=True, raise_if_used=True):
             immutable value to an existing immutable value.
         raise_if_used (bool): If True, ValueError will be raised if trying to
             config a value which has already been used.
+        sole_init (bool): If True, the config value can only be set once. Any
+            previous or future calls will raise a ValueError. This is helpful
+            in enforcing a singular point of initialization, thus eliminating
+            any potential side effects from possible prior or future overrides.
+            This flag overrides the mutable flag.
```
hnyu: This flag only overwrites the mutable flag if it's True?

QuantuMope: Right. Good catch. I've clarified this.

hnyu (approved):

PR #1663, opened 2024-07-01
Add override_config logic to support student-teacher training setups

QuantuMope:
Please review PR #1662 before this one.

This PR adds `alf.override_config` which is a wrapper for `alf.config(..., override_all=True)`.

This function will let users set a config value regardless of the protected flags on it (e.g., `mutable` or `sole_init`). The use of this function should be **heavily discouraged** and used only when absolutely necessary (e.g., a student-teacher learning setup where the student must override some teacher inherited settings). Note that when the `override_all` flag is True, only the value of the config is changed. In other words, the previous `mutable` and `sole_init` settings are unchanged.

### DIFF
--- a/alf/config_util.py
+++ b/alf/config_util.py
@@ -35,6 +35,7 @@ __all__ = [
     'get_operative_configs',
     'import_config',
     'load_config',
+    'override_config',
     'pre_config',
     'reset_configs',
     'validate_pre_configs',
@@ -59,6 +60,7 @@ def config(prefix_or_dict,
            mutable=True,
            raise_if_used=True,
            sole_init=False,
+           override_all=False,
            **kwargs):
     """Set the values for the configs with given name as suffix.
 
@@ -117,6 +119,13 @@ def config(prefix_or_dict,
             This flag overrides the mutable flag if True. For users wanting this
             to be the default behavior, the ALF_SOLE_CONFIG env variable can be
             set to 1.
+        override_all (bool): If True, the value of the config will be set regardless
+            of any preeixsting mutable or sole_init settings. This should be used
+            only when absolutely necessary (e.g., a teacher-student training loop,
+            where the student must override certain configs inheritted from the
+            teacher). Otherwise, use ``mutable`` or ``sole_init`` instead.
+            NOTE: if override_all is True, only the value of the config will be
+            set. In other words, the mutable and sole_init flags will be ignored.
         **kwargs: only used if ``prefix_or_dict`` is a str.
     """
     if isinstance(prefix_or_dict, str):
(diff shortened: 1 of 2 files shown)


alf/config_util.py
```diff
@@ -117,6 +119,13 @@ def func(self, a, b):
             This flag overrides the mutable flag if True. For users wanting this
             to be the default behavior, the ALF_SOLE_CONFIG env variable can be
             set to 1.
+        override_all (bool): If True, the value of the config will be set regardless
+            of any preeixsting mutable or sole_init settings. This should be used
+            only when absolutely necessary (e.g., a teacher-student training loop,
+            where the student must override certain configs inheritted from the
+            teacher). Otherwise, use ``mutable`` or ``sole_init`` instead.
+            NOTE: if override_all is True, only the value of the config will be
```
emailweixu: I don't quite understand this comment. Why "only"? What is the possible alternative?
And why is "_all" suffix needed? What does "all" mean here?

QuantuMope: The other alternative is that `alf.config(..., override_all=True)` will also set the `mutable` and `sole_init` flags of that particular config along with the value. This is probably not what the user wants though. As an example, imagine that a user sets a config value to be immutable. The user then wants to one-time override that value by calling `alf.config(..., override_all=True)`. This call would set `mutable=True` by default, which has now changed that config's original setting. To me, `alf.config(..., override_all=True)` should only be for changing the value of a config and respect other parameters since it's used in very nuanced situations. The other flipside is that we allow the call to modify `mutable` and `sole_init`, but if the user wanted to respect the previous settings, it would require them to somehow find them out.

The `_all` suffix is meant to mean that we change the value of a config regardless of "all" of its protections: `mutable` and `sole_init`.

emailweixu: > The other alternative is that `alf.config(..., override_all=True)` will also set the `mutable` and `sole_init` flags of that particular config along with the value. This is probably not what the user wants though. As an example, imagine that a user sets a config value to be immutable. The user then wants to one-time override that value by calling `alf.config(..., override_all=True)`. This call would set `mutable=True` by default, which has now changed that config's original setting. To me, `alf.config(..., override_all=True)` should only be for changing the value of a config and respect other parameters since it's used in very nuanced situations. The other flipside is that we allow the call to modify `mutable` and `sole_init`, but if the user wanted to respect the previous settings, it would require them to somehow find them out.

How about just explicitly comment that overrid_all=True does not change the value of mutable or sole_init? This is clearer to me.

> 
> The `_all` suffix is meant to mean that we change the value of a config regardless of "all" of its protections: `mutable` and `sole_init`.

`_all` suffix seems unnecessary.  "override" would mean the same thing.

QuantuMope: Sounds good. Done.

emailweixu (approved):

PR #1666, opened 2024-07-16
[Small] Fix minor bugs and errors in dist_utils.py

QuantuMope:
This PR simply cleans up some mistakes in `dist_utils.py`, such as spelling mistakes, incorrect arg types in docstrings, and a minor bug in a nested function arg.

### DIFF
--- a/alf/utils/dist_utils.py
+++ b/alf/utils/dist_utils.py
@@ -654,14 +654,14 @@ def _get_transform_builders_params(transforms):
         return transform.__class__
 
     def _get_transform_params(transform):
-        if hasattr(transforms, 'params') and transforms.params is not None:
+        if hasattr(transform, 'params') and transform.params is not None:
             # We assume that if a td.Transform has attribute 'params', then they are the
             # parameters we'll extract and store.
             assert isinstance(
-                transforms.params,
+                transform.params,
                 dict), ("Transform params must be provided as a dict! "
-                        f"Got {transforms.params}")
-            return transforms.params
+                        f"Got {transform.params}")
+            return transform.params
         return {}  # the transform doesn't have any parameter
 
     if isinstance(transforms, td.Transform):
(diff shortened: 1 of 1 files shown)


alf/utils/dist_utils.py
```diff
@@ -1215,14 +1220,14 @@ def get_base_dist(dist):
         NotImplementedError: if ``dist`` or its based distribution is not
             ``td.Normal``, ``td.Independent`` or ``td.TransformedDistribution``.
     """
-    if isinstance(dist, (td.Normal, td.Categorical, StableCauchy, Beta,
-                         TruncatedDistribution)):
+    dist_type = type(dist)
+    if dist_type in _base_dist_types:
         return dist
```
hnyu: This is not right? The base dist of a Diag is not itself. This function serves to find the lowest-level distribution.

QuantuMope: I guess I misunderstood the purpose of this function then? To me the base dist should be the outermost non-transformed distribution. 

Otherwise, by that same logic, if this function is trying to find the lowest-level (in terms of inheritance or attribute) distribution, then why would it stop at `StableCauchy` and not continue onto `td.Cauchy`?

emailweixu: base_dist is the innermost wrapped distribution.

QuantuMope: I've reverted this and changed the purpose of this PR to the arg bug fix and minor fixes.

alf/utils/dist_utils.py
```diff
@@ -654,14 +654,14 @@ def _get_transform_builder(transform):
         return transform.__class__
 
     def _get_transform_params(transform):
-        if hasattr(transforms, 'params') and transforms.params is not None:
+        if hasattr(transform, 'params') and transform.params is not None:
```
emailweixu, line 657: This bug seems to be a serious one that affect the training. @hnyu, does your previous experiments on normalized flow depend on this?

QuantuMope: It's my understanding that this is more of a silent bug as the logic of the function does not technically change.

hnyu: > This bug seems to be a serious one that affect the training. @hnyu, does your previous experiments on normalized flow depend on this?

I also believe it's just a minor bug. If you look at the code, when _get_transform_params is called, `transform` is always the same with `transforms`

hnyu (approved):

PR #1667, opened 2024-07-16
extra affine transform params

hnyu:
Extract the params for AffineTransform. This could be useful if we want to implement a function to split a distribution at an index, in which case we need to split the affine params.

### DIFF
--- a/alf/utils/dist_utils.py
+++ b/alf/utils/dist_utils.py
@@ -70,6 +70,11 @@ class AffineTransform(get_invertible(td.AffineTransform)):
     compatible with ``DistributionSpec.build_distribution()``.
     """
 
+    def __init__(self, loc, scale, event_dim=0, cache_size=0):
+        loc = torch.as_tensor(loc)
+        scale = torch.as_tensor(scale)
+        super().__init__(loc, scale, event_dim, cache_size)
+
     def get_builder(self):
         return functools.partial(
             AffineTransform, loc=self.loc, scale=self.scale)
@@ -472,8 +477,6 @@ class AffineTransformedDistribution(td.TransformedDistribution):
             loc (Tensor or float): Location parameter.
             scale (Tensor or float): Scale parameter.
         """
-        loc = torch.as_tensor(loc)
-        scale = torch.as_tensor(scale)
         super().__init__(
             base_distribution=base_dist,
             transforms=AffineTransform(loc, scale))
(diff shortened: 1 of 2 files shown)


alf/utils/dist_utils.py
```diff
@@ -468,6 +472,8 @@ def __init__(self, base_dist: td.Distribution, loc, scale):
             loc (Tensor or float): Location parameter.
             scale (Tensor or float): Scale parameter.
         """
+        loc = torch.as_tensor(loc)
+        scale = torch.as_tensor(scale)
```
QuantuMope: I think you need to add this to `AffineTransform` as well in case a scalar float is passed.

QuantuMope (commented):
Hey Haonan, thanks for the feature. Just one comment.

alf/utils/dist_utils.py
hnyu: > I think you need to add this to `AffineTransform` as well in case a scalar float is passed.

Right. Updated.

alf/utils/dist_utils.py
```diff
@@ -695,16 +701,19 @@ def _get_transformed_builder(obj: td.TransformedDistribution):
...
     builder, params = _get_builder(obj.base_dist)
-    new_builder = functools.partial(_builder_affine_transformed, builder,
-                                    obj.loc, obj.scale)
-    return new_builder, params
+    new_builder = functools.partial(_builder_affine_transformed, builder)
+    new_params = {
+        "params_": params,
+        "transforms_params_": obj.transforms[0].params
+    }
+    return new_builder, new_params
```
emailweixu, line 719: What is the reason for making this change?

hnyu: > What is the reason for making this change?

Need an easy way to extract all transform params, including the special AffineTransformedDistribution. @QuantuMope is writing the code to split an action distribution into two for the one-bit movement PR. That one-bit is merged into the agent's action space for the policy. When later we split a distribution params, we also need to split its transform's params.

alf/utils/dist_utils.py
```diff
@@ -70,6 +70,11 @@ class AffineTransform(get_invertible(td.AffineTransform)):
     compatible with ``DistributionSpec.build_distribution()``.
     """
 
+    def __init__(self, loc, scale, event_dim=0, cache_size=0):
```
QuantuMope, line 73: Nit: If we never use `event_dim` and `cache_size`, perhaps we can just remove them from the arg list. That way, we'll always use pytorch's defaults?

alf/utils/dist_utils.py
emailweixu: Shouldn't the original code for this part still work?

hnyu:
Closing this PR as I realized that the extracted transform params don't have a batch dim. When later accumulating training distributions (params) along time axis, this will cause errors. Not sure about an easy way to fix this.

PR #1668, opened 2024-07-22
multi-nodes multi-gpus training

sqchai:


alf/bin/train.py
```diff
@@ -77,12 +77,13 @@ def _define_flags():
     flags.DEFINE_bool(
         'force_torch_deterministic', True,
         'torch.use_deterministic_algorithms when random_seed is set')
-    flags.DEFINE_bool('store_snapshot', True,
+    flags.DEFINE_bool('store_snapshot', False,
```
emailweixu: why change this?

alf/environments/process_environment.py
```diff
@@ -475,4 +479,4 @@ def render(self, mode='human'):
         Raises:
             NotImplementedError: If the environment does not support rendering.
         """
-        return self.call('render', mode)()
+        return self.call('render', mode)()
```
emailweixu: add new line

alf/environments/process_environment.py
```diff
@@ -150,8 +152,9 @@ def _worker(conn: multiprocessing.connection,
             env = alf.get_env()
         else:
             env = env_constructor(env_id=env_id)
-        if not alf.get_config_value("TrainerConfig.sync_progress_to_envs"):
-            disallow_scheduler()
+        #TODO fix this disallow_scheduler in ddp context
+        # if not alf.get_config_value("TrainerConfig.sync_progress_to_envs"):
+        #     disallow_scheduler()
```
emailweixu: what is the problem with the original code?

alf/bin/train.py
```diff
@@ -239,23 +292,48 @@ def main(_):
...
-                training_worker(0, world_size, conf_file, root_dir,
-                                paras_queue)
-                context.join()
+        except KeyboardInterrupt:
+            pass
+        except Exception as e:
+            # ``e`` has been printed in the subprocess, so here we won't print it
+            # again. But we raise another error so that we will have a correct
+            # exit code for the program.
+            raise ChildProcessError(f'Training failed on subprocess exception')
```
emailweixu: What are reasons for making these changes?

The previous code make sure that rank-0 is the main process, which is necessary for debugging.

alf/bin/train.py
```diff
@@ -239,23 +292,48 @@ def main(_):
...
+        
+    elif FLAGS.distributed == 'multi-node-multi-gpu':
```
emailweixu, line 327: Can you add documentation on how multi-node-multi-gpu mode is supposed to be used, i.e. how to start a multi-node-multi-gpu job?

And how do the processes on different machines know how to communicate with the other machines?

emailweixu (changes requested):
Also please fix the code style following https://github.com/HorizonRobotics/alf/blob/pytorch/docs/contributing.rst

alf/bin/train.py
sqchai: Fixed 'store_snapshot'. Previously each node will try to write a snapshot to the same root directory in multi-node multi-gpu mode. Restored 'store_snapshot' default value to True. Modified such that only node rank = 0 process will write a snapshot.

alf/bin/train.py
sqchai: Previous change to train.py on line 255 to line 311 (multi-gpu part) is not necessary. Restored to original code.

alf/environments/process_environment.py
sqchai: For the 'disallow_scheduler' part in 'process_environment.py'. For some reasons my previous local code base do not have this 'disallow_scheduler' function, and does not import this function in the script. Therefore, this part of code always fails. I fixed those things according to the latest alf implementation.

PR #1669, opened 2024-07-22
Add argument to optionally create environment when parsing conf files

Haichao-Zhang:
The option is useful in the cases when we only want to parse conf files but does not want to (re-)create environments and/or set random seeds.

For example, this can be used in the async mode of hobot project, where we want to parse the confs but does not want to touch anything related to environments.

### DIFF
--- a/alf/config_helpers.py
+++ b/alf/config_helpers.py
@@ -192,7 +192,7 @@ def adjust_config_by_multi_process_divider(ddp_rank: int,
         config1('TrainerConfig.evaluate', False, raise_if_used=False)
 
 
-def parse_config(conf_file, conf_params):
+def parse_config(conf_file, conf_params, set_random_seed=True):
     """Parse config file and config parameters
 
     Note: a global environment will be created (which can be obtained by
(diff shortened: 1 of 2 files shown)


emailweixu:
Why is this necessary?

Haichao-Zhang:
> Why is this necessary?

The option is useful in the cases when we only want to parse conf files but does not want to (re-)create environments and/or set random seeds.

For example, this can be used in the async mode of hobot project, where we want to parse the confs but does not want to touch anything related to environments.

alf/config_helpers.py
```diff
@@ -192,7 +192,7 @@ def adjust_config_by_multi_process_divider(ddp_rank: int,
         config1('TrainerConfig.evaluate', False, raise_if_used=False)
 
 
-def parse_config(conf_file, conf_params):
+def parse_config(conf_file, conf_params, set_random_seed=True):
```
emailweixu: It seems to me create_env is better than set_random_seed.

Haichao-Zhang: Changed argname to create_env

emailweixu (approved):

PR #1670, opened 2024-07-23
Remove equality checks to allow for arbitrary config value datatypes

QuantuMope:
Previously, `alf.config` calls only went through if the value of the call differed from its previous value. This is checked using the equality operator. This check poses a problem for datatypes that do not implement the equality operator (e.g., numpy arrays), which results in an error. This PR removes the equality checks, which now allows for any arbitrary datatype to be a config value. 

The only logic change occurring from this is that previously duplicated `alf.config` calls setting the same value would be ignored. Now, they will trigger the logic of `mutable` or `raise_if_used` even if the value is identical. I believe this is preferable to the prior anyways, since users should be aware of exactly which point of code, they are setting a certain config.

### DIFF
--- a/alf/config_util.py
+++ b/alf/config_util.py
@@ -386,8 +386,7 @@ def config1(config_name,
     """
     config_node = _get_config_node(config_name)
 
-    if (raise_if_used and config_node.is_used()
-            and config_node.get_effective_value() != value):
+    if raise_if_used and config_node.is_used():
         raise ValueError(
             "Config '%s' has already been used. You should config "
             "its value before using it." % config_name)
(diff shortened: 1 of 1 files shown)


alf/config_util.py
```diff
@@ -386,8 +386,7 @@ def config1(config_name,
     """
     config_node = _get_config_node(config_name)
 
-    if (raise_if_used and config_node.is_used()
-            and config_node.get_effective_value() != value):
+    if raise_if_used and config_node.is_used():
```
breakds, line 389: So the behavior will be different but IMHO in a good way.

Just FYI there might be "config-after-use-but-with-same-value" lurking in the codebase previously, and this change will make them fail. If they exists it is most likely unintentional and we should fix them.

QuantuMope: Right, that's exactly what I'm thinking as well. I think this behavior is preferable and we should fix all unintentional calls in our codebase moving forward. Should be relatively quick.

breakds (approved):

Haichao-Zhang (approved):
Thanks for the change

PR #1679, opened 2024-08-01
add option "keep order" for py_pack_sequence_as

hnyu:
Add an option to keep the fields order when packing. This is useful if we want to pack a flattened list in the original order of the fields. For simplicity, this option is only for py version currently. The cnest version still doesn't support.

### DIFF
--- a/alf/nest/nest.py
+++ b/alf/nest/nest.py
@@ -268,12 +268,13 @@ def is_unnamedtuple(value):
     return isinstance(value, tuple) and not is_namedtuple(value)
 
 
-def extract_fields_from_nest(nest):
+def extract_fields_from_nest(nest, keep_order=False):
     """Extract fields and the corresponding values from a nest if it's either
     a ``namedtuple`` or ``dict``.
 
     Args:
         nest (nest): a nested structure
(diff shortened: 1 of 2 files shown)


alf/nest/nest.py
```diff
@@ -539,8 +542,18 @@ def fast_map_structure(func, *structure):
...
-def py_pack_sequence_as(nest, flat_seq):
-    """Returns a given flattened sequence packed into a given structure."""
+def py_pack_sequence_as(nest, flat_seq, keep_fields_order=False):
+    """Returns a given flattened sequence packed into a given structure.
+
+    Args:
+        nest: a nested structure
+        flat_seq: a flattened sequence to be packed into the same structure as
+            ``nest``
+        keep_fields_order: if true, do not sort the fields when packing for
```
emailweixu, line 561: Shouldn't we have `keep_fields_order` for flatten() too in order to restore the original nest?

hnyu: > Shouldn't we have `keep_fields_order` for flatten() too in order to restore the original nest?

Added; for now only py_flatten supports this order arg.

emailweixu (approved):

PR #1680, opened 2024-08-02
multi-node DDP support

sqchai:
1. Fixed 'store_snapshot'. Previously each node will try to write a snapshot to the same root directory in multi-node multi-gpu mode. Restored 'store_snapshot' default value to True. Modified such that only node rank = 0 process will write a snapshot.
2. Previous change to train.py on line 255 to line 311 (multi-gpu part) is not necessary. Restored to original code.
3. For the 'disallow_scheduler' part in 'process_environment.py'. For some reasons my previous local code base do not have this 'disallow_scheduler' function, and does not import this function in the script. Therefore, this part of code always fails. I fixed those things according to the latest alf implementation.
4. Modified README.md to include instructions to launch multi-node multi-gpu training.
5. Fixed code style problem.

### DIFF
--- a/alf/bin/train.py
+++ b/alf/bin/train.py
@@ -223,7 +223,7 @@ def training_worker(rank: int,
             interpreted as "non distributed mode".
         conf_file (str): Path to the training configuration.
         root_dir (str): Path to the directory for writing logs/summaries/checkpoints.
-        paras_queue (mp.Queue): a shared Queue for checking the consistency of model parameters
+        paras_queue (Queue): a shared Queue for checking the consistency of model parameters
             in different worker processes, if multi-gpu training is used.
     """
     try:
@@ -288,7 +288,7 @@ def training_worker_multi_node(local_rank: int,
             interpreted as "non distributed mode".
         conf_file (str): Path to the training configuration.
         root_dir (str): Path to the directory for writing logs/summaries/checkpoints.
-        paras_queue: a shared Queue for checking the consistency of model parameters
+        paras_queue (Queue): a shared Queue for checking the consistency of model parameters
             in different worker processes, if multi-gpu training is used.
     """
     try:
@@ -423,7 +423,6 @@ def main(_):
         local_rank = int(os.environ['LOCAL_RANK'])
         rank = int(os.environ['RANK'])
         world_size = int(os.environ['WORLD_SIZE'])
-        print("local_rank: {} | rank: {} | world_size: {}".format(local_rank, rank, world_size))
 
         try:
             # Create a shared queue for checking the consistency of the parameters
--- a/alf/bin/train.py
+++ b/alf/bin/train.py
@@ -429,7 +429,14 @@ def main(_):
             manager = mp.Manager()
             paras_queue = manager.Queue()
             CUDA_VISIBLE_DEVICES = os.environ.get('CUDA_VISIBLE_DEVICES')
-            os.environ['CUDA_VISIBLE_DEVICES'] = str(local_rank)
+            if CUDA_VISIBLE_DEVICES is None:
+                num_devices = torch.cuda.device_count()
+                devices = [d for d in range(num_devices)]
+            else:
+                devices = CUDA_VISIBLE_DEVICES.split(',')
+                devices = [int(d) for d in devices]
+            assert local_rank < len(devices)
+            os.environ['CUDA_VISIBLE_DEVICES'] = str(devices[local_rank])
             training_worker_multi_node(
                 local_rank=local_rank,
                 rank=rank,
--- a/alf/bin/train.py
+++ b/alf/bin/train.py
@@ -78,7 +78,7 @@ def _define_flags():
     flags.DEFINE_bool(
         'force_torch_deterministic', True,
         'torch.use_deterministic_algorithms when random_seed is set')
-    flags.DEFINE_bool('store_snapshot', False,
+    flags.DEFINE_bool('store_snapshot', True,
                       'Whether store an ALF snapshot before training')
     flags.DEFINE_enum(
         'distributed', 'none', ['none', 'multi-gpu', 'multi-node-multi-gpu'],
@@ -189,6 +189,13 @@ def _train(root_dir, local_rank=-1, rank=0, world_size=1):
         raise ValueError("Unsupported ml_type: %s" % trainer_conf.ml_type)
 
     trainer.train()
+
+
+def _training_worker_helper(rank: int, *args, **kwargs):
+    # Helper to start the training worker with the correct rank
+    # so that rank 0 is from the main process and the rest are
+    # from the spawned processes.
+    training_worker(rank + 1, *args, **kwargs)
 
 
 def training_worker(rank: int,
@@ -318,7 +325,7 @@ def main(_):
 
     conf_file = common.get_conf_file()
 
-    if FLAGS.store_snapshot:
+    if FLAGS.store_snapshot and int(os.environ['RANK']) == 0:
         common.generate_alf_snapshot(common.alf_root(), conf_file, root_dir)
 
     # FLAGS.distributed is guaranteed to be one of the possible values.
@@ -342,7 +349,7 @@ def main(_):
             # in different work processes.
             manager = mp.Manager()
             paras_queue = manager.Queue()
-            with common.get_unused_port(12360) as port:
+            with common.get_unused_port(12355) as port:
                 # The other process will communicate with the authoritative
                 # process via network protocol on localhost:port.
                 os.environ['MASTER_PORT'] = str(port)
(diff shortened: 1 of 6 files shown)


alf/bin/train.py
```diff
@@ -176,13 +179,12 @@ def training_worker(rank: int,
             # Specialization for distributed mode
             dist.init_process_group('nccl', rank=rank, world_size=world_size)
             # Recover the flags when spawned as a sub process
-            if rank > 0:
```
JingyuQian: this `if` condition, if present during "multi-gpu" mode, will raise a `"root_dir" has been defined twice` error. Removing it seems to solve the problem. Need further confirmation?

emailweixu: On the contrary, `if` condition is needed so that flags will not be redefined for rank0 because rank0 process is the main process, which already defined the flags.

alf/bin/train.py
```diff
@@ -215,7 +279,7 @@ def main(_):
 
     conf_file = common.get_conf_file()
 
-    if FLAGS.store_snapshot:
+    if FLAGS.store_snapshot and int(os.environ['RANK']) == 0:
```
emailweixu: There may not be RANK environment variable for single process / single node job.

alf/bin/train.py
```diff
@@ -264,6 +328,40 @@ def main(_):
             # exit code for the program.
             raise ChildProcessError(f'Training failed on subprocess exception')
 
+    elif FLAGS.distributed == 'multi-node-multi-gpu':
+        local_rank = int(os.environ['LOCAL_RANK'])
```
emailweixu, line 422: why not use FLAGS.local_rank?

alf/bin/train.py
```diff
@@ -264,6 +328,40 @@ def main(_):
...
+        rank = int(os.environ['RANK'])
+        world_size = int(os.environ['WORLD_SIZE'])
```
emailweixu, line 424: And I couldn't find documentation about RANK or WORLD_SIZE environment variable for torch.distributed.launch. Can you give the link for the documentation?

alf/utils/per_process_context.py
```diff
@@ -78,6 +81,10 @@ def is_distributed(self):
     def ddp_rank(self):
         return self._ddp_rank
 
+    @property
+    def local_rank(self):
```
emailweixu, line 86: since now there are two rank, ddp_rank and local_rank, it is better to clearly document what their meanings are.

alf/utils/spawned_process_utils.py
```diff
@@ -31,6 +31,7 @@ class SpawnedProcessContext(NamedTuple):
     """
     ddp_num_procs: int
     ddp_rank: int
+    local_rank: int
```
emailweixu, line 38: comment ddp_rank and local_rank

alf/bin/train.py
```diff
@@ -264,6 +328,40 @@ def main(_):
...
+            training_worker(
+                rank=0, world_size=1, conf_file=conf_file, root_dir=root_dir)
+            return
+
+        try:
+            # Create a shared queue for checking the consistency of the parameters
+            # in different work processes.
+            manager = mp.Manager()
+            paras_queue = manager.Queue()
+            training_worker_multi_node(
```
emailweixu, line 440: I guess Queue does not work across multiple machines. If so,  need to update Trainer._check_dpp_paras_consistency for multi-node mode. Perhaps change `if self._rank > 0:` to `if local_rank > 0:`. If so, need to update the doc-string for TrainerConfig.ddp_paras_check_interval to say that it only check the consistency within one node. And please test if TraomerConfig.check_dpp_paras_consistency=True works properly.

alf/bin/train.py
```diff
@@ -191,7 +194,68 @@ def training_worker(rank: int,
...
+            interpreted as "non distributed mode".
+        conf_file (str): Path to the training configuration.
+        root_dir (str): Path to the directory for writing logs/summaries/checkpoints.
+        paras_queue: a shared Queue for checking the consistency of model parameters
+            in different worker processes, if multi-gpu training is used.
+    """
+    try:
+        _setup_logging(log_dir=root_dir, rank=rank)
+        _setup_device(local_rank)
+        if world_size > 1:
```
emailweixu: it's guaranteed that `world_size > 1`, right? So there is no need for the `if`

alf/bin/train.py
```diff
@@ -133,6 +135,8 @@ def _train(root_dir, rank=0, world_size=1):
 
     if trainer_conf.ml_type == 'rl':
         ddp_rank = rank if world_size > 1 else -1
+        if ddp_rank > -1 and local_rank > -1:
+            ddp_rank = local_rank
```
emailweixu: what is the reason for this?

alf/bin/train.py
emailweixu: Actually, I think `paras_queue` doesn't work at all. In multi-gpu mode, `paras_queue` can work because it is created by the main process. All the worker will use the same paras_queue. Here different worker will use the different queue, which makes no sense.
So it's better just to report unsupported error for TraomerConfig.check_dpp_paras_consistency=True in multi-node-multi-gpu mode.

alf/bin/train.py
sqchai: The 'LOCAL_RANK' is automatically assigned by PyTorch launcher. I'm not sure how to pre-assign this from a shell script

alf/bin/train.py
sqchai: This is updated in the new README

emailweixu:
Need resolve conflict first.

emailweixu:
Still need to resolve conflict

alf/bin/train.py
emailweixu: Can you address this?

alf/bin/train.py
```diff
@@ -190,6 +203,13 @@ def _train(root_dir, rank=0, world_size=1):
     trainer.train()
 
 
+def _training_worker_helper(rank: int, *args, **kwargs):
+    # Helper to start the training worker with the correct rank
+    # so that rank 0 is from the main process and the rest are
+    # from the spawned processes.
+    training_worker(rank + 1, *args, **kwargs)
```
emailweixu: no longer used.

alf/bin/train.py
```diff
@@ -147,12 +159,13 @@ def _setup_remote_configs_if_needed():
         })
 
 
-def _train(root_dir, rank=0, world_size=1):
+def _train(root_dir, local_rank=-1, rank=0, world_size=1):
```
emailweixu, line 162: local_rank is not used.

alf/trainers/policy_trainer.py
```diff
@@ -808,6 +808,12 @@ def _check_dpp_paras_consistency(self, iter_num: int,
             return
 
         proc_cxt = PerProcessContext()
+        if (proc_cxt.is_distributed and proc_cxt.local_rank >= 0
```
emailweixu: This error can be reported earlier in train.py

alf/bin/train.py
sqchai: I'm currently addressing this in policy_trainer.py. I'm not sure how to get this flag value in train.py.  It seems like the config is not yet parsed from file in train.py?

emailweixu (dismissed):

sqchai:
1. Switched the launching method to torchrun
2. The setup_device() function was modified in a previous commit to support Taichi, which caused problems to multi-node DDP launching. I updated the multi-node training worker so they're compatible now.

alf/bin/train.py
```diff
@@ -428,13 +428,20 @@ def main(_):
             # in different work processes.
             manager = mp.Manager()
             paras_queue = manager.Queue()
+            CUDA_VISIBLE_DEVICES = os.environ.get('CUDA_VISIBLE_DEVICES')
+            os.environ['CUDA_VISIBLE_DEVICES'] = str(local_rank)
```
emailweixu: If the original CUDA_VISIBLE_DEVICES is set, it needs to be set based on the original CUDA_VISIBLE_DEVICES here, similar to what single node DDP does.

sqchai: Yes I tried this, but If the CUDA_VISIBLE_DEVICES is set to 2,3, and the local_rank will be assigned as logical rank 0,1. In this case, using the physical device id can cause problems.

emailweixu: > Yes I tried this, but If the CUDA_VISIBLE_DEVICES is set to 2,3, and the local_rank will be assigned as logical rank 0,1. In this case, using the physical device id can cause problems.

What is the problem? If CUDA_VISIBLE_DEVICES is not supported, it should report error.

emailweixu (approved):

PR #1681, opened 2024-08-03
additional kwargs for Algorithm.state_dict()

hnyu:
Sometimes other libs will call `state_dict()` with additional args for old pytorch versions. For example, `torch.jit.trace` will call `state_dict(keep_vars=True)`. Thus we need to make our implementation compatible with older version code.

keep_vars is only used in torch<0.4.0. See https://stackoverflow.com/questions/57534801/what-is-the-meaning-of-keep-vars-in-state-dict

### DIFF
--- a/alf/algorithms/algorithm.py
+++ b/alf/algorithms/algorithm.py
@@ -889,7 +889,7 @@ class Algorithm(AlgorithmInterface):
         return r
 
     @common.add_method(nn.Module)
-    def state_dict(self, destination=None, prefix='', visited=None):
+    def state_dict(self, destination=None, prefix='', visited=None, **kwargs):
         """Get state dictionary recursively, including both model state
         and optimizers' state (if any). It can handle a number of special cases:
 
--- a/alf/algorithms/algorithm.py
+++ b/alf/algorithms/algorithm.py
@@ -904,7 +904,8 @@ class Algorithm(AlgorithmInterface):
                 state dictionary.
             visited (set): a set keeping track of the visited objects.
             kwargs: additional keyword arguments for backward compatibility (
-                e.g., 'keep_vars'). They are not used in newer pytorch versions.
+                e.g., 'keep_vars' for torch<0.4). They are not used in newer
+                pytorch versions.
 
         Returns:
             OrderedDict: the dictionary including both model state and optimizers'
--- a/alf/algorithms/algorithm.py
+++ b/alf/algorithms/algorithm.py
@@ -903,6 +903,8 @@ class Algorithm(AlgorithmInterface):
                 (modules, params, algorithms etc) as the key used in the
                 state dictionary.
             visited (set): a set keeping track of the visited objects.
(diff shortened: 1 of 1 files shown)


alf/algorithms/algorithm.py
```diff
@@ -889,7 +889,7 @@ def get_initial_train_state(self, batch_size):
         return r
 
     @common.add_method(nn.Module)
-    def state_dict(self, destination=None, prefix='', visited=None):
+    def state_dict(self, destination=None, prefix='', visited=None, **kwargs):
```
QuantuMope, line 892: Maybe add a short comment explaining why kwargs is needed here.

emailweixu:
> Sometimes other libs will call `state_dict()` with additional args. For example, `torch.jit.trace` will call `state_dict(keep_vars=True)`. Thus we need to make our implementation compatible with others.

If keev_vars is set, do we need our `state_dict()` to handle it?

hnyu:
> > Sometimes other libs will call `state_dict()` with additional args. For example, `torch.jit.trace` will call `state_dict(keep_vars=True)`. Thus we need to make our implementation compatible with others.
> 
> If keev_vars is set, do we need our `state_dict()` to handle it?

I think keep_vars is only used by older pytorch. Since our current code already doesn't use it and our code has been running fine 

> > Sometimes other libs will call `state_dict()` with additional args. For example, `torch.jit.trace` will call `state_dict(keep_vars=True)`. Thus we need to make our implementation compatible with others.
> 
> If keev_vars is set, do we need our `state_dict()` to handle it?

keep_vars is only used in torch<0.4.0. See https://stackoverflow.com/questions/57534801/what-is-the-meaning-of-keep-vars-in-state-dict 
We will never use torch<0.4.0, so it's OK to just ignore it?

alf/algorithms/algorithm.py
hnyu: > Maybe add a short comment explaining why kwargs is needed here.

added

alf/algorithms/algorithm.py
```diff
@@ -903,6 +903,8 @@ def state_dict(self, destination=None, prefix='', visited=None, **kwargs):
                 (modules, params, algorithms etc) as the key used in the
                 state dictionary.
             visited (set): a set keeping track of the visited objects.
+            kwargs: additional keyword arguments for backward compatibility (
+                e.g., 'keep_vars'). They are not used in newer pytorch versions.
```
QuantuMope: We can mention `They are not used in newer pytorch versions > 0.4.0.` so that users know right away their code will be compatible.

hnyu: > We can mention `They are not used in newer pytorch versions > 0.4.0.` so that users know right away their code will be compatible.

Updated. But ALF never supports torch<0.4.0. So a user won't even know or worry about this... Other code has this arg for backward compatibility only; they don't actually use this.

emailweixu (dismissed):

QuantuMope (approved):

PR #1685, opened 2024-08-09
fix histogram_continuous

hnyu:
Currently our histogram_continuous will call tensorboard's histogram function which relies on np.histogram to count. However, np.histogram will not include data points that are exactly equal to the rightmost edge. For example,

```python
import torch
import numpy as np
import alf


data = torch.tensor([1.7] * 4)
bucket_min = -1.38
bucket_max = 1.7

bucket_count = 10
bins = (
        bucket_min +
        (torch.arange(bucket_count + 1, dtype=torch.float64) / bucket_count) *
        (bucket_max - bucket_min))

print(np.histogram(data.numpy(), bins=bins.numpy())[0])
```
This will output all bins with zero counts. 

So we need to slightly extend the boundary of the histogram.

For our case of summarizing continuous actions, if the batch size is very small and all actions have the same value that are equal to the rightmost edge, the summary will fail. 

The actions equal to the edge are possible if we clip the policy actions by the action range, or if we are summarizing rollout actions that are generated by some demonstration policy.

### DIFF
--- a/alf/utils/summary_utils.py
+++ b/alf/utils/summary_utils.py
@@ -71,6 +71,7 @@ def histogram_continuous(name,
                          bucket_min=None,
                          bucket_max=None,
                          bucket_count=DEFAULT_BUCKET_COUNT,
+                         edge_inclusive=True,
                          step=None):
     """histogram for continuous data.
 
@@ -82,6 +83,8 @@ def histogram_continuous(name,
         bucket_max (float|None): represent bucket max value,
             if None value, ``data.max()`` will be used
         bucket_count (int):  positive ``int``. The output will have this many buckets.
+        edge_inclusive (bool): whether to include a data point that lies exactly
+            on the leftmost/rightmost edge.
         step (None|Tensor): step value for this summary. this defaults to
             ``alf.summary.get_global_counter()``
     """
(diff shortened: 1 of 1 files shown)


alf/utils/summary_utils.py
```diff
@@ -98,7 +101,8 @@ def histogram_continuous(name,
         bucket_min +
         (torch.arange(bucket_count + 1, dtype=torch.float64) / bucket_count) *
         (bucket_max - bucket_min))
-    data = data.clamp(bucket_min, bucket_max)
+    eps = 1e-6 if edge_inclusive else 0
+    data = data.clamp(bucket_min + eps, bucket_max - eps)
```
QuantuMope: Wouldn't it be better to add eps to the bucket min and max instead? Though since it's such a small number, shouldn't matter either way.

QuantuMope (dismissed):
Thanks for the fix!

alf/utils/summary_utils.py
hnyu: > Wouldn't it be better to add eps to the bucket min and max instead? Though since it's such a small number, shouldn't matter either way.

I think they are equivalent.

QuantuMope: I was thinking this would be an issue if mean values are also repeated by mousing over the histogram bars, but we report mean separately anyways so should be fine.

emailweixu: I also feel changing the boundary slightly is better. The change will be simpler and the resulted histogram is more likely to be same as the desired one.

le-horizon (approved):

alf/utils/summary_utils.py
hnyu: > I also feel changing the boundary slightly is better. The change will be simpler and the resulted histogram is more likely to be same as the desired one.

Now updated to extend the leftmost/rightmost edge.

QuantuMope (approved):

PR #1695, opened 2024-08-26
Stop async evaluation immediately for Evaluator.close()

emailweixu:
When error happens in the training, ALF will try to close all the evaluator. However, some evaluation takes long time. This causes the main process just wait and we cannot see the error message from exception until the evaluation is over.

With this change, the evaluation will stop immediately so we can see the error messages.

### DIFF
--- a/alf/trainers/evaluator.py
+++ b/alf/trainers/evaluator.py
@@ -139,6 +139,40 @@ def _define_flags():
 FLAGS = flags.FLAGS
 
 
+class PeekableQueue(object):
+    """A queue that supports peeking the first element without removing it.
+
+    Note that this can only be used for a queue with one consumer.
+    """
+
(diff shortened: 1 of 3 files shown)


alf/trainers/evaluator.py
```diff
@@ -131,6 +139,40 @@ def _define_flags():
...
 
+class PeekableQueue(object):
+    """A queue that supports peeking the first element without removing it.
+
+    Note that this can only be used for a queue with one consumer.
+    """
+
+    def __init__(self, queue: mp.Queue):
+        self._queue = queue
+        self._elements = []
```
hnyu: Maybe add a comment saying that this stores the peeked element and its max len is 1.

hnyu (approved):

PR #1698, opened 2024-08-28
add support for tensorrt 8

hnyu:
Add support for TensorRT 8.6. Several important differences with TensorRT 10:
1. execute_async_v3 might not exist. It does for 8.6.2 (Jetson nano) but not for 8.6.1. Even it exists, the function might be buggy (all zero output). So for safety concern, we use v2 for TRT 8.
2. Some API differences when creating an engine.
3. A big trouble of TensorRT 8 is that its input/output args order might not respect that of the ONNX model! So we need to manually keep track of the correspondence when memcopying between host/device. For tensorrt 10, there is no such an issue as we can explicitly set the IO tensor mapping.

Since there is no backward compatibility, we need to implement two classes to handle them separately.

According to my test, TRT8 and TRT10 have the same performance. Also tested on Jetson nano and it passed the test case.

PS: onnxruntime can handle different TRT versions, so it always behaves normally.

### DIFF
--- a/alf/utils/tensorrt_utils.py
+++ b/alf/utils/tensorrt_utils.py
@@ -311,7 +311,6 @@ class OnnxRuntimeEngine(object):
         return self._onnx_wrapper.recover_module_output(outputs)
 
 
-@alf.configurable(whitelist=['memory_limit_gb', 'fp16'])
 class TensorRTEngine(object):
     def __init__(self,
                  module: torch.nn.Module,
@@ -321,7 +320,8 @@ class TensorRTEngine(object):
                  memory_limit_gb: float = 1.,
                  fp16: bool = False,
                  example_args: Tuple[Any] = (),
-                 example_kwargs: Dict[str, Any] = {}):
+                 example_kwargs: Dict[str, Any] = {},
+                 validate_args: bool = False):
         """Class for converting a torch.nn.Module to TensorRT engine for fast
         inference, via ONNX model as the intermediate representation.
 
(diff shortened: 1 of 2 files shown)


alf/utils/tensorrt_utils.py
```diff
@@ -450,6 +460,82 @@ def __call__(self, *args, **kwargs):
...
+
+    def __call__(self, *args, **kwargs):
+        flat_all_args = _dtype_conversions(alf.nest.flatten([args, kwargs]))
+
+        for i in range(len(flat_all_args)):
+            im = self._input_mem[i]
+            arg = flat_all_args[self._input_idx[i]]
+            cuda.memcpy_dtod_async(im,
+                                   arg.contiguous().data_ptr(),
+                                   self._get_bytes(arg), self._stream)
```
emailweixu, line 552: synchronization is not needed here?

alf/utils/tensorrt_utils.py
```diff
@@ -450,6 +460,82 @@ def __call__(self, *args, **kwargs):
...
+        self._output_mem = []
+        self._output_idx = []
+        self._bindings = []
+        # TRT8: This order might be different from the order of the onnx model!!
+        for i in range(engine.num_io_tensors):
+            name = engine.get_tensor_name(i)
+            shape = tuple(engine.get_tensor_shape(name))
+            idx = int(name.split('-')[1])
+            dtype = trt.nptype(engine.get_tensor_dtype(name))
+            host_mem = cuda.pagelocked_empty(shape, dtype)
```
emailweixu: It seems that this memory allocation is just for calculating the size?
why not use `self._get_bytes()`

alf/utils/tensorrt_utils.py
hnyu: > synchronization is not needed here?

Actually this is still needed. I've added

alf/utils/tensorrt_utils.py
hnyu: > It seems that this memory allocation is just for calculating the size? why not use `self._get_bytes()`

Changed to use _get_bytes().

hnyu:
I've changed the instruction to install TRT8 locally by default, to match the version on Jetson nano.

emailweixu (approved):

PR #1699, opened 2024-09-06
add file cache for tensorrt engine

hnyu:
Added an option of file cache for the tensorrt engine. The idea is that if we want to repeated launch an inference process for the same model, we could save the compiling time by exploiting this file cache option. 

Added a new test case and passed all tensorrt test cases locally.

### DIFF
--- a/alf/utils/tensorrt_utils.py
+++ b/alf/utils/tensorrt_utils.py
@@ -351,16 +351,14 @@ class TensorRTEngine(object):
             engine_file: if provided, this class will first check if such a file
                 exists. If so, it will load the engine from the file and skip the
                 build process. If not, the built engine will be saved to this file.
-                When a file is provided, it is the *user's responsibility* to ensure
-                that the loaded engine will work correctly in the current context;
-                no check will be performed by this class.
-                NOTE: This option is mainly for an inter-process cache and its
-                priority is lower than the in-process engine caching mechanism.
-                Within a process when the same compiled method is called multiple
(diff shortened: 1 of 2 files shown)


alf/utils/tensorrt_utils.py
```diff
@@ -344,6 +348,19 @@ def __init__(self,
...
+                exists. If so, it will load the engine from the file and skip the
+                build process. If not, the built engine will be saved to this file.
+                When a file is provided, it is the *user's responsibility* to ensure
+                that the loaded engine will work correctly in the current context;
+                no check will be performed by this class.
+                NOTE: This option is mainly for an inter-process cache and its
+                priority is lower than the in-process engine caching mechanism.
+                Within a process when the same compiled method is called multiple
+                times with the same signature, engine file loading and ``force_build_engine``
+                won't have any effect.
```
le-horizon: by "won't have any effect" do you mean it will always build the engine?

le-horizon (commented):
Thanks for the nice enhancement Haonan.  Almost good, just one clarification.

Le

alf/utils/tensorrt_utils.py
hnyu: > by "won't have any effect" do you mean it will always build the engine?

whether it will or not depends on the in-process cache mechanism. There is always a cache in the process memory so the file cache won't even be triggered.

le-horizon (approved):

PR #1703, opened 2024-09-18
[Bug Fix] Add missing sync_progress call for parallel play

QuantuMope:
Noticed a bug where certain schedulers had their progresses set to 0 incorrectly when `--parallel_play > 1`. After debugging, I found that this only occurred for process specific code (e.g., tasks) and did not affect the main process's schedulers (e.g., DeltaControl). Furthermore, when setting parallel_play to 0, all schedulers were properly set, which heightened my suspicion of there being a syncing issue when creating parallel processes.

This PR solves this issue by adding an `env_sync_progress()` call before the play loop starts. Now, all process-specific schedulers are properly initialized and parallel_play works as expected.

### DIFF
--- a/alf/trainers/policy_trainer.py
+++ b/alf/trainers/policy_trainer.py
@@ -1120,6 +1120,8 @@ def play(root_dir,
     else:
         selective_criteria_func = None
 
+    # Sync the progress for all environments in case parallel_play > 1
+    env.sync_progress()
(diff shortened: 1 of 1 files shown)


emailweixu (approved):
I should have mentioned this earlier. I vaguely thought this might be related, but somehow didn't talked about it.

PR #1705, opened 2024-09-20
Change override to override_sole_init in alf.config

QuantuMope:
This PR changes the `override` flag of `alf.config` to `override_sole_init`. The implementation before allowed `alf.override_config` to override a value of a config regardless of the `mutable` flag. Unfortunately, this breaks the logic of setting parameters via `--conf_params` or `alf.pre_config`. It also causes issues in evaluator processes that may set certain config values with `mutable=False` and then read from the conf file as it overwrites the desired eval setting.

Therefore, this PR now changes `alf.override_config` to `alf.override_sole_config`, which respects immutability. On our local hobot code, I've now verified that all the above problems have been remedied and singular config calls can now be enforced when using `ALF_SOLE_CONFIG=1`, greatly reducing chance for bugs via nested / unintentional overwrites.

Additional test cases for edge cases has also been provided in `config_utils_test.py`.

Once this gets merged, I can submit PRs for `ALF_SOLE_CONFIG=1` compatible hobot confs.

### DIFF
--- a/alf/config_helpers.py
+++ b/alf/config_helpers.py
@@ -153,7 +153,8 @@ def adjust_config_by_multi_process_divider(ddp_rank: int,
     config1(
         tag,
         math.ceil(num_parallel_environments / multi_process_divider),
-        raise_if_used=False)
+        raise_if_used=False,
+        override_sole_init=True)
 
     # Adjust the mini_batch_size. If the original configured value is 64 and
     # there are 4 processes, it should mean that "jointly the 4 processes have
@@ -165,7 +166,8 @@ def adjust_config_by_multi_process_divider(ddp_rank: int,
         config1(
             tag,
             math.ceil(mini_batch_size / multi_process_divider),
-            raise_if_used=False)
+            raise_if_used=False,
+            override_sole_init=True)
 
     # If the termination condition is num_env_steps instead of num_iterations,
     # we need to adjust it as well since each process only sees env steps taking
@@ -176,20 +178,26 @@ def adjust_config_by_multi_process_divider(ddp_rank: int,
         config1(
             tag,
             math.ceil(num_env_steps / multi_process_divider),
-            raise_if_used=False)
+            raise_if_used=False,
+            override_sole_init=True)
 
     tag = 'TrainerConfig.initial_collect_steps'
     init_collect_steps = get_config_value(tag)
     config1(
         tag,
         math.ceil(init_collect_steps / multi_process_divider),
-        raise_if_used=False)
+        raise_if_used=False,
+        override_sole_init=True)
 
     # Only allow process with rank 0 to have evaluate. Enabling evaluation for
     # other parallel processes is a waste as such evaluation does not offer more
     # information.
     if ddp_rank > 0:
-        config1('TrainerConfig.evaluate', False, raise_if_used=False)
+        config1(
+            'TrainerConfig.evaluate',
+            False,
+            raise_if_used=False,
+            override_sole_init=True)
 
 
 def parse_config(conf_file, conf_params, create_env=True):
--- a/alf/config_helpers.py
+++ b/alf/config_helpers.py
@@ -285,7 +285,11 @@ def get_env():
             # A 'None' random seed won't set a deterministic torch behavior.
             for _ in range(PerProcessContext().ddp_rank):
                 random_seed = random.randint(0, 2**32)
-        config1("TrainerConfig.random_seed", random_seed, raise_if_used=False)
+        config1(
+            "TrainerConfig.random_seed",
+            random_seed,
+            raise_if_used=False,
+            override_sole_init=True)
 
         # We have to call set_random_seed() here because we need the actual
         # random seed to call create_environment.
--- a/alf/config_util.py
+++ b/alf/config_util.py
@@ -35,7 +35,7 @@ __all__ = [
     'get_operative_configs',
     'import_config',
     'load_config',
-    'override_config',
+    'override_sole_config',
     'pre_config',
     'reset_configs',
     'validate_pre_configs',
@@ -147,7 +147,7 @@ def config(prefix_or_dict,
                 override_sole_init)
 
 
-def override_config(prefix_or_dict, **kwargs):
+def override_sole_config(prefix_or_dict, **kwargs):
     """Wrapper function for configuring a config with override_sole_init=True.
 
     This call allows a user to attempt to overwrite a config's value even
--- a/alf/config_util.py
+++ b/alf/config_util.py
@@ -60,7 +60,7 @@ def config(prefix_or_dict,
            mutable=True,
            raise_if_used=True,
            sole_init=False,
-           override=False,
+           override_sole_init=False,
            **kwargs):
     """Set the values for the configs with given name as suffix.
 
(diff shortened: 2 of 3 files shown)


alf/config_util_test.py
```diff
@@ -182,6 +182,63 @@ def sole_init_test_env(x):
         self.assertEqual(alf.get_config_value("sole_init_test_twice.x"), 1)
         self.assertEqual(alf.get_config_value("sole_init_test_env.x"), 1)
 
+        # Test override_config doesn't doesn't overwrite for immutable values.
```
le-horizon: remove one "doesn't"

alf/config_util_test.py
```diff
@@ -182,6 +182,63 @@ def sole_init_test_env(x):
...
+        @alf.configurable
+        def override_on_immutable(x):
+            pass
+
+        alf.config("override_on_immutable", x=0, mutable=False)
+        alf.override_config("override_on_immutable", x=1)
```
le-horizon: Is it possible to raise an error on overriding immutable config?  Silently ignoring override can be hard to debug?

alf/config_util_test.py
```diff
@@ -182,6 +182,63 @@ def sole_init_test_env(x):
...
+            alf.get_config_value("override_on_immutable_and_sole_init.x"), 0)
+
+        # Test that pre_config calls work before a sole_init config call
+        @alf.configurable
+        def pre_config_before(x):
+            pass
+
+        alf.pre_config({"pre_config_before.x": 0})
+        alf.config("pre_config_before", x=1)
+        alf.override_config("pre_config_before", x=1)
```
le-horizon: We use pre_config in deployment app to specify deploy mode etc..  In order to avoid them from being overridden, should we specify them as immutable?  Or should this override_config call throw error?

alf/config_util_test.py
```diff
@@ -182,6 +182,63 @@ def sole_init_test_env(x):
...
+        with self.assertRaises(RuntimeError) as context:
+            alf.pre_config({"pre_config_after.x": 0})
+
+        # Test that calling override_config doesn't affect previous sole_init calls.
+        @alf.configurable
+        def override_no_affect_sole_init(x):
+            pass
+
+        alf.config("override_no_affect_sole_init", x=1, sole_init=True)
+        alf.override_config("override_no_affect_sole_init", x=2)
```
le-horizon: should this also throw error?

alf/config_util.py
```diff
@@ -60,7 +60,7 @@ def config(prefix_or_dict,
            mutable=True,
            raise_if_used=True,
            sole_init=False,
-           override=False,
+           override_sole_init=False,
```
le-horizon, line 63: please also add tests for override_sole_init.

alf/config_util.py
```diff
@@ -112,19 +112,18 @@ def func(self, a, b):
...
+            This is helpful in enforcing a singular point of initialization,
+            thus eliminating any potential side effects from possible future
+            overrides. For users wanting this to be the default behavior, the
+            ALF_SOLE_CONFIG env variable can be set to 1.
+        override_sole_init (bool): If True, the value of the config will be set
+            regardless of any previous ``sole_init`` setting. This should be used
+            only when absolutely necessary (e.g., a teacher-student training loop,
+            where the student must override certain configs inherited from the
+            teacher). If the config is immutable, a warning will be declared with
+            no changes made.
```
le-horizon, line 126: instead of quietly ignoring the config, should we just throw error for this case?

le-horizon (commented):
Thanks for the much cleaner changes and tests, Andrew.  A few questions regarding alf config behavior.

Le

alf/config_util.py
QuantuMope: `alf.override_config` is simply a wrapper for `alf.config(override_sole_init=True)`.

alf/config_util.py
QuantuMope: So the reason we cannot do this is because `alf.pre_config` and `--conf_param` set values with `mutable=False` (that is how those values avoid getting overwritten). Therefore, there may be times we want to override our conf files using one of the two options and raising an error in this situation is problematic. A workaround for this was to possibly introduce a `pre_configged` state variable for configs, but I thought that may be unnecessary. Let me know what you think.

alf/config_util_test.py
QuantuMope: Thanks! Done.

alf/config_util_test.py
QuantuMope: Refer to my reply above for why this cannot be done currently without adding more state parameters. There is a logger warning that gets printed in this scenario, though it's easy for users to not see it.

alf/config_util_test.py
QuantuMope: `pre_config` already sets all values as `mutable=False`. This is how it and `--conf_params` prevent their config values from being overwritten.

alf/config_util_test.py
QuantuMope: It shouldn't throw an error since `override_config` allows us to set a config's value despite an earlier `sole_init=True` call.

QuantuMope (commented):
Thanks for the really speedy review Le. Addressed your comments. Let me know if there are any questions or clarifications I can make.

alf/config_util.py
le-horizon: I see.  You are right.  Let's go with current way for the PR.

Separately, are there some examples of these cases?  It would be good to understand a bit more, and see if we can design better solutions.

alf/config_util_test.py
le-horizon: Maybe let's color code the warning message a bright yellow, so it's harder to miss?

alf/config_util_test.py
le-horizon: Ah, my mistake.

alf/config_util_test.py
```diff
@@ -182,6 +182,63 @@ def sole_init_test_env(x):
...
+        with self.assertRaises(RuntimeError) as context:
+            alf.config("override_no_affect_sole_init", x=3)
```
le-horizon, line 240: test that x's value is 2

alf/config_util_test.py
le-horizon: Oh, I see.  Maybe rename the function to alf.override_sole_config()?  Just as how you renamed override to override_sole_init.

le-horizon (commented):
Thanks for answering the questions.  Only a few nits.

Thanks,
Le

alf/config_helpers.py
```diff
@@ -285,7 +285,11 @@ def get_env():
             # A 'None' random seed won't set a deterministic torch behavior.
             for _ in range(PerProcessContext().ddp_rank):
                 random_seed = random.randint(0, 2**32)
-        config1("TrainerConfig.random_seed", random_seed, raise_if_used=False)
+        config1(
```
emailweixu, line 296: there are several other config1 calls in config_helpers.py that may need fix too.

QuantuMope: Thanks for the catch. This has been fixed for the other calls.

alf/config_util.py
QuantuMope: Are you referring to examples of the teacher-student setup?

One is our own agent training. Let's say you train expert with `ALF_SOLE_CONFIG=1`. When we train the agent with `ALF_SOLE_CONFIG=1` as well, it will inevitably raise errors as agent often overwrites some of the expert's configs. Here, we can use the `alf.override_config` function. Not only does this minimize config calls, but it also explicitly tells the coder which inherited calls are being overwritten.

alf/config_util_test.py
QuantuMope: Right now it prints as red via `absl.logging.warning`. Do you think yellow is better?

alf/config_util_test.py
QuantuMope: Good idea. Done.

alf/config_util_test.py
QuantuMope: Thanks. Done.

QuantuMope (commented):
Thanks for the comments. PTAL

emailweixu (approved):

le-horizon (approved):

PR #1706, opened 2024-09-27
Enable play with supervised algorithms

Haichao-Zhang:
This PR enables alf.bin.play with non-rl algorithms (e.g. supervised algorithms).
With this, given a proper environment, we can also evaluate policies trained in a non-standard RL fashion, e.g. via offline supervised training.

### DIFF
--- a/alf/bin/play.py
+++ b/alf/bin/play.py
@@ -28,6 +28,7 @@ from absl import app
 from absl import flags
 from absl import logging
 import copy
+import inspect
 import os
 import subprocess
 import sys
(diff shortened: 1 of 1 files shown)


alf/bin/play.py
```diff
@@ -155,11 +156,19 @@ def play():
...
+        'observation_spec': observation_spec,
+        'action_spec': env.action_spec(),
+        'config': config
+    }
+    # only the algorithms derived from ``RLAlgorithm`` need reward_spec
+    if 'reward_spec' in signature.parameters:
+        kwargs['reward_spec'] = env.reward_spec()
+
+    algorithm = algorithm_ctor(**kwargs)
+    algorithm.set_path('')
```
emailweixu: duplicated with the line 172

Haichao-Zhang: Thanks for the catch and now removed.

emailweixu (approved):

PR #1708, opened 2024-10-09
update tensorrt cheatsheet

hnyu:
Added a note about `rsample()` for tensorrt.

### DIFF
--- a/alf/utils/README.md
+++ b/alf/utils/README.md
@@ -73,8 +73,7 @@ ORT_ONNX_BACKEND_EXCLUDE_PROVIDERS=TensorrtExecutionProvider
 ```
 
 ### `rsample()`
-TensorRT will report "ERROR: Network must have at least one output" when there is any `rsample()` in the eager mode code. Fortunately, for inference we can usually use the distribution's mode
-instead to avoid this issue.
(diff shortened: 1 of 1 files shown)


alf/utils/README.md
```diff
@@ -72,6 +72,10 @@ If you have to use a dynamic shape, you can choose to use the CUDA backend by se
 ORT_ONNX_BACKEND_EXCLUDE_PROVIDERS=TensorrtExecutionProvider
 ```
 
+### `rsample()`
+TensorRT will report "ERROR: Network must have at least one output" when there is any `rsample()` in the eager mode code. Fortunately, for inference we can usually use the distribution's mode
```
emailweixu: Can `sample()` be used?

hnyu: > Can `sample()` be used?

verified that sample() is OK.

emailweixu: Update the doc to say that inference we can usually use `sample()` or the distribution's mode.

hnyu: > Update the doc to say that inference we can usually use `sample()` or the distribution's mode.

updated

emailweixu (approved):

PR #1709, opened 2024-10-09
fix loading trt engine

hnyu:
Currently, when loading a saved trt engine file, there is an error "getPluginCreator could not find plugin InstanceNormalization_TRT version 1" on some machines. Need to manually init libnvinfer before using tensorrt.


Haichao-Zhang (commented):
On which of the machines do we have this issue?

hnyu:
> On which of the machines do we have this issue?

hyaku laptop

Haichao-Zhang (approved):

PR #1710, opened 2024-10-15
option of instance norm in MoNetUNet

hnyu:
Add a flag for turning off InstanceNorm layers in the MoNet U-Net creation. 

Sometimes if we don't want to have distinct feature values across the spatial domain, we might want to turn the normalization off. 

For example, if we want to predict a depth map with similar values over the field, enforcing instance norm might result in something like below:
Screenshot 2024-10-15 at 10 24 10 AM

The white block on the top right seems a trick done by the network for instance norm.

### DIFF
--- a/alf/algorithms/monet_algorithm.py
+++ b/alf/algorithms/monet_algorithm.py
@@ -51,6 +51,7 @@ class MoNetUNet(alf.networks.Network):
                  filters: Tuple[int],
                  nonskip_fc_layers: Tuple[int],
                  output_channels: int,
+                 instancenorm: bool = True,
                  name: str = "MoNetUNet"):
         """
         Args:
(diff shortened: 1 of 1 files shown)


alf/algorithms/monet_algorithm.py
```diff
@@ -51,6 +51,7 @@ def __init__(self,
                  filters: Tuple[int],
                  nonskip_fc_layers: Tuple[int],
                  output_channels: int,
+                 instancenorm: bool = True,
```
Haichao-Zhang: add an underscore  (``use_instance_norm``) in the name to be consistent with other arguments?

hnyu: > add an underscore (`use_instance_norm`) in the name to be consistent with other arguments?

updated the name

Haichao-Zhang (approved):

PR #1711, opened 2024-11-20
Implement flow matching for generation and imitation 

hnyu:
Added a simple flow matching algorithm for generative modeling and imitation learning, as described in:

> "π0: A Vision-Language-Action Flow Model for General Robot Control", Physical Intelligence, 2024.

Tested on MNIST and CIFAR10. Generating two samples for each of 10 classes.

MNIST:

flow_match_generation

CIFAR10:

flow_match_generation_cifar

Since CIFAR10's images have far more variation within each class, the training data (6k per class) is not enough. So the generation results usually don't resemble any training sample.


emailweixu (commented):
Did you try the midpoint solver to see if it generate better? According to the original FM paper, it has better generation quality given same number of function evaluation. And it is easy to implement too:
I^{t+dt/2} = I^t + v(I^t, t)dt/2
I^{t+dt} = I^t + v(I^{t+dt/2}, t+dt/2) dt

hnyu:
> Did you try the midpoint solver to see if it generate better? According to the original FM paper, it has better generation quality given same number of function evaluation. And it is easy to implement too: I^{t+dt/2} = I^t + v(I^t, t)dt/2 I^{t+dt} = I^t + v(I^{t+dt/2}, t+dt/2) dt

Tried but the visual quality difference is hard to tell.

alf/algorithms/flow_matching_algorithm.py
```diff
@@ -0,0 +1,223 @@
...
+                 return_intermediate_steps: bool = False):
+        """Generate new outputs.
+
+        Args:
+            cond_input: the conditional input. If provided, its batch size will
+                be used instead of ``batch_size``.
+            batch_size: the batch size of the generated outputs.
+            return_intermediate_steps: whether to return the intermediate outputs.
+                If True, return all ``self._int_steps`` outputs in a list; otherwise
+                return the final output as a list.
```
emailweixu: Why return the final output as an one-element list? It is more intuitive to just return the final tensor.

hnyu: > Why return the final output as an one-element list? It is more intuitive to just return the final tensor.

Updated to return a single tensor

hnyu:
Removed flow_matching_algorithm_test.py from build.sh because on the CI server, there is always an issue of downloading the MNIST dataset:

```bash
Downloading http://yann.lecun.com/exdb/mnist/train-images-idx3-ubyte.gz
Failed to download (trying next):
HTTP Error 403: Forbidden
```

emailweixu (approved):

PR #1712, opened 2024-12-03
distributed off-policy algorithm

hnyu:
This PR adds the first version of distributed off-policy algorithm. The idea is to support unrollers (responsible for interacting with an environment and collecting data) and a trainer (responsible for training the policy) on different machines.

1. Unrollers and the trainer can have different ips or the same ip.
2. Unrollers will send collected data to the trainer, while the trainer will send the updated model params to the unrollers. The data collection and params update are async. Basically we can think of splitting a typical off-policy algorithm into two complementary parts (unroll and train) that run on different machines.
3. A new unroller can joint at any time of the training. Also an existing unroller can be killed at any time. 
4. The trainer supports mutli-gpu training with DDP. Each DDP rank has its own replay buffer whose data gets routed from the unrollers evenly. Also the jobs of param syncing with the unrollers are assigned to DDP ranks evenly.
5. Currently only supports envs with batch_size=1 for each unroller for simplifying data routing decisions.

`DistributedOffPolicyAlgorithm` technically is not a new algorithm. Instead, it's just a wrapper around any existing off-policy algorithm (e.g., SAC, DDPG, etc). The wrapper is mainly used for establishing the communication protocol between unrollers and trainer.

A diagram of the pipeline can be found at https://horizonrobotics.feishu.cn/wiki/BFPewZ1mzi4GkMkOgUXcSsHHnEc

I tested this pipeline on cart_pole as in `distributed_sac_cartpole_conf.py`:

Training (two DDP ranks):
```bash
python -m alf.bin.train --root_dir /tmp/trainer --conf  distributed_sac_cartpole_conf.py --as_remote_trainer --conf_param="make_ddp_performer.find_unused_parameters=1" --distributed multi-gpu --conf_param="DistributedTrainer.push_params_every_n_iters=10"
```

Unrolling (in another terminal):
```bash
python -m alf.bin.train --root_dir /tmp/unroller --conf distributed_sac_cartpole_conf.py --as_remote_unroller
```

A second unroller (in a third terminal) 
```python
python -m alf.bin.train --root_dir /tmp/unroller1 --conf distributed_sac_cartpole_conf.py --as_remote_unroller
```

image

TODOs in the future:
- [ ] loading/writing persistent replay buffer on disk
- [ ] support env with batch_size>1 for the unroller

### DIFF
--- a/alf/algorithms/distributed_off_policy_algorithm.py
+++ b/alf/algorithms/distributed_off_policy_algorithm.py
@@ -37,6 +37,8 @@ from alf.utils import dist_utils
 class UnrollerMessage(object):
     # unroller indicates end of experience for the current segment
     EXP_SEG_END = 'unroller: last_seg_exp'
+    # confirmation
+    OK = 'unroller: ok'
 
 
 def get_local_ip():
@@ -248,6 +250,7 @@ def pull_params_from_trainer(shared_dict: dict, unroller_id: str):
     while True:
         shared_dict['params'] = socket.recv()
         shared_dict['params_updated'] = True
+        socket.send_string(UnrollerMessage.OK)
 
 
 @alf.configurable(whitelist=[
@@ -328,17 +331,36 @@ class DistributedTrainer(DistributedOffPolicyAlgorithm):
     def is_main_ddp_rank(self):
         return self._ddp_rank == 0
 
-    def _send_params_to_unroller(self, unroller_id: str):
-        # Need to first receive a message from the unroller so that
-        # send_multipart
+    def _send_params_to_unroller(self, unroller_id: str) -> bool:
+        """Send model params to a specified unroller.
+
+        Args:
+            unroller_id: id (bytes str) of the unroller.
+
+        Returns:
+            bool: True if the unroller is still alive.
+        """
         # Get all parameters/buffers in a state dict and send them out
         buffer = io.BytesIO()
         torch.save(self._core_alg.state_dict(), buffer)
         self._params_socket.send_multipart(
             [unroller_id + b'_params',
              buffer.getvalue()])
-        logging.debug(
-            f"[worker-0] Params sent to unroller {unroller_id.decode()}.")
+        success = False
+        for _ in range(100):  # 1s in total for acknowledgement
+            try:
+                # In case some unrollers might die, we don't want to block forever
+                _, message = self._params_socket.recv_multipart(
+                    flags=zmq.NOBLOCK)
+                assert message == UnrollerMessage.OK.encode()
+                logging.debug(
+                    f"[worker-0] Params sent to unroller {unroller_id.decode()}."
+                )
+                success = True
+                break
+            except zmq.Again:
+                time.sleep(0.01)
+        return success
 
     def _create_unroller_registration_thread(self):
         self._new_unroller_ips_and_ports = mp.Queue()
@@ -445,8 +467,13 @@ class DistributedTrainer(DistributedOffPolicyAlgorithm):
         if (self.is_main_ddp_rank and alf.summary.get_global_counter() %
                 self._push_params_every_n_iters == 0):
             # Sending params to all the connected unrollers.
+            dead_unrollers = []
             for unroller_id in self._connected_unrollers:
-                self._send_params_to_unroller(unroller_id)
+                if not self._send_params_to_unroller(unroller_id):
+                    dead_unrollers.append(unroller_id)
+            # remove dead unrollers
+            for unroller_id in dead_unrollers:
+                self._connected_unrollers.remove(unroller_id)
 
         return steps
 
--- a/alf/algorithms/distributed_off_policy_algorithm.py
+++ b/alf/algorithms/distributed_off_policy_algorithm.py
@@ -33,6 +33,7 @@ from alf.experience_replayers.replay_buffer import ReplayBuffer
 from alf.data_structures import Experience, make_experience
 from alf.utils.per_process_context import PerProcessContext
 from alf.utils import dist_utils
+from alf.utils.summary_utils import record_time
 
 
 class UnrollerMessage(object):
@@ -170,7 +171,6 @@ class DistributedOffPolicyAlgorithm(OffPolicyAlgorithm):
     ###############################
     ######### Forward calls #######
     ###############################
-    @alf.utils.common.mark_eval
     def predict_step(self, inputs, state):
         return self._core_alg.predict_step(inputs, state)
 
@@ -186,6 +186,13 @@ class DistributedOffPolicyAlgorithm(OffPolicyAlgorithm):
     def preprocess_experience(self, root_inputs, rollout_info, batch_info):
         return self._core_alg.preprocess_experience(root_inputs, rollout_info,
                                                     batch_info)
+
+    def transform_experience(self, experience: Experience):
+        # Global data transformer
+        experience = super().transform_experience(experience)
+        # In the case where core_alg has in-alg data transformer
+        experience = self._core_alg.transform_experience(experience)
+        return experience
 
     def after_update(self, root_inputs, info):
         return self._core_alg.after_update(root_inputs, info)
@@ -528,8 +535,9 @@ class DistributedTrainer(DistributedOffPolicyAlgorithm):
             logging.debug(f"Rank {self._ddp_rank} sends params to unrollers "
                           f"{self._unrollers_to_update_params}")
             for unroller_id in self._unrollers_to_update_params:
-                if not self._send_params_to_unroller(unroller_id):
-                    dead_unrollers.append(unroller_id)
+                with record_time("time/trainer_send_params_to_unroller"):
+                    if not self._send_params_to_unroller(unroller_id):
+                        dead_unrollers.append(unroller_id)
             # remove dead unrollers
             for unroller_id in dead_unrollers:
                 self._unrollers_to_update_params.remove(unroller_id)
--- a/alf/algorithms/distributed_off_policy_algorithm.py
+++ b/alf/algorithms/distributed_off_policy_algorithm.py
@@ -1,0 +1,541 @@
+# Copyright (c) 2024 Horizon Robotics and ALF Contributors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import Callable
+import threading
+import multiprocessing as mp
+import zmq
+import time
+import io
+import subprocess
+from absl import logging
+
+import torch
+
+import alf
+from alf.algorithms.off_policy_algorithm import OffPolicyAlgorithm
+from alf.algorithms.config import TrainerConfig
+from alf.environments.alf_environment import AlfEnvironment
+from alf.experience_replayers.replay_buffer import ReplayBuffer
+from alf.data_structures import Experience, make_experience
+from alf.trainers import policy_trainer
+from alf.utils.per_process_context import PerProcessContext
+from alf.utils import dist_utils
+
+
+def get_local_ip():
+    """Get the ip address of the local machine."""
+    return subprocess.check_output(["hostname",
+                                    "-I"]).decode().strip().split()[0]
+
+
+@alf.configurable
+class UnrollerAddrConfig(object):
+    """A simple class for configuring the address of the unroller."""
+
+    def __init__(self, ip: str = 'localhost', port: int = 50000):
+        """
+        Args:
+            ip: ip address of the unroller.
+            port: port number used by the unroller.
+        """
+        self.ip = ip
+        self.port = port
+
+
+_unroller_addr_config = UnrollerAddrConfig()
+
+
+def create_zmq_socket(type: int, ip: str, port: int, id: str = None):
+    """A helper function for creating a ZMQ socket.
+
+    Args:
+        type: type of the socket.
+        ip: ip address. If it's '*', then `socket.bind()` will be used.
+        port: port number.
+        id: identity of the socket (optional). Only required for DEALER
+            sockets.
+
+    Returns:
+        tuple:
+        - socket: used for sending/receiving messages
+        - ZMQ context
+    """
+    cxt = zmq.Context()
+    socket = cxt.socket(type)
+    if id is not None:
+        socket.identity = id.encode('utf-8')
+    addr = 'tcp://' + ':'.join([ip, str(port)])
+    if ip == '*':
+        socket.bind(addr)
+    else:
+        socket.connect(addr)
+    return socket, cxt
+
+
+class DistributedOffPolicyAlgorithm(OffPolicyAlgorithm):
+    def __init__(self,
+                 *args,
+                 port: int = 50000,
+                 core_alg_ctor: Callable = OffPolicyAlgorithm,
+                 env: AlfEnvironment = None,
+                 config: TrainerConfig = None,
+                 optimizer: alf.optimizers.Optimizer = None,
+                 checkpoint: str = None,
+                 debug_summaries: bool = False,
+                 name: str = "DistributedOffPolicyAlgorithm",
+                 **kwargs):
+        """
+        Args:
+            config: the global ``TrainerConfig`` instance. The user is required
+                to always specify this argument.
+            port: port number for communication on the *current* machine.
+            core_alg_ctor: creates the algorithm to be wrapped by this class.
+            env: The environment to interact with. Its batch size must be 1.
+            optimizer: optimizer for the training the core algorithm.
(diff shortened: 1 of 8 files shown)


alf/algorithms/distributed_off_policy_algorithm.py
```diff
@@ -0,0 +1,541 @@
...
+        self._current_worker = 0
+
+    @property
+    def has_offline(self):
+        """Hardcode this flag to train without creating an online replay buffer.
+
+        Because the unroller never creates an online replay buffer, we need this
+        hacked flag for to perform training in ``_train_iter_off_policy()``.
+        """
+        return True
```
emailweixu: Instead of overriding has_offline() and train_from_replay_buffer(), I feel just override train_iter() is easier to understand.

alf/algorithms/distributed_off_policy_algorithm.py
```diff
@@ -0,0 +1,541 @@
...
+    Args:
+        m: a torch module whose parameters will be sent to the unroller.
+        port: port number used by the trainer.
+        lock: a lock to prevent concurrent read/write on the params.
+    """
+    socket, _ = create_zmq_socket(zmq.REP, '*', port)
+    while True:
+        try:
+            message = socket.recv_string(flags=zmq.NOBLOCK)
+            if message == "unroller: update":
```
emailweixu: It's better to use a global variable to store the message name, so that it can be shared at difference code locations.

alf/algorithms/distributed_off_policy_algorithm.py
```diff
@@ -0,0 +1,541 @@
...
+        steps = super()._train_iter_off_policy()
+        self._total_updates += self._config.num_updates_per_train_iter
+        return steps
+
+    def _backward_and_gradient_update(self, loss):
+        if self.is_main_ddp_rank:
+            # Lock the params to avoid sending them to the unroller while being updated.
+            # We only need to lock for the main DDP rank because the other DDP ranks
+            # will sync with it.
+            with self._params_lock:
```
emailweixu: This is not reliable. There are other things besides parameters which can be changed during model training (e.g. various buffers). Instead of allowing sending parameters at anytime, it is more reliable to only send at the end of an iteration.

alf/algorithms/distributed_off_policy_algorithm.py
```diff
@@ -0,0 +1,541 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import Callable
+import threading
+import multiprocessing as mp
```
emailweixu: Should be torch.multiprocessing?

alf/algorithms/distributed_off_policy_algorithm.py
```diff
@@ -0,0 +1,541 @@
...
+            f'worker-0: {PerProcessContext().num_processes}, {trainer_ip}, {self._port}'
+        )
+        message = socket.recv_string()
+        assert message == 'unroller: ok'
+        socket.close()
+        cxt.term()
+
+    @property
+    def utd(self):
+        with self._total_exps_received.get_lock():
```
emailweixu: the total number of experiences can be obtained from `replay_buffer.get_current_position().sum()`

alf/algorithms/distributed_off_policy_algorithm.py
```diff
@@ -0,0 +1,541 @@
...
+        trainer_ip = get_local_ip()
+        socket.send_string(
+            f'worker-0: {PerProcessContext().num_processes}, {trainer_ip}, {self._port}'
+        )
+        message = socket.recv_string()
+        assert message == 'unroller: ok'
+        socket.close()
+        cxt.term()
+
+    @property
```
emailweixu: I feel @property should be used only for some very simple stuff.

alf/algorithms/distributed_off_policy_algorithm.py
hnyu: > Instead of allowing sending parameters at anytime, it is more reliable to only send at the end of an iteration.

I tried this method in the first place. But there is something tricky here. Imagine that rank 0 is responsible for sending the params to the unroller at an end of some training iter. When some other rank gets stuck due to `replay_buffer_not_ready` or `utd_exceeded`, it won't be able to perform backward. So rank 0 might also get stuck in DDP backward sync. If at this moment the unroller requests the param update, it will get stuck waiting for the reply from rank 0. Now if the unroller gets stuck, other ranks won't get more unroll data, and they won't get out of `replay_buffer_not_ready` or `utd_exceeded` loop. So the entire system have a deadlock. 

I haven't figured out a way how to solve this issue except using a separate thread to respond to the unroller.

hnyu: > This is not reliable. There are other things besides parameters which can be changed during model training (e.g. various buffers). Instead of allowing sending parameters at anytime, it is more reliable to only send at the end of an iteration.

Now changed to trainer pushing params to unrollers and each unroller has a thread to receive params, without blocking unroll().

hnyu:
Added several new majors changes:
1. Trainer pushes params at some interval, so there is no race condition between param update and param sending. Unroller uses a thread to receive the param, without blocking unroll().
2. Assuming only one trainer machine but multiple unroller machines, now we specify the trainer ip instead of unroller ip.
3. Support multiple unrollers. Because two unrollers might want to send exp data to the same DDP rank at the same time, each DDP rank maintains a table of unrollers' exp data. Only when one unroller finishes the current episode, its data is stored into the replay buffer. 

I have updated the PR description to include the new changes.

alf/algorithms/distributed_off_policy_algorithm.py
```diff
@@ -0,0 +1,658 @@
...
+            daemon=True)
+        process.start()
+
+    def utd(self):
+        total_exps = int(self._replay_buffer.get_current_position().sum())
+        if total_exps == 0:
+            return 0
+        return self._total_updates / total_exps
+
+    def _train_iter_off_policy(self):
```
emailweixu, line 501: Can just override train_iter() since train_iter is documented in algorithm_interface.py as part of the interface.

alf/algorithms/distributed_off_policy_algorithm.py
```diff
@@ -0,0 +1,676 @@
...
+        cxt.term()
+
+    def _create_pull_params_subprocess(self):
+        # Compute the total size of the params
+        buffer = io.BytesIO()
+        torch.save(self._opt_free_state_dict(), buffer)
+        size = len(buffer.getvalue())
+        # Create a shared memory object to store the new params
+        # The first char indicates whether the params have been updated
+        self._params = SharedMemory(create=True, size=size + 1, name='params')
```
emailweixu: It's better to make a more unique name to void conflict.

alf/algorithms/distributed_off_policy_algorithm.py
```diff
@@ -0,0 +1,676 @@
...
+            self._current_worker = (
+                self._current_worker + 1) % self._num_trainer_workers
+
+    def _check_paramss_update(self) -> bool:
+        """Returns True if params have been updated.
+        """
+        # Check if the params have been updated
+        if bytes(self._params.buf[:1]) == b'1':
+            params = bytes(self._params.buf[1:])
+            buffer = io.BytesIO(params)
```
emailweixu: `bytes()` will copy the memory. Use `io.BytesIO(self._params.buf[1:]` to avoid copy.

alf/algorithms/distributed_off_policy_algorithm.py
```diff
@@ -0,0 +1,676 @@
...
+            state_dict = torch.load(buffer, map_location='cpu')
+            # We might only update part of the params
+            self._core_alg.load_state_dict(state_dict, strict=False)
+            logging.debug("Params updated from the trainer.")
+            self._params.buf[:1] = b'0'
```
emailweixu: Can use `self._params.buf[:1]=b'\0'` here, and use `if self._params.buf[0]==1` at line 647

alf/algorithms/distributed_off_policy_algorithm.py
```diff
@@ -0,0 +1,676 @@
...
+        for i in range(self._num_trainer_workers):
+            register_socket.send_string('init')
+
+        register_socket.close()
+        cxt.term()
+
+    def _create_pull_params_subprocess(self):
+        # Compute the total size of the params
+        buffer = io.BytesIO()
+        torch.save(self._opt_free_state_dict(), buffer)
```
emailweixu, line 642: Does `torch.save` guarantee same size every time?

alf/algorithms/distributed_off_policy_algorithm.py
```diff
@@ -0,0 +1,676 @@
...
+
+        steps = super()._train_iter_off_policy()
+        self._total_updates += self._config.num_updates_per_train_iter
+
+        if (self.is_main_ddp_rank and alf.summary.get_global_counter() %
+                self._push_params_every_n_iters == 0):
+            # Sending params to all the connected unrollers.
+            dead_unrollers = []
+            for unroller_id in self._connected_unrollers:
+                if not self._send_params_to_unroller(unroller_id):
```
emailweixu: Distribute sending among workers: `if urroller_id % num_workers == ddp_rank`

hnyu: > Distribute sending among workers: `if urroller_id % num_workers == ddp_rank`

We will detect if an unroller dies in `self._send_params_to_unroller`. If yes, the trainer should no longer send params to it. So It's relatively easier to manage this information in just one rank.

alf/algorithms/distributed_off_policy_algorithm.py
hnyu: > Does `torch.save` guarantee same size every time?

Without optimizer state, I think so.

alf/algorithms/distributed_off_policy_algorithm.py
hnyu: > Distribute sending among workers: `if urroller_id % num_workers == ddp_rank`

Updated. But this minor feature actually requires a significant code change as I need to let each unroller know which DDP rank (port) to connect to for params syncing. This port number is transmitted from the DDP rank during registration.

alf/algorithms/distributed_off_policy_algorithm.py
```diff
@@ -0,0 +1,705 @@
...
+_trainer_addr_config = TrainerAddrConfig()
+_params_port_offset = 100
+_unroller_port_offset = 1000
+
+
+def create_zmq_socket(type: int, ip: str, port: int, id: str = None):
+    """A helper function for creating a ZMQ socket.
+
+    Args:
+        type: type of the socket.
```
emailweixu: Comment what are the valid values for `type`

alf/algorithms/distributed_off_policy_algorithm.py
```diff
@@ -0,0 +1,705 @@
...
+        # 2. params port: self._port + _params_port_offset
+
+        self._max_utd_ratio = max_utd_ratio
+
+        # overwrite ``observe_for_replay`` to make sure it is never called
+        # by the parent ``RLAlgorithm``
+        self.observe_for_replay = self._observe_for_replay
+
+        self._params_socket, _ = create_zmq_socket(
+            zmq.ROUTER, '*', self._port + _params_port_offset + self._ddp_rank)
```
emailweixu, line 350: This port is different from the comment at line 327

alf/algorithms/distributed_off_policy_algorithm.py
```diff
@@ -0,0 +1,705 @@
...
+        process.start()
+
+    def utd(self):
+        total_exps = int(self._replay_buffer.get_current_position().sum())
+        if total_exps == 0:
+            return 0
+        return self._total_updates / total_exps
+
+    def _train_iter_off_policy(self):
+        if not self._daemons_started:
```
emailweixu: For continual training from an existing checkpoint, the checkpoint is loaded after the first `train_iter()` call in policy trainer. See `policy_trainer.Trainer._restore_checkpoint()`

alf/algorithms/distributed_off_policy_algorithm.py
```diff
@@ -0,0 +1,705 @@
...
+                            unroller_id, first_time=True)
+
+                    total_unrollers += 1
+
+        thread = threading.Thread(target=_wait_unroller_registration)
+        thread.daemon = True
+        thread.start()
+
+    def _create_data_receiver_subprocess(self):
+        """Create a proc to receive experience data from unrollers.
```
emailweixu: proc -> process

alf/algorithms/distributed_off_policy_algorithm.py
```diff
@@ -0,0 +1,705 @@
...
+        # we need to explicitly set the start method to fork, so that the daemon
+        # subprocess can share torch modules.
+        mp.set_start_method('fork', force=True)
+        # start the data receiver subprocess
+        process = mp.Process(
+            target=receive_experience_data,
+            args=(self._replay_buffer, self._new_unroller_ips_and_ports,
+                  self._ddp_rank),
+            daemon=True)
+        process.start()
```
emailweixu, line 493: It seems that the subprocess will keep running even after the main process is terminated?
Need to have a mechanism for terminating it.

alf/algorithms/distributed_off_policy_algorithm.py
```diff
@@ -0,0 +1,705 @@
...
+        # Initialize the update char to False (not updated)
+        self._shared_alg_params.buf[0] = 0
+
+        mp.set_start_method('fork', force=True)
+        process = mp.Process(
+            target=pull_params_from_trainer,
+            args=(self._shared_alg_params.name, self._id,
+                  self._params_socket_k),
+            daemon=True)
+        process.start()
```
emailweixu, line 657: need to have mechanism for terminating the subprocess after the main process ends

alf/algorithms/distributed_off_policy_algorithm.py
```diff
@@ -0,0 +1,705 @@
...
+                unroller_exps_buffer[unroller_id].append(exp_params)
+        else:
+            time.sleep(0.1)
+
+
+def pull_params_from_trainer(memory_name: str, unroller_id: str,
+                             params_socket_k: int):
+    """ Once new params arrive, we put it in the shared memory and mark updated.
+    Later after the current unroll finishes, the unroller can load the
+    new params.
```
emailweixu, line 267: what is params_socket_k?

alf/algorithms/distributed_off_policy_algorithm.py
hnyu: > For continual training from an existing checkpoint, the checkpoint is loaded after the first `train_iter()` call in policy trainer. See `policy_trainer.Trainer._restore_checkpoint()`

Updated to create the thread and subprocess at iter=1.

alf/algorithms/distributed_off_policy_algorithm.py
hnyu: > It seems that the subprocess will keep running even after the main process is terminated? Need to have a mechanism for terminating it.

The daemon subprocess will terminate whenever the main process exits?

alf/algorithms/distributed_off_policy_algorithm.py
hnyu: > what is params_socket_k?

Updated the variable name and docstring.

alf/algorithms/distributed_off_policy_algorithm.py
emailweixu: record the time spent on sending parameter using `with record_time()` outside of the `if` statement

alf/algorithms/distributed_off_policy_algorithm.py
emailweixu: Both Google search and ChatGPT says the contrary.

hnyu: > Both Google search and ChatGPT says the contrary.

It's easy to verify this:

```python
import time
import multiprocessing as mp


def worker():
    while True:
        print("worker")
        time.sleep(1)


proc = mp.Process(target=worker, daemon=True)
proc.start()

time.sleep(2.)
print("Main process terminates")
```

And I've been using this feature for quite a long time.

alf/algorithms/distributed_off_policy_algorithm.py
hnyu: > record the time spent on sending parameter using `with record_time()` outside of the `if` statement

Added

alf/algorithms/distributed_off_policy_algorithm.py
emailweixu: You are right. It seems that there are two types of daemon processes.

alf/algorithms/distributed_off_policy_algorithm.py
```diff
@@ -528,8 +535,9 @@ def _train_iter_off_policy(self):
             logging.debug(f"Rank {self._ddp_rank} sends params to unrollers "
                           f"{self._unrollers_to_update_params}")
             for unroller_id in self._unrollers_to_update_params:
-                if not self._send_params_to_unroller(unroller_id):
-                    dead_unrollers.append(unroller_id)
+                with record_time("time/trainer_send_params_to_unroller"):
```
emailweixu: should be outside the "if" at line 532 so that it can correctly compute the average time spent for each iteration.

emailweixu (approved):

PR #1713, opened 2024-12-04
SplitBatchRunner

emailweixu:
Sometime we want to use a large batch for training but the memory is not enough. In this kind of situation, we can split the large batch into smaller batches and use lean_function to run the small batches. This can reduces the memory usage because the intermediate tensors will all be released.

So SplitBatchRunner is implemented to make this process very simple.

### DIFF
--- a/alf/utils/lean_function.py
+++ b/alf/utils/lean_function.py
@@ -306,8 +306,12 @@ def _infer_device_type(*args):
 class SplitBatchRunner(torch.nn.Module):
     """Split the input into smaller batches and run the model on each batch.
 
+    ``SplitBatchRunner`` uses ``LeanFunction`` to run the model on smaller batches
+    so that the intermediate results can be discarded to save memory.
+
     Note that models using random number generators (e.g. DropOut) are not supported
-    for training.
+    for training. For models involving calculating statistics on a batch, using
+    SplitBatchRunner may lead to different results.
 
     Example:
 
@@ -318,7 +322,9 @@ class SplitBatchRunner(torch.nn.Module):
 
     Args:
         model (nn.Module): the model to run
-        max_batch_size (int): the maximum batch size to run the model.
+        max_batch_size (int): the maximum batch size to run the model. If negative,
+            the model will not be split into smaller batches and the original
+            forward of the model will be called.
     """
 
     def __init__(self, model, max_batch_size):
--- a/alf/utils/lean_function.py
+++ b/alf/utils/lean_function.py
@@ -301,3 +301,56 @@ def _infer_device_type(*args):
         return "cuda"
     else:
         return device_types[0]
+
+
+class SplitBatchRunner(torch.nn.Module):
+    """Split the input into smaller batches and run the model on each batch.
+
+    Note that models using random number generators (e.g. DropOut) are not supported
+    for training.
+
+    Example:
+
+    .. code-block:: python
+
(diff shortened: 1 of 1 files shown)


alf/utils/lean_function.py
```diff
@@ -301,3 +301,56 @@ def _infer_device_type(*args):
...
+        batch_size = alf.nest.get_nest_batch_size((args, kwargs))
+        if self._max_batch_size <= 0 or batch_size <= self._max_batch_size:
+            return self._model._original_forward_for_lean_function(
+                *args, **kwargs, **non_batched_inputs)
+
+        outputs = []
+        for i in range(0, batch_size, self._max_batch_size):
+            batch_args, batch_kwargs = alf.nest.map_structure(
+                lambda x: x[i:i + self._max_batch_size], (args, kwargs))
+            outputs.append(
```
qiang-sean-liu, line 352: are we supposed to have the same results from the full batch and the split batch? what if there are dropouts and batch norm in the model?

emailweixu: As the doc-string notes, dropout is not supported for training.
BatchNorm will also be affected. But if the batch_size is not too small, its effect on training should be small.

alf/utils/lean_function.py
```diff
@@ -301,3 +301,56 @@ def _infer_device_type(*args):
...
+    Example:
+
+    .. code-block:: python
+
+            model = SplitBatchRunner(model, max_batch_size=256)
+            output = model(input, non_batched_inputs={'mask': mask})
+
+    Args:
+        model (nn.Module): the model to run
+        max_batch_size (int): the maximum batch size to run the model.
```
hnyu: Need to add a comment that if max_batch_size <=0, what will happen.

alf/utils/lean_function.py
```diff
@@ -301,3 +301,56 @@ def _infer_device_type(*args):
...
+        outputs = []
+        for i in range(0, batch_size, self._max_batch_size):
+            batch_args, batch_kwargs = alf.nest.map_structure(
+                lambda x: x[i:i + self._max_batch_size], (args, kwargs))
+            outputs.append(
+                self._model(*batch_args, **batch_kwargs, **non_batched_inputs))
+
+        return alf.nest.map_structure(lambda *x: torch.cat(x, dim=0), *outputs)
+
+    def original_forward(self, *args, **kwargs):
```
hnyu, line 357: Where is this function used?

alf/utils/lean_function.py
```diff
@@ -301,3 +301,56 @@ def _infer_device_type(*args):
         return "cuda"
     else:
         return device_types[0]
+
+
+class SplitBatchRunner(torch.nn.Module):
+    """Split the input into smaller batches and run the model on each batch.
```
hnyu, line 307: Need to clarify that lean_function will be used.

emailweixu: comment added

alf/utils/lean_function.py
emailweixu: added

alf/utils/lean_function.py
emailweixu: added notes

alf/utils/lean_function.py
emailweixu: This is used by user of this class.

hnyu (approved):

PR #1714, opened 2024-12-09
Use override=True for config1 in adjust_config_by_multi_process_divid…

emailweixu:
…er()

Sometimes, those configurations are provided through commandline. In general, configurations provided through commandline will not be changed by config1(). However, for DDP training, even if these configurations are from commandline, we still need to change them.

### DIFF
--- a/alf/config_util.py
+++ b/alf/config_util.py
@@ -395,7 +395,8 @@ def config1(config_name,
         override_all (bool): If True, the value of the config will be set regardless
             of any pre-existing ``mutable`` or ``sole_init`` settings. This should
             be used only when absolutely necessary (e.g., adjusting certain configs
-            such as mini_batch_size for DDP workers.)."""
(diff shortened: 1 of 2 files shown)


alf/config_util.py
```diff
@@ -391,15 +392,33 @@ def config1(config_name,
             where the student must override certain configs inherited from the
             teacher). If the config is immutable, a warning will be declared with
             no changes made.
-    """
+        override_all (bool): If True, the value of the config will be set regardless
+            of any pre-existing ``mutable`` or ``sole_init`` settings. This should
+            be used only when absolutely necessary (e.g., adjusting certain configs
+            such as mini_batch_size for DDP workers.)."""
```
QuantuMope: Formatting Nit: Add a newline before """

QuantuMope (dismissed):
LGTM

QuantuMope (approved):

PR #1715, opened 2024-12-16
Speedup replay buffer add_batch

emailweixu:
Specialize it for the case when all the evironments unroll the same number of steps.

This change reduces time for observe_for_replay from 5.59s to 1.08s on total training time of 131s. And about 20% speedup on an actual job.

### DIFF
--- a/alf/utils/data_buffer.py
+++ b/alf/utils/data_buffer.py
@@ -118,6 +118,7 @@ class RingBuffer(nn.Module):
             self._enqueued = None
 
         buffer_id = [0]
+        self._env_ids = torch.arange(self._num_envs, device=device)
 
         def _create_buffer(spec_path, tensor_spec):
             buf = tensor_spec.zeros((num_environments, max_length))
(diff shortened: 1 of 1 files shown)


alf/utils/data_buffer.py
```diff
@@ -245,22 +246,38 @@ def _enqueue(self, batch, env_ids=None):
...
+                # fast path for the common case
+                pos = self.circular(current_pos)
 
-            current_pos = self._current_pos[env_ids]
-            indices = env_ids * self._max_length + self.circular(current_pos)
-            alf.nest.map_structure(
-                lambda buf, bat: buf.__setitem__(indices, bat.detach()),
-                self._flattened_buffer, batch)
+                def _set(buf, bat):
+                    buf[:, pos] = bat
```
le-horizon: bat.detach() to be consistent with the other branch?

alf/utils/data_buffer.py
```diff
@@ -245,22 +246,38 @@ def _enqueue(self, batch, env_ids=None):
...
-            # Make sure that there is no duplicate in `env_id`
-            # torch.unique(env_ids, return_counts=True)[1] is the counts for each unique item
-            assert torch.unique(
-                env_ids, return_counts=True)[1].max() == 1, (
-                    "There are duplicated ids in env_ids %s" % env_ids)
+            current_pos = self._current_pos[0].item()
+            if (env_ids.numel() == self._num_envs
+                    and (self._current_pos == current_pos).all()
+                    and (env_ids == self._env_ids).all()):
+                # fast path for the common case
```
le-horizon: explain a bit more in the comment, that this common case is when all envs have the same current_pos, and when all envs unroll the same number of steps.

le-horizon (dismissed):
Nice speed up.  Only two nits.

le-horizon (approved):
LG

PR #1716, opened 2024-12-16
Speedup metrics.py

emailweixu:
When there are lots of env_info items, this can take a lot time.

The change reduces time of observe_for_metrics from 3.47s to 2.54s on total 131s training time

Speedup mainly comes from directly implementing the queue logic in MetricBuffer instead of relying on DataBuffer, which is more complex due to its general functionality.


le-horizon (approved):
LG.  Could you briefly comment on where the speed up comes from, in the PR desc?  Thanks.

PR #1719, opened 2024-12-20
set a global flag for remote training

hnyu:
Sometimes we want to have the information of which remote mode a conf will use, and write the configuration code accordingly. So this PR defines such a config value.

### DIFF
--- a/alf/bin/train.py
+++ b/alf/bin/train.py
@@ -122,9 +122,9 @@ def _setup_device(rank: int = 0):
 def _setup_remote_configs_if_needed():
     """Preconfig some configurations for remote training and unrolling.
 
-    There will also be a global config flag named '_USER.remote_training' set in
-    this function, n case the user needs this info in the conf. The flag has one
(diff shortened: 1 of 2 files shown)


alf/bin/train.py
```diff
@@ -121,6 +121,13 @@ def _setup_device(rank: int = 0):
 
 def _setup_remote_configs_if_needed():
     """Preconfig some configurations for remote training and unrolling.
+
+    There will also be a global config flag named '_USER.remote_training' set in
```
emailweixu: _USER should be used for user defined configs. Perhaps we can put this into TrainerConfig.

alf/bin/train.py
```diff
@@ -121,6 +121,13 @@ def _setup_device(rank: int = 0):
...
+    this function, n case the user needs this info in the conf. The flag has one
```
emailweixu: n case => in case

emailweixu (approved):

PR #1720, opened 2024-12-24
[Small] Update yapf precommit hook

QuantuMope:
The current yapf repo we are using is deprecated and as of `python3.12`, fails to work. Updating the pre-commit yaml to a more modern, supported version.


emailweixu:
Need to update the docker to use newer pre-commit

QuantuMope:
Had to run `pre-commit run --all-files` because the style seems to have changed significantly from v0.25 to v0.43

emailweixu:
I notice that Hobot is using yapf 0.28.0.  We should use the same for both. Otherwise it will create headache since both of them will be in the same python environment.

emailweixu:
> I notice that Hobot is using yapf 0.28.0. We should use the same for both. Otherwise it will create headache since both of them will be in the same python environment.

Also https://github.com/HorizonRobotics/alf/blob/pytorch/docs/contributing.rst says yapf 0.28.0.

QuantuMope:
> I notice that Hobot is using yapf 0.28.0. We should use the same for both. Otherwise it will create headache since both of them will be in the same python environment.

Hobot should also be using yapf 0.43. I changed it a while ago

https://github.com/HorizonRoboticsInternal/Hobot/blob/5e7a6b0db524980af484159b6092c7f92544de04/.pre-commit-config.yaml#L7-L11

emailweixu:
> > I notice that Hobot is using yapf 0.28.0. We should use the same for both. Otherwise it will create headache since both of them will be in the same python environment.
> 
> Hobot should also be using yapf 0.43. I changed it a while ago
> 
> https://github.com/HorizonRoboticsInternal/Hobot/blob/5e7a6b0db524980af484159b6092c7f92544de04/.pre-commit-config.yaml#L7-L11

But the CICD docker uses 0.28.0: https://github.com/HorizonRoboticsInternal/gail-container/blob/main/hobot_cicd/requirements.txt

QuantuMope:
> > > I notice that Hobot is using yapf 0.28.0. We should use the same for both. Otherwise it will create headache since both of them will be in the same python environment.
> > 
> > 
> > Hobot should also be using yapf 0.43. I changed it a while ago
> > https://github.com/HorizonRoboticsInternal/Hobot/blob/5e7a6b0db524980af484159b6092c7f92544de04/.pre-commit-config.yaml#L7-L11
> 
> But the CICD docker uses 0.28.0: https://github.com/HorizonRoboticsInternal/gail-container/blob/main/hobot_cicd/requirements.txt

Oh I see. Actually I believe that since we specify a non-local repo for the yapf version in pre-commit, the CICD will use that one as well. I.e., the local version pip installed (0.28.0) is never used.

This is likely to be true since the CICD complained about all the code in this PR just from [dcf5583](https://github.com/HorizonRobotics/alf/pull/1720/commits/dcf558313d772c25dba8edbd9735e1dd3e69b1f1)

You can see below CICD seems to pull yapf from the google repo.

image

@hnyu maybe you can confirm?

emailweixu (dismissed):

QuantuMope:
Sorry forgot to merge and had to rebase.

emailweixu (approved):

PR #1721, opened 2025-01-13
add max size for tmp exp buffer

hnyu:
To prevent CPU mem overflow, this PR adds a cap to the tmp exp buffer for each unroller. When the cap is reached, earliest exps will be dropped with a warning message.

### DIFF
--- a/alf/algorithms/distributed_off_policy_algorithm.py
+++ b/alf/algorithms/distributed_off_policy_algorithm.py
@@ -204,8 +204,7 @@ class DistributedOffPolicyAlgorithm(OffPolicyAlgorithm):
 
 def receive_experience_data(replay_buffer: ReplayBuffer,
                             new_unroller_ips_and_ports: mp.Queue,
-                            worker_id: int,
-                            max_tmp_buffer_size: int = 2000) -> None:
+                            worker_id: int, max_tmp_buffer_size: int) -> None:
     """A worker function for consistently receiving experience data from
     unrollers.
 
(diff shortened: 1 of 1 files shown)


alf/algorithms/distributed_off_policy_algorithm.py
```diff
@@ -221,6 +223,11 @@ def receive_experience_data(replay_buffer: ReplayBuffer,
             new unrollers.
         worker_id: the id of the worker; used by each unroller to route the
             experience data.
+        max_tmp_buffer_size: the maximum size of the temporary buffer for storing
+            experience data from each unroller. If the buffer is full, the
+            experience data will be dropped from the beginning and a warning
+            message will be output. This is for CPU memory safety consideration.
+            When the warning occurs, we should reconsider unroller's
```
emailweixu: incomplete sentence

alf/algorithms/distributed_off_policy_algorithm.py
```diff
@@ -203,7 +204,8 @@ def after_train_iter(self, root_inputs, rollout_info):
 
 def receive_experience_data(replay_buffer: ReplayBuffer,
                             new_unroller_ips_and_ports: mp.Queue,
-                            worker_id: int) -> None:
+                            worker_id: int,
+                            max_tmp_buffer_size: int = 2000) -> None:
```
emailweixu: It seems this cannot be changed if it turns out 2000 is not enough. Since the RL-in-rea environment will not have concept of episode, we cannot wait until episode end to store data to replay buffer.

hnyu: > It seems this cannot be changed if it turns out 2000 is not enough. Since the RL-in-rea environment will not have concept of episode, we cannot wait until episode end to store data to replay buffer.

I discussed with Haichao today. There will be a future change of getting the step type from rollout_info on the unroller side if the algorithm wants to set that information. For now we assume the env will set the LAST step type. 

I've made this variable configurable.

hnyu:
Closing this PR as I will let the unroller periodically set StepType.LAST to switch between trainer workers, instead of always waiting for env's last step.

PR #1722, opened 2025-01-16
Pr rlpd

runjerry:
Extended two algorithms from SAC algorithm
- RLPD: the pure off-policy version of RLPD
- RLPD2: RLPD with critics optimization uncertainty estimation and training reweighting

Both verified on the hopper:run environment of DM Control.


runjerry:
Open a separate [PR](https://github.com/HorizonRobotics/alf/pull/1727) for RLPD. Will PR RLPD2 separately after further validation.

PR #1725, opened 2025-01-17
distunroller set last step periodically

hnyu:
This PR let the DistributedUnroller to truncate the experience stream on its own. The stream is truncated if either the predefined max episode length is reached, or the env returns a LAST step.

After the stream is truncated, the unroller will switch to sending exps to a different trainer worker (if available).

This PR is dependent on PR #1723 which fixes a ReplayBuffer sharing issue among processes. Now I've added a minimal test to the init of DistributedTrainer to make sure that ReplayBuffer can correctly shared with a subprocess.

### DIFF
--- a/alf/algorithms/distributed_off_policy_algorithm.py
+++ b/alf/algorithms/distributed_off_policy_algorithm.py
@@ -255,6 +255,8 @@ def receive_experience_data(replay_buffer: ReplayBuffer,
 
             if int(exp_params.step_type) == StepType.LAST:
                 # Add the temp exp buffer to the replay buffer
+                # ``DistributedOffPolicyAlgorithm`` assumes batch_size=1
+                exp_params.env_id.zero_()
                 for i, exp_params in enumerate(
                         unroller_exps_buffer[unroller_id]):
                     replay_buffer.add_batch(exp_params, exp_params.env_id)
--- a/alf/algorithms/distributed_off_policy_algorithm.py
+++ b/alf/algorithms/distributed_off_policy_algorithm.py
@@ -362,7 +362,10 @@ class DistributedTrainer(DistributedOffPolicyAlgorithm):
         self._num_earliest_frames_ignored = self._core_alg._num_earliest_frames_ignored
 
         # We always test tensor sharing among processes, because
-        # we rely on undocumented features of PyTorch
+        # we rely on undocumented features of PyTorch:
+        # 1. tensors will automatically be moved to shared memory, even without
+        #    ``Module.share_memory()`` or ``Tensor.share_memory_()`` being called.
+        # 2. only a 'spawned' subprocess is reliable for tensor sharing.
         _test_tensor_sharing()
 
     def _observe_for_replay(self, exp: Experience):
--- a/alf/algorithms/distributed_off_policy_algorithm.py
+++ b/alf/algorithms/distributed_off_policy_algorithm.py
@@ -16,6 +16,7 @@ from absl import logging
 from typing import Callable
 import time
 import io
+import queue
 import random
 import threading
 import subprocess
@@ -24,21 +25,21 @@ import zmq
 import torch
 import torch.multiprocessing as mp
 from multiprocessing.shared_memory import SharedMemory
+from multiprocessing import Manager
 
 import alf
 from alf.algorithms.off_policy_algorithm import OffPolicyAlgorithm
 from alf.algorithms.config import TrainerConfig
 from alf.environments.alf_environment import AlfEnvironment
 from alf.experience_replayers.replay_buffer import ReplayBuffer
-from alf.data_structures import Experience, make_experience
+from alf.data_structures import Experience, make_experience, StepType
 from alf.utils.per_process_context import PerProcessContext
 from alf.utils import dist_utils
 from alf.utils.summary_utils import record_time
+from alf.utils.common_test import _test_tensor_sharing
 
 
 class UnrollerMessage(object):
-    # unroller indicates end of experience for the current segment
-    EXP_SEG_END = 'unroller: last_seg_exp'
     # confirmation
     OK = 'unroller: ok'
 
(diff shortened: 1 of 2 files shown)


alf/algorithms/distributed_off_policy_algorithm.py
```diff
@@ -662,6 +677,35 @@ def observe_for_replay(self, exp: Experience):
...
+                exp = alf.nest.set_field(
+                    exp, 'time_step.step_type',
+                    torch.tensor([StepType.FIRST], dtype=torch.int32))
+            self._is_first_step = False
+
+        if episode_end:
+            # One episode finishes; move to the next worker
+            # We need to make sure a whole episode is always sent to the same
+            # worker so that the temporal information is preserved.
+            exp = alf.nest.set_field(
```
emailweixu, line 706: In the case of a single trainer workers, we don't need to change the step type to LAST.

alf/algorithms/distributed_off_policy_algorithm.py
```diff
@@ -672,26 +716,21 @@ def observe_for_replay(self, exp: Experience):
         buffer = io.BytesIO()
         torch.save(exp_params, buffer)
 
-        worker_id = f'worker-{self._current_worker}'
+        if self._exp_socket is None:
+            self._exp_socket, _ = create_zmq_socket(zmq.ROUTER, '*',
+                                                    self._port, self._id)
+
         try:
```
emailweixu, line 728: should send only for LAST step or episode length reached.

alf/algorithms/distributed_off_policy_algorithm.py
```diff
@@ -358,6 +358,13 @@ def __init__(self,
         # may be incremented every mini-batch
         self._num_train_iters = 0
 
+        # respect core_alg's replay buffer setting
+        self._num_earliest_frames_ignored = self._core_alg._num_earliest_frames_ignored
+
+        # We always test tensor sharing among processes, because
+        # we rely on undocumented features of PyTorch
```
emailweixu: Explain what the undocumented feature is.

alf/utils/common_test.py
```diff
@@ -73,5 +75,66 @@ def test_get_all_parameters(self):
...
+
+    start_method = mp.get_start_method()
+    mp.set_start_method('spawn', force=True)
+    # Change ``m`` in the child process
+    process = mp.Process(target=_test_worker, args=(m, ))
+    process.start()
+    process.join()
+
+    # numpy array should not be modified
+    assert np.allclose(m.y, np.zeros([2]))
```
emailweixu: should be equal instead of close

alf/algorithms/distributed_off_policy_algorithm.py
hnyu: > In the case of a single trainer workers, we don't need to change the step type to LAST.

If there are multiple unrollers, we still need to set LAST. But it's not straightforward for an unroller to know if there is any other unroller, unless via the trainer. So for simplicity, here we always set LAST.

alf/algorithms/distributed_off_policy_algorithm.py
hnyu: > should send only for LAST step or episode length reached.

Right now, we always send on a per-exp basis, instead of waiting for a long traj. The trainer is responsible for maintaining the traj integrity. The reason is for latency concern, because sending a very long traj might take a long time (especially with images), blocking the unroller.

alf/algorithms/distributed_off_policy_algorithm.py
hnyu: > Explain what the undocumented feature is.

added explanation

alf/utils/common_test.py
hnyu: > should be equal instead of close

updated

alf/algorithms/distributed_off_policy_algorithm.py
```diff
@@ -241,21 +239,26 @@ def receive_experience_data(replay_buffer: ReplayBuffer,
...
+            # Use a temp buffer to store the received exps
+            if unroller_id not in unroller_exps_buffer:
+                unroller_exps_buffer[unroller_id] = []
+            unroller_exps_buffer[unroller_id].append(exp_params)
+
+            if int(exp_params.step_type) == StepType.LAST:
                 # Add the temp exp buffer to the replay buffer
-                for exp_params in unroller_exps_buffer[unroller_id]:
+                for i, exp_params in enumerate(
+                        unroller_exps_buffer[unroller_id]):
```
emailweixu, line 261: If the batch size of the replay buffer is 1. env_id has to be 0 at the next line

hnyu: > If the batch size of the replay buffer is 1. env_id has to be 0 at the next line

This is true for the current assumption. But since exp_params always contains env_id, we can just use it. Do you mean we should assert it's equal to 0?

emailweixu: We can just set it to 0 here?

hnyu: > We can just set it to 0 here?

updated

alf/algorithms/distributed_off_policy_algorithm.py
```diff
@@ -241,21 +239,27 @@ def receive_experience_data(replay_buffer: ReplayBuffer,
...
+            # Use a temp buffer to store the received exps
+            if unroller_id not in unroller_exps_buffer:
+                unroller_exps_buffer[unroller_id] = []
+            unroller_exps_buffer[unroller_id].append(exp_params)
+
+            if int(exp_params.step_type) == StepType.LAST:
                 # Add the temp exp buffer to the replay buffer
-                for exp_params in unroller_exps_buffer[unroller_id]:
-                    replay_buffer.add_batch(exp_params, exp_params.env_id)
+                env_id = torch.zeros([1], dtype=torch.int32)
```
emailweixu: device="cpu"

emailweixu: Should change `exp_arams.env_id` instead, since exp_params.env_id will be stored in replay buffer and we dont' want inconsistency.
```python
exp_params.env_id.zero_()
```

emailweixu (approved):

PR #1727, opened 2025-02-03
RLPD algorithm

runjerry:
Extend SAC to RLPD. Tested on 5 DM control tasks with 4 seeds each to confirmed its performance improvement over SAC.

### DIFF
--- a/alf/algorithms/rlpd_algorithm.py
+++ b/alf/algorithms/rlpd_algorithm.py
@@ -57,7 +57,7 @@ class RlpdAlgorithm(SacAlgorithm):
                  normalize_entropy_reward=False,
                  calculate_priority=False,
                  num_critic_replicas=2,
-                 num_critic_targets=2,
+                 num_sampled_critic_targets=2,
                  env=None,
                  config: TrainerConfig = None,
                  critic_loss_ctor=None,
@@ -78,12 +78,10 @@ class RlpdAlgorithm(SacAlgorithm):
                  name="RlpdAlgorithm"):
         # **kwargs):
         """
-        Refer to SacAlgorithm for more details for kwargs
+        Refer to SacAlgorithm for more details for other arguments.
 
-        Args:
-            name (str): The name of this algorithm.
-            num_critic_targets (int): Number of sampled subset of target critics
-                for computing TD target in critic training.
+            num_sampled_critic_targets (int): Number of sampled subset of target 
+                critics for computing TD target in critic training.
         """
         super().__init__(
             observation_spec=observation_spec,
@@ -120,8 +118,11 @@ class RlpdAlgorithm(SacAlgorithm):
 
         assert self._act_type == ActionType.Continuous, (
             "RLPD algorithm only supports continuous action spaces.")
+        assert num_sampled_critic_targets <= num_critic_replicas, (
+            "The number of sampled target critics should be less than or equal to"
+            "the number of replicas.")
 
-        self._num_critic_targets = num_critic_targets
+        self._num_sampled_critic_targets = num_sampled_critic_targets
 
     def _compute_critics(self,
                          critic_net,
@@ -131,6 +132,16 @@ class RlpdAlgorithm(SacAlgorithm):
                          replica_consensus='mean',
                          sample_subset=False,
                          apply_reward_weights=True):
+        """
+        The following two arguments are different from the super class.
+
+            replica_consensus (str): the methods to consensus the possibly
+                multiple (due to critic replicas) critic_net outputs. Options
+                are ['mean', 'min', None].
+            sample_subset (bool): whether to sample a subset of critics outputs 
+                before applying the ``replica_consensus``. This is only used
+                for sampling a subset of target critics for computing TD target. 
+        """
         observation = (observation, action)
         # critics shape [B, replicas]
         critics, critics_state = critic_net(observation, state=critics_state)
@@ -143,17 +154,12 @@ class RlpdAlgorithm(SacAlgorithm):
             critics = critics.reshape(-1, self._num_critic_replicas,
                                       *self._reward_spec.shape,
                                       *remaining_shape)
-            if self._act_type == ActionType.Discrete:
-                # permute: [B, replicas, reward_dim, num_actions]
-                #       -> [B, replicas, num_actions, reward_dim]
-                order = [0, 1, -1] + list(
-                    range(2, 2 + len(self._reward_spec.shape)))
-                critics = critics.permute(*order)
 
-        if sample_subset:
+        if sample_subset and (self._num_sampled_critic_targets <
+                              self._num_critic_replicas):
             critics = critics[:,
-                              torch.randperm(self._num_critic_replicas
-                                             )[:self._num_critic_targets], ...]
+                              torch.randperm(self._num_critic_replicas)
+                              [:self._num_sampled_critic_targets], ...]
 
         if replica_consensus == 'min':
             if self.has_multidim_reward():
--- a/alf/algorithms/rlpd_algorithm.py
+++ b/alf/algorithms/rlpd_algorithm.py
@@ -479,6 +479,6 @@ class RlpdAlgorithm(SacAlgorithm):
             priority=priority,
             extra=critic_loss / float(self._num_critic_replicas))
 
-    def after_train_iter(self, inputs: TimeStep, info: RlpdInfo):
-        self._periodic_reset()
+    def after_update(self, root_inputs, info: RlpdInfo):
         self._update_train_mode()
+        super().after_update(root_inputs, info)
--- a/alf/algorithms/rlpd_algorithm.py
+++ b/alf/algorithms/rlpd_algorithm.py
@@ -168,8 +168,11 @@ class RlpdAlgorithm(SacAlgorithm):
             else:
                 critics = critics.min(dim=1)[0]
         elif replica_consensus == 'mean':
-            sign = self.reward_weights.sign()
-            critics = (critics * sign).mean(dim=1) * sign
+            if self.has_multidim_reward():
+                sign = self.reward_weights.sign()
+                critics = (critics * sign).mean(dim=1) * sign
+            else:
+                critics = critics.mean(dim=1)
 
         if apply_reward_weights and self.has_multidim_reward():
             critics = self._apply_reward_weights(critics)
--- a/alf/algorithms/rlpd_algorithm.py
+++ b/alf/algorithms/rlpd_algorithm.py
@@ -1,0 +1,213 @@
+# Copyright (c) 2024 Horizon Robotics and ALF Contributors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""RLPD Algorithm."""
+
+import torch
+from typing import Callable, Optional, Union
+
+import alf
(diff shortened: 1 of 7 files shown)


alf/algorithms/rlpd_algorithm.py
```diff
@@ -0,0 +1,213 @@
...
+
+        # For multi-dim reward, do
+        # [B, replicas * reward_dim] -> [B, replicas, reward_dim]
+        # For scalar reward, do nothing
+        if self.has_multidim_reward():
+            remaining_shape = critics.shape[2:]
+            critics = critics.reshape(-1, self._num_critic_replicas,
+                                      *self._reward_spec.shape,
+                                      *remaining_shape)
+            if self._act_type == ActionType.Discrete:
```
emailweixu: According to `__init__()`, only `ActionType.Continuous` is supported

alf/algorithms/rlpd_algorithm.py
```diff
@@ -0,0 +1,213 @@
...
+            if self.has_multidim_reward():
+                sign = self.reward_weights.sign()
+                critics = (critics * sign).min(dim=1)[0] * sign
+            else:
+                critics = critics.min(dim=1)[0]
+        elif replica_consensus == 'mean':
+            if self.has_multidim_reward():
+                sign = self.reward_weights.sign()
+                critics = (critics * sign).mean(dim=1) * sign
+            else:
```
emailweixu: This branch is unnecessary for 'mean'

alf/algorithms/rlpd_algorithm.py
```diff
@@ -0,0 +1,213 @@
...
+                 dqda_clipping=None,
+                 actor_optimizer=None,
+                 critic_optimizer=None,
+                 alpha_optimizer=None,
+                 checkpoint=None,
+                 debug_summaries=False,
+                 name="RlpdAlgorithm"):
+        # **kwargs):
+        """
+        Refer to SacAlgorithm for more details for kwargs
```
emailweixu: for other arguments

alf/algorithms/rlpd_algorithm.py
```diff
@@ -0,0 +1,213 @@
...
+
+        Args:
+            name (str): The name of this algorithm.
```
emailweixu: can be removed.

alf/algorithms/rlpd_algorithm.py
```diff
@@ -0,0 +1,213 @@
...
+
+        self._num_critic_targets = num_critic_targets
+
+    def _compute_critics(self,
+                         critic_net,
+                         observation,
+                         action,
+                         critics_state,
+                         replica_consensus='mean',
+                         sample_subset=False,
```
Haichao-Zhang, line 256: can add some comments to ``sample_subset`` argument, and the reason using ``True`` for target critics and ``False`` for critics. Seems that using ``False`` for critics is to ensure all the critics have gradients.

alf/algorithms/rlpd_algorithm.py
```diff
@@ -0,0 +1,213 @@
...
+                 checkpoint=None,
+                 debug_summaries=False,
+                 name="RlpdAlgorithm"):
+        # **kwargs):
+        """
+        Refer to SacAlgorithm for more details for kwargs
+
+        Args:
+            name (str): The name of this algorithm.
+            num_critic_targets (int): Number of sampled subset of target critics
```
Haichao-Zhang: ``num_critic_targets``: maybe can change its name to reflect the ``sampled`` aspect? Currently, it is very close to the meaning of ``num_critic_replicas``

Haichao-Zhang: Should ensure that ``num_critic_targets <= num_critic_replicas``.

alf/algorithms/rlpd_algorithm.py
```diff
@@ -0,0 +1,213 @@
...
+                                      *remaining_shape)
+            if self._act_type == ActionType.Discrete:
+                # permute: [B, replicas, reward_dim, num_actions]
+                #       -> [B, replicas, num_actions, reward_dim]
+                order = [0, 1, -1] + list(
+                    range(2, 2 + len(self._reward_spec.shape)))
+                critics = critics.permute(*order)
+
+        if sample_subset:
+            critics = critics[:,
```
Haichao-Zhang, line 281: when ``self._num_critic_targets`` and ``num_critic_replicas`` are equal, the `randperm`` is not necessary and can be removed to save computation.

alf/algorithms/rlpd_algorithm.py
Haichao-Zhang: Also curious the motivation of the ``mean`` branch? Any benefits?

runjerry: > This branch is unnecessary for 'mean'

I feel that it is needed in order to get the mean critic values across all critics for actor training?

runjerry: > Also curious the motivation of the `mean` branch? Any benefits?

This is the default setting of RLPD to get a concensus q_value for actor training. Given that RLPD might use more than two critics, taking a `min` as in SAC would be too conservative. Moreover, from my previous experience, as long as critics are trained with conservative target critic values, there is no need to be conservative in actor training.

alf/algorithms/rlpd_algorithm.py
runjerry: Good point, updated.

alf/algorithms/rlpd_algorithm.py
runjerry: Removed.

alf/algorithms/rlpd_algorithm.py
runjerry: Updated.

alf/algorithms/rlpd_algorithm.py
runjerry: Updated.

alf/algorithms/rlpd_algorithm.py
runjerry: Removed.

alf/algorithms/rlpd_algorithm.py
runjerry: Updated.

alf/algorithms/rlpd_algorithm.py
Haichao-Zhang: I see. Thanks!

Haichao-Zhang (dismissed):
LGTM

alf/algorithms/rlpd_algorithm.py
emailweixu: > > This branch is unnecessary for 'mean'
> 
> I feel that it is needed in order to get the mean critic values across all critics for actor training?

`(critics * sign).mean(dim=1) * sign` is same as `critics.mean(dim=1)`

runjerry: Oh I see, thanks, updated.

runjerry: I had to revert to keep this branch since `self.reward_weights` by default is `None` for scalar reward.

emailweixu: Why can't you change the whole branch as the following?
```python
elif replica_consensus == 'mean':
    critics = critics.mean(dim=1)
```

runjerry: Oh yes, that should work well.

alf/algorithms/rlpd_algorithm.py
```diff
@@ -0,0 +1,219 @@
...
+                 target_update_period: Union[int, Scheduler] = 1,
+                 parameter_reset_period: Union[int, Scheduler] = -1,
+                 dqda_clipping=None,
+                 actor_optimizer=None,
+                 critic_optimizer=None,
+                 alpha_optimizer=None,
+                 checkpoint=None,
+                 debug_summaries=False,
+                 name="RlpdAlgorithm"):
+        # **kwargs):
```
Haichao-Zhang: can remove this line?

runjerry: Removed.

runjerry:
Pushed an updated version that works consistently better than previous version. Please take a look.

Haichao-Zhang (approved):
LGTM

alf/algorithms/rlpd_algorithm.py
```diff
@@ -34,13 +51,46 @@ class RlpdAlgorithm(SacAlgorithm):
...
+
+    2. Besides critics UTD, the actor UTD is also configurable. The actor and the critics
+    are updated alternatively, with each being updated successively for (its) UTD number
+    of times. This is realized by designing three train modes for RLPD, as defined
+    by ``TrainMode``. For each train_iter, train_step can only be in one mode,
+
+    - By default, both actor_utd and critic_utd are None, train_mode is initialized as 
+      ``standard``, and will be kept fixed throughout the entire training process. 
+      Under the ``standard`` mode, both the actor and the critics will be updated
+      simultaneously once per train_iter.
```
emailweixu, line 70: once -> TrainerConfig.num_updates_per_iter

alf/algorithms/rlpd_algorithm.py
```diff
@@ -122,7 +183,69 @@ def __init__(self,
             "The number of sampled target critics should be less than or equal to"
             "the number of replicas.")
 
+        if actor_utd is None and critic_utd is None:
+            self._train_mode = TrainMode.standard
+        else:
+            total_utd = alf.config_util.get_config_value(
+                "num_updates_per_train_iter")
```
emailweixu, line 190: Use "TrainerConfig.num_updates_per_train_iter" to be safer

alf/algorithms/rlpd_algorithm.py
```diff
@@ -122,7 +183,69 @@ def __init__(self,
...
+            if critic_utd is not None:
```
emailweixu, line 191: assert actor_utd is None

alf/algorithms/rlpd_algorithm.py
```diff
@@ -122,7 +183,69 @@ def __init__(self,
...
+                assert critic_utd < total_utd, (
+                    "critic_utd should be less than num_updates_per_train_iter"
+                )
+                actor_utd = total_utd - critic_utd
+            else:
```
emailweixu, line 196: assert critic_utd is None

alf/algorithms/rlpd_algorithm.py
```diff
@@ -214,6 +339,146 @@ def _critic_train_step(self, observation, target_observation,
...
+        There are two modifications compared with SAC's ``_calc_critic_loss``:
+
+        1. Skip the critic loss computation if in actor train mode.
+
+        2. Apply bootstrap mask to critic loss if ``use_bootstrap_critic``.
+        """
+        if self._train_mode == TrainMode.actor:
+            return LossInfo()
+
+        if self._use_entropy_reward:
```
emailweixu, line 440: Perhaps we can use super()._calc_critic_loss() and apply mask to its result.

alf/algorithms/rlpd_algorithm.py
runjerry: I am working on another PR and have updated here.

alf/algorithms/rlpd_algorithm.py
runjerry: Updated in the new PR.

alf/algorithms/rlpd_algorithm.py
runjerry: The idea was to let critic_utd (if provided) override actor_utd, I have updated the docstring in the beginning of this algorithm  to reflect this design choice.

alf/algorithms/rlpd_algorithm.py
runjerry: If entering this branch, critic_utd must be None and actor _critic must not?

alf/algorithms/rlpd_algorithm.py
runjerry: A concern I had for calling the SacAlgorithm._calc_critic_loss() was that some unnecessary computations will be executed, say if ``self._use_entropy_reward`` or ``self._calculate_priority`` is True.

PR #1728, opened 2025-02-18
request video clips

Haichao-Zhang:
This PR adds supports to alf for requesting video clips online.

image

image


alf/trainers/policy_trainer.py
```diff
@@ -317,6 +317,14 @@ def train(self):
                 "Use `kill -%s %s` to request checkpoint during training." %
                 (int(signal.SIGUSR2), self._pid))
 
+        self._video_clip_requested = False
+        if threading.current_thread() == threading.main_thread():
+            signal.signal(signal.SIGRTMIN, self._request_video_clip)
+            # kill -34 PID
+            logging.info(
+                "Use `kill -%s %s` to request video-clip during training." %
```
emailweixu: Also include information about where is the video file located.

Haichao-Zhang: Thanks for the comment. Added. Now will output info as follows:

image

emailweixu (approved):

PR #1729, opened 2025-02-21
Improve prev_action docstring

Haichao-Zhang:
Make the meaning of the ``time_step.prev_action`` more clear.

### DIFF
--- a/alf/data_structures.py
+++ b/alf/data_structures.py
@@ -99,9 +99,9 @@ class TimeStep(
     - reward: a ``Tensor`` of reward values from executing 'prev_action'.
     - discount: A discount value in the range :math:`[0, 1]`.
     - observation: A (nested) ``Tensor`` for observation.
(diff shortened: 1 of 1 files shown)


alf/data_structures.py
```diff
@@ -99,7 +99,9 @@ class TimeStep(
     - reward: a ``Tensor`` of reward values from executing 'prev_action'.
     - discount: A discount value in the range :math:`[0, 1]`.
     - observation: A (nested) ``Tensor`` for observation.
-    - prev_action: A (nested) ``Tensor`` for action from previous time step.
+    - prev_action: A (nested) ``Tensor`` for action from previous time step. Note that
```
hnyu: "for action generated for the previous time step"

Haichao-Zhang: Updated comments

hnyu (approved):

PR #1730, opened 2025-03-05
unroller_only option for DistributedUnroller

Haichao-Zhang:
Add an option to skip the training related steps and do unroll only. Therefore, it won't be blocked by the trainer and is useful for visualizing the unroll behaviors without training and without the need to start a distributed trainer.

### DIFF
--- a/alf/algorithms/distributed_off_policy_algorithm.py
+++ b/alf/algorithms/distributed_off_policy_algorithm.py
@@ -561,7 +561,8 @@ class DistributedTrainer(DistributedOffPolicyAlgorithm):
         return steps
 
 
-@alf.configurable(whitelist=['episode_length', 'name', 'optimizer'])
+@alf.configurable(
+    whitelist=['episode_length', 'name', 'optimizer', 'unroller_only'])
 class DistributedUnroller(DistributedOffPolicyAlgorithm):
     def __init__(self,
                  core_alg_ctor: Callable,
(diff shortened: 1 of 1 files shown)


alf/algorithms/distributed_off_policy_algorithm.py
```diff
@@ -784,5 +795,6 @@ def train_iter(self):
             torch.cuda.empty_cache()
         # Experience will be sent to the trainer in this function
         self._unroll_iter_off_policy()
-        self._check_paramss_update()
+        if not self._unroller_only:
+            self._check_paramss_update()
```
hnyu: What about making the logic simpler? 
```python
# Copied from super().train_iter()
if self._config.empty_cache:
    torch.cuda.empty_cache()
if self._unroller_only:
    self._unroller_iter_off_policy()
    return 0

if not self._registered:
    ....
self._unroll_iter_off_policy()
self._check_params_update()
return 0
```

Haichao-Zhang: Good suggestion. Done.

hnyu (approved):

alf/algorithms/distributed_off_policy_algorithm.py
```diff
@@ -776,7 +776,15 @@ def train_iter(self):
         There is actually no training happening in this function. But the unroller
         will check if there are updated params available.
         """
-        if not self._registered and not self._unroller_only:
+        # Copied from super().train_iter()
+        if self._config.empty_cache:
+            torch.cuda.empty_cache()
+
+        if self._unroller_only:
+            self._unroller_iter_off_policy()
```
hnyu, line 784: Should be `self._unroll_iter_off_policy()`

PR #1732, opened 2025-03-13
shared_train_eval_env

Haichao-Zhang:
Provides an option to train and eval on the same shared env, mimicking the case such as training in real with only one physical env.

### DIFF
--- a/alf/trainers/evaluator.py
+++ b/alf/trainers/evaluator.py
@@ -74,10 +74,14 @@ class Evaluator(object):
                       pre_configs, num_envs, config.root_dir, seed))
             self._worker.start()
         else:
-            self._env = create_environment(
-                for_evaluation=True,
-                num_parallel_environments=num_envs,
-                seed=seed)
+            if config.shared_train_eval_env:
+                self._env = alf.get_env()
+                self._env.reset()
+            else:
+                self._env = create_environment(
+                    for_evaluation=True,
+                    num_parallel_environments=num_envs,
+                    seed=seed)
             self._evaluator = SyncEvaluator(self._env, config)
 
     def eval(self, algorithm: RLAlgorithm, step_metric_values: Dict[str, int]):
--- a/alf/trainers/evaluator.py
+++ b/alf/trainers/evaluator.py
@@ -75,6 +75,7 @@ class Evaluator(object):
             self._worker.start()
         else:
             if config.shared_train_eval_env:
+                assert not self._async, "should not use async_eval in shared_train_eval_env mode"
                 self._env = alf.get_env()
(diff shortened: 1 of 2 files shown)


alf/trainers/evaluator.py
```diff
@@ -74,10 +74,14 @@ def __init__(self, config: TrainerConfig, conf_file: str):
                       pre_configs, num_envs, config.root_dir, seed))
             self._worker.start()
         else:
-            self._env = create_environment(
-                for_evaluation=True,
-                num_parallel_environments=num_envs,
-                seed=seed)
+            if config.shared_train_eval_env:
+                self._env = alf.get_env()
+                self._env.reset()
```
le-horizon, line 80: assert async_eval = False?

Haichao-Zhang: Good point. Added assertion.

le-horizon: Commit not pushed to the right remote?

Haichao-Zhang: ah, yes, that was what happened ... now pushed

le-horizon: Happened to me also.  It's hard to remember, especially now that we don't change alf that often.  We can probably remove the other remote.

alf/trainers/evaluator.py
```diff
@@ -74,10 +74,15 @@ def __init__(self, config: TrainerConfig, conf_file: str):
...
             self._worker.start()
         else:
-            self._env = create_environment(
-                for_evaluation=True,
-                num_parallel_environments=num_envs,
-                seed=seed)
+            if config.shared_train_eval_env:
+                assert not self._async, "should not use async_eval in shared_train_eval_env mode"
+                self._env = alf.get_env()
+                self._env.reset()
```
le-horizon, line 80: Curious, why do we call env.reset() here but not in the other branch.  Maybe add a comment in the code?

alf/trainers/evaluator.py
```diff
@@ -74,10 +74,15 @@ def __init__(self, config: TrainerConfig, conf_file: str):
                       pre_configs, num_envs, config.root_dir, seed))
             self._worker.start()
         else:
-            self._env = create_environment(
-                for_evaluation=True,
-                num_parallel_environments=num_envs,
-                seed=seed)
+            if config.shared_train_eval_env:
```
emailweixu, line 77: Need to set the step_type in the replay buffer just before evaluation started to StepType.LAST.

emailweixu: Also need to set the next step type for training to FIRST

PR #1733, opened 2025-03-13
 Add prune_exp_replay_env_info function and prune env_info for distributed off-policy RL

Haichao-Zhang:
With this, we can have the flexibility to use customized env_info on the unroller side, without breaking the spec assumption on the trainer side.

### DIFF
--- a/alf/algorithms/distributed_off_policy_algorithm.py
+++ b/alf/algorithms/distributed_off_policy_algorithm.py
@@ -248,6 +248,14 @@ def receive_experience_data(replay_buffer: ReplayBuffer,
 
             buffer = io.BytesIO(message)
             exp_params = torch.load(buffer, map_location='cpu')
+            # we prune env_info according to the replay buffer for the following reasons:
+            # 1) avoid env_info mismatch and allow the distributed unroller to have
+            # a customized env_info for tb summarization,
+            # 2) reduce memory usage for the replay buffer
+            exp_params = alf.utils.common.prune_exp_replay_env_info(
+                exp_params,
+                env_info_spec=replay_buffer.data_spec.time_step.env_info)
+
             # Use a temp buffer to store the received exps
             if unroller_id not in unroller_exps_buffer:
                 unroller_exps_buffer[unroller_id] = []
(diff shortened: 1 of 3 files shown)


alf/algorithms/distributed_off_policy_algorithm.py
```diff
@@ -491,6 +491,12 @@ def _create_data_receiver_subprocess(self):
         exp = alf.utils.common.prune_exp_replay_state(
             exp, self._use_rollout_state, self.rollout_state_spec,
             self.train_state_spec)
+        # we do not save env_info to replay buffer for the following reasons:
+        # 1) avoid env_info mismatch and allow the distributed unroller to have
+        # a customized env_info for tb summarization, 2) reduce communication overhead,
+
+        # 3) reduce memory usage for the replay buffer
+        exp = alf.utils.common.prune_exp_replay_env_info(exp)
```
hnyu: There is no need to call it here? This exp is obtained from the env itself, so no extra env_info fields exist here.

Haichao-Zhang: Now we remove env_info completely before saving into replay buffer, not only the mis-match part.

hnyu: > Now we remove env_info completely before saving into replay buffer, not only the mis-match part.

Not sure about this, but did we preserve env_info in the replay buffer for a general usage scenario?

Haichao-Zhang: Now switched to prune, so only the additional part of env_info will be removed.

hnyu (approved):
It's slightly better to do the pruning on the unroller side so that we can save some communication time. But the current PR will also work.

Haichao-Zhang:
> It's slightly better to do the pruning on the unroller side so that we can save some communication time. But the current PR will also work.

That was the logic of my first design (when pruning all), but did not keep it as it is a bit less reliable in the case when not pruning all. In this case, it is safer to do this close to the end consumer side, as that is where the lease amount of assumptions on the specs required.

PR #1734, opened 2025-03-24
Support bf16 for AMP

emailweixu:
The range of fp16 is very limited. Sometimes we need bf16.


hnyu:
Should we not use GradScaler if dtype is bfloat16?

emailweixu:
> Should we not use GradScaler if dtype is bfloat16?

Now change to use GradScaler only for fp16 training.

hnyu (approved):

PR #1736, opened 2025-03-24
Use CUDA_VISIBLE_DEVICES instead of set_device to set device for DDP worker

emailweixu:
We use CUDA_VISIBLE_DEVICES instead torch.cuda.set_device(rank) in the training_worker to specify the GPU because we want to actually only make the specified GPU visible to the process. This is because some other user of cuda (e.g. Taichi) may only use the first GPU. If multiple GPUs are visible to the process, all the processes will compete for the first GPU.

### DIFF
--- a/alf/bin/train.py
+++ b/alf/bin/train.py
@@ -53,6 +53,8 @@ replace the "--gin_file" option with "--conf", and "--gin_param" with "--conf_pa
 from absl import app
 from absl import flags
 from absl import logging
+import datetime
+import multiprocessing
 import os
 import sys
 import torch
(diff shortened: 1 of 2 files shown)


alf/utils/distributed.py
```diff
@@ -115,7 +115,7 @@ def make_ddp_performer(module: torch.nn.Module,
     print(f'find_unused_parameters={find_unused_parameters}')
     return DDP(
         _MethodPerformer(module=module, perform=method),
-        device_ids=[ddp_rank],
+        device_ids=None,
```
hnyu, line 117: Maybe remove the argument 'ddp_rank' since it's no longer use?

alf/bin/train.py
```diff
@@ -120,7 +124,6 @@ def _setup_device(rank: int = 0):
     """
     if torch.cuda.is_available():
         alf.set_default_device('cuda')
```
hnyu, line 122: Remove the arg 'rank'?

alf/utils/distributed.py
emailweixu: removed

hnyu (dismissed):

hnyu (approved):

PR #1738, opened 2025-03-24
disable_amp option for CategoricalProjectionNetwork

emailweixu:
Add an option to disable amp for CategoricalProjectionNetwork as we may need to use float32 for it.

### DIFF
--- a/alf/networks/projection_networks.py
+++ b/alf/networks/projection_networks.py
@@ -39,6 +39,7 @@ class CategoricalProjectionNetwork(Network):
                  logits_init_output_factor=0.1,
                  weight_opt_args=None,
                  bias_opt_args=None,
+                 disable_amp: bool = False,
                  name="CategoricalProjectionNetwork"):
         """Creates a categorical projection network that outputs a discrete
         distribution over a number of classes.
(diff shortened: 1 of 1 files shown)


alf/networks/projection_networks.py
```diff
@@ -82,16 +84,24 @@ def __init__(self,
...
-        if len(self._output_shape) > 1:
-            return td.Independent(
-                td.Categorical(logits=logits),
-                reinterpreted_batch_ndims=len(self._output_shape) - 1), state
-        else:
-            return td.Categorical(logits=logits), state
+        amp_enabled = torch.is_autocast_enabled()
+        if self._disable_amp and amp_enabled:
+            inputs = alf.layers.to_float32(inputs)
+            amp_enabled = False
```
hnyu, line 94: I wonder did you have negative experience with bfloat16 for this network?

emailweixu: Sometime the range of fp16 may not be enough. But the precision of bf16 may be not enough for some other parts of the model

hnyu: > Sometime the range of fp16 may not be enough. But the precision of bf16 may be not enough for some other parts of the model

So you want to use fp16 for other places but exclude this network from using it?

emailweixu: yes

hnyu (approved):

PR #1739, opened 2025-03-27
Fix initial timestep for continual training from checkpoint

emailweixu:
Previously, when continuing training from a checkpoint, ALF will first run one iteration of training before loading the checkpoint. After the checkpoint is loaded, it will reset the environment and then start actual training. However, inside RLAlgorithm,         _current_time_step, _current_policy_state, _current_transform_state are for the last step of the initial iteration. They doesn't match the reseted environment. So we need clear _current_time_step, _current_policy_state, _current_transform_state so that they can obtain the correct value.


Haichao-Zhang (approved):
Thanks for the fix!

PR #1742, opened 2025-03-27
Pass conf and env into the core algorithm in DistributedOffPolicyAlgorithm

Haichao-Zhang:
This is necessary since the core algorithm sometimes need them (e.g. to get some values from config)


hnyu:
The reason I didn't pass is that once they are passed into the core, some data will be created inside the core alg. E.g, replay buffer. So there is some cost of doing that.

Haichao-Zhang:
> The reason I didn't pass is that once they are passed into the core, some data will be created inside the core alg. E.g, replay buffer. So there is some cost of doing that.

This was not documented in the original doc string.
Now changed to only pass in the config, since that is useful while does not have a bit side effect. It is used in the hub metric summary.

hnyu (approved):

PR #1743, opened 2025-03-28
env-step mode for evaluation

Haichao-Zhang:
Currently there is only an episode mode for evaluation.

This PR adds a new mode (env-step mode) for evaluation during training, by specifying env steps that should be spent on each evaluation.

### DIFF
--- a/alf/algorithms/config.py
+++ b/alf/algorithms/config.py
@@ -52,6 +52,7 @@ class TrainerConfig(object):
                  epsilon_greedy=0.,
                  eval_uncertainty=False,
                  num_eval_episodes=10,
+                 num_eval_steps=0,
                  num_eval_environments: int = 1,
                  async_eval: bool = True,
                  save_checkpoint_for_best_eval: Optional[Callable] = None,
@@ -225,7 +226,8 @@ class TrainerConfig(object):
                 help prevent a dead loop in some deterministic environment like
                 Breakout. Only used for evaluation.
             eval_uncertainty (bool): whether to evaluate uncertainty after training.
-            num_eval_episodes (int) : number of episodes for one evaluation.
+            num_eval_episodes (int): number of episodes for one evaluation.
+            num_eval_steps (int): number of environmental steps for evaluation.
             num_eval_environments: the number of environments for evaluation.
             async_eval: whether to do evaluation asynchronously in a different
                 process. Note that this may use more memory.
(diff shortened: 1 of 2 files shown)


alf/trainers/evaluator.py
```diff
@@ -410,18 +424,37 @@ def evaluate(env: AlfEnvironment,
...
+        env_episodes = torch.zeros(batch_size, dtype=torch.int32)
+        total_num = num_episodes
+    else:
+        episode_mode = False
+        num_eval_steps = num_steps()
+        assert num_eval_steps > 0
+        steps_per_env = (num_eval_steps + batch_size - 1) // batch_size
+        total_num = steps_per_env
+        # additional steps required after ``steps_per_env`` parallel steps over ``batch_size`` envs
+        additional_steps = num_eval_steps - steps_per_env * batch_size
```
emailweixu: It seems that this can never be positive? And I feel there is no need to handle this. The extra complexity does not really gain anything meaningful.

Haichao-Zhang: > It seems that this can never be positive? And I feel there is no need to handle this. The extra complexity does not really gain anything meaningful.

Agree. Simplified the logic in the new commit.

emailweixu (approved):

alf/algorithms/config.py
```diff
@@ -225,7 +226,8 @@ def __init__(self,
                 help prevent a dead loop in some deterministic environment like
                 Breakout. Only used for evaluation.
             eval_uncertainty (bool): whether to evaluate uncertainty after training.
-            num_eval_episodes (int) : number of episodes for one evaluation.
+            num_eval_episodes (int): number of episodes for one evaluation.
+            num_eval_steps (int): number of environmental steps for evaluation.
```
le-horizon, line 230: It may make more sense to let the user specify the number of steps during invoking the on-demand evaluation function.

Haichao-Zhang: Online eval request is not the whole story.

This PR (#1743) was implemented earlier and mainly served for the purpose of enabling **the auto evaluation** (the evaluation curve parallel to the training) even without the concept of ``episode`` on the env side.


The online eval request feature was something that came later (#1744). I decided not to change the interface of the Evaluator too much therefore I didn't pass in a number of steps for now. I can try to see if there is a better way to achieve this goal.

PR #1747, opened 2025-04-01
[Small] Fix function call mispell in DistributedUnroller

QuantuMope:
Ran into a small bug running prior on latest alf. This PR fixes a function mispell.


hnyu:
thx for the fix

hnyu (approved):

PR #1748, opened 2025-04-01
record finer-grained time for unroller

Haichao-Zhang:
This PR adds some finer-grained timers for unroller, in order to track the time spent on each step.

### DIFF
--- a/alf/algorithms/distributed_off_policy_algorithm.py
+++ b/alf/algorithms/distributed_off_policy_algorithm.py
@@ -769,13 +769,14 @@ class DistributedUnroller(DistributedOffPolicyAlgorithm):
         with self._shared_mem_lock:
             if self._shared_alg_params.buf[0] == 1:
                 with record_time(
-                        "time/unroller_params_update/1_buffer_from_io"):
+                        "time/dist_unroller_params_update/1_buffer_from_io"):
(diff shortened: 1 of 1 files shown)


hnyu:
These timers should be more like local debugging code?

Haichao-Zhang:
> These timers should be more like local debugging code?

Not only useful locally. I believe these metrics are very useful in general, since these are quite important to understand the system's bottleneck, and the time-wise performance of each component is not fixed, but changing according to settings (NN size, env implementations etc.)

hnyu:
> > These timers should be more like local debugging code?
> 
> Not only useful locally. I believe these metrics are very useful in general, since these are quite important to understand the system's bottleneck, and the time-wise performance of each component is not fixed, but changing according to settings (NN size, env implementations etc.)

OK can you change the summary name from "unroller" to "dist_unroller"?

Haichao-Zhang:
> > > These timers should be more like local debugging code?
> > 
> > 
> > Not only useful locally. I believe these metrics are very useful in general, since these are quite important to understand the system's bottleneck, and the time-wise performance of each component is not fixed, but changing according to settings (NN size, env implementations etc.)
> 
> OK can you change the summary name from "unroller" to "dist_unroller"?

Done.

hnyu (approved):

PR #1750, opened 2025-04-03
[Small] Support non-Scheduler num_steps in evaluate

QuantuMope:
Saw a minor bug going through alf where `num_steps` is assumed to be a `Scheduler`. This PR handles the case where it is an `int`.

### DIFF
--- a/alf/trainers/evaluator.py
+++ b/alf/trainers/evaluator.py
@@ -438,7 +438,10 @@ def evaluate(env: AlfEnvironment,
         total_num = num_episodes
     else:
         episode_mode = False
-        num_eval_steps = num_steps()
+        if isinstance(num_steps, Scheduler):
+            num_eval_steps = num_steps()
+        else:
+            num_eval_steps = num_steps
         assert num_eval_steps > 0
         # adjust the ``num_eval_steps`` so that all the envs will have the same number of steps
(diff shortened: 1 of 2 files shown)


Haichao-Zhang (commented):
I'm curious how is the num_steps specified in your case? Typically we specify it via conf, in which case it will be converted to scheduler already:
https://github.com/HorizonRobotics/alf/blob/5ae9db1aa64a3b3eee1ed86f90b81a1f89071eb0/alf/algorithms/config.py#L390

But I guess the type hint might not be consistent.
Another simpler way to fix (if we still want to keep the same type hint) is to add a single line: num_steps = as_scheduler(num_steps)

QuantuMope:
> I'm curious how is the num_steps specified in your case? Typically we specify it via conf, in which case it will be converted to scheduler already:
> 
> https://github.com/HorizonRobotics/alf/blob/5ae9db1aa64a3b3eee1ed86f90b81a1f89071eb0/alf/algorithms/config.py#L390
> 
> But I guess the type hint might not be consistent. Another simpler way to fix (if we still want to keep the same type hint) is to add a single line: num_steps = as_scheduler(num_steps)

I just saw the typehint mismatch show up on my IDE as a warning. Added the `as_scheduler` change.

Haichao-Zhang (approved):

PR #1751, opened 2025-04-04
Update setup.py and pre-commit

QuantuMope:
This PR updates the `setup.py` for alf so that installation is seamless for Ubuntu22.04 and 24.04 users. It relaxes version requirements on many packages to minimize using deprecated older software. I've also taken the liberty of updating some old pre-commit hooks and moving documentation based packages (`sphinx`) as optional extras. 

## Testing:

I've confirmed that this works on my personal Ubuntu 22.04 system on a fresh venv.
I've confirmed that the new setup.py works for fresh minimal versions of Ubuntu 22.04 (python3.10) and Ubuntu 24.04 (python3.12) using docker.


hnyu:
I think the question is, will this ALF work for older docker images with ubuntu 20.04?

QuantuMope:
> I think the question is, will this ALF work for older docker images with ubuntu 20.04?

Does this affect older docker images that are already confirmed to be working? I think should only affect local installs.

hnyu:
> > I think the question is, will this ALF work for older docker images with ubuntu 20.04?
> 
> Does this affect older docker images that are already confirmed to be working? I think should only affect local installs.

What I meant is that if we want to build a docker with Ubuntu 20.02 and ALF in it, probably the two won't be compatible.

QuantuMope:
Added static version numbers. Confirmed works on ubuntu 22 and 24 docker.

emailweixu (approved):

runjerry (approved):
Thank you for this very useful update!

PR #1752, opened 2025-04-04
Record the time cost by policy_step in unroll

Haichao-Zhang:


alf/algorithms/rl_algorithm.py
```diff
@@ -662,20 +662,25 @@ def _sync_unroll(self, unroll_length: int):
...
+            t0 = time.time()
             policy_step = self.rollout_step(transformed_time_step,
                                             policy_state)
+            policy_step_time += time.time() - t0
 
             action = common.detach(policy_step.output)
 
             t0 = time.time()
-            next_time_step = self._env.step(action)
+            with record_time("time/_sync_unroll/2_env_step"):
```
hnyu: Why is this called "2_env_step"?

Haichao-Zhang: It indicates that there is the second component within the _sync_unroll block. There should be another summary before this. Added.

hnyu (approved):

alf/algorithms/rl_algorithm.py
```diff
@@ -662,20 +662,26 @@ def _sync_unroll(self, unroll_length: int):
...
         for _ in range(unroll_length):
             policy_state = common.reset_state_if_necessary(
                 policy_state, initial_state, time_step.is_first())
             transformed_time_step, trans_state = self.transform_timestep(
                 time_step, trans_state)
-            policy_step = self.rollout_step(transformed_time_step,
-                                            policy_state)
+
+            t0 = time.time()
+            with record_time("time/_sync_unroll/1_per_rollout_step"):
```
hnyu, line 675: Actually just realized that `record_time` computes the average time spent of this code block for one training iteration. For example, if there are 100 rollout steps per iter, then `record_time` computes the sum of 100 rollout steps. So '1_per_rollout_step' name is confusing or did you actually mean to accumulate the time over rollout steps?

PR #1754, opened 2025-04-09
Make sure global_step is a torch tensor for torch >= 2.6.0

QuantuMope:
Starting from `torch >= 2.6.0`, `torch.load` sets `weights_only` from False to True by default. This results in the following error when attempting to load a model containing numpy arrays.

```
In PyTorch 2.6, we changed the default value of the weights_only argument in torch.load
from False to True. Re-running torch.load with weights_only set to False will
likely succeed, but it can result in arbitrary code execution. Do it only if you got
the file from a trusted source.
(2) Alternatively, to load with weights_only=True please check the recommended steps
in the following error message.
WeightsUnpickler error: Unsupported global: GLOBAL numpy.core.multiarray._reconstruct
was not an allowed global by default. Please use torch.serialization.add_safe_globals([_reconstruct])
or the torch.serialization.safe_globals([_reconstruct]) context manager to allowlist
this global if you trust this class/function.
```

After some investigation, I've found that this is caused by `global_step` not being stored as a torch tensor. This PR fixes this issue.

Note that older checkpoints will still not be able to run without running into this error. For those, people have to set `TORCH_FORCE_NO_WEIGHTS_ONLY_LOAD=1`.

### DIFF
--- a/alf/utils/checkpoint_utils.py
+++ b/alf/utils/checkpoint_utils.py
@@ -258,6 +258,7 @@ class Checkpointer(object):
         map_location = torch.device('cpu')
 
         checkpoint = torch.load(f_path, map_location=map_location)
+        checkpoint['global_step'] = checkpoint['global_step'].numpy()
         if including_optimizer:
             opt_checkpoint = torch.load(f_path + '-optimizer',
                                         map_location=map_location)
--- a/alf/utils/checkpoint_utils.py
+++ b/alf/utils/checkpoint_utils.py
@@ -364,7 +364,7 @@ class Checkpointer(object):
             optimizer_state[k] = opts
             replay_buffer_state[k] = rs
 
-        model_state['global_step'] = global_step
+        model_state['global_step'] = torch.tensor(global_step)
 
         torch.save(model_state, f_path)
         torch.save(optimizer_state, f_path + '-optimizer')


alf/utils/checkpoint_utils.py
```diff
@@ -17,13 +17,19 @@
...
 from torch import nn
 from typing import Optional, List
 import warnings
 
 import alf
 from alf.nest import map_structure
 
+# Starting from torch 2.6, torch.load() sets weights_only=True by default.
+# If a model contains numpy arrays, this will result in an error during loading.
+# Here, we'll allow numpy's _recontruct function.
```
hnyu: I thought torch Module doesn't track numpy arrays at all? Basically treating them as normal objects like others?

QuantuMope: Seems like that's not the case. This is a backward compatibility breaking change in torch 2.6.

https://dev-discuss.pytorch.org/t/bc-breaking-change-torch-load-is-being-flipped-to-use-weights-only-true-by-default-in-the-nightlies-after-137602/2573

I've added several "key" numpy classes and functions to the allow list, but as soon as a module uses something out of this list, we'll have to append it.

The other option is just to allow this list (which should suffice for most models) and then use `TORCH_FORCE_NO_WEIGHTS_ONLY_LOAD=1` for cases where it doesn't work.

hnyu: > Seems like that's not the case. This is a backward compatibility breaking change in torch 2.6.
> 
> https://dev-discuss.pytorch.org/t/bc-breaking-change-torch-load-is-being-flipped-to-use-weights-only-true-by-default-in-the-nightlies-after-137602/2573
> 
> I've added several "key" numpy classes and functions to the allow list, but as soon as a module uses something out of this list, we'll have to append it.
> 
> The other option is just to allow this list (which should suffice for most models) and then use `TORCH_FORCE_NO_WEIGHTS_ONLY_LOAD=1` for cases where it doesn't work.

I think my question is that we only call `torch.save` on model state dict which should contain tensors. If not, then there is some bug in obtaining the state dict. If we call `model.state_dict()` the returned dict shouldn't contain any numpy array because they are not tracked by torch.nn.Module.

QuantuMope: Will look into this.

runjerry:
> Starting from `torch >= 2.6.0`, `torch.load` sets `weights_only` from False to True by default. This results in the following error when attempting to load a model containing numpy arrays.
> 
> ```
> In PyTorch 2.6, we changed the default value of the weights_only argument in torch.load
> from False to True. Re-running torch.load with weights_only set to False will
> likely succeed, but it can result in arbitrary code execution. Do it only if you got
> the file from a trusted source.
> (2) Alternatively, to load with weights_only=True please check the recommended steps
> in the following error message.
> WeightsUnpickler error: Unsupported global: GLOBAL numpy.core.multiarray._reconstruct
> was not an allowed global by default. Please use torch.serialization.add_safe_globals([_reconstruct])
> or the torch.serialization.safe_globals([_reconstruct]) context manager to allowlist
> this global if you trust this class/function.
> ```
> 
> This PR allows numpy arrays during `torch.load()`.

Could you possibly also describe how this fix has been tested? I assume loading our previous ckpts work well with this PR on torch >= 2.6.0.

alf/utils/checkpoint_utils.py
QuantuMope: You were right. There was a numpy array culprit causing the issue. `global_step` was being saved as a non-tensor. Updated PR to simply convert this to a torch tensor which solves the issue.

QuantuMope:
> > Starting from `torch >= 2.6.0`, `torch.load` sets `weights_only` from False to True by default. This results in the following error when attempting to load a model containing numpy arrays.
> > ```
> > In PyTorch 2.6, we changed the default value of the weights_only argument in torch.load
> > from False to True. Re-running torch.load with weights_only set to False will
> > likely succeed, but it can result in arbitrary code execution. Do it only if you got
> > the file from a trusted source.
> > (2) Alternatively, to load with weights_only=True please check the recommended steps
> > in the following error message.
> > WeightsUnpickler error: Unsupported global: GLOBAL numpy.core.multiarray._reconstruct
> > was not an allowed global by default. Please use torch.serialization.add_safe_globals([_reconstruct])
> > or the torch.serialization.safe_globals([_reconstruct]) context manager to allowlist
> > this global if you trust this class/function.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This PR allows numpy arrays during `torch.load()`.
> 
> Could you possibly also describe how this fix has been tested? I assume loading our previous ckpts work well with this PR on torch >= 2.6.0.

Tested by running unroller via `alf.bin.train` and saving a checkpoint and then trying to run again, which will result in an error. Since prior has no actual weights, we know that the culprit must be some alf-related metrics. Confirmed after some digging that it is because our `global_step` parameter is not stored as a torch tensor. I've updated the PR to now solve this. You should be able to now run models without having to specify `load_weights=False`. Let me know if you can test on your side please.

alf/utils/checkpoint_utils.py
hnyu: > You were right. There was a numpy array culprit causing the issue. `global_step` was being saved as a non-tensor. Updated PR to simply convert this to a torch tensor which solves the issue.

Need to make sure after loading, the returned tensor global step is compatible with current code?

QuantuMope: It is. I tested this.

hnyu (dismissed):

alf/utils/checkpoint_utils.py
emailweixu: During load, it's better to change the loaded global_step to numpy array to avoid any potential problem.

QuantuMope: Done.

emailweixu (approved):

PR #1756, opened 2025-04-11
Replace torch.cuda.amp.autocast with torch.amp.autocast('cuda')

QuantuMope:
As of torch 2.6, `torch.cuda.amp.autocast` is now deprecated. This PR replaces all calls with `torch.amp.autocast('cuda')`. Confirmed that this is backward compatible with `torch >= 2.1` which is used by CICD.

### DIFF
--- a/alf/algorithms/algorithm.py
+++ b/alf/algorithms/algorithm.py
@@ -1888,8 +1888,9 @@ class Algorithm(AlgorithmInterface):
             weight (float): weight for this batch. Loss will be multiplied with
                 this weight before calculating gradient.
         """
-        with torch.cuda.amp.autocast(self._config.enable_amp,
-                                     dtype=self._config.amp_dtype):
+        with torch.amp.autocast('cuda',
+                                enabled=self._config.enable_amp,
+                                dtype=self._config.amp_dtype):
             train_info, loss_info = self._compute_train_info_and_loss_info(
                 experience)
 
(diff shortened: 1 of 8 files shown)


runjerry:
I feel a tricky thing is that some of our cluster docker images were built with older pytorch versions where torch has no amp attribute.

QuantuMope:
> I feel a tricky thing is that some of our cluster docker images were built with older pytorch versions where torch has no amp attribute.

I think the oldest torch version is 2.1 on our cluster dockers, which I've confirmed works with this change.

Are there dockers that have older versions than this?

runjerry:
> > I feel a tricky thing is that some of our cluster docker images were built with older pytorch versions where torch has no amp attribute.
> 
> I think the oldest torch version is 2.1 on our cluster dockers, which I've confirmed works with this change.
> 
> Are there dockers that have older versions than this?

I think "horizonrobotics/cuda:11.3-cudnn8-py3.8-ubuntu20.04_kincpp" is using torch 1.11+cu11.3, this is the docker I am currently running for env v1, but I think I will switch to v2 soon. I also noticed some @Haichao-Zhang 's running jobs using this docker image, but not sure if he is still actively using it.

QuantuMope:
> > > I feel a tricky thing is that some of our cluster docker images were built with older pytorch versions where torch has no amp attribute.
> > 
> > 
> > I think the oldest torch version is 2.1 on our cluster dockers, which I've confirmed works with this change.
> > Are there dockers that have older versions than this?
> 
> I think "horizonrobotics/cuda:11.3-cudnn8-py3.8-ubuntu20.04_kincpp" is using torch 1.11+cu11.3, this is the docker I am currently running for env v1, but I think I will switch to v2 soon. I also noticed some @Haichao-Zhang 's running jobs using this docker image, but not sure if he is still actively using it.

Got it. We can just hold off on this PR for now and merge it when we need it.

PR #1758, opened 2025-04-17
a context manager to restore nn.Module functions

hnyu:
ALF overwrites some nn.Module functions for some purpose. However, if we want to load modern HF models, the original nn.Module functions are needed. Otherwise there will be some undefined behaviors or even errors for loading. 

ALF's custom nn.Module functions were written for very old pytorch versions. We probably need to update them in the future to take into account new changes. For now, this context manager can help us avoid any issue appearing when loading an HF model.

How to test:
Without this PR, when I tried to load an HF model, there would either be a function signature mismatch, or some uninitialized weights/buffers warning even after the model is loaded. 

With this PR, errors/warnings are gone.

### DIFF
--- a/alf/module.py
+++ b/alf/module.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from contextlib import contextmanager
 from torch.nn import Module
 
 # ALF Algorithm will overwrite these functions, we save the original ones.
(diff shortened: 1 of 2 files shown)


emailweixu (commented):
Perhaps we can simply remove module.py as it causes problems once a while.

hnyu:
> Perhaps we can simply remove module.py as it causes problems once a while.

I've removed the attribute settings. Still keep the old state dict related functions in module.py.

le-horizon (dismissed):
works nicely, thanks for the fix!

alf/module.py
```diff
@@ -11,54 +11,11 @@
...
-    old_add_module(self, name, module)
-    object.__setattr__(self, name, module)
-
-
-Module.add_module = _new_add_module
+# ALF Algorithm will overwrite these functions, we save the original ones.
+old_state_dict = Module.state_dict
+old_load_state_dict = Module.load_state_dict
+old__save_to_state_dict = Module._save_to_state_dict
+old__load_from_state_dict = Module._load_from_state_dict
```
emailweixu, line 22: We can just move these close to `original_torch_module_functions()`. It will be easy to maintain and read.

hnyu: > We can just move these close to `original_torch_module_functions()`. It will be easy to maintain and read.

No guarantee that utils.common will be evaluated before Algorithm? If not, then these old functions are not right.

le-horizon: Then maybe move the original_torch_module_functions function here?

le-horizon (approved):

PR #1759, opened 2025-04-17
only transfer trainable params

hnyu:
For distributed traininer and unroller, this PR makes sure that only the following two types of tensors are transferred between them:
1. buffers
2. parameters that require grad

(2) can save lots of communication overhead, especially when inside the core alg we load a large pretrained model.

### DIFF
--- a/alf/algorithms/distributed_off_policy_algorithm.py
+++ b/alf/algorithms/distributed_off_policy_algorithm.py
@@ -154,16 +154,20 @@ class DistributedOffPolicyAlgorithm(OffPolicyAlgorithm):
         self._ddp_rank = max(0, PerProcessContext().ddp_rank)
         self._num_ranks = PerProcessContext().num_processes
 
-    def _opt_free_state_dict(self) -> dict:
-        """Return `self._core_alg` state dict without optimizers.
+    def _distributed_state_dict(self) -> dict:
+        """Return `self._core_alg` state dict for distributed training.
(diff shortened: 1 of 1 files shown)


hnyu:
@Haichao-Zhang This PR is needed for efficient training if a large pretrained model (like success model) has been loaded by the algorithm.

alf/algorithms/distributed_off_policy_algorithm.py
```diff
@@ -154,16 +154,20 @@ def __init__(self,
...
+        Also we should exclude other parameters such as those of pretrained models.
         """
+        # Note that self._core_alg won't create a relay buffer so we don't have
+        # to worry about including it in the state dict.
         return {
             k: v
             for k, v in self._core_alg.state_dict().items()
-            if '_optimizers.' not in k
+            if (('_optimizers.' not in k) and (
+                not isinstance(v, torch.nn.Parameter) or v.requires_grad))
```
Haichao-Zhang, line 170: can you explain why ``not isinstance(v, torch.nn.Parameter)``?

hnyu: > can you explain why `not isinstance(v, torch.nn.Parameter)`?

`requires_grad` is only available to Parameter. We should also sync buffers.

Haichao-Zhang (approved):

PR #1764, opened 2025-05-01
Fix fused_linear_act pytorch version requirement

JingyuQian:
This PR fixes the pytorch version requirement of `ext/` module.

the `lazyInitDevice` used in the cuda code was introduced in the 2.6.0-rc1 version of pytorch [see here](https://github.com/pytorch/pytorch/blob/v2.6.0-rc1/aten/src/ATen/Context.h#L111). On lower versions, this extension will fail to build. 

The PR enriches build conditions to:
1. CUDA is available;
2. `torch.__version__ >= 2.6.0-rc1`;
3. environment variable `ALF_DISABLE_CUDA_EXT != 1`

### DIFF
--- a/alf/ext/act_backward.py
+++ b/alf/ext/act_backward.py
@@ -16,10 +16,21 @@ import torch
 from torch.utils.cpp_extension import load
 import pathlib
 import os
+from packaging.version import parse as parse_version
 
 DIR = pathlib.Path(__file__).parent.absolute()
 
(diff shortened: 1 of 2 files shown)


emailweixu:
Perhaps you can try removing lazyInitDevice() call?

JingyuQian:
> Perhaps you can try removing lazyInitDevice() call?

OK I'll see what I can do.

PR #1767, opened 2025-05-07
skip compiling fused linear act if CUDA_HOME is undefined

hnyu:
If other repos such as Hobot depend on ALF but its env has no CUDA_HOME defined, we also should skip compiling fused linear act. This requirement might be due to higher version of torch.

```bash
File "/app/venv/lib/python3.12/site-packages/torch/utils/cpp_extension.py", line 2525, in _join_cuda_home
[127](https://github.com/HorizonRoboticsInternal/Hobot/actions/runs/14872078984/job/41762363599?pr=1688#step:7:128)
    raise OSError('CUDA_HOME environment variable is not set. '
[128](https://github.com/HorizonRoboticsInternal/Hobot/actions/runs/14872078984/job/41762363599?pr=1688#step:7:129)
OSError: CUDA_HOME environment variable is not set. Please set it to your CUDA install root.
```

### DIFF
--- a/alf/ext/act_backward.py
+++ b/alf/ext/act_backward.py
@@ -14,6 +14,7 @@
 
 import torch
 from torch.utils.cpp_extension import load
+from absl import logging
 import pathlib
 import os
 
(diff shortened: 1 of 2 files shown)


QuantuMope (dismissed):
LGTM

emailweixu (dismissed):

alf/ext/act_backward.py
```diff
@@ -26,7 +26,10 @@
                     verbose=True)
 
         relu_backward_cuda = _ext.relu_backward
-    except ImportError:
+    except (ImportError, OSError):
```
emailweixu: perhaps we can print the exception in a warning so that people knows what to do if fused_linear_act is needed.

emailweixu (approved):

PR #1768, opened 2025-05-07
Post Process Experience with Customizable Modes

Haichao-Zhang:
**Post Process Experience**
preprocess_unroll_experience feature, useful for a number of scenarios. For example:
1) many kinds of trajectory labeling (e.g. hindsight success relabeling)
2) trajectory filtering (e.g. excluding some trajectories from being stored into the buffer).

In some cases, it is also a necessary component to use if we want to use in-algorithm procedures to overwrite quantities in the timestep and record it into the buffer (e.g. step type), so that all the other logics in ALF are fully respected (e.g.masking out the loss for the LAST step based on step type).

~~**Synced Traning**
Note that although one may think step-based filtering (e.g. excluding tasks) can also be done on the replay buffer side, the training dynamics are not the same.
This PR ensures synced training, meaning we won't do train step for those invalid/to be excluded steps.
In contrast, replay buffer based filtering cannot ensure synced training.~~

**Customizable Modes**
   The behavior could be customized by the user. Some examples:
    (1) per-step saving without delay: saving each step of unroll experience into the replay buffer as we get it. 
    (2) all-step saving with delay: saving all the steps of unroll experience into the replay buffer with delay. This can happen in the case where we want to annotate an trajectory based on some quantities that are not immediately available in the current step (e.g. task success/failure). 
    (3) selective saving: exclude some of the unroll experiences and only save the rest. This could be useful in the case where there are transitions that are irrelevant to the training (e.g. in the multi-task case, where we want to exclude data from certain subtasks).

### DIFF
--- a/alf/algorithms/algorithm.py
+++ b/alf/algorithms/algorithm.py
@@ -1426,9 +1426,7 @@ class Algorithm(AlgorithmInterface):
         return shape[0] * shape[1]
 
     @common.mark_replay
-    def train_from_replay_buffer(self,
-                                 effective_unroll_steps,
-                                 update_global_counter=False):
+    def train_from_replay_buffer(self, update_global_counter=False):
         """This function can be called by any algorithm that has its own
         replay buffer configured. There are several parameters specified in
         ``self._config`` that will affect how the training is performed:
@@ -1482,8 +1480,7 @@ class Algorithm(AlgorithmInterface):
         # training is not started yet, ``_replay_buffer`` will be None since it
         # is only lazily created later when online RL training started.
         if (self._replay_buffer and self._replay_buffer.total_size
-                < config.initial_collect_steps) or (effective_unroll_steps
-                                                    == 0):
+                < config.initial_collect_steps):
             assert (
                 self._replay_buffer.num_environments *
                 self._replay_buffer.max_length >= config.initial_collect_steps
--- a/alf/algorithms/algorithm.py
+++ b/alf/algorithms/algorithm.py
@@ -1426,7 +1426,9 @@ class Algorithm(AlgorithmInterface):
         return shape[0] * shape[1]
 
     @common.mark_replay
-    def train_from_replay_buffer(self, update_global_counter=False):
+    def train_from_replay_buffer(self,
+                                 effective_unroll_steps,
+                                 update_global_counter=False):
         """This function can be called by any algorithm that has its own
         replay buffer configured. There are several parameters specified in
         ``self._config`` that will affect how the training is performed:
@@ -1469,6 +1471,7 @@ class Algorithm(AlgorithmInterface):
                 ``True``, it will affect the counter only if
                 ``config.update_counter_every_mini_batch=True``.
         """
+
         config: TrainerConfig = self._config
 
         # returns 0 if haven't started training yet, when ``_replay_buffer`` is
@@ -1479,7 +1482,8 @@ class Algorithm(AlgorithmInterface):
         # training is not started yet, ``_replay_buffer`` will be None since it
         # is only lazily created later when online RL training started.
         if (self._replay_buffer and self._replay_buffer.total_size
-                < config.initial_collect_steps):
+                < config.initial_collect_steps) or (effective_unroll_steps
+                                                    == 0):
             assert (
                 self._replay_buffer.num_environments *
                 self._replay_buffer.max_length >= config.initial_collect_steps
@@ -1493,6 +1497,7 @@ class Algorithm(AlgorithmInterface):
             # ``_replay_buffer`` for training.
             # TODO: If this function can be called asynchronously, and using
             # prioritized replay, then make sure replay and train below is atomic.
+            effective_num_updates_per_train_iter = config.num_updates_per_train_iter
             with record_time("time/replay"):
                 mini_batch_size = config.mini_batch_size
                 if mini_batch_size is None:
@@ -1500,14 +1505,14 @@ class Algorithm(AlgorithmInterface):
                 if config.whole_replay_buffer_training:
                     experience, batch_info = self._replay_buffer.gather_all(
                         ignore_earliest_frames=True)
-                    num_updates = config.num_updates_per_train_iter
+                    num_updates = effective_num_updates_per_train_iter
                 else:
                     assert config.mini_batch_length is not None, (
                         "No mini_batch_length is specified for off-policy training"
                     )
                     experience, batch_info = self._replay_buffer.get_batch(
                         batch_size=(mini_batch_size *
-                                    config.num_updates_per_train_iter),
+                                    effective_num_updates_per_train_iter),
                         batch_length=config.mini_batch_length)
                     num_updates = 1
             return experience, batch_info, num_updates, mini_batch_size
--- a/alf/algorithms/algorithm.py
+++ b/alf/algorithms/algorithm.py
@@ -1469,7 +1469,6 @@ class Algorithm(AlgorithmInterface):
                 ``True``, it will affect the counter only if
                 ``config.update_counter_every_mini_batch=True``.
         """
-
         config: TrainerConfig = self._config
 
         # returns 0 if haven't started training yet, when ``_replay_buffer`` is
@@ -1494,7 +1493,6 @@ class Algorithm(AlgorithmInterface):
             # ``_replay_buffer`` for training.
             # TODO: If this function can be called asynchronously, and using
             # prioritized replay, then make sure replay and train below is atomic.
-            effective_num_updates_per_train_iter = config.num_updates_per_train_iter
             with record_time("time/replay"):
                 mini_batch_size = config.mini_batch_size
                 if mini_batch_size is None:
@@ -1502,14 +1500,14 @@ class Algorithm(AlgorithmInterface):
                 if config.whole_replay_buffer_training:
                     experience, batch_info = self._replay_buffer.gather_all(
                         ignore_earliest_frames=True)
-                    num_updates = effective_num_updates_per_train_iter
+                    num_updates = config.num_updates_per_train_iter
                 else:
                     assert config.mini_batch_length is not None, (
                         "No mini_batch_length is specified for off-policy training"
                     )
                     experience, batch_info = self._replay_buffer.get_batch(
                         batch_size=(mini_batch_size *
-                                    effective_num_updates_per_train_iter),
+                                    config.num_updates_per_train_iter),
                         batch_length=config.mini_batch_length)
                     num_updates = 1
             return experience, batch_info, num_updates, mini_batch_size
--- a/alf/algorithms/rl_algorithm.py
+++ b/alf/algorithms/rl_algorithm.py
@@ -604,13 +604,15 @@ class RLAlgorithm(Algorithm):
         effective_unroll_iters = 1 if unroll_length == 0 else effective_unroll_steps // unroll_length
         return experience, effective_unroll_iters
 
-    def post_process_experience(self, rollout_info, step_type: StepType,
-                                experiences: Experience) -> Tuple[List, int]:
-        """A function for postprocessing experience. By default, it returns the input
+    def preprocess_unroll_experience(
+            self, rollout_info, step_type: StepType,
+            experiences: Experience) -> Tuple[List, float]:
+        """A function for processing the experience obtained from an unroll step before
+        being saved into the replay buffer. By default, it returns the input
         experience unmodified. Users can customize this function in the derived
         class to achieve different effects. For example:
         - per-step processing: return the current step of experience unmodified (by default)
-            or a modified version according to the customized ``post_process_experience``.
+            or a modified version according to the customized ``preprocess_unroll_experience``.
             As another example, task filtering can be simply achieved by returning ``[]``
             for that particular task.
         - per-episode processing: this can be achieved by returning a list of processed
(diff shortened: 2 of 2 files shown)


alf/algorithms/algorithm.py
```diff
@@ -1426,7 +1426,9 @@ def train_from_unroll(self, experience, train_info):
         return shape[0] * shape[1]
 
     @common.mark_replay
-    def train_from_replay_buffer(self, update_global_counter=False):
+    def train_from_replay_buffer(self,
+                                 effective_unroll_steps,
```
runjerry: Add docstring for this arg?

alf/algorithms/rl_algorithm.py
```diff
@@ -235,11 +239,14 @@ def __init__(self,
...
         self._current_policy_state = None
         self._current_transform_state = None
-
+        self._cached_exp = []  # for lazy observation
         if self._env is not None and not self.on_policy:
             replay_buffer_length = adjust_replay_buffer_length(
                 config, self._num_earliest_frames_ignored)
 
+            if self._episodic_annotation:
+                assert self._env.batch_size == 1, "only support non-batched environment"
```
runjerry: Add this to the docstring of ``episodic_annotation``?

alf/algorithms/rl_algorithm.py
```diff
@@ -598,19 +605,62 @@ def _async_unroll(self, unroll_length: int):
 
         return experience
 
+    def should_post_process_episode(self, rollout_info, step_type: StepType):
+        """A function that determines whether the ``post_process_episode`` function should
+        be applied to the current list of experiences.
+        """
```
runjerry: This is an interface mainly used for subclasses? Maybe mention this. Same for ``post_process_episode``.

alf/algorithms/rl_algorithm.py
```diff
@@ -598,19 +605,62 @@ def _async_unroll(self, unroll_length: int):
...
+            self._cached_exp.append(exp)
+            if self.should_post_process_episode(rollout_info,
+                                                time_step.step_type):
+
+                # 1) process
+                annotated_exp_list = self.post_process_episode(
+                    self._cached_exp)
+                effective_number_of_unroll_steps = len(annotated_exp_list)
+                # 2) observe
+                if not self.on_policy:
```
runjerry: Maybe this condition check should be performed earlier, since it seems a waste to do all the ``post_process_episode`` if ``self.on_policy``?

alf/algorithms/algorithm.py
```diff
@@ -1479,7 +1482,8 @@ def train_from_replay_buffer(self, update_global_counter=False):
         # training is not started yet, ``_replay_buffer`` will be None since it
         # is only lazily created later when online RL training started.
         if (self._replay_buffer and self._replay_buffer.total_size
-                < config.initial_collect_steps):
+                < config.initial_collect_steps) or (effective_unroll_steps
+                                                    == 0):
```
runjerry: Is there any situation that ``train_from_replay_buffer`` will be called with ``effective_unroll_steps=0``?

alf/algorithms/rl_algorithm.py
```diff
@@ -846,11 +901,13 @@ def _train_iter_off_policy(self):
...
-        if unrolled:
-            with record_time("time/after_train_iter"):
-                self.after_train_iter(root_inputs, rollout_info)
+        steps = 0
+        for i in range(effective_unroll_steps):
+            steps += self.train_from_replay_buffer(effective_unroll_steps=1,
+                                                   update_global_counter=True)
+            if unrolled:
+                with record_time("time/after_train_iter"):
+                    self.after_train_iter(root_inputs, rollout_info)
```
runjerry: I feel that this update fundamentally changes the off-policy update logic w.r.t. its actual unroll in the env. Previously, between every call of ``self._unroll_iter_off_policy``, the policy gets an "update" from ``self.train_from_replay_buffer``. Now if ``self._episodic_annotation``, policy training only happens after each episode, though the UTD stays the same. I feel that the episodic annotation function should be configurable independently of the choice of such unroll/update logic. Ideally, we may want to keep the previous version here while achieving the same effect of the change of above lines by configuring ``unroll_length`` and ``num_updates_per_train_iter``.

alf/algorithms/algorithm.py
Haichao-Zhang: This arg is now removed

alf/algorithms/rl_algorithm.py
Haichao-Zhang: The assertion is not necessary here so remove also

alf/algorithms/rl_algorithm.py
Haichao-Zhang: Good point. Added comments. Also for post_process_episode

alf/algorithms/rl_algorithm.py
Haichao-Zhang: Updated

alf/algorithms/algorithm.py
Haichao-Zhang: effective_unroll_steps is now removed from this function

alf/algorithms/rl_algorithm.py
Haichao-Zhang: If ``self._episodic_annotation`` is False, everything is the same as before.
If  ``self._episodic_annotation`` is True, by default (with the new commit), also reduces to the original logic, so everything is the same after before (policy training only happens after each time step, not after each episode)

In the derived class, it is up to the user for determining what kind of annotation function he/she wants to implement and use.

alf/algorithms/rl_algorithm.py
```diff
@@ -633,6 +638,7 @@ def _process_unroll_step(self, policy_step, action, time_step,
                               alf.layers.to_float32(policy_state))
         effective_number_of_unroll_steps = 1
         if self._episodic_annotation:
+            assert not self.on_policy, "only support episodic annotation for off policy training"
```
runjerry: Maybe assert this in the ``__init__`` function?

runjerry:
Thank you Haichao for addressing all my comments. Just one more minor question.

alf/algorithms/rl_algorithm.py
```diff
@@ -605,12 +637,36 @@ def _process_unroll_step(self, policy_step, action, time_step,
...
+                annotated_exp_list = self.post_process_episode(
+                    self._cached_exp)
+                effective_number_of_unroll_steps = len(annotated_exp_list)
+                # 2) observe
+                t0 = time.time()
+                for exp in annotated_exp_list:
+                    self.observe_for_replay(exp)
+                store_exp_time = time.time() - t0
+                # clean up the exp cache
+                self._cached_exp = []
```
le-horizon: This seems to assume that all envs end on the same step?  What if some envs are LAST, some are MID?  cached_exp will be cleared even for those with MID steps?

Even when doing this for an env with batch_size 1, this annotation mode will delay experience from being stored into the replay buffer.

Ok to submit the change as is, but may need to do two things:

1) rename the feature to something like ``store_experience_on_episode_end``, and document its behavior clearly in the docstr.
  experience relabel should be done when reading data out of replay buffer as in [hindsight relabel](https://github.com/HorizonRobotics/alf/blob/911d9573866df41e9e3adf6cdd94ee03016bf5a8/alf/algorithms/data_transformer.py#L672).

2) assert that batch_size is 1 when enabled.

Also, delaying train_step because of delayed experience storage can have unexpected side effects, e.g. if episodes are 100 steps long, and unroll once per train iter, then summary will only happen every 100 train iters.  It will also shift the distribution of the data training sees due to the delay.

Overall I think doing this episode level relabeling at the DataTransformer stage, after reading from replay_buffer is perhaps a better way, and a cleaner way as well (less scattered code).  That would require the replay buffer to keep track of episode begin and end, which I think it already does.

alf/algorithms/rl_algorithm.py
```diff
@@ -846,11 +907,12 @@ def _train_iter_off_policy(self):
             return 0
 
         self.train()
-        steps = self.train_from_replay_buffer(update_global_counter=True)
-
-        if unrolled:
-            with record_time("time/after_train_iter"):
-                self.after_train_iter(root_inputs, rollout_info)
+        steps = 0
+        for i in range(effective_unroll_steps):
```
le-horizon: unroll_steps is the wrong name?  It should be called unroll_iterations to indicate training iterations, not env steps?

also rename effective_number_of_unroll_steps to be effective_unroll_iters to be consistent.  (i.e. remove "number_of_")

le-horizon (commented):
some high and low level comments if they make sense.

alf/algorithms/rl_algorithm.py
Haichao-Zhang: > This seems to assume that all envs end on the same step? What if some envs are LAST, some are MID? cached_exp will be cleared even for those with MID steps?

There is no such assumption. It is totally up to the users to inject their own assumptions.
By default, the behavior is the same as before. 
Sorry that the function names are a bit mis-leading and their role has been extended to handle per-step case as well. Changed the function names and added more comments.

> 
> Even when doing this for an env with batch_size 1, this annotation mode will delay experience from being stored into the replay buffer.

No it won't. By default, the behavior is the same as before. 



> 
> Ok to submit the change as is, but may need to do two things:
> 
> 1. rename the feature to something like `store_experience_on_episode_end`, and document its behavior clearly in the docstr.
The suggested name is not appropriate. 


>    experience relabel should be done when reading data out of replay buffer as in [hindsight relabel](https://github.com/HorizonRobotics/alf/blob/911d9573866df41e9e3adf6cdd94ee03016bf5a8/alf/algorithms/data_transformer.py#L672).

Different use cases. This is an alternative interface that can support more than pure relabeling (e.g. excluding data), which is not directly supported by the replay buffer hindsight relabel.


> 2. assert that batch_size is 1 when enabled.
There is no such assumption in the current PR. It is up to the user.


> 
> Also, delaying train_step because of delayed experience storage can have unexpected side effects, e.g. if episodes are 100 steps long, and unroll once per train iter, then summary will only happen every 100 train iters. It will also shift the distribution of the data training sees due to the delay.
There is no delay.

> 
> Overall I think doing this episode level relabeling at the DataTransformer stage, after reading from replay_buffer is perhaps a better way, and a cleaner way as well (less scattered code). That would require the replay buffer to keep track of episode begin and end, which I think it already does.
As explained, it is more than pure relabeling.

alf/algorithms/algorithm.py
```diff
@@ -1493,21 +1494,22 @@ def _replay():
...
+            effective_num_updates_per_train_iter = config.num_updates_per_train_iter
             with record_time("time/replay"):
                 mini_batch_size = config.mini_batch_size
                 if mini_batch_size is None:
                     mini_batch_size = self._replay_buffer.num_environments
                 if config.whole_replay_buffer_training:
                     experience, batch_info = self._replay_buffer.gather_all(
                         ignore_earliest_frames=True)
-                    num_updates = config.num_updates_per_train_iter
+                    num_updates = effective_num_updates_per_train_iter
```
emailweixu: why do you need to make this change?

alf/algorithms/rl_algorithm.py
Haichao-Zhang: Thanks for the comments. Changed.

alf/algorithms/rl_algorithm.py
```diff
@@ -596,21 +598,80 @@ def _async_unroll(self, unroll_length: int):
 
         self._current_transform_state = common.detach(trans_state)
 
-        return experience
+        effective_unroll_iters = effective_unroll_steps // unroll_length
+        return experience, effective_unroll_iters
+
+    def should_post_process_experience(self, rollout_info,
```
emailweixu: This is unnecessary. We can always call post_process_experience

alf/algorithms/rl_algorithm.py
```diff
@@ -596,21 +598,80 @@ def _async_unroll(self, unroll_length: int):
...
+        be called. Users can customize this pair of functions in the derived class to achieve
+        different effects. For example:
+        - per-step processing: ``should_post_process_experience``
+            returns True for all the steps (by default), and ``post_process_experience``
+            returns the current step of experience unmodified (by default) or a modified version
+            according to their customized ``post_process_experience`` function.
+            As another example, task filtering can be simply achieved by returning ``[]``
+            in ``post_process_experience`` for that particular task.
+        - per-episode processing: ``should_post_process_experience`` returns True on episode
+            end and ``post_process_experience`` can return a list of cached and processed
```
emailweixu: no need to mention "cached". It will confuse the user.

alf/algorithms/algorithm.py
Haichao-Zhang: not necessary anymore. removed

alf/algorithms/rl_algorithm.py
Haichao-Zhang: Correct. Remove this function

alf/algorithms/rl_algorithm.py
Haichao-Zhang: Updated the docstring.

alf/algorithms/rl_algorithm.py
```diff
@@ -846,11 +894,12 @@ def _train_iter_off_policy(self):
             return 0
 
         self.train()
-        steps = self.train_from_replay_buffer(update_global_counter=True)
-
-        if unrolled:
-            with record_time("time/after_train_iter"):
-                self.after_train_iter(root_inputs, rollout_info)
+        steps = 0
+        for i in range(effective_unroll_iters):
```
emailweixu, line 954: it's possible the effective_unroll_iters is always smaller than 1 in the case of num_envs > 1.

alf/algorithms/rl_algorithm.py
```diff
@@ -605,11 +634,18 @@ def _process_unroll_step(self, policy_step, action, time_step,
...
                               alf.layers.to_float32(policy_state))
-
+        effective_unroll_steps = 1
         store_exp_time = 0
         if not self.on_policy:
+            rollout_info = policy_step.info
+            # 1) process
+            post_processed_exp_list = self.post_process_experience(
+                rollout_info, time_step.step_type, exp)
+            effective_unroll_steps = len(post_processed_exp_list)
```
emailweixu: In order really get this right, need to do: `sum(exp.step_type.shape[0] for exp in post_processed_exp_list) / exp.step_type.shape[0]`

Haichao-Zhang: Updated.

alf/algorithms/rl_algorithm.py
Haichao-Zhang: Good point. Now also handles the fractional unroll case.

Haichao-Zhang:
Close for now since this is mostly used in a side project. Will reopen if this becomes a general feature for other usecases.

Haichao-Zhang:
Reopen as seems other people are also trying to use this feature.

Haichao-Zhang:
Pushed a commit 94a50bfc52071f4d7df3f7f0e4e7dff499904fe5 to give the user the flexibility to customize ``effective_unroll_steps`` for achieving different effects.

alf/algorithms/rl_algorithm.py
```diff
@@ -596,20 +599,59 @@ def _async_unroll(self, unroll_length: int):
 
         self._current_transform_state = common.detach(trans_state)
 
-        return experience
+        # if the input unroll_length is 0 (e.g. fractional unroll), then this it treated as
+        # an effective unroll iter
+        effective_unroll_iters = 1 if unroll_length == 0 else effective_unroll_steps // unroll_length
+        return experience, effective_unroll_iters
+
+    def post_process_experience(self, rollout_info, step_type: StepType,
```
hnyu: This name is confusing with the existing function `preprocess_experience` which might suggest that this happens after that but in fact this happens before training.

alf/algorithms/rl_algorithm.py
```diff
@@ -544,6 +544,7 @@ def _async_unroll(self, unroll_length: int):
         store_exp_time = 0.
         step_time = 0.
         max_step_time = 0.
+        effective_unroll_steps = 0
```
hnyu, line 547: I think we lack a formal definition of "effective" in the code document.

alf/algorithms/rl_algorithm.py
```diff
@@ -723,7 +770,10 @@ def _sync_unroll(self, unroll_length: int):
         self._current_policy_state = common.detach(policy_state)
         self._current_transform_state = common.detach(trans_state)
 
-        return experience
+        # if the input unroll_length is 0 (e.g. fractional unroll), then this it treated as
+        # an effective unroll iter
+        effective_unroll_iters = 1 if unroll_length == 0 else effective_unroll_steps // unroll_length
```
hnyu, line 823: It's strange to call unroll "iter"? The original definition is that each training iter we have one unroll. So what does unroll iters mean in this context?

hnyu:
>  This PR ensures synced training, meaning we won't do train step for those invalid/to be excluded steps.

I think the name "synced training" is somewhat confusing to me. The second half of the sentence doesn't relate to synchronization?

alf/algorithms/rl_algorithm.py
Haichao-Zhang: Added comments. One ``effective_unroll_iter`` refers to the ``unroll_length`` times of calling of ``rollout_step`` in the unroll phase.

alf/algorithms/rl_algorithm.py
Haichao-Zhang: Added more comments with examples, especially in ``preprocess_unroll_experience``

alf/algorithms/rl_algorithm.py
Haichao-Zhang: changed to ``preprocess_unroll_experience``

Haichao-Zhang:
> > This PR ensures synced training, meaning we won't do train step for those invalid/to be excluded steps.
> 
> I think the name "synced training" is somewhat confusing to me. The second half of the sentence doesn't relate to synchronization?



> > This PR ensures synced training, meaning we won't do train step for those invalid/to be excluded steps.
> 
> I think the name "synced training" is somewhat confusing to me. The second half of the sentence doesn't relate to synchronization?

Yeah, this part of description is out-dated. Updated with new one.

PR #1770, opened 2025-05-09
Make sure GradScaler.update gets called after every step

QuantuMope:
May be due to a new torch version, but running `alf.bin.train` with `enable_amp=True` and `amp_dtype=torch.float16`, runs into the following error

```
  File "/home/asjchoi/Desktop/alf/alf/algorithms/algorithm.py", line 1307, in _backward_and_gradient_update
    self._grad_scaler.step(optimizer)
  File "/home/asjchoi/venvs/horizon310/lib/python3.10/site-packages/torch/amp/grad_scaler.py", line 392, in step
    raise RuntimeError(
RuntimeError: step() has already been called since the last update().
```

This PR solves this by making sure update is called after every step call.

Also remove deprecated init api in favor of supported one.

### DIFF
--- a/alf/algorithms/algorithm.py
+++ b/alf/algorithms/algorithm.py
@@ -169,7 +169,7 @@ class Algorithm(AlgorithmInterface):
             self.use_rollout_state = config.use_rollout_state
             if (config.enable_amp and torch.cuda.is_available()
                     and config.amp_dtype == torch.float16):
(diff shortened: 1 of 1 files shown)


QuantuMope:
This was due to a bug in my own code by using the same optimizer for multiple parameter groups.

PR #1771, opened 2025-05-12
Warn if using offline data buffer without Agent wrapper

QuantuMope:
This PR now warns users if they are using the offline data buffer without the `Agent` wrapper.

### DIFF
--- a/alf/algorithms/algorithm.py
+++ b/alf/algorithms/algorithm.py
@@ -29,7 +29,7 @@ import torch.nn as nn
 from torch.nn.modules.module import _IncompatibleKeys, _addindent
 
 import alf
-from alf.data_structures import AlgStep, LossInfo, StepType, TimeStep
+from alf.data_structures import AlgStep, LossInfo, StepType, TimeStep, BasicRolloutInfo
 from alf.experience_replayers.replay_buffer import BatchInfo, ReplayBuffer
 from alf.optimizers.utils import GradientNoiseScaleEstimator
 from alf.utils.checkpoint_utils import (is_checkpoint_enabled,
@@ -1368,6 +1368,8 @@ class Algorithm(AlgorithmInterface):
               customized training.
         """
         try:
+            if isinstance(rollout_info, BasicRolloutInfo):
+                rollout_info = rollout_info.rl
             return self.train_step(inputs, state, rollout_info)
         except:
             # the default train_step is not compatible with the
--- a/alf/algorithms/algorithm.py
+++ b/alf/algorithms/algorithm.py
@@ -1369,6 +1369,11 @@ class Algorithm(AlgorithmInterface):
         """
         try:
             if isinstance(rollout_info, BasicRolloutInfo):
+                logging.log_first_n(
+                    logging.WARNING,
+                    "Detected offline buffer training without Agent wrapper. "
+                    "For best compatibility, it is advised to use the Agent wrapper.",
+                    n=1)
                 rollout_info = rollout_info.rl
             return self.train_step(inputs, state, rollout_info)
         except:
--- a/alf/algorithms/sac_algorithm.py
+++ b/alf/algorithms/sac_algorithm.py
@@ -27,13 +27,13 @@ import alf
 from alf.algorithms.config import TrainerConfig
 from alf.algorithms.off_policy_algorithm import OffPolicyAlgorithm
 from alf.algorithms.one_step_loss import OneStepTDLoss
-from alf.algorithms.rl_algorithm import RLAlgorithm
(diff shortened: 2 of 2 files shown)


alf/algorithms/sac_algorithm.py
```diff
@@ -847,6 +847,10 @@ def _select_q_value(self, action, q_values):
     def _critic_train_step(self, observation, target_observation,
                            state: SacCriticState, rollout_info: SacInfo,
                            action, action_distribution):
+
+        if isinstance(rollout_info, BasicRolloutInfo):
+            rollout_info = rollout_info.rl
```
hnyu: This should be put outside of this function. The general principle is, the algorithm should always receive what it's supposed to receive. In this case, this means that the rollout_info passed in should already be SacInfo.

emailweixu: > This should be put outside of this function. The general principle is, the algorithm should always receive what it's supposed to receive. In this case, this means that the rollout_info passed in should already be SacInfo.

+1

QuantuMope: Thanks. Fixed.

alf/algorithms/sac_algorithm.py
```diff
@@ -845,11 +845,9 @@ def _select_q_value(self, action, q_values):
         return q_values.gather(2, action).squeeze(2)
 
     def _critic_train_step(self, observation, target_observation,
-                           state: SacCriticState, rollout_info: SacInfo,
-                           action, action_distribution):
-
-        if isinstance(rollout_info, BasicRolloutInfo):
-            rollout_info = rollout_info.rl
+                           state: SacCriticState,
+                           rollout_info: SacInfo | BasicRLInfo, action,
```
hnyu: Still should always be SacInfo? If it's BasicRLInfo, the algorithm will crash.

QuantuMope: Offline buffer data is stored as `BasicRLInfo` which comprises of just (s,a,r) data.

QuantuMope: I guess, I could convert `BasicRLInfo` into `SacInfo` with some fields empty? Not sure which is a better design. Lmk which one you think is cleaner and I can change.

hnyu: > Offline buffer data is stored as `BasicRLInfo` which comprises of just (s,a,r) data.

If you look at SAC's train_step(), it will get access to `rollout_info.repr`. This means that SAC is currently incompatible with offline training.

QuantuMope: I'm using a frozen encoder so not training a repr.

QuantuMope: Wait, it seems that `repr` is stored in `BasicRolloutInfo`. Not sure how this code was running then. I'll take a look

hnyu: It's using elastic_namedtuple so any missing field returns (). Anyway, a little weird but it works.

hnyu (dismissed):

alf/algorithms/sac_algorithm.py
Haichao-Zhang: Ideally, we should not include ``BasicRLInfo`` here as it could confuse the pure sac users. The better alternative might be comply with the Agent assumption and possibly extend it.

QuantuMope: Removed typehints. Also added a warning message advising users to use `Agent`. Before the code would simply crash due to interface conflict.

alf/algorithms/algorithm.py
```diff
@@ -1369,6 +1369,11 @@ def train_step_offline(self, inputs, state, rollout_info, pre_train=False):
         """
         try:
             if isinstance(rollout_info, BasicRolloutInfo):
+                logging.log_first_n(
+                    logging.WARNING,
+                    "Detected offline buffer training without Agent wrapper. "
+                    "For best compatibility, it is advised to use the Agent wrapper.",
+                    n=1)
```
hnyu, line 1376: This warning won't work. When using Agent, we still get rollout_info as `BasicRolloutInfo`.

QuantuMope: It will. When using `Agent`, it properly feeds the nested `BasicRLInfo` to this function instead.

QuantuMope: In other words, there was never a bug with hybrid RL training, just that it was never meant to be used without using `Agent`.

hnyu: > It will. When using `Agent`, it properly feeds the nested `BasicRLInfo` to this function instead.

you're right. I didn't know `Agent` overwrites this function itself.

PR #1772, opened 2025-05-22
Add num_gpus_per_ddp_worker flag to alf.bin.train

QuantuMope:
Before, alf always assigns one DDP worker for each available gpu.  Sometimes, it may be preferable to assign groups of gpus to a DDP worker instead though. For example, if a certain alf job requires loading in several memory intensive models, it may be necessary to assign 2 or more gpus to a single DDP worker.

This PR adds the `num_gpus_per_ddp_worker` flag to `alf.bin.train` for this purpose.

Now, the number of DDP workers created will be `N = CUDA_VISIBLE_DEVICES // num_gpus_per_ddp_worker` with each DDP worker having its own `num_gpus_per_ddp_worker` amount of gpus.

### DIFF
--- a/alf/bin/train.py
+++ b/alf/bin/train.py
@@ -83,6 +83,12 @@ def _define_flags():
     flags.DEFINE_enum(
         'distributed', 'none', ['none', 'multi-gpu', 'multi-node-multi-gpu'],
         'Set whether and how to run training in distributed mode.')
+    flags.DEFINE_integer(
+        'num_gpus_per_ddp_worker', 1,
+        "The number of gpus per DDP worker. If specified will create N DDP workers where each worker "
+        "has num_gpus_per_ddp_worker gpus assigned to it. N will be CUDA_VISIBLE_DEVICES // num_gpus_per_ddp_worker. "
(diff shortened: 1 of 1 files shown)


alf/bin/train.py
```diff
@@ -354,7 +360,20 @@ def main(_):
...
+        if CUDA_VISIBLE_DEVICES is None:
+            devices = [str(i) for i in range(torch.cuda.device_count())]
+            device_list = ','.join(devices)
+            CUDA_VISIBLE_DEVICES = device_list
+            os.environ['CUDA_VISIBLE_DEVICES'] = device_list
+        else:
+            devices = CUDA_VISIBLE_DEVICES.split(',')
+
+        num_devices = len(devices)
+        assert num_devices % FLAGS.num_gpus_per_ddp_worker == 0, \
```
emailweixu, line 374: assert that FLAGS.num_gpus_per_ddp_worker > 1 is not supported for FLAGS.distributed == 'multi-node-multi-gpu'

QuantuMope: Done.

emailweixu (approved):

PR #1773, opened 2025-05-22
Lazy load for alf.ext

emailweixu:
torch.utils.cpp_extension.load() is slow and cause problems frequently. So we don't load the extension unless it is used.

### DIFF
--- a/alf/ext/act_backward.py
+++ b/alf/ext/act_backward.py
@@ -17,22 +17,15 @@ from torch.utils.cpp_extension import load
 import pathlib
 import os
 
-DIR = pathlib.Path(__file__).parent.absolute()
+_ext = None
 
-if torch.cuda.is_available():
-    try:
-        _ext = load(name="act_backward",
-                    sources=[os.path.join(DIR, "act_backward.cu")],
-                    verbose=True)
 
-        relu_backward_cuda = _ext.relu_backward
-    except ImportError:
-        # There is a bug in pybind11 currently where pybind11 will
-        # incorrectly use the system python instead of the virtualenv python.
-        # This can result in a python version mismatch error.
-        # For now, we'll just catch this and ignore it.
-        # See https://github.com/pybind/pybind11/issues/5626
-        pass
+def _load_ext():
+    global _ext
+    DIR = pathlib.Path(__file__).parent.absolute()
+    _ext = load(name="act_backward",
+                sources=[os.path.join(DIR, "act_backward.cu")],
+                verbose=True)
 
 
 def relu_backward(output, grad_output):
@@ -57,7 +50,9 @@ def relu_backward(output, grad_output):
     assert output.dtype.is_floating_point
     if output.is_cuda and output.is_contiguous() and grad_output.is_contiguous(
     ):
-        return relu_backward_cuda(output, grad_output)
+        if _ext is None:
+            _load_ext()
+        return _ext.relu_backward(output, grad_output)
     else:
         return grad_output * (output > 0).float()
 
--- a/alf/ext/fused_linear_act.py
+++ b/alf/ext/fused_linear_act.py
@@ -20,21 +20,16 @@ import pathlib
 import os
 from .act_backward import act_backward
 
-DIR = pathlib.Path(__file__).parent.absolute()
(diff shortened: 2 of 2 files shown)


alf/ext/act_backward.py
```diff
@@ -17,22 +17,15 @@
...
-                    verbose=True)
 
-        relu_backward_cuda = _ext.relu_backward
-    except ImportError:
-        # There is a bug in pybind11 currently where pybind11 will
-        # incorrectly use the system python instead of the virtualenv python.
-        # This can result in a python version mismatch error.
-        # For now, we'll just catch this and ignore it.
-        # See https://github.com/pybind/pybind11/issues/5626
-        pass
```
QuantuMope: Seems like we should still keep this error handling in the `_load_ext()` func.

Also, in the case that it does trigger, we would need to make sure it doesn't keep trying to recompile the cuda code on every iteration.

alf/ext/fused_linear_act.py
```diff
@@ -20,21 +20,16 @@
...
+_ext = None
+
+
+def _load_ext():
+    global _ext
+
+    DIR = pathlib.Path(__file__).parent.absolute()
+    _ext = load(name="fused_matmul_act",
+                sources=[os.path.join(DIR, "fused_matmul_act.cu")],
+                verbose=True)
```
QuantuMope, line 33: Maybe create a util function in a `utils.py` used across both files?

```python
def _load_ext(name: str):
     # With try except added
    DIR = pathlib.Path(__file__).parent.absolute()
    _ext = load(name=name,
                sources=[os.path.join(DIR, f"{name}.cu")],
                verbose=True)            
```

QuantuMope (commented):
Hey Wei, some comments. Also looks like needs to be rebased on main.

alf/ext/act_backward.py
emailweixu: Now _ext is loaded when it is used. So it should fail if it cannot be loaded.

alf/ext/fused_linear_act.py
emailweixu: I wanted to have soemthing like that. But the tricky thing is `global _ext`, it can only be used for the module where the function is defined.

alf/ext/act_backward.py
QuantuMope: Right, but don't we want to make it so it will not fail if it cannot be loaded (similar to how it works now).

alf/ext/fused_linear_act.py
QuantuMope: Perhaps we can use a global dict in a separate file?

```python
# In file A
_exts = {}

def _load_ext(name: str): # With try except added 
    global _exts
    DIR = pathlib.Path(__file__).parent.absolute() 
    ext = load(name=name, 
               sources=[os.path.join(DIR, 
               f"{name}.cu")], verbose=True)
    _exts[name] = getattr(ext, name)

```

```python
# In file B
from A import _load_ext, _exts

name = "fused_matmul_act"
if name not in _exts:
    _load_ext(name)
    
    # can call like this
    _exts[name](...)
```

Maybe it's a bit complicated but would help to not repeat code if the number of .cu files starts to grow.

QuantuMope (commented):
Hey Wei, some thoughts.

alf/ext/act_backward.py
emailweixu: If it needs to be used, it should fail. For this change, loading happens only when it needs to be used.

alf/ext/fused_linear_act.py
emailweixu: There can be more elegant way. We can implement something called `lazy_load` to replace `load`, which will only do the actual loading when the returned object is used (i.e. accessing its attribute). But it will take more time. We can implement the more elegant way when new extension module is added.

alf/ext/act_backward.py
QuantuMope: Oh I see, I had originally thought these compiled modules would always be used by default for an alf model.

QuantuMope (dismissed):
LGTM

QuantuMope (approved):

PR #1775, opened 2025-06-02
[Bug Fix] Fix summarize_grads_and_vars for hybrid RL training

QuantuMope:
Found a minor bug for when `summarize_grads_and_vars=True` for hybrid RL training where return isn't unrolled properly. This PR fixes that.

### DIFF
--- a/alf/algorithms/algorithm.py
+++ b/alf/algorithms/algorithm.py
@@ -1265,7 +1265,11 @@ class Algorithm(AlgorithmInterface):
         Args:
             loss (Tensor): an aggregated scalar loss
         Returns:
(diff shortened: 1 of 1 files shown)


Haichao-Zhang (commented):
Seems to be introduced by #1351 . Suggest to update the docstr of ``_backward_and_gradient_update`` function, especially the return field.

QuantuMope:
> Seems to be introduced by #1351 . Suggest to update the docstr of `_backward_and_gradient_update` function, especially the return field.

Good catch. Added docstring

Haichao-Zhang (approved):
Thanks for the fix!

PR #1776, opened 2025-06-05
Allow unique offline replay buffers for DDP workers

QuantuMope:
This PR adds some additional logic to `TrainerConfig.offline_buffer_dir` so that if a directory of replay buffers is provided and DDP is being used, then a unique replay buffer is assigned to each DDP worker. If a directory is provided and we're not using DDP, then all the buffers will be used.

Note that although the variable is called "offline_buffer_dir", it actually previously expected a path to a file or a list of paths to files. Now this variable can accept a directory, a file, or a list of files.

Tested via a cluster job where unique buffers were loaded for each DDP worker.

### DIFF
--- a/alf/algorithms/config.py
+++ b/alf/algorithms/config.py
@@ -314,8 +314,21 @@ class TrainerConfig(object):
                 ``ReplayBuffer``.
             priority_replay_eps (float): minimum priority for priority replay.
             offline_buffer_dir (str|[str]): path to the offline replay buffer
-                checkpoint to be loaded. If a list of strings provided, each
-                will represent the directory to one replay buffer checkpoint.
+                checkpoint to be loaded. Has several scenarios:
+                - If a single string is provided and that path is to a replay buffer,
+                  then the buffer will be loaded.
+                - If a list of strings is provided, each string will be assumed
+                  to be a replay buffer and each will be loaded and concatenated.
+                - If a single string is provided and that path is to a directory,
+                  then there are two further cases:
+                    - If we're not using DDP, then all the replay buffers from the
+                      directory will be loaded.
+                    - If we are using DDP, then a unique replay buffer from the directory
+                      will be loaded for each DDP worker. The replay buffers being used
+                      for each worker will be determined by the worker number indexing
+                      into the sorted buffer names. This means that the directory must
+                      contain at least one replay buffer for each DDP worker.
+                - Will error if you provide multiple directories.
             offline_buffer_length (int): the maximum length will be loaded
                 from each replay buffer checkpoint. Therefore the total
                 buffer length is offline_buffer_length * len(offline_buffer_dir).
--- a/alf/algorithms/config.py
+++ b/alf/algorithms/config.py
@@ -327,10 +327,11 @@ class TrainerConfig(object):
                       will be loaded for each DDP worker. The replay buffers being used
                       for each worker will be determined by the worker number indexing
                       into the sorted buffer names. This means that the directory must
-                      contain at least one replay buffer for each DDP worker.
+                      contain at least one replay buffer for each DDP worker. Any remaining
+                      replay buffers will simply be ignored.
                 - Will error if you provide multiple directories.
             offline_buffer_length (int): the maximum length will be loaded
-                from each replay buffer checkpoint. Therefore the total
+                from each replay buffer checkpoint. Therefore, the total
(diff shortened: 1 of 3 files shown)


alf/algorithms/config.py
```diff
@@ -314,8 +314,21 @@ def __init__(self,
...
+                  to be a replay buffer and each will be loaded and concatenated.
+                - If a single string is provided and that path is to a directory,
+                  then there are two further cases:
+                    - If we're not using DDP, then all the replay buffers from the
+                      directory will be loaded.
+                    - If we are using DDP, then a unique replay buffer from the directory
+                      will be loaded for each DDP worker. The replay buffers being used
+                      for each worker will be determined by the worker number indexing
+                      into the sorted buffer names. This means that the directory must
+                      contain at least one replay buffer for each DDP worker.
```
emailweixu: Also comment that the remaining files will be ignored.

Haichao-Zhang:
> Note that although the variable is called "offline_buffer_dir", it actually previously expected a path to a file. Now this variable can accept a directory, a file, or a list of files.

Not true. It is called ``offline_buffer_dir`` because it can already support a list of files:

https://github.com/HorizonRobotics/alf/blob/8954925fb2ee66c6648aaec53d3a4967df1899e6/alf/examples/hybrid_rl/hybrid_sac_pendulum.gin#L25-L29

QuantuMope:
> > Note that although the variable is called "offline_buffer_dir", it actually previously expected a path to a file. Now this variable can accept a directory, a file, or a list of files.
> 
> Not true. It is called `offline_buffer_dir` because it can already support a list of files:
> 
> https://github.com/HorizonRobotics/alf/blob/8954925fb2ee66c6648aaec53d3a4967df1899e6/alf/examples/hybrid_rl/hybrid_sac_pendulum.gin#L25-L29

Right. I meant that it expects a list of files or a file. Not to a directory. Editted PR descrip

alf/algorithms/config.py
QuantuMope: Done.

alf/algorithms/config.py
```diff
@@ -314,10 +314,24 @@ def __init__(self,
...
+                - If a single string is provided and that path is to a replay buffer,
+                  then the buffer will be loaded.
+                - If a list of strings is provided, each string will be assumed
+                  to be a replay buffer and each will be loaded and concatenated.
+                - If a single string is provided and that path is to a directory,
+                  then there are two further cases:
+                    - If we're not using DDP, then all the replay buffers from the
+                      directory will be loaded.
+                    - If we are using DDP, then a unique replay buffer from the directory
+                      will be loaded for each DDP worker. The replay buffers being used
```
Haichao-Zhang, line 327: Not fully sure about the best design here, but it could be possible that people use a DDP number for offline data collection and online training. For example, to speedup offline data collection, not only a different DDP number can be used, but also possible to do that with multiple machines.  All these seems to suggest that it is a bit restrictive for assuming there much be one replay buffer for each DDP worker in the directory.

Also, this make it impossible to adjust the DDP number for training, and instead, is tied to the offline buffer structure?

QuantuMope: Could you clarify your scenario? Not sure what you mean by use a DDP number for offline data collection and online training. For offline data collection, `offline_buffer_dir` would not be used.

Haichao-Zhang: > Could you clarify your scenario? Not sure what you mean by use a DDP number for offline data collection and online training. For offline data collection, `offline_buffer_dir` would not be used.

I mean:
1) you have to specially structure the replay buffer such that the directory contains exact certain number of them
2) unable to change/tune the DDP number for training once the offline buffer is given

QuantuMope: I see. Right, it is a bit annoying because if I want to change number of DDP workers and have same data coverage, I need to concatenate or partition buffers manually. 

Actually, maybe a better idea is to forward any combination of buffers and if there are multiple gpus, alf will concatenate the buffers, partition them into num_gpus chunks, and save the partitions in a tmp directory. It can release the large buffer before forking processes and then each ddp worker will load the individual chunk. What do you think of this approach?

QuantuMope: After some discussion offline, we'll keep this simple design for now and think of improvements in the future.

emailweixu (approved):

PR #1777, opened 2025-06-06
lazy_load_extension, and use it for penv

emailweixu:
1. Make the lazy loading easy to use.
2. Use it to load parallel_environment.cpp so that we don't need to run make_penv.py any more.

### DIFF
--- a/alf/environments/process_environment.py
+++ b/alf/environments/process_environment.py
@@ -22,11 +22,12 @@ import atexit
 from enum import Enum
 from functools import partial
 import multiprocessing
+import os
 import sys
 import threadpoolctl
 import torch
 import traceback
-from typing import Dict, Any, Callable, List, Tuple
+from typing import Any, Callable, List, Tuple
 
 import alf
 import alf.nest as nest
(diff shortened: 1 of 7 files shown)


alf/utils/common.py
```diff
@@ -1836,3 +1838,70 @@ def save_video(frames: List[np.ndarray],
...
+    The extension will be loaded at the time when its attribute is accessed. This
+    is useful for extensions that are not always being used. It will save a lot
+    of time for compiling the extension at the beginning of the program.
+
+    Args:
+        **kwargs: keyword arguments to be passed to
+            ``torch.utils.cpp_extension.load()``. The most important one is
+            ``name``, which should be the name of the extension.
+    """
+    return LazyExtention(**kwargs)
```
QuantuMope: Nit: How about we make the args `name, **kwargs` here to match `LazyExtension` since `name` is required.

alf/environments/process_environment.py
```diff
@@ -22,19 +22,28 @@
...
-from . import _penv
+from alf.utils.common import lazy_load_extension
+import pathlib
+
+DIR = pathlib.Path(__file__).parent.absolute()
+_penv = lazy_load_extension(
+    name="penv",
+    sources=[os.path.join(DIR, "parallel_environment.cpp")],
+    extra_cflags=['-O3'],
+    verbose=True)
```
QuantuMope: Let's remove `ext_modules` from `setup.py`?

Can also match the preexisting compile flags `'-O3', '-Wall', '-std=c++17', '-fPIC', '-fvisibility=hidden'`

QuantuMope (commented):
Hey Wei, just two comments. Thanks

QuantuMope (approved):
LGTM

PR #1778, opened 2025-06-09
fix loss = ()

hnyu:
Sometimes we have loss=() and gradient update will be skipped.

### DIFF
--- a/alf/algorithms/algorithm.py
+++ b/alf/algorithms/algorithm.py
@@ -1208,8 +1208,10 @@ class Algorithm(AlgorithmInterface):
 
         loss_info = self._aggregate_loss(loss_info, valid_masks, batch_info)
 
-        all_params, gns = self._backward_and_gradient_update(loss_info.loss *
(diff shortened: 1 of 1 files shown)


alf/algorithms/algorithm.py
```diff
@@ -1208,8 +1208,10 @@ def update_with_gradient(self,
 
         loss_info = self._aggregate_loss(loss_info, valid_masks, batch_info)
 
-        all_params, gns = self._backward_and_gradient_update(loss_info.loss *
-                                                             weight)
+        loss = loss_info.loss
+        if loss != ():
+            loss = loss * weight
+        all_params, gns = self._backward_and_gradient_update(loss)
```
emailweixu: should we skip _backward_and_gradient_update() in this case?

emailweixu: Maybe warning_once for this too.

emailweixu (approved):

PR #1779, opened 2025-06-09
option for optimizer ignore params not requiring grad

hnyu:
Add an option to TrainerConfig for an optimizer to ignore adding params that have `requires_grad=False`. This is very helpful for saving memory in the case we use LoRA layers to finetune a huge base model. We don't want the (Adam) optimizer to keep track of the state  for params not being trained.


emailweixu (dismissed):

alf/optimizers/optimizers.py
```diff
@@ -131,6 +132,15 @@ def __init__(self,
...
                 a small part of model.
+            grad_accumulation_steps (int): If >1, accumulating (averaging) so
+                many gradients from ``loss.backward()`` before triggering an
+                optimizer step. Any optimizer ``step()`` call before the accumuation
+                is done will have no effect on the parameters. This flag can be
+                used to simulate training with a large batch size but with limited
+                memory. Note 1) with this option, grad clipping should only happen
+                after the accumulation is done; 2) all other schedulers will be
+                calculated based on the actual effective optimizer steps instead
+                of gradient steps.
```
emailweixu: How can schedulers be based on the actual effective optimizer steps?

hnyu: > How can schedulers be based on the actual effective optimizer steps?

I was trying to say that the scheduler value only takes effect at optimizer steps. Updated the docstring.

emailweixu (approved):

PR #1781, opened 2025-06-13
RL with MC return

Haichao-Zhang:
Implemented Monte Carlo (MC) return-based RL, with usage example.
Can potentially help with the value learning and therefore make the learning faster.

1. example:  ``sac_cart_pole_conf.py``
2. also added summarization, so that in tensorboard under the ``critic_loss`` panel, there will be an additional curve (or distribution/histogram) for ``episodic_discounted_return``
3. support multi-step td (``mini_batch_length>2``), see figure below

cart_pole_mc_return


Haichao-Zhang:
@emailweixu  added support to multi-step learning. Updated PR description and figure.

emailweixu (approved):

PR #1783, opened 2025-06-17
data structure support for computing `log P_current (rollout action)` in train_step for PPO

hnyu:
This PR adds a data structure support for PPO computing `log P_current(rollout_action)` in `train_step()` instead of when computing importance ratio. 

Currently, we store the current action distribution and delay computing log probs until importance ratio. However, sometimes, it might not be easy to store such a distribution. If it's easier to precompute the log probs in `train_step()`, then we should do so.

### DIFF
--- a/alf/utils/value_ops.py
+++ b/alf/utils/value_ops.py
@@ -28,7 +28,8 @@ def action_importance_ratio(action_distribution,
                             log_prob_clipping,
                             check_numerics,
                             debug_summaries,
-                            rollout_log_prob=None):
+                            rollout_log_prob=(),
+                            log_prob=()):
     """ ratio for importance sampling, used in PPO loss and vtrace loss.
 
         Caller has to save alf.summary.scope() and pass scope to this function.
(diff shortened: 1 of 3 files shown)


emailweixu (commented):
the info from ActorCriticAlgorithm already has log_prob

hnyu:
> the info from ActorCriticAlgorithm already has log_prob

That one is the log prob of actions sampled from the current policy. Also PPO's `train_step` will receive (from process_experience) and return `PPOInfo` instead of `ActorCriticInfo` so there should be a `log_prob` in `PPOInfo` to support such a feature.

emailweixu (commented):
PPOInfo has a field called rollout_log_prob

hnyu:
> PPOInfo has a field called rollout_log_prob

Yeah that's `log P_rollout (rollout action)` but not `log P_current (rollout action)`. The latter is computed during importance ratio currently, provided that `P_current` is passed all the way to importance ratio function. 

```python
def action_importance_ratio(...):
        action_log_prob = dist_utils.compute_log_probability(
            current_policy_distribution, action)
```

alf/utils/value_ops.py
```diff
@@ -61,22 +62,29 @@ def action_importance_ratio(action_distribution,
             check_numerics (bool):  If true, adds checks to help find
                 ``NaN``/``Inf`` values. For debugging only.
             debug_summaries (bool): If true, output summary metrics to tensorboard.
-            rollout_log_prob (nested tensor): the log probability of the action
+            rollout_log_prob (nested tensor): the rollout log probability of the action.
+                If empty, will be computed from ``rollout_action_distribution``
+                and ``action``.
+            log_prob (nested tensor): the log probability of the action. If empty,
```
emailweixu: log_prob is not nested tensor.

hnyu: fixed the type

emailweixu (approved):

PR #1785, opened 2025-06-19
option for non-strict preloading ckpt

hnyu:
While currently we allow the user to configure whether a ckpt is strictly loaded or not in `policy_trainer.py` for loading an algorithm's entire ckpt, there  is no such an option for in-alg ckpt preloading. This PR adds a flag to `Algorithm` to allow partial loading. This is useful if we want to ignore loading some model weights that can be downloaded as pretrained ones.

### DIFF
--- a/alf/algorithms/algorithm.py
+++ b/alf/algorithms/algorithm.py
@@ -62,6 +62,7 @@ def _flatten_module(module):
         return [module]
 
 
+@alf.configurable(whitelist=['checkpoint_preload_strict'])
 class Algorithm(AlgorithmInterface):
     """Base implementation for AlgorithmInterface."""
 
@@ -72,6 +73,7 @@ class Algorithm(AlgorithmInterface):
                  is_on_policy=None,
                  optimizer=None,
                  checkpoint=None,
+                 checkpoint_preload_strict=True,
                  config: TrainerConfig = None,
                  debug_summaries=False,
                  name="Algorithm"):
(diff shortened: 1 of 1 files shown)


alf/algorithms/algorithm.py
```diff
@@ -62,6 +62,7 @@ def _flatten_module(module):
         return [module]
 
 
+@alf.configurable(whitelist=['checkpoint_preload_strict'])
```
emailweixu, line 65: I feel it is better to pass this argument from the subclass algorithms and config for the subclass algorithm so we don't make global change for the behavior of Algorithm.

hnyu: > I feel it is better to pass this argument from the subclass algorithms and config for the subclass algorithm so we don't make global change for the behavior of Algorithm.

I think the question is down to which subalgorithm? If all the way down to each leaf algorithm, there will be lots of argument change lines across the entire repo. 

Or I can add this option to TrainerConfig so that it's easier for the change.

emailweixu: > > I feel it is better to pass this argument from the subclass algorithms and config for the subclass algorithm so we don't make global change for the behavior of Algorithm.
> 
> I think the question is down to which subalgorithm? If all the way down to each leaf algorithm, there will be lots of argument change lines across the entire repo.

Just change the specific algorithm that needs this behavior at this moment.

hnyu:
Closing this PR as this feature is no longer needed on my side.

PR #1787, opened 2025-06-25
More device support for metrics

emailweixu:
Added device argument for AverageReturnMetric since we may want to use gpu for it.


nemonswx (approved):
Looks good to me

nemonswx (approved):

le-horizon (approved):

PR #1792, opened 2025-07-10
support fractional num_updates

hnyu:
This PR supports fractional number of updates, as a complement to the existing fractional unroll length. This option is mainly for whole replay buffer training where we currently don't have an option of adjusting the amount of unroll data to train.

### DIFF
--- a/alf/algorithms/algorithm.py
+++ b/alf/algorithms/algorithm.py
@@ -1517,8 +1517,8 @@ class Algorithm(AlgorithmInterface):
                         "No mini_batch_length is specified for off-policy training"
                     )
                     experience, batch_info = self._replay_buffer.get_batch(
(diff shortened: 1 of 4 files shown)


emailweixu (commented):
The logic becomes very complicated. I wonder if there's way to make it simpler and more understandable.
An alternative for achieving this for a particular algorithm without complicating alf is to implement the filter_experience() hook.

hnyu:
> The logic becomes very complicated. I wonder if there's way to make it simpler and more understandable. An alternative for achieving this for a particular algorithm without complicating alf is to implement the filter_experience() hook.

Sounds good. No algorithm is actually using this feature for now. I briefly tried fractional num_updates for OpenVLA + PPO but the wall-clock time didn't improve.

PR #1795, opened 2025-07-18
Summarize the first importance ratio for PPO

emailweixu:
The importance ratio for the first gradient update in each iteration should be exact 1 (or very close to it). Nonzero importance ratio indicates something is wrong. So it is useful to summarize it to uncover some hidden bugs.

### DIFF
--- a/alf/algorithms/ppo_loss.py
+++ b/alf/algorithms/ppo_loss.py
@@ -138,11 +138,18 @@ class PPOLoss(ActorCriticLoss):
         if alf.summary.get_grad_step_counter() == 0:
             # For the first gradient step in one iteration, the importance ratios
             # should be 1. Summarize them so that we can notice something is wrong
-            # if they are not 1.
-            with alf.summary.record_if(lambda: True), scope:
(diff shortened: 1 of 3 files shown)


alf/algorithms/ppo_loss.py
```diff
@@ -135,6 +135,15 @@ def _pg_loss(self, info, advantages):
             log_prob_clipping=self._log_prob_clipping,
             check_numerics=self._check_numerics,
             debug_summaries=self._debug_summaries)
+        if alf.summary.get_grad_step_counter() == 0:
+            # For the first gradient step in one iteration, the importance ratios
+            # should be 1. Summarize them so that we can notice something is wrong
+            # if they are not 1.
+            with alf.summary.record_if(lambda: True), scope:
```
runjerry: So this first gradient step importance ratio summary is regardless of the ``TrainerConfig.summary_interval``?

emailweixu: Good catch. Fixed it.

runjerry (approved):
Looks good to me, thanks for adding this useful summary.

PR #1797, opened 2025-07-29
skip adding param group when no param

hnyu:
Fixed a bug introduced in PR #1779 . Sometimes after filtering out params without grad, the params list is empty. We should not add this param group in this case.


le-horizon (approved):
Thanks for the fix, Haonan.

PR #1798, opened 2025-08-01
Update contributing.rst

le-horizon:


Haichao-Zhang (commented):
Thanks for the update. Maybe useful to provide a comment saying "refer to https://github.com/HorizonRobotics/alf/blob/71a1aa2e32ed82261456068143ee2d44072180a7/setup.py#L22-L53 for the exact version"  ?

docs/contributing.rst
```diff
@@ -11,10 +11,10 @@ Workflow
...
 
-    pip install pre-commit==1.17.0
-    pip install cpplint==1.4.4
+    pip install pre-commit==4.2.0
+    pip install cpplint==2.0.1
     pip install pydocstyle==4.0.0
-    pip install pylint==2.3.1
+    pip install pylint==3.3.6
     pip install yapf==0.43.0
     sudo apt install clang-format
```
emailweixu: They should be same as the versions used by the CI docker (.ci-cd/Dockerfile_py3.11_torch2.2.cpu) . Otherwise may cause CI issue.

emailweixu (changes requested):

le-horizon (commented):
Done both.  Also updated comments in all three places to warn to keep them consistent.

PTAL.
Le

emailweixu (approved):

PR #1799, opened 2025-08-12
Set EXE_MODE_EVAL for alf.bin.play

QuantuMope:
PR simply sets alf's exe_mode to `EXE_MODE_EVAL` when using `alf.bin.play`.

This allows us to do eval-specific procedures using `alf.utils.common.is_eval()` such as merging LoRA weights.

### DIFF
--- a/alf/bin/play.py
+++ b/alf/bin/play.py
@@ -108,9 +108,8 @@ def _define_flags():
 FLAGS = flags.FLAGS
 
 
+@common.mark_eval
 def play():
-    alf.utils.common.set_exe_mode(alf.utils.common.EXE_MODE_EVAL)
-
     if torch.cuda.is_available():
         alf.set_default_device("cuda")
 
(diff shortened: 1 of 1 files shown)


alf/bin/play.py
```diff
@@ -109,6 +109,8 @@ def _define_flags():
 
 
 def play():
```
emailweixu, line 112: @common.mark_eval

hnyu:
`policy_trainer.play()` is already decorated with the eval mode. Ckpt loading happens inside it but not outside, so you should be able to directly check the mode in that function?

QuantuMope:
> `policy_trainer.play()` is already decorated with the eval mode. Ckpt loading happens inside it but not outside, so you should be able to directly check the mode in that function?

Oh that must mean I'm merging the weights before they're loaded by the checkpoint manager. No errors since checkpoint loading strict is set to False. Need to investigate this.

hnyu:
Maybe you are using in-alg ckpt and merging the weights inside init.

QuantuMope:
> Maybe you are using in-alg ckpt and merging the weights inside init.

Right. That was the issue. No need for this PR

PR #1803, opened 2025-08-28
add http_server:8080/render; fix capture frame; add more logging

le-horizon:
See http://gail8.local:8080/render

image

Sometimes recorder is receiving a single frame.  Add defense against that.

### DIFF
--- a/alf/trainers/policy_trainer.py
+++ b/alf/trainers/policy_trainer.py
@@ -1023,7 +1023,8 @@ def _step(algorithm,
     policy_step = algorithm.predict_step(transformed_time_step, policy_state)
 
     if recorder and selective_criteria_func is None:
-        recorder.capture_frame(policy_step.info, time_step.is_last())
+        recorder.capture_frame(policy_step.info,
+                               torch.all(time_step.is_last()))
 
     elif recorder and selective_criteria_func is not None:
         env_frame = recorder.capture_env_frame()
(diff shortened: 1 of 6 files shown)


alf/trainers/policy_trainer.py
```diff
@@ -1023,7 +1023,8 @@ def _step(algorithm,
     policy_step = algorithm.predict_step(transformed_time_step, policy_state)
 
     if recorder and selective_criteria_func is None:
-        recorder.capture_frame(policy_step.info, time_step.is_last())
+        recorder.capture_frame(policy_step.info,
+                               torch.all(time_step.is_last()))
```
Haichao-Zhang, line 1027: should not use torch.all here? instead, we should assert batch size is 1 before calling capture_frame?

le-horizon: The assert already happens in play() in this change.  We allow recording with batch size > 1 now, in which case, the env will stack the frames into one.

I don't know if we want to do anything special about is_last_step in the batched case.

Haichao-Zhang: What I mean is recording for a batch size > 1 is not well defined. 

Yes, I saw the assert added in play. With that, do you still need torch.all? 

I feel currently the assumption on which scenario (B=1, B>1) is under consideration is not clear and a bit mixed.

le-horizon: Yes, the assert in play basically allows rendering with bs 1 or when the env is batched, in which case, the env is responsible for stacking the images, but the time_step is still batched, hence requires torch.all().

Not sure how best to make this clearer.

Haichao-Zhang (dismissed):

Haichao-Zhang (approved):

PR #1805, opened 2025-09-17
Fix async_eval error "pidfd_getfd: Operation not permitted"

emailweixu:
This happens on certain combination of pytorch and os configuration. The reason is that shared cuda tensor need to use pidfd_getfd, which is not allowed by default in some system.

### DIFF
--- a/alf/trainers/evaluator.py
+++ b/alf/trainers/evaluator.py
@@ -14,6 +14,7 @@
 
 from absl import logging
 from absl import flags
+import ctypes
 import math
 import torch.multiprocessing as mp
 import os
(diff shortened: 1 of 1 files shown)


alf/trainers/evaluator.py
```diff
@@ -41,6 +42,30 @@
                      defaults=[None] * 4)
 
 
+def _allow_child_to_ptrace(child_pid: int) -> None:
+    """
+    In the *parent* process: allow a given child to ptrace us.
+    This relaxes Yama's ptrace restriction for this specific relationship.
+
+    Requires: kernel.yama.ptrace_scope <= 1 (default on many distros).
```
le-horizon, line 50: maybe add assert instead of just comment about this requirement?

alf/trainers/evaluator.py
```diff
@@ -41,6 +42,30 @@
...
+    The value of kernel.yama.ptrace_scope can be checked with:
+        cat /proc/sys/kernel/yama/ptrace_scope
+    """
+    libc = ctypes.CDLL(ctypes.util.find_library("c"), use_errno=True)
+
+    # prctl constants from <linux/prctl.h>:
+    PR_SET_PTRACER = 0x59616D61
```
le-horizon, line 57: maybe comment a bit more whether this const is specific to versions of the kernel or general.

le-horizon (approved):
LGTM.  Optional comments.

PR #1807, opened 2025-10-02
forward spawn context to create shared replay buffer

le-horizon:
This change first avoids changing global multiprocess context in distributed off policy alg: ``mp.set_start_method('spawn', force=True)``
and instead uses ``ctx = mp.get_context('spawn')`` to start the receive_experience process.

Due to the change, we also need to pass context into replay buffer, so that the process Lock objects inside the replay buffer can be created using the same multiprocess context.

From now on, the creation of the processes interacting with the replay buffer needs to happen together with replaybuffer initialization, so that the mp context can be passed correctly.

Tested distributed trainer and unroller.  They no longer crash due to mp context and lock being created in a different context.  They still crash, due to other unrelated reason (probably due to running openvlappo agent with distributed unroller and trainer for the first time):
  File "/home/horizon/Desktop/alf/alf/algorithms/distributed_off_policy_algorithm.py", line 306, in pull_params_from_trainer
    params.buf[1:] = data
ValueError: memoryview assignment: lvalue and rvalue have different structures

(Since we are not going to use distributed unroll for now, I'll leave this as TODO.)

### Testing:

To run as trainer, add --as_remote_trainer (to your alf.bin.train command)

To run as unroller, add --as_remote_unroller  --conf_param="TrainerAddrConfig.ip='10.40.0.117'" (to your alf.bin.train command.  Use the IP address of the trainer machine here if it's on a different machine; defaults to localhost).

### DIFF
--- a/alf/utils/common.py
+++ b/alf/utils/common.py
@@ -17,6 +17,8 @@ from absl import flags
 from absl import logging
 import contextlib
 import copy
+import ctypes
+import ctypes.util
 from fasteners.process_lock import InterProcessLock
 from filelock import FileLock
 from functools import wraps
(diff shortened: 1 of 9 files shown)


alf/utils/common.py
```diff
@@ -95,6 +97,31 @@ def wrapper(*args, **kwargs):
     return decorator
 
 
+def allow_child_to_ptrace(child_pid: int) -> None:
+    """In the *parent* process: allow a given child to ptrace us.
```
emailweixu, line 101: Add explanation why we need to allow child to ptrace us. That is, avoiding "pidfd_getfd: Operation not permitted" for Tensor sharing.

le-horizon: Done.  PTAL.

emailweixu (approved):

PR #1808, opened 2025-10-03
Add entropy field to ActorCriticInfo and PPOInfo

QuantuMope:
Adds an explicit entropy field that can be used for entropy regularization. Useful if the entropy must be computed from some custom distribution that is not in `dist_utils.py`'s `_get_builder_map`.

### DIFF
--- a/alf/algorithms/actor_critic_loss.py
+++ b/alf/algorithms/actor_critic_loss.py
@@ -216,8 +216,14 @@ class ActorCriticLoss(Loss):
 
         entropy_loss = ()
         if self._entropy_regularization is not None:
-            entropy, entropy_for_gradient = dist_utils.entropy_with_fallback(
-                info.action_distribution, return_sum=False)
+            # If entropy is explicitly provided, we'll use it.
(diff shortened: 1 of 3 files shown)


alf/algorithms/actor_critic_loss.py
```diff
@@ -216,8 +216,14 @@ def _summarize(v, r, adv, suffix):
 
         entropy_loss = ()
         if self._entropy_regularization is not None:
-            entropy, entropy_for_gradient = dist_utils.entropy_with_fallback(
-                info.action_distribution, return_sum=False)
+            # If entropy is explicitly provided, we'll use it.
+            # Otherwise, we will compute it from the provided action_distribution.
+            if info.entropy is not ():
+                entropy = info.entropy
+                entropy_for_gradient = info.entropy
```
emailweixu, line 223: you can use info.log_prob in this case?

QuantuMope: I guess this would result in a monte carlo estimate of the entropy, right? For me, I have access to the analytical entropy.

emailweixu (approved):

PR #1815, opened 2025-10-17
add uv package management support to alf, update README

le-horizon:
Tested locally with ``uv run python -m unittest alf/algorithms/agent_test.py``

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,86 @@
 [build-system]
-requires = ["setuptools>=42", "wheel", "pybind11"]
+requires = [
+    "setuptools>=42",
+    "wheel",
+    "pybind11",
+    "tomli>=1.1; python_version < \"3.11\"",
+]
 build-backend = "setuptools.build_meta"
+
+[project]
+name = "alf"
+version = "0.1.0"
+description = "Agent Learning Framework for flexible RL research."
+readme = "README.md"
+requires-python = ">=3.10,<3.13"
+license = { file = "LICENSE" }
+authors = [{ name = "Horizon Robotics" }]
+dependencies = [
+    "absl-py==2.2.2",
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -1,3 +1,86 @@
...
+    "pybullet==2.5.0",
+    "pydocstyle==6.3.0",
+    "pylint==2.17.7",
+    "pyglet==1.3.2",
+    "rectangle-packer==2.0.4",
+    "tensorboard==2.19.0",
+    "threadpoolctl==3.6.0",
+    "torch==2.8.0+cu128",
+    "torchtext==0.18.0",
+    "torchvision==0.23.0+cu128",
```
QuantuMope, line 49: If changing any package versions, we should make sure that alf install into Hobot install doesn't break.

le-horizon: Yes, see https://github.com/HorizonRoboticsInternal/Hobot/pull/1862
I've tested on 5090, as well as 3090 with cuda 12.4. Both work.

QuantuMope: Currently the package versions match those that are in our cluster and CICD docker. Should we try to keep these as close as possible for max repeatability?

le-horizon: > Currently the package versions match those that are in our cluster and CICD docker. Should we try to keep these as close as possible for max repeatability?

Current versions are too old, doesn't support newer GPUs like 5090.  Perhaps we should upgrade docker instead?

QuantuMope: Sure, that's also an option. Just keep track of all the changed versions and I can update the docker. Is it just torch?

le-horizon: I guess if we switch over to uv for cluster training, then we don't need to worry about this.  Let's see.

emailweixu (approved):

PR #1816, opened 2025-10-17
Allow multiple alf.define_config calls for nested confs

QuantuMope:
Quality of life feature for nested confs.

Let's say I have a conf file A with `alf.define_config("x", 4)`.

To reduce duplication, I use `alf.import_config("A")`, in conf file B, but in this conf file, the default arg of "x" should be changed. Currently, multiple `alf.define_config` calls will error. With this PR, we can now do something like

```python
x = alf.define_config("x", 3)
# A defines "x" too, but it is ignored since we define it first in the outer conf
conf = alf.import_config("A")
print(x) -> 3
```

### DIFF
--- a/alf/config_util.py
+++ b/alf/config_util.py
@@ -21,7 +21,6 @@ from inspect import Parameter
 import os
 import pprint
 import runpy
-import shutil
 
 __all__ = [
     'config',
@@ -994,6 +993,9 @@ def define_config(name, default_value):
 
     Its value can be retrieved by ``get_config_value("_CONFIG._USER.{name}")``.
 
+    If the configurable has already been defined, subsequent define_config calls
+    will the same name will be ignored and the already set default value will be returned.
+
     Args:
         name (str): name of the configurable value
         default_value (Any): default value
@@ -1002,8 +1004,15 @@ def define_config(name, default_value):
     """
     node = _Config()
     node.set_default_value(default_value)
-    _add_to_conf_tree(['_CONFIG'], '_USER', name, node)
-    _DEFINED_CONFIGS.append('_CONFIG._USER.' + name)
+    try:
+        _add_to_conf_tree(['_CONFIG'], '_USER', name, node)
+        _DEFINED_CONFIGS.append('_CONFIG._USER.' + name)
+    except ValueError:
+        already_set_default_value = get_config_value("_CONFIG._USER." + name)
+        logging.warning(
+            f"Config {name} has already been configured. Provided value {default_value} will be ignored "
+            f"in favor of {already_set_default_value}")
+
     return get_config_value("_CONFIG._USER." + name)
 
 
--- a/alf/config_util_test.py
+++ b/alf/config_util_test.py
@@ -342,6 +342,11 @@ class ConfigTest(alf.test.TestCase):
         test_partial = partial(TestPositionalArgs, y=0)
         test_partial()
 
+    def test_define_config(self):
+        val = alf.define_config("foobar", 3)
+        new_val = alf.define_config("foobar", 4)
+        self.assertEqual(val, new_val)
+
 
 if __name__ == '__main__':
     alf.test.main()


emailweixu (approved):

le-horizon:
> Quality of life feature for nested confs.
> 
> Let's say I have a conf file A with `alf.define_config("x", 4)`.
> 
> To reduce duplication, I use `alf.import_config("A")`, in conf file B, but in this conf file, the default arg of "x" should be changed. Currently, multiple `alf.define_config` calls will error. With this PR, we can now do something like
> 
> ```python
> x = alf.define_config("x", 3)
> # A defines "x" too, but it is ignored since we define it first in the outer conf
> conf = alf.import_config("A")
> print(x) -> 3
> ```

Perhaps not so safe to do this.  If the two define_conf default values are different, then the value would be different for different parts of the code.  Having only one define_config forces us to move the definition to the right place.

In the example, would it be possible to import A before the define_config, and refer to x as conf.x?

QuantuMope:
> > Quality of life feature for nested confs.
> > Let's say I have a conf file A with `alf.define_config("x", 4)`.
> > To reduce duplication, I use `alf.import_config("A")`, in conf file B, but in this conf file, the default arg of "x" should be changed. Currently, multiple `alf.define_config` calls will error. With this PR, we can now do something like
> > ```python
> > x = alf.define_config("x", 3)
> > # A defines "x" too, but it is ignored since we define it first in the outer conf
> > conf = alf.import_config("A")
> > print(x) -> 3
> > ```
> 
> Perhaps not so safe to do this. If the two define_conf default values are different, then the value would be different for different parts of the code. Having only one define_config forces us to move the definition to the right place.
> 
> In the example, would it be possible to import A before the define_config?

This feature is to enable having a default value specific to each conf. Otherwise, you're forced to use the default value of config A. If you were to import config A first, and config A defines `alf.define_config("x", 4)` when in the current config, I need a default value of 3, then the value would stay 4.

Oh by the way, by setting a new default value before the import config, the config A will use the value of 3, not whatever is defined within the conf.
```python
# Inside conf A
x = alf.define_config("x", 4)
alf.config("Something", x=x)
```

```python
x = alf.define_config("x", 3)
conf = alf.import_config("A")  <---- if conf A uses "x", the value of 3 will be used.
print(x) -> 3
```

le-horizon:
> > > Quality of life feature for nested confs.
> > > Let's say I have a conf file A with `alf.define_config("x", 4)`.
> > > To reduce duplication, I use `alf.import_config("A")`, in conf file B, but in this conf file, the default arg of "x" should be changed. Currently, multiple `alf.define_config` calls will error. With this PR, we can now do something like
> > > ```python
> > > x = alf.define_config("x", 3)
> > > # A defines "x" too, but it is ignored since we define it first in the outer conf
> > > conf = alf.import_config("A")
> > > print(x) -> 3
> > > ```
> > 
> > 
> > Perhaps not so safe to do this. If the two define_conf default values are different, then the value would be different for different parts of the code. Having only one define_config forces us to move the definition to the right place.
> > In the example, would it be possible to import A before the define_config?
> 
> This feature is to enable having a default value specific to each conf. Otherwise, you're forced to use the default value of config A. If you were to import config A first, and config A defines `alf.define_config("x", 4)` when in the current config, I need a default value of 3, then the value would stay 4.
> 
> Oh by the way, by setting a new default value before the import config, the config A will use the value of 3, not whatever is defined within the conf.
> 
> ```python
> # Inside conf A
> x = alf.define_config("x", 4)
> alf.config("Something", x=x)
> ```
> 
> ```python
> x = alf.define_config("x", 3)
> conf = alf.import_config("A")  <---- if conf A uses "x", the value of 3 will be used.
> print(x) -> 3
> ```

I see, looks good then.  I typically pre_config "x" to a new value (e.g. via command line argument).

QuantuMope:
> > > > Quality of life feature for nested confs.
> > > > Let's say I have a conf file A with `alf.define_config("x", 4)`.
> > > > To reduce duplication, I use `alf.import_config("A")`, in conf file B, but in this conf file, the default arg of "x" should be changed. Currently, multiple `alf.define_config` calls will error. With this PR, we can now do something like
> > > > ```python
> > > > x = alf.define_config("x", 3)
> > > > # A defines "x" too, but it is ignored since we define it first in the outer conf
> > > > conf = alf.import_config("A")
> > > > print(x) -> 3
> > > > ```
> > > 
> > > 
> > > Perhaps not so safe to do this. If the two define_conf default values are different, then the value would be different for different parts of the code. Having only one define_config forces us to move the definition to the right place.
> > > In the example, would it be possible to import A before the define_config?
> > 
> > 
> > This feature is to enable having a default value specific to each conf. Otherwise, you're forced to use the default value of config A. If you were to import config A first, and config A defines `alf.define_config("x", 4)` when in the current config, I need a default value of 3, then the value would stay 4.
> > Oh by the way, by setting a new default value before the import config, the config A will use the value of 3, not whatever is defined within the conf.
> > ```python
> > # Inside conf A
> > x = alf.define_config("x", 4)
> > alf.config("Something", x=x)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```python
> > x = alf.define_config("x", 3)
> > conf = alf.import_config("A")  <---- if conf A uses "x", the value of 3 will be used.
> > print(x) -> 3
> > ```
> 
> I see, looks good then. I typically pre_config "x" to a new value (e.g. via command line argument).

Right. That's how I was also doing it but it got pretty hard to track after several values diverged.

PR #1821, opened 2025-10-29
downgrade torch to 2.7.1, tensorboard to 2.15, remove pathos to avoid conflict with openpi & lerobot

le-horizon:
as title


emailweixu (commented):
uv sync

emailweixu (dismissed):

emailweixu:
> uv sync

never mind

le-horizon:
Turns out some more is needed to be compatible with openpi (tensorboard change)

emailweixu (approved):

PR #1824, opened 2025-12-10
[Bug Fix] Make DDP workers set absl logging as default

QuantuMope:
Discovered a bug recently where DDP worker log files were being created but not written to.

Ultimately discovered that this is caused by spawned subprocesses using python's logging module by default as they create a new interpreter. Adding `logging.use_absl_handler()` (suggested by Wei) solves this issue.

One mystery that remains is that hobot1 projects didn't seem to have this issue with no recent code changes to `train.py` offering any hints. Low priority figuring this out for now as this new change fixes the current main.

### DIFF
--- a/alf/bin/train.py
+++ b/alf/bin/train.py
@@ -118,17 +118,15 @@ def check_valid_launch():
         assert not extra, f"Unexpected environment variables for non-distributed launch: {extra}"
 
 
-def _setup_logging(rank: int, log_dir: str):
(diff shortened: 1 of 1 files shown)


emailweixu (dismissed):

QuantuMope:
Sorry, needs a restamp. Found out that pretty file names results in duplicate log files so removed it.

emailweixu (approved):

PR #1826, opened 2026-01-21
Do not include non_persistent buffer in state_dict

emailweixu:
Per the original implementation of state_dict for pytorch Module, non persistent buffers are not included in state_dict.

Background: state_dict usually include both parameters and buffers. However, if a buffer is registered with persistent=False, it will not be included in state_dict (the default value for persistent is True).

### DIFF
--- a/alf/algorithms/algorithm.py
+++ b/alf/algorithms/algorithm.py
@@ -1050,7 +1050,8 @@ class Algorithm(AlgorithmInterface):
                 visited.add(param)
                 destination[prefix + name] = param.detach()
         for name, buf in self._buffers.items():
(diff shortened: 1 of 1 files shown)


le-horizon (approved):
LG.

Could you add a bit more explanation in the PR desc, which case this fix is for?  E.g. two modules sharing parameter? or the other case?

Thanks,
Le

emailweixu:
> LG.
> 
> Could you add a bit more explanation in the PR desc, which case this fix is for? E.g. two modules sharing parameter? or the other case?
> 
> Thanks, Le

Updated

PR #1829, opened 2026-03-13
Add average_all_summaries context manager

QuantuMope:
A useful context manager for computing summary averages over an interval.

Nice to have because several alf algorithms have averaging off as default.
If using these summaries with a small mini-batch-size, we can selectively control which nested summaries get averaged without having to overwrite the source code.

### DIFF
--- a/alf/summary/summary_ops.py
+++ b/alf/summary/summary_ops.py
@@ -102,51 +102,43 @@ class scope(object):
 
 
 @contextlib.contextmanager
-def average_all_summaries(cond: Callable, target_names: List[str] = None):
+def average_all_summaries(cond: Callable, target_names: List[str]):
     """
-    Context manager that sets all nested scalar summaries to average.
-    It also disables any nested recording interval logic.
-    Scalar summaries with an explicit average_over_summary_interval=False
-    will not be overridden.
-    
+    Context manager that sets selected nested scalar summaries to average.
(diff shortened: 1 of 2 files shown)


alf/summary/summary_ops.py
```diff
@@ -99,6 +101,57 @@ def __exit__(self, type, value, traceback):
...
+
+        return wrapped
+
+    @contextlib.contextmanager
+    def _disabled_record_if(*args, **kwargs):
+        yield
+
+    alf.summary.scalar = _wrap(orig_scalar)
+    alf.summary.record_if = _disabled_record_if
+    alf.summary.should_record_summaries = lambda: True
```
emailweixu: Why do these two need to be changed?

QuantuMope: `record_if` must be changed so that the recording interval set by the user is actually used.

`should_record_summaries` actually shouldn't be overridden as doing so is redundant. I've removed this check. Unit test still passes successfully.

QuantuMope: Hey @emailweixu, I've pushed a fix concerning our offline discussion. Now the user has to provide a set of target names, so summaries that shouldn't be messed with don't get caught by accident. It also ensures that record_if logic does not get altered for any nested summary not included in target_names.

You can take a look at the updated unit test to see the correct behavior.

emailweixu (approved):

PR #1830, opened 2026-03-27
Add compute_episodic_return_on_last_step

QuantuMope:
Currently, MC returns only get computed when encountering a discount of 0. This makes it so that MC returns cannot be computed if we do not terminate on success. This PR adds a flag `compute_episodic_return_on_last_step` to instead compute returns upon `step_type.LAST` instead.

### DIFF
--- a/alf/experience_replayers/replay_buffer.py
+++ b/alf/experience_replayers/replay_buffer.py
@@ -66,6 +66,7 @@ class ReplayBuffer(RingBuffer):
                  mp_context=None,
                  keep_episodic_info=None,
                  record_episodic_return=False,
+                 compute_episodic_return_on_last_step=False,
                  default_return=-1000.,
                  gamma=.99,
                  reward_clip=None,
@@ -113,6 +114,11 @@ class ReplayBuffer(RingBuffer):
                     ``ReplayBuffer.reward_clip=(-1,1)``.
                 2) Discount ``gamma`` needs to be set consistent with ``TDLoss.gamma``.
                 3) Assumes ``keep_episodic_info`` to be True.
+            compute_episodic_return_on_last_step (bool): If True, compute episodic
+                return when a LAST step is encountered regardless of the discount factor.
+                Default behavior when False is to compute the return when a discount factor
+                of 0 is encountered. If True, keep_episodic_return and record_episodic_return
+                must also be True.
             default_return (float): The default values of ``discounted_return``
                 when the episode has not ended.  For value target lower bounding,
                 default_return should not be bigger than the smallest possible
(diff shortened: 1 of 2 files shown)


emailweixu:
> Currently, MC returns only get computed when encountering a discount of 0. This makes it so that MC returns cannot be computed if we do not terminate on success. This PR adds a flag `compute_episodic_return_on_last_step` to instead compute returns upon `step_type.LAST` instead.

Why don't we set discount to 0 for LAST step?

Haichao-Zhang:
> This makes it so that MC returns cannot be computed if we do not terminate on success

1. keep adding more flags might not be ideal and might be a bit confusing to users? 
2. ``This makes it so that MC returns cannot be computed if we do not terminate on success`` --> not sure what kind of use case you are considering, but typically, upon task success, we set step type as LAST + discount as 0?

QuantuMope:
> > Currently, MC returns only get computed when encountering a discount of 0. This makes it so that MC returns cannot be computed if we do not terminate on success. This PR adds a flag `compute_episodic_return_on_last_step` to instead compute returns upon `step_type.LAST` instead.
> 
> Why don't we set discount to 0 for LAST step?

This is for RL setups where we don't terminate on success and only terminate upon timeout.
For EmbodiedGen resets, batch resetting all environments at the same step is often more efficient and sometimes the only possible strategy (e.g., if lighting randomization is turned on, cannot reset individual envs on GPU).

QuantuMope:
> > This makes it so that MC returns cannot be computed if we do not terminate on success
> 
> 1. keep adding more flags might not be ideal and might be a bit confusing to users?
> 2. `This makes it so that MC returns cannot be computed if we do not terminate on success` --> not sure what kind of use case you are considering, but typically, upon task success, we set step type as LAST + discount as 0?

1. Would making this the default behavior be better?
2. Most of our OpenVLAPPO, PPOFlow, SACFlow RL training uses timeout termination only. This would model an infinite horizon RL formulation.

emailweixu:
Why can't we set discount to 0 at the time of setting step type to LAST?

QuantuMope:
> Why can't we set discount to 0 at the time of setting step type to LAST?

_Accidentally editted the question comment instead of replying. Simply fixing that._

I think in practice we could, but it would be less correct to do so? We're essentially modeling an infinite time horizon problem. Applying a discount of 0 at time out when there were numerous success idle steps prior may cause confusion to the model?

QuantuMope:
After some investigating, this flag is necessary to maintain correctness during TD bootstrapping for the infinite horizon RL formulation.

There is an important caveat where if `compute_episodic_return_on_last_step==True`, the computed MC return can be heavily biased. Added an explanation of why in commit [35a6b6d](https://github.com/HorizonRobotics/alf/pull/1830/commits/35a6b6d93708b409f495556dc615ce714b513705).

emailweixu (approved):

PR #1833, opened 2026-03-30
Make unroll_length schedulable

QuantuMope:
This PR allows for a scheduled unroll length if we are running synced off-policy RL training:
1) `async_unroll=False`
2) `whole_replay_buffer_training=False`

It also allows for a scheduled value of 0, which in turn skips unrolling to train from the replay buffer.

This allows us to "simulate" very diverse training strategies. E.g.,
```python
unroll_length = StepScheduler("iterations", [
    # unroll to collect an "offline" dataset
    (1, int(initial_collect_steps / num_para_envs)),        
    # perform offline training iterations. no unroll 
    (offline_training_iters, 0),     
    # continue with online RL                                    
    (offline_training_iters + 1, desired_unroll_length)])
```

Codex cleverly makes a minimal change with full backward compatibility by adding the following code
```python
    @property
    def unroll_length(self):
        return self._unroll_length()

    @unroll_length.setter
    def unroll_length(self, value):
        self._unroll_length = as_scheduler(value)
```

### DIFF
--- a/alf/algorithms/config.py
+++ b/alf/algorithms/config.py
@@ -15,7 +15,7 @@
 from typing import Optional, Callable
 import torch
 import alf
-from alf.utils.schedulers import as_scheduler
+from alf.utils.schedulers import ConstantScheduler, as_scheduler
 
 
 @alf.configurable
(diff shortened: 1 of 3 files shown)


alf/algorithms/config.py
```diff
@@ -389,6 +394,16 @@ def __init__(self,
         self.unroll_with_grad = unroll_with_grad
         self.use_root_inputs_for_after_train_iter = use_root_inputs_for_after_train_iter
         self.async_unroll = async_unroll
+        if not isinstance(self._unroll_length, ConstantScheduler):
```
Haichao-Zhang, line 397: ConstantScheduler --> should check against a base class, e.g. Scheduler?

QuantuMope: We need to check against `ConstantScheduler` here because a scalar input will be converted to one before this check due to the setter function on line 479.

Any non-constant scheduler should then raise an error if we're doing on-policy or async unroll.

QuantuMope (commented):
Hey Haichao, responded to your comment. Let me know if I misunderstood it.

QuantuMope:
Gentle reminder for review. Thanks

Haichao-Zhang (approved):

PR #1835, opened 2026-05-27
Add several fixes to distributed training

QuantuMope:
This PR adds several fixes for distributed training. The first 5 are in the first commit d2813241f06c8c4c665214b95c0c040456052637 while the last one is in the second commit 46887083425c45c12538e200333c44635b037079.

1. Removes optimizers from distributed wrappers and instead handles state dicts according to the wrapped core alg. This makes it so that checkpoints are agnostic to whether they are being trained with distributed trainer.
2. Add missing `summarize_metrics(self)` call so that summaries from the algorithm are recorded.
3. Add `weights_only=False` to `torch.load` for `torch.__version__ >= 2.6`.
4. Increase unroller acknowledge timeout from 3 to 60 seconds. Three seconds is generally too fast for transmitting VLA weights.
5. Bug fix in `DistributedUnroller` triggered by `i % 0`.
6. Checkpoint loading support for `DistributedTrainer`.

### DIFF
--- a/alf/algorithms/distributed_off_policy_algorithm.py
+++ b/alf/algorithms/distributed_off_policy_algorithm.py
@@ -503,26 +503,41 @@ class DistributedTrainer(DistributedOffPolicyAlgorithm):
         thread.daemon = True
         thread.start()
 
-    def _create_data_receiver_subprocess(self):
-        """Create a process to receive experience data from unrollers.
+    def _create_replay_buffer_sample_experience(self):
         """
-        # First create the replay buffer in the main process. For this, we need
-        # to create a dummy experience to set up the replay buffer.
+        Create a sample experience used to initialize the replay buffer.
+        """
         time_step = self._env.current_time_step()
         rollout_state = self.get_initial_rollout_state(self._env.batch_size)
         alg_step = self.rollout_step(time_step, rollout_state)
         exp = make_experience(time_step, alg_step, rollout_state)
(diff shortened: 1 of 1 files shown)


le-horizon (commented):
Thanks for the fixes, Andrew.

A couple points from codex if they make sense:
1) alf/algorithms/distributed_off_policy_algorithm.py:155
DistributedOffPolicyAlgorithm.state_dict() now returns only self._core_alg.state_dict(). The trainer replay buffer lives on the distributed wrapper, not the core alg, so distributed checkpoints saved after this change will not include _replay_buffer.* keys. That makes the new checkpoint-loading setup at lines 519-569 ineffective for replay data: resume creates an empty multiprocessing replay buffer, then loads only core model/optimizer state. It also cannot load older distributed checkpoints that do contain wrapper _replay_buffer.* keys because load_state_dict() delegates straight to the core alg.

Need to throw error when resuming a ckpt encounters an empty replay buffer.
Need to add unittest coverage for the correct ckpt and reload.

2) alf/algorithms/distributed_off_policy_algorithm.py:211
summarize_metrics() now only calls self._core_alg.summarize_metrics(). The wrapper owns the env metrics created by RLAlgorithm.__init__, and unroller rollout updates those wrapper metrics via inherited observe_for_metrics(). This drops normal unroller env metric summaries such as episode count, env steps, return, and episode length. This should likely call super().summarize_metrics() as well.

